PR #419, opened 2014-11-19
WIP: Adds support for scale out groups

andreasohlund:
### Things to do

- [ ] Add EndpointInstance doc
- [ ] Plugin needs to distinguish between a scaleout group not found and a group with no online instances
- [x] Rename to connect/disconnect
- [ ] Retries need to use scale out  
[This code](https://github.com/Particular/ServiceControl/blob/e292cdae819c2451b34aaf1455edc346ac09afbe/src/ServiceControl/Contracts/Operations/ImportFailedMessage.cs#L31) checks the FailedQ header, so we need to update [this code](https://github.com/Particular/ServiceControl/blob/e292cdae819c2451b34aaf1455edc346ac09afbe/src/ServiceControl/MessageFailures/Handlers/PerformRetryHandler.cs#L65) to do a lookup in the scaleout group, we need latest v5 because we need the translator impl.


andreasohlund:
Code has been moved to the new capability repo: https://github.com/Particular/Capability.Routing lets continue the work over there

PR #437, opened 2015-01-09
Issue-432

gbiellem:
Changed MSMQ bulk importer to forward message that cannot be converted to a transport message due to bad header serialization to the ServiceControl errors queue (e.g. particular.servicecontrol.errors) rather than discarding them


gbiellem:
@SimonCropp  implemented as discussed.

PR #441, opened 2015-01-28
[WIP] Stabilize

SzymonPobiega:
The purpose of this pull is to stabilize the SC so that we can start moving forward with composite architecture approach

### DIFF
--- a/src/ServiceControl/ConfigTransportConfig.cs
+++ b/src/ServiceControl/ConfigTransportConfig.cs
@@ -9,7 +9,7 @@ namespace Particular.ServiceControl
         {
             return new TransportConfig
             {
-
+                MaximumMessageThroughputPerSecond = 350,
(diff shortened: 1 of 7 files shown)


SzymonPobiega:
@gbiellem @andreasohlund what else do we need to call it done and move to splitting the codebase and revamping the installer?

src/ServiceControl/ConfigTransportConfig.cs
```diff
@@ -9,7 +9,7 @@ public TransportConfig GetConfiguration()
         {
             return new TransportConfig
             {
-
+                MaximumMessageThroughputPerSecond = 350,
```
andreasohlund: Do we need to expose this as a setting in app.config since this value might be different depending on hardware?

andreasohlund:
I'm good, lets release this?

PR #443, opened 2015-01-30
Using Nowin (https://github.com/Bobris/Nowin/blob/master/README.md) as the host

johnsimons:
This avoids to have to configure URLALCs

Issues:
- It seems that we would need to expose raven in a different port (I couldn't get it working)
- Custom VirtualDirectory is not supported at the moment (can be fixed)


johnsimons:
@Particular/servicecontrol thoughts?

andreasohlund:
Looks interesting. 

I'd say

>Current status is usable for testing, not for production, nobody did any security review, you have been warned

makes it a non-starter for now but lets keep an eye on it?

synhershko:
I may be a little bit late to the party, but considering Nowin is "Current status is usable for testing, not for production" as per the README, why not just auto-register URLACLs? Raven does that and nancy self host supports that as well

gbiellem:
@synhershko 

I understand @johnsimons frustration with URLACL.  Some common issues with it:

- Auto registration fails if the process isn't running with admin rights.   That's why nancy self host allows it be turned off. 

- Registrations have to be changed when the hostname is changed or we want to listen beyond localhost.  

- Overlapping registrations cause HTTP 500 errors but the host process has no idea there is a problem

gbiellem:
@johnsimons - Based on @andreasohlund  [comment](https://github.com/Particular/ServiceControl/pull/443#issuecomment-72170499) I think we should kill this off for now

PR #449, opened 2015-02-16
Fix reimporting

SzymonPobiega:
This pull tries to address the issue in reimport code that caused loss of header values when attempting to reimport
 * Instead of creating a failed import documents, forward messages to special queues (error.failedimports and audit.failedimports)
 * If there are existing failed import documents in the DB, log a warning instructing user go through them in the maintenance mode and delete if they are no longer relevant

### DIFF
--- a/src/ServiceControl/Operations/Msmq/MsmqAuditQueueImporter.cs
+++ b/src/ServiceControl/Operations/Msmq/MsmqAuditQueueImporter.cs
@@ -27,12 +27,6 @@ namespace ServiceControl.Operations
             this.store = store;
             this.builder = builder;
             enabled = receiver is MsmqDequeueStrategy;
-
-            importFailuresHandler = new SatelliteImportFailuresHandler(store,
-                Path.Combine(Settings.LogPath, @"FailedImports\Audit"), tm => new FailedAuditImport
-                {
-                    Message = tm,
-                });
         }
 
         public IBus Bus { get; set; }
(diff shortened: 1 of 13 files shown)


SzymonPobiega:
@gbiellem @indualagarsamy please review

src/ServiceControl/Operations/Msmq/MsmqAuditQueueImporter.cs
```diff
@@ -27,12 +27,6 @@ public MsmqAuditQueueImporter(IDocumentStore store, IBuilder builder, IDequeueMe
             this.store = store;
             this.builder = builder;
             enabled = receiver is MsmqDequeueStrategy;
```
andreasohlund: Since we've looked the throughput to 300 msg/s should we consider removing the batching all together since we don't actually need it? (ie go with the plain vanilla msmq dequeue strategy)

andreasohlund: cc @johnsimons

SzymonPobiega: Makes sense to me but I wanted this particular change to be as small as possible without making any other assumptions outside of the actual area being fixed

andreasohlund: Agreed, lets touch base with @johnsimons and treat that as a seprate task if needed.

I shouldn't have brought it up here. My bad!

andreasohlund: Separate issue raised: https://github.com/Particular/ServiceControl/issues/450

gbiellem:
@SzymonPobiega Is this PR dead? :ghost:

SzymonPobiega:
@gbiellem I think it is. Will be obsoleted by whatever we build as part of https://github.com/Particular/Requirements/issues/170. I don't think there is much to salvage from here.

gbiellem:
OK @SzymonPobiega closing this

PR #474, opened 2015-06-03
[WIP] 272 group failures by exception type

danielmarbach:
@wolfbyte turned it into PR


danielmarbach:
@wolfbyte You seem not to be using the index. Is this a leftover?

wolfbyte:
GetAllExceptionGroups uses it. ExceptionGroups is not stored in RavenDB

danielmarbach:
@wolfbyte wouldn't it be better to have an index which groups the failed messages by exception type and  then essentially use that to determine the exceptions groups filtering on a specific group. Because now you are using a dynamic index

wolfbyte:
@danielmarbach I'm not entirely certain what you mean. This is really day 1 for me with RavenDB :) If you can point me in the right direction I will fix it up tomorrow.

danielmarbach:
@wolfbyte 

Something like this:

```
public class MessageFailureHistory_ByExceptionGroup : AbstractIndexCreationTask<MessageFailureHistory, MessageFailureHistory_ByExceptionGroup.ReduceResult>
        {
            public class ReduceResult
            {
                public string ExceptionGroup { get; set; }
                public int Count { get; set; }
                public string[] HistoryIds { get; set; }
            }

            public MessageFailureHistory_ByExceptionGroup()
            {
                Map = messages => from message in messages
                                  let exceptionGroup = message.ProcessingAttempts.Last().ExceptionType
                                  select new
                                             {
                                                 ExceptionGroup = exceptionGroup,
                                                 Count = 1,
                                                 HistoryIds = new[] { message.Id }
                                             };

                Reduce = results => from result in results
                                    group result by result.ExceptionGroup
                                        into g
                                        select new
                                                   {
                                                       ExceptionGroup = g.Key,
                                                       Count = g.Sum(x => x.Count),
                                                       HistoryIds = g.SelectMany(m => m.HistoryIds)
                                                   };
            }
        }
```

This index creates the groups, the counts and adds all ids of the `MessageFailureHistory` into the HistoryIds. Then when a user collects you can bulk load the history with the Ids in a paged way. You can actually load multiple ids on the same call if you want. 

Then you can just query it:

```

            var stuff = session.Query<MessageFailureHistory_ByExceptionGroup.ReduceResult, MessageFailureHistory_ByExceptionGroup>()
                .Where(r => r.ExceptionGroup == typeof(StackOverflowException).FullName)
                .ToList();
```
I did a quick run with 300 0000 failed messages in 3 groups

Here are the index stats
image

Here the index result

image

This is the suite I used:

```
[Test]
        public void Foo()
        {
            var store = new DocumentStore { Url = "http://localhost:8083", DefaultDatabase = "FooTest" };
            store.Initialize(true);
            IndexCreation.CreateIndexes(typeof(FooTest).Assembly, store);

            var bulkInsert = store.BulkInsert();
            for (int i = 0; i < 100000; i++)
            {
                
                var history1 = new MessageFailureHistory();
                history1.ProcessingAttempts.Add(new MessageFailureHistory.ProcessingAttempt { ExceptionType = typeof(InvalidOperationException).FullName });
                history1.ProcessingAttempts.Add(new MessageFailureHistory.ProcessingAttempt { ExceptionType = typeof(InvalidOperationException).FullName });
                bulkInsert.Store(history1);
            }

            for (int i = 0; i < 100000; i++)
            {
                var history2 = new MessageFailureHistory();
                history2.ProcessingAttempts.Add(new MessageFailureHistory.ProcessingAttempt
                {
                    ExceptionType = typeof(StackOverflowException).FullName
                });
                history2.ProcessingAttempts.Add(new MessageFailureHistory.ProcessingAttempt
                {
                    ExceptionType = typeof(StackOverflowException).FullName
                });
                bulkInsert.Store(history2);
            }

            for (int i = 0; i < 100000; i++)
            {
                var history2 = new MessageFailureHistory();
                history2.ProcessingAttempts.Add(new MessageFailureHistory.ProcessingAttempt
                {
                    ExceptionType = typeof(ArgumentException).FullName
                });
                history2.ProcessingAttempts.Add(new MessageFailureHistory.ProcessingAttempt
                {
                    ExceptionType = typeof(ArgumentException).FullName
                });
                bulkInsert.Store(history2);
            }

            var session = store.OpenSession();

            var stuff = session.Query<MessageFailureHistory_ByExceptionGroup.ReduceResult, MessageFailureHistory_ByExceptionGroup>()
                .Where(r => r.ExceptionGroup == typeof(StackOverflowException).FullName)
                .ToList();
        }
```

danielmarbach:
Here an example how you can then easily load 2000 messages by id:

```
            var stuff = session.Query<MessageFailureHistory_ByExceptionGroup.ReduceResult, MessageFailureHistory_ByExceptionGroup>()
                .Where(r => r.ExceptionGroup == typeof(StackOverflowException).FullName)
                .ToList();

            var all = session.Load<MessageFailureHistory>(stuff.First().HistoryIds.Take(2000));
```

wolfbyte:
Applied the updated index and used a ResultTransformer when I don't need the Id's (as they become a large %age of the payload).  As I need to be able to filter the results on Time/MessageType these fields may end up in the Result too. @danielmarbach I'd appreciate it if you could take a look.

wolfbyte:
As the existing Archive/Retry functions rely on the UniqueMessageId (which is not in the index) I am forced to load them all, iterate through them all and extract the Id needed. Is that OK or should I add this Id to the Index and do it in one pass?

danielmarbach:
Just add it to the index. The index is feature specific anyway, right?

On Thu, Jun 4, 2015 at 7:54 AM, Mike Minutillo <notifications@github.com>
wrote:

> As the existing Archive/Retry functions rely on the UniqueMessageId (which
> is not in the index) I am forced to load them all, iterate through them all
> and extract the Id needed. Is that OK or should I add this Id to the Index
> and do it in one pass?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Particular/ServiceControl/pull/474#issuecomment-108737746>
> .
>



-- 

Daniel Marbach

Solution Architect, Particular Software

Email:  daniel.marbach@particular.net

Web:    www.particular.net
<http://t.signauxneuf.com/e1t/c/5/f18dQhb0S7lC8dDMPbW2n0x6l2B9nMJN7t5XX4RzDjlVS9QB88pTc_dW3Ljg7x56dWyBdgBXs402?t=http%3A%2F%2Fwww.particular.net%2F&si=4896787407044608&pi=33cc6202-1f31-4040-8d6c-e70641f44bba>

danielmarbach:
Should this PR be closed in favour of https://github.com/Particular/ServiceControl/pull/475?

danielmarbach:
@janovesk @weralabaj @wolfbyte see https://github.com/Particular/ServiceControl/pull/474#issuecomment-109889782 would be good to clean up a little bit so that it is easier for us to follow where the effort is taking place

PR #475, opened 2015-06-05
Generic group spike

janovesk:
Let's see if a more generic approach is better.


weralabaj:
Based on the discussion with @janovesk - there's a missing element that allows to identify grouper.
For now we assume there will be only one, but seems like it might be easier to support it now, rather than find workarounds later for supporting multiple groupers at a time and adding this information to existing groups.
Questions: Any concerns? Is it YAGNI? Should the grouper type also be part of the index?

wolfbyte:
Thoughts on Retry All: The current implementation is naive and will likely still fail when the numbers get high (i.e. It will flood the queue with Retry messages which in turn will kick off another set of messages). 

What if we had a GroupRetryPolicy which Issued N retries and then listened for the results. As each one comes back (regardless of success/failure) the DB is re-queried to get the next Unresolved error in the group (1), issue a retry and carry on. When there is no Next Unresolved error then the Saga is closed and the user is informed that the Group Retry is completed.

This will limit the number of Retry messages in the Queue at any one time to N and acts like a throttle on the whole process. It's not going to speed it up but it will keep it consistent and prevent the Retry from taking over all resources within SC. It also allows for the Retry All to be cancelled and the Saga could even publish messages indicating how far through it is for the UI to report to the user the long-running process.

(1) I'd also store the DateTime started in the saga and limit the query to only those items which have no Processing attempts after that time. This prevents infinite retries and stops the saga from getting stuck on a few continually failing messages. 

NOTE: This REQUIRES that the Auditing feature is switched on so that SC can record successes or it will start a Retry All and hang there as it never hears about any successes. 

Thoughts?

weralabaj:
@wolfbyte I think that should be discussed also with regard to desk case @danielmarbach is working on now. Would that be good enough solution for now? 
Also possibly out of scope of this story?

wolfbyte:
Load Testing
Method: Shut down SC, delete raven database and log file. Load 100,000 failures into the error queue (4 different exception types in a rainbow stripe). Start SC and then examine the logs to determine how long it took to process all 100K messages.
Baseline (No Grouping): 24 minutes 22 seconds to process (14.6ms per message). Database is 5.49GB (57.57kB per failure)
Grouping: 37 minutes 54 seconds (22.7ms per message which is +8.1ms). Database is 5.71GB (59.87KB per message which is +2.3)

In processing time that's a 30% increase. For storage the additional space is expected and not all that problematic. 

I'm looking at SC under a profiler today to see where it's taking it's time.

wolfbyte:
Pulling out the extra Load call being made to RavenDB when adding failure groups improves the time for processing 100K messages with Grouping switched on to 26 minutes 30 seconds (15.9ms per message which is +1.3ms per message).

johnsimons:
Superseded by #486

PR #479, opened 2015-06-24
Bulk retries

johnsimons:


src/ServiceControl.Core/Recoverability/Retries/Retryer.cs
```diff
@@ -0,0 +1,78 @@
...
+                        Value = (int) FailedMessageStatus.RetryIssued
+                    }, 
+                    new PatchRequest
+                    {
+                        Type = PatchCommandType.Set, 
+                        Name = "RetryId", 
+                        Value = batchId
+                    }
+                },
+                allowStale: true);
```
johnsimons: If we allow staleness, wouldn't this get us into concurrency issues ?

src/ServiceControl.Core/Recoverability/Retries/Retryer.cs
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        public string StartRetryForIndex(string indexName, string query = null)
+        {
+            var batchId = RetryBatch.MakeId(Guid.NewGuid().ToString());
+
+            // TODO: Issue a message. Turn Batch into a Saga and let it handle it's state
+            using (var session = Store.OpenSession())
+            {
+                session.Store(new RetryBatch
```
johnsimons: do we need this initial save ?

src/ServiceControl.Core/Recoverability/Retries/Retryer.cs
```diff
@@ -0,0 +1,78 @@
...
+                session.SaveChanges();
+            }
+
+            var indexQueryText = String.Format("RetryId:[[NULL_VALUE]] AND Status:{0}", (int) FailedMessageStatus.Unresolved);
+            if (query != null)
+            {
+                indexQueryText = indexQueryText + " AND " + query;
+            }
+
+            var operation = Store.DatabaseCommands.UpdateByIndex(indexName,
```
johnsimons: I thought we were splitting the batch in sets of 50 ?

src/ServiceControl.Core/Recoverability/Retries/Retryer.cs
```diff
@@ -0,0 +1,78 @@
...
+                },
+                allowStale: true);
+
+            operation.WaitForCompletionAsync().ContinueWith(result =>
+            {
+                // TODO: Publish a message. Probably make the RetryBatch a saga and have it take care of it's status. 
+                // That way if a shutdown happens we can issue this status change on a timer 
+                using (var session = Store.OpenSession())
+                {
+                    var retryBatch = session.Load<RetryBatch>(batchId);
```
johnsimons: No need to load again

wolfbyte:
I updated the Batch marking and now it's quite fast (comparatively).

We load 1000 docs from the query at a time and fire 1000 patch requests to Raven to create new FailureRetry documents which has an Id based on the original message Id. These docs contain a reference to the Failure and the Batch. If a record already exists with this ID the patch request does nothing (guaranteeing that only one batch at a time can retry).

As we go we collect the document ids of these FailureRetry documents and when we update the status of the batch we also patch in these ids (So a batch is now an Id, a Status, and a list of Retries).

When we pull the batch out at the other end we can query the batches and eagerly fetch the corresponding Retry documents. As we have their Id no index is needed and this is a single load operation. We then filter out any whose batch Ids don't match (these were grabbed by another batch before got to it). Then we can bulk load the Failure documents by Id. This has been a total of 2 calls to Raven and the second one is a Load.

TODO: Power goes out part way through scenario. At startup (before normal processing starts), if there are any Retry Batches in the Marking status, query the FailureRetry documents with corresponding batchId (will have to wait for non stale results on this one but new ones shouldn't be arriving) and update the status and Ids all together (just as we would when we complete the batch online). Then resume normal processing.

danielmarbach:
* Do I see correctly that the retries are not covered by ATTs?
* The `Retryer` starts a new task which is not tracked. What happens if a user starts a clean shutdown of SC?
* What is the purpose of the `Parallel.For` happening on the `Retryer` task inside `StageRetryByUniqueMessageIds`
* Don't see the value of the `Retries` namespace. I'd just put the infrastructure alongside with the handlers.
* Can the number of orphaned batches be greater then the query limit of the session? Do we need `WaitForNonStaleResultsAsOfNow` in there?

wolfbyte:
> * Do I see correctly that the retries are not covered by ATTs?
Acceptance Test? Not yet. That's a task for this week.
> * The `Retryer` starts a new task which is not tracked. What happens if a user starts a clean shutdown of SC?
Good call. I'll put some cancellation logic in there. It is specifically designed so that if it fails partway through it will get fixed on the next startup so all that needs to happen is that the thread checks for closure on a regular basis.
> * What is the purpose of the `Parallel.For` happening on the `Retryer` task inside `StageRetryByUniqueMessageIds`
We are firing hundreds (if not thousands) of patch requests at Raven. Each one blocks until the patch completes. The Parallel.For speeds up the process by executing the requests in parallel.
> * Don't see the value of the `Retries` namespace. I'd just put the infrastructure alongside with the handlers.
The handlers are in the namespace as well. Not really sure what's being suggested here. All of the retries code is in the Recoverability.Retries namespace. All of the grouping code is in the Recoverability.Groups namespace.
> * Can the number of orphaned batches be greater than the query limit of the session? Do we need `WaitForNonStaleResultsAsOfNow` in there?
This is a recovery task that occurs on startup and only happens if Service Control stops before the background thread above completes. In that case we DO want to wait for non stale results and update their status accordingly as no other task ever will. Given that `Retryer` will need to be notified of shutdown to cleanly cancel background tasks it would actually make sense to move this task there and do it as a startup task.

johnsimons:
Superseded by #486

PR #488, opened 2015-07-27
Configure maximum message body size for audit storage optimisation

petersgiles:
A new setting was added to ServiceControl that allows administrators to control the maximum size of message bodies that ServiceControl stores in its own internal database.
This facilitates systems that were affected by ServiceControl not storing audited message above 100Kb, which in turn affects the debugging experience when using ServiceInsight.

Up until version 1.6 ServiceControl only stores bodies of audit messages that are smaller than 100Kb by default. After version 1.6 Increase this number to store messages with larger bodies. Messages that have a larger message body in bytes than MaxBodySizeToStore are not stored for audit. This is to ensure that the majority of our users enjoy the best level of performance. For users with special analysis needs, edit MaxBodySizeToStore in ServiceControl.exe.config to increase the size of storeable audit messages

### DIFF
--- a/src/ServiceControl.AcceptanceTests/AcceptanceTest.cs
+++ b/src/ServiceControl.AcceptanceTests/AcceptanceTest.cs
@@ -24,6 +24,8 @@
     [TestFixture]
     public abstract class AcceptanceTest
     {
+        protected Dictionary<string, string> AppConfigurationSettings = new Dictionary<string, string>();
+
         public AcceptanceTest()
         {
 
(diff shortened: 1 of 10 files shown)


src/ServiceControl.AcceptanceTests/AcceptanceTest.cs
```diff
@@ -334,11 +345,28 @@ protected byte[] DownloadData(string url)
                 }
 
                 Console.Out.Write(urlToMessageBody);
-
+       
                 return client.DownloadData(urlToMessageBody);
             }
         }
 
+        //protected byte[] DownloadData(string url)
```
johnsimons: we can just delete it

johnsimons:
Looks good :+1:

johnsimons:
Actually @petersgiles u forgot to update https://github.com/Particular/ServiceControl/blob/develop/src/ServiceControl/Hosting/Help.txt with the new setting

src/ServiceControl.AcceptanceTests/AcceptanceTest.cs
petersgiles: ok have done that now

petersgiles:
fixed the help file

petersgiles:
https://github.com/Particular/FeatureDevelopment/issues/296

PR #491, opened 2015-08-04
Return nulls when TimeSent is missing

weralabaj:
Allow for nullable dates (SentTime header) for messages list and failed messages.

### DIFF
--- a/src/ServiceControl.UnitTests/CompositeViews/FailedMessageTest.cs
+++ b/src/ServiceControl.UnitTests/CompositeViews/FailedMessageTest.cs
@@ -13,7 +13,7 @@
     public class FailedMessagesTests 
     {
         [Test]
-        public void Should_allow_errors_with_no_metadata()
+        public void Should_not_provide_defaults_for_missing_metadata()
         {
             using (var session = documentStore.OpenSession())
             {
@@ -54,7 +54,10 @@
                         Console.Out.WriteLine("Checking result");
                         Assert.AreEqual(1, results.Count);
 
-                        Assert.AreEqual(DateTime.MinValue,results.First().TimeSent);
+                        Assert.AreEqual(null, results.First().TimeSent);
+                        Assert.AreEqual(null, results.First().MessageType);
+                        Assert.AreEqual(null, results.First().ReceivingEndpoint);
+                        Assert.AreEqual(null, results.First().SendingEndpoint);
                     }
                 }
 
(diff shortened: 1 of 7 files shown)


weralabaj:
@johnsimons This is internal change, I think we don't need include it in an announcement, given that we'll provide this information in [SP](https://github.com/Particular/ServicePulse/pull/180) and SI (next version) releases.

weralabaj:
That can be released in 1.6. @johnsimons Can I merge it?

johnsimons:
To me this looks good, @wolfbyte thoughts ?

wolfbyte:
It looks good to me. My only concern is that any records in the DB which
already have `DateTime.MinValue` will carry on doing so.

johnsimons:
@weralabaj :+1: to merge

johnsimons:
Actually we need to talk @weralabaj, hold on the merge

weralabaj:
@wolfbyte the issue was casting to DateTime.Min when results were returned, in db they were still nulls. The only problem might be sagas, but that is out of scope.

johnsimons:
@weralabaj it looks like u missed more casts:
- https://github.com/Particular/ServiceControl/blob/84dde512246ded6b67b9a5e00d5bbcc2e9579780/src/ServiceControl/CompositeViews/Messages/MessagesViewIndex.cs#L39
- https://github.com/Particular/ServiceControl/blob/84dde512246ded6b67b9a5e00d5bbcc2e9579780/src/ServiceControl/CompositeViews/Messages/MessagesViewIndex.cs#L62

johnsimons:
@weralabaj can we add an acceptance test as well ?

weralabaj:
@johnsimons ok I'll revert my merge, sorry for that.

weralabaj:
@johnsimons The first omission only changes the order in which null timesent is treated, e.g. before it was casted to DateTime.Min during sorting, so it was first message in asc order, after making it nullable it'll be the last message in that sort.
I'm not sure what the second change does really, thus I don't know how to test it. Is it only impacting sorting as well?

weralabaj:
@johnsimons Daniel said that if we change the types then Raven MIGHT do re-indexing, it depends on how it is implemented. I need to investigate this further, so I reverted the changes from develop. 
Apart from Raven forums is there anybody/anywhere I could consult?

johnsimons:
Daniel and Mauro should be able to help.
Yes, we need to be cautious with reindexing large collections.

On Tue, 4 Aug 2015 at 20:09 Weronika Łabaj <notifications@github.com> wrote:

> @johnsimons <https://github.com/johnsimons> Daniel said that if we change
> the types then Raven MIGHT do re-indexing, it depends on how it is
> implemented. I need to investigate this further, so I reverted the changes
> from develop.
> Apart from Raven forums is there anybody/anywhere I could consult?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Particular/ServiceControl/pull/491#issuecomment-127554311>
> .
>
-- 

*John Simons*
*Engineer, Particular Software*

Email*:*  john.simons@particular.net <karen.fruchtman@particular.net>
Web:    www.particular.net

PR #492, opened 2015-08-04
Revert "Return nulls when TimeSent is missing"

weralabaj:
Reverts Particular/ServiceControl#491

### DIFF
--- a/src/ServiceControl.UnitTests/CompositeViews/FailedMessageTest.cs
+++ b/src/ServiceControl.UnitTests/CompositeViews/FailedMessageTest.cs
@@ -13,7 +13,7 @@
     public class FailedMessagesTests 
     {
         [Test]
-        public void Should_not_provide_defaults_for_missing_metadata()
+        public void Should_allow_errors_with_no_metadata()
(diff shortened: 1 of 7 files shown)


wolfbyte:
Doing a deeper investigation I came across this.

Whenever we construct (at least in v4.6 on NServiceBus) a `TransportMessage` the `TimeSent` header is filled in to `UtcNow`. https://github.com/Particular/NServiceBus/blob/support-4.6/src/NServiceBus.Core/TransportMessage.cs#L32

That means when we ingest an error that does not have that error set [here](https://github.com/Particular/ServiceControl/blob/develop/src/ServiceControl/Operations/ErrorQueueImport.cs#L29), it is set to the time of ingestion.

It should never be `null`.

weralabaj:
Exactly it was done in [Jan 2014](https://github.com/Particular/NServiceBus/commit/301bf4ed20ba7f2ad6132c6608e40cd41b01992d). But this is only applied for outgoing messages, the problem is for messages coming from external systems to NSB.

PR #493, opened 2015-08-11
Don't cast nulls to DateTime.Min.

weralabaj:


weralabaj:
@gbiellem @danielmarbach If there're no objections, I'd like to merge this one.

johnsimons:
Because we changed transformers (the one that concerns me most is `MessagesViewTransformer`) wouldn't this cause a re-index ? It is my understanding these are done server side right, @danielmarbach @mauroservienti please comment ?

mauroservienti:
Luckily no, a transformer is executed server side at query time each time the query is issued by the client.
The flow is:

* A client sends the query
* An index is triggered/used to execute the query
* A transformer is applied, if requested by the client query, to the index results before results are sent back to the client

So, the answer is no, changing a transformer definition won't cause a re-index.

weralabaj:
@johnsimons Reindexing would be caused by the changes you pointed out I missed last time (in indexes definitions). I wrote a test to make sure nobody "fixes" it to make consistent. 
But those changes here won't force reindexing.

johnsimons:
Thank you @mauroservienti for the clarification.
It looks good @weralabaj :+1:

johnsimons:
@weralabaj would the rest api now instead of a datetime.min send a null value ?
And if so, then would it make sense to also have a AT for it ?

weralabaj:
@johnsimons You mean other tests apart from what is already in PR? If you think I should test it on another level/in a different way I'll need some help ;)

johnsimons:
No worries @weralabaj I'm here to help
Yes, I mean that we should also test it at the REST level from a consumer perspective.

Have u looked at our acceptance tests?

weralabaj:
Yep, I thought this is not suitable test because I wasn't sure how to send a message without NSB. When we set up tests we do sth like ".WithEndpoint<Receiver>(b => b.Given(bus => bus.SendLocal(new MyMessage())))" but in that way I'll never get a null date. It only happens for messages from 3rd party systems.

johnsimons:
There are ways to craft a message and send it to SC, see https://github.com/Particular/ServiceControl/blob/11734c9f6b74dd81b22373f162b28559c1afa569/src/ServiceControl.AcceptanceTests/MessageFailures/When_a_retry_for_a_empty_body_message_is_successful.cs#L102-L120

And then u just need to validate the REST response

weralabaj:
@johnsimons Thanks, I'll look into that later today.

PR #496, opened 2015-08-19
Grouping of messages with non-standard stack traces fail

janovesk:
### Symptoms
The current ExceptionTypeAndStackTraceMessageGrouper will not group a failed message that has a non-standard stack trace. Non-standard meaning custom exceptions that override the `StackTrace` property, for example random text. Combined with the fact that we make an assumption that all failed messages WILL be grouped, leads to the following issues:

1) Any failed messages with non-standard stack traces that were present before upgrade to 1.6+ will be regrouped on every ServiceControl restart. They will never be successfully grouped.

2) If there are no groups, but at least one message with non-standard stack trace ServicePulse will incorrectly show "We are still grouping your failed messages, it shouldn't be much longer." on the grouping tab forever.

image

### Who's affected
ServiceControl users that have custom exceptions that override the `StackTrace` property with non-standard text.


janovesk:
Database is on gdrive.

janovesk:
Inital reporter confirmed that they were stripping of the stack trace information in their custom exceptions.

janovesk:
Need a another set of eyes on this before we push it out. @johnsimons ? @wolfbyte ? @andreasohlund ?

johnsimons:
looks good, I'll push it out

PR #518, opened 2015-09-03
[WIP] Clean up message bodies

wolfbyte:
Changes related to https://github.com/Particular/PlatformDevelopment/issues/430

Message body is stored as Raven Attachment IF:
* The message is an error
OR
* The message is an audit AND the message is below the size threshold AND the message is binary

Message body is stored as message metadata IF:
* The message is below the size threshold AND the message is NOT binary

The NotStored flag is only set if the message body is not stored in either place.

Raven Attachment is deleted when corresponding Audit Message is deleted for expiry.

### DIFF
--- a/src/ServiceControl/Operations/BodyStorage/BodyStorageEnricher.cs
+++ b/src/ServiceControl/Operations/BodyStorage/BodyStorageEnricher.cs
@@ -1,6 +1,7 @@
 ﻿namespace ServiceControl.Operations.BodyStorage
 {
     using System.IO;
+    using System.Text;
     using Contracts.Operations;
     using NServiceBus;
     using ServiceBus.Management.Infrastructure.Settings;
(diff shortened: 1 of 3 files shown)


src/ServiceControl/Operations/BodyStorage/BodyStorageEnricher.cs
```diff
@@ -11,51 +12,103 @@ public class BodyStorageEnricher : ImportEnricher
...
+        static bool ShouldStoreInBodyStorage(ImportMessage message, int bodySize, string contentType)
+        {
+            if (message is ImportFailedMessage)
             {
-                StoreBody(message, bodyId, contentType, bodySize);  
+                return true;
             }
-            else
+
+            if (bodySize <= Settings.MaxBodySizeToStore && contentType.Contains("binary"))
```
johnsimons: You don't need `contentType.Contains("binary")`

johnsimons: We want to store it in regardless of type

src/ServiceControl/Operations/BodyStorage/BodyStorageEnricher.cs
```diff
@@ -11,51 +12,103 @@ public class BodyStorageEnricher : ImportEnricher
...
+            }
 
-            var bodyId = message.MessageId;
+            message.Metadata.Add("BodyUrl", bodyUrl);
+
+            return stored;
+        }
 
-            if (message is ImportFailedMessage || bodySize <= Settings.MaxBodySizeToStore)
+        static bool ShouldStoreInBodyStorage(ImportMessage message, int bodySize, string contentType)
```
petersgiles: just a matter of style but couldn't this method be written as a single return statement
something like this

return message is ImportFailedMessage || (bodySize <= Settings.MaxBodySizeToStore && contentType.Contains("binary"));

src/ServiceControl/Operations/BodyStorage/BodyStorageEnricher.cs
```diff
@@ -11,51 +12,103 @@ public class BodyStorageEnricher : ImportEnricher
...
-                message.Metadata.Add("BodyNotStored", true);
+                return true;
             }
 
-            // Issue #296 Body Storage Enricher config
-            if (!contentType.Contains("binary") && bodySize <= Settings.MaxBodySizeToStore)
+            return false;
+        }
+
+        static bool ShouldStoreBodyInMessageMetadata(int bodySize, string contentType)
```
petersgiles: again a single line if that's compatible with code style

src/ServiceControl/Operations/BodyStorage/BodyStorageEnricher.cs
wolfbyte: It can be. I went for clarity :)

petersgiles:
looking at @johnsimons comments i see mine could be refined further
i.e.

return message is ImportFailedMessage || bodySize <= Settings.MaxBodySizeToStore;

src/ServiceControl/Operations/BodyStorage/BodyStorageEnricher.cs
wolfbyte: Why? If it's non-binary we're going to store it in the message metadata. What benefit to we gain from putting it in body storage too?

src/ServiceControl/Operations/BodyStorage/BodyStorageEnricher.cs
petersgiles: yep, its fine either way

wolfbyte:
There was a bug where oversized error messages were not being ingested. This is inconsistent with the prior behavior which ALWAYS ingests errors.

wolfbyte:
Closed in favor of https://github.com/Particular/ServiceControl/pull/518 (which is a renamed version)

Issue #308, opened 2014-04-11
Azure ServiceBus & Storage Queues acceptance tests failing with timeouts


synhershko:
On the build server - probably because the build server (EC2) is remote to the Azure data-center we use.

This shouldn't be that slow, but for some reason it apparently is.

So if the Azure-based transports happen to have a good connection they will pass otherwise the tests will fail with a timeout

/cc @andreasohlund @yvesgoeleven @gbiellem @johnsimons

SeanFeldman:
When SC is used with a transport that requires HTTP, `ServicePointManager.DefaultConnectionLimit` needs to be increased. If not, default 10 connections are used up quickly causing threads that need to make an http web request to block.

PR #527, opened 2015-09-04
Allow to configure DefaultConnectionLimit for HTTP

SeanFeldman:
## Analysis
* SC is using RavenDB client which is using HttpWebRequests. 
* Transports such ASQ is using HttpWebRequests to work as well.
* The default .NET limit is set by `DefaultConnectionLimit` is 10.
* With several satelites and transport concurrency, SC easily exceeds the maximum `DefaultConnectionLimit` when used with ASQ or ASB. 
* This causes blocking behaviour which in conjunction with default transport retry policies can cause blocking calls to stall up to several minutes.

## Solution
Solution to this is to increase `DefaultConnectionLimit` for the entire application. The default set by SC will be 100 with an option to override through configuration file using `HttpDefaultConnectionLimit` setting.

The reasoning behind making it a configurable setting is the following:

Depending on the configured concurreny setting of the transport users will need to be able to increase the `DefaultConnectionLimit`

Has to be set in ServiceControl upon start-up. Any time later the settings will not be applied.

---
https://github.com/Particular/ServiceControl/issues/308

### DIFF
--- a/src/ServiceControl/Bootstrapper.cs
+++ b/src/ServiceControl/Bootstrapper.cs
@@ -3,6 +3,7 @@ namespace Particular.ServiceControl
     using System;
     using System.Diagnostics;
     using System.IO;
+    using System.Net;
     using System.ServiceProcess;
     using Autofac;
     using global::ServiceControl.Infrastructure.SignalR;
@@ -29,6 +30,9 @@ namespace Particular.ServiceControl
 
         public Bootstrapper(ServiceBase host = null, HostArguments hostArguments = null, BusConfiguration configuration = null)
         {
+            // .NET default limit is 10. RavenDB in conjunction with transports that use HTTP exceeds that limit.
+            ServicePointManager.DefaultConnectionLimit = Settings.HttpDefaultConnectionLimit;
+
             Settings.ServiceName = DetermineServiceName(host, hostArguments);
             ConfigureLogging();
             var containerBuilder = new ContainerBuilder();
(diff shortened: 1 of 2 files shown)


SeanFeldman:
:+1 @danielmarbach for helping me with Ants Profiler :)

SeanFeldman:
Fixes #308

SeanFeldman:
.NET 4.5 HttpRuntime sets it to `Int32.MaxValue`, so we're safe with 100

SeanFeldman:
@Particular/servicecontrol please review

johnsimons:
@SeanFeldman are you able to make the PR description more release notes friendly ?

Also, would you consider this a feature ?

SeanFeldman:
@johnsimons no, this is not a feature, this is a bug. The inverse is not applicable here :)

> are you able to make the PR description more release notes friendly ?

Are you referring to the subject or body?

johnsimons:
> Are you referring to the subject or body?

@SeanFeldman probably both

SeanFeldman:
@johnsimons added general description and changed the title.

johnsimons:
@SeanFeldman we should probably add a link to the docs ? So users know how to do it

SeanFeldman:
Very good point @johnsimons. Thanks for catching that.
Done.

johnsimons:
@SeanFeldman I've modified it slightly to comply with the required sections.
thank you

danielmarbach:
Awesome team work over 3 time zones!

PR #558, opened 2015-09-19
When retrying failed messages, previously archived failed messages are also retried

johnsimons:
## Who's affected

- Anyone using ServiceControl version 1.6.0 and up

## Symptoms

When retrying failed messages at `Error Group` page using `Retry a Group` or `Retry All Failed Messages`, incorrect number of messages is retried (failed and archived messages are retried all together).

---

## Steps to Repro

1. Run the ErrorHandling sample to have two messages to fail
2. Archive the 2 messages in ServiceControl via ServicePulse
3. Have another 2 messages error out. 
4. Invoke the RetryAll method in ServiceControl via the RetryAll button in the FailedMessages page in ServicePulse. 
5. Instead of retrying just the 2 recently failed messages, ServiceControl will retries all messages including those that were archived before. 

Related Desk Case: https://nservicebus.desk.com/agent/case/13323

--

**Customer Communication**
## Notice of a Critical Bug in the Retry Failed Messages Functionality in Service Pulse / Service Control version 1.6.x
When we released the new version of 1.6.0 of ServiceControl, we introduced a new feature where error messages can be grouped together based on the exception type in Service Pulse. This allowed the retry all functionality to become simpler. However, we discovered a bug today with our new error grouping functionality. When retrying either all messages or error messages within the same error group in ServicePulse, ServiceControl also incorrectly includes the previously archived failed messages as well. 

We wanted to make sure you knew what the situation was, if you were affected, and how to handle things in the meantime while we work to provide you with a fix.

## How to know if you're affected
If you are using ServiceControl versions 1.6.0 and above.
This bug does not affect if you are using ServiceControl version 1.5.3 and below. 
In Service Pulse, you can verify the version by looking at the bottom of the dashboard in ServicePulse. For example:
version

## What to do if you are affected 
If you are using version 1.6.0 and above, 
   - Do not use the `Retry All Failed Messages` button
   - Do not use the `Retry Group` button within each failure group under the Action column 
   - If you’ve previously used these functions, please review the endpoint’s logs to ensure that those messages did not cause any adverse side-effects. If that’s the case, please contact us via support (http://particular.net/support) and we will work with you to get your situation remedied. 

## The plan going forwards
We are actively working on a hot fix for this, and we will release it as soon as we have this bug fixed. We rely heavily on our automated unit tests before we release new features. And we are very sorry that our acceptance test suite, had missed this use case. We will be adding this test to our suite of existing tests.
We appreciate your understanding and patience while we work through this. 

With thanks for all your support.

The Particular Engineering team

### DIFF
--- a/src/ServiceControl.AcceptanceTests/MessageFailures/When_all_messages_are_retried.cs
+++ b/src/ServiceControl.AcceptanceTests/MessageFailures/When_all_messages_are_retried.cs
@@ -1,0 +1,142 @@
+﻿namespace ServiceBus.Management.AcceptanceTests.Recoverability.Groups
+{
+    using System;
+    using NServiceBus;
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.Config;
+    using NServiceBus.Features;
+    using NUnit.Framework;
+    using ServiceBus.Management.AcceptanceTests.Contexts;
+    using ServiceControl.Infrastructure;
+    using ServiceControl.MessageFailures;
+
+    public class When_all_messages_are_retried : AcceptanceTest
+    {
+        [Test]
+        public void Only_unresolved_issues_should_be_retried()
+        {
+            var context = new MyContext();
+
+            FailedMessage messageToBeRetriedAsPartOfRetryAll = null;
+            FailedMessage messageToBeArchived = null;
+
+            Scenario.Define(context)
+                .WithEndpoint<ManagementEndpoint>(c => c.AppConfig(PathToAppConfig))
+                .WithEndpoint<Receiver>(b => b.Given(bus =>
+                {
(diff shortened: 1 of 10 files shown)


johnsimons:
It seems we already have the `Status` on both indexes used by this code.
The indexes we are using are:
- `FailedMessages_ByGroup`
- `FailedMessageViewIndex`

So because of that we should just need to add an extra where clause to the queries and there won't be any re-index necessary.

@wolfbyte thoughts ?

indualagarsamy:
@johnsimons - did you review the associated commits?

johnsimons:
Yes I did @indualagarsamy, hence the reason for my comment

johnsimons:
Also, I will add an integration test

johnsimons:
See different proposal in #559

PR #559, opened 2015-09-20
Add filter by unresolved to where clause

johnsimons:
Fixes Issue #558


johnsimons:
@colin-higgins @indualagarsamy @wolfbyte 
This is what I propose instead to fix #558.
Thoughts ?

PR #560, opened 2015-09-22
Tool to detect messages which may have been affected for issue 558

wolfbyte:
@Particular/servicecontrol please review. 

This tool is for us to send to our customers to see if #558 has affected them. It works by examining the EventLog history looking for cases where SC 1.6 has submitted a message for Retry. If the event immediately before is not a Message Failed then the customer is (likely) affected. 

It should work even if Audit Ingestion is switched off.

### DIFF
--- a/tools/Issue558Detector/Issue558Detector.Tests/Issue558Detector.Tests.csproj
+++ b/tools/Issue558Detector/Issue558Detector.Tests/Issue558Detector.Tests.csproj
@@ -1,0 +1,89 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <ProjectGuid>{6411725A-8B6F-4CE8-9547-B3827601F70A}</ProjectGuid>
+    <OutputType>Library</OutputType>
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <RootNamespace>Issue558Detector.Tests</RootNamespace>
+    <AssemblyName>Issue558Detector.Tests</AssemblyName>
+    <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
+    <FileAlignment>512</FileAlignment>
+    <ProjectTypeGuids>{3AC096D0-A1C2-E12C-1390-A8335801FDAB};{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}</ProjectTypeGuids>
+    <VisualStudioVersion Condition="'$(VisualStudioVersion)' == ''">10.0</VisualStudioVersion>
+    <VSToolsPath Condition="'$(VSToolsPath)' == ''">$(MSBuildExtensionsPath32)\Microsoft\VisualStudio\v$(VisualStudioVersion)</VSToolsPath>
+    <ReferencePath>$(ProgramFiles)\Common Files\microsoft shared\VSTT\$(VisualStudioVersion)\UITestExtensionPackages</ReferencePath>
+    <IsCodedUITest>False</IsCodedUITest>
+    <TestProjectType>UnitTest</TestProjectType>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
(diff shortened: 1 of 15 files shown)


wolfbyte:
To operate:
1. Stop ServiceControl
2. Start ServiceControl in Maintenance Mode (step 1 from [here](http://docs.particular.net/servicecontrol/db-compaction#step-1-start-servicecontrol-in-the-maintenance-mode))
3. Run the tool on the machine that is running ServiceControl (we'll need to allow customers to specify where their SC database is if they have a non-standard installation)
4. Restart ServiceControl

The results will tell you if you definitely have, maybe have or definitely have not been affected by #558.

weralabaj:
I run a few tests, starting from 1.4.4 then upgrading to 1.6.0, comments:
- I don't see archived messages retried when retrying a group (which is great!).
- The tool detects as suspicious situation when I have a message that was multiple failure in 1.4.4. If that is intended then it would be good to provide some background for that scenario in the email. To me it doesn't seem like a failure though, because the message is a failure and should be retried. Is there some specific edge case when that is an error?

Besides, I think I've found an additional bug. In 1.4.4 if the message is retried we keep the failed message with Unresolved status, so we'll end up with duplicated messages (or am I doing something wrong?!). To retry I used an old version of SP (before grouping). This is how it looks in SI:
capture
As you can see I have two failed messages, but one of them is already successfully retried.
When in 1.6.0 I hit "Retry all", both messages are retried and this is what I end up with:
capture2

weralabaj:
Would it make sense to have some explanations of what possible timelines mean?
E.g. this one is a bit confusing:
capture3
It seems like MessageFailureResolvedByRetry wasn't emitted before 1.6.0 (can that be true?!).
In the instruction we could give an example that it's a timeline for a previously successful retry, that was incorrectly retried again. So even though the final status is "Ok" it might not be actually good.

wolfbyte:
@weralabaj that's a good point. Everything after a NotOK is technically suspect and using the term Ok for it is misleading. Each timeline event is either expected (Ok), unexpected (NotOk) or Unknown. Once an unexpected message is encountered, everything after that is Unknown I think.

wolfbyte:
I added some unit tests covering the likely set of scenarios. Please review and see if I've missed any important ones (or if you disagree with how I think the tool should behave)

wolfbyte:
In discussion with @indualagarsamy we decided that if the first event is a Retry then that is Ok. We are only interested in capturing Retries that occur immediately after Archives, Resolutions, and Retry Issues (where we haven't yet heard if it was successful or a failure).

We also agreed to get rid of the term "Unknown" and use "May be affected" instead. This will be the status for every event after an Affected event or after an unexpected event.

indualagarsamy:
@weralabaj - I agree on the console output redirect to the file. We can easily achieve that by a DOS output `>` redirect to a file. I think the instructions can cover it as part of the instructions on using the tool.

wolfbyte:
First pass of a Readme file for the tool is https://github.com/Particular/ServiceControl/blob/issue-558-detector/tools/Issue558Detector/readme.md

weralabaj:
@wolfbyte The timeline for "A Bulk Retry after an Outstanding Retry" is more ambiguous than we say.
For example I got the same timeline (MessageFailed => MessageSubmittedForRetry => MessageSubmittedForRetry => MessageResolvedByRetry) for the following cases:
- Repeated failure => retried 
- Retry issued, endpoint down => retried
- Successful retry => retried
I generated data with 1.4.4 and then migrated to 1.6.0, where I retried all successfully.
If all retries fail I get the same results, only timeline doesn't contain the last event (i.e. it is MessageFailed => MessageSubmittedForRetry => MessageSubmittedForRetry).

I think because of this it's better if we do it backwards, i.e. "You got this timeline? Here is what it might mean". Now, if we retried again after successful retry then in SI they'll be able to see duplicated message (because of the bug I mentioned above). I don't have an idea how to distinguish between the other two though.

So I see the following cases:
- FailedMessageArchived > MessagesSubmittedForRetry => incorrectly retried archived message
- MessageResolvedByRetry > MessagesSubmittedForRetry => incorrectly retried successfully retried message
- MessageResolvedByRetry > MessageResolvedByRetry + duplicated message in SI  => incorrectly retried successfully retried message
- others => not clear what happened: either outstanding retry or repeated failure; I'm looking if we can get something more on that either from SI or directly from Raven

weralabaj:
Actually I have an idea - if at the moment of upgrade the message was in the failed state (i.e. either failed or multiple failure) it'll have a failure group. All other retried messages won't have it.
See here (before retry all and after retry all):
5scenarios-retryall-raven-before
5scenarios-retryall-raven-after

indualagarsamy:
This code has been moved to a new repository. Please see:
https://github.com/Particular/ServiceControl.Issue558Detector

PR #561, opened 2015-09-23
Also expire SagaHistories

SimonCropp:
* Use RavenLastModified date for expiring documents
 * Add a scoped raven last modified helper for testing
 * remove MessageStatus.ResolvedSuccessfully : MessageStatus.Successful
usage since those are the only statuses we query on it is redundant data
to store in the index
 * dont store MessageId in the index, we only use __document_id so
MessageId is redundant
 * simplify and speed up tests but only applying the expiry bundle to the
test that require it

### DIFF
--- a/src/ServiceControl/Infrastructure/RavenDB/Expiration/ExpiredDocumentsCleaner.cs
+++ b/src/ServiceControl/Infrastructure/RavenDB/Expiration/ExpiredDocumentsCleaner.cs
@@ -3,7 +3,6 @@
 
     using System;
     using System.Collections.Generic;
-    using System.ComponentModel.Composition;
     using System.Diagnostics;
     using System.Globalization;
     using System.Linq;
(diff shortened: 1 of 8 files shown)


SimonCropp:
@wolfbyte can u review this?

src/ServiceControl/Infrastructure/RavenDB/Expiration/ExpiredDocumentsCleanerBundle.cs
```diff
@@ -0,0 +1,44 @@
...
+            var deleteFrequencyInSeconds = Settings.ExpirationProcessTimerInSeconds;
+
+            if (deleteFrequencyInSeconds == 0)
+            {
+                return;
+            }
+
+            logger.Info("Expired Documents every {0} seconds", deleteFrequencyInSeconds);
+            logger.Info("Deletion Batch Size: {0}", deletionBatchSize);
+            logger.Info("Retention Period: {0}", Settings.HoursToKeepMessagesBeforeExpiring);
```
wolfbyte: {0} hours

src/ServiceControl/Infrastructure/RavenDB/Expiration/ExpiredDocumentsCleanerBundle.cs
```diff
@@ -0,0 +1,44 @@
...
+
+            timer = new PeriodicExecutor(executor => ExpiredDocumentsCleaner.RunCleanup(Settings.ExpirationProcessBatchSize, database), TimeSpan.FromSeconds(deleteFrequencyInSeconds));
```
wolfbyte: Why pass in the ExpirationProcessBatchSize here but then read the Setting for Hours to Keep inside the ExpiredDocumentsCleaner? I'd consider reading all of the settings here and passing everything in to the RunCleanup method.

SimonCropp: yep will fix

src/ServiceControl/Infrastructure/RavenDB/Expiration/ExpiredDocumentsCleaner.cs
```diff
@@ -14,88 +13,70 @@
...
+                    "LastModified",
                     "MessageMetadata"
                 },
                 SortedFields = new[]
                 {
-                    new SortedField("ProcessedAt")
+                    new SortedField("LastModified")
                     {
-                        Field = "ProcessedAt",
+                        Field = "LastModified",
```
wolfbyte: The Constructor Parameter sets this to `LastModified` as well.

src/ServiceControl/Infrastructure/RavenDB/Expiration/ExpiredDocumentsCleaner.cs
```diff
@@ -14,88 +13,70 @@
...
-            var currentExpiryThresholdTime = currentTime.AddHours(-Settings.HoursToKeepMessagesBeforeExpiring);
-            logger.Debug("Trying to find expired documents to delete (with threshold {0})", currentExpiryThresholdTime.ToString(Default.DateTimeFormatsToWrite, CultureInfo.InvariantCulture));
-            const string queryString = "Status:3 OR Status:4";
             var query = new IndexQuery
             {
                 Start = 0,
                 PageSize = deletionBatchSize,
-                Cutoff = currentTime,
-                Query = queryString,
+                Cutoff = DateTime.UtcNow,
```
wolfbyte: This is only necessary if we are checking the `IsStale` flag

src/ServiceControl/Infrastructure/RavenDB/Expiration/ExpiredDocumentsCleaner.cs
```diff
@@ -111,71 +92,66 @@ void Delete(PeriodicExecutor executor)
...
+                catch (OperationCanceledException)
                 {
-                    logger.Debug("No expired documents found");
+                    //Ignore
                 }
-                else
+
+                logger.Debug("Batching deletion of {0} documents.", items.Count);
+
+                docsToExpire += items.Count;
```
wolfbyte: `docsToExpire` will always be 0 at this point. You could move the declaration here and just set it to `items.Count`

src/ServiceControl/Infrastructure/RavenDB/Expiration/ExpiredDocumentsCleaner.cs
```diff
@@ -111,71 +92,66 @@ void Delete(PeriodicExecutor executor)
...
+                database.TransactionalStorage.Batch(accessor =>
                 {
-                    logger.Debug("Deleted {0} out of {1} expired documents batch - Execution time:{2}ms", deletionCount, docsToExpire, stopwatch.ElapsedMilliseconds);
-                }
+                    foreach (var attach in attachments)
+                    {
+                        accessor.Attachments.DeleteAttachment(attach, null);
+                    }
+                });
+                deletionCount = results.Count(x => x.Deleted == true);
```
wolfbyte: What is this used for?

src/ServiceControl/Infrastructure/RavenDB/Expiration/ExpiredDocumentsCleaner.cs
```diff
@@ -111,71 +92,66 @@ void Delete(PeriodicExecutor executor)
...
+                items.Clear();
```
wolfbyte: Is this necessary? The collection is about to go out of scope anyway

wolfbyte:
Review done.

It's unlikely but if we ever add fields for `MessageMetadata.BodyNotStored` or `MessageMetadata.MessageId` to `SagaHistory` then we might try and delete attachments that we didn't intend. I don't know that there's a nice way around that without adding a flag to the index to indicate the type of message and given that it's an unlikely scenario if it is worth that added complexity.

src/ServiceControl/Infrastructure/RavenDB/Expiration/ExpiredDocumentsCleaner.cs
SimonCropp: docsToExpire is scoped outside the database.Query so it increments on each run.

Am i missing something or is your suggestion invalid?

src/ServiceControl/Infrastructure/RavenDB/Expiration/ExpiredDocumentsCleaner.cs
SimonCropp: correct. will fix

src/ServiceControl/Infrastructure/RavenDB/Expiration/ExpiredDocumentsCleaner.cs
SimonCropp: it is used for logging below

src/ServiceControl/Infrastructure/RavenDB/Expiration/ExpiredDocumentsCleaner.cs
wolfbyte: Isn't this update outside of the query scope as well? There is no iteration around this update.

src/ServiceControl/Infrastructure/RavenDB/Expiration/ExpiredDocumentsCleaner.cs
wolfbyte: Ah. Missed that. I thought it had been replaced by the `docsToExpire` but now I see that both are used.

src/ServiceControl/Infrastructure/RavenDB/Expiration/ExpiredDocumentsCleaner.cs
SimonCropp: yeah the scope is weird there. i will move the logging inside the cancellation token using and that will make it cleaner

SimonCropp:
@mikeminutillo think i addressed all your suggestions.

Issue #501, opened 2015-08-26
Retry Batching consumes too many resources and crashes SC


cjoergensen:
## Who's affected

Users of ServiceControl 1.6.0 and up with large numbers of failed messages to retry.

## Symptoms

When retrying a large number of messages, ServiceControl maxes the HDD utilization and then memory until a Raven error occurs. This is caused by an unbounded parallel execution during retry batch creation. 

--
This fix pushes all retry requests into a background queue and processes them one at a time.
If this issue persists then the embedded RavenDB may require some tweaking. Add the following line to the ServiceControl config file:

`<add key="Raven/Esent/MaxVerPages" value="1024" />`

The value is the size of the "version store" used for open transactions in megabytes. The default value is 512MB. You can increase this to 2048 or 4096 if necessary.

cjoergensen:
Im sorry, the proper version is 1.6.1

wolfbyte:
@cjoergensen - it sounds like with that many messages that the embedded Raven instance may need some tweaking. 

Can you provide some information about the machine that ServiceControl is running on? 

Which operating system?
32 or 64 bit?
RAM?
What else is running on the machine?

If you are able to send me a backup of your database I can try a few tweaks to see if I can help you resolve it. mike.minutillo-at-particular.net

mikeminutillo:
Original Issue Details

Hi

We are retrying a lot of failed message (200.000+) and ServiceControl keeps on stopping.

When investigation the Windows Event log I found this entry just before the service stops:
```
------------------------------------------------------------------------
Raven (13472) 1-szk08-E:\RavenDB\ServiceControl.1.4.2\Data: The version store for this instance (0) has reached its maximum size of 511Mb. It is likely that a long-running transaction is preventing cleanup of the version store and causing it to build up in size. Updates will be rejected until the long-running transaction has been completely committed or rolled back. 
Possible long-running transaction: 
	SessionId: 0x000000387F184640 
	Session-context: 0x00000000 
	Session-context ThreadId: 0x0000000000004488 
	Cleanup: 1 
	Session-trace: 
57573@20:04:35
------------------------------------------------------------------------
```
The ServiceControl log file shows these entries:

NServiceBus.Unicast.Transport.TransportReceiver|Failed to process message
Microsoft.Isam.Esent.Interop.EsentVersionStoreOutOfMemoryException: Version store out of memory (cleanup already attempted)

mikeminutillo:
[...]

PR #562, opened 2015-09-30
Move batch creation to a single background thread

mikeminutillo:
Related to https://github.com/Particular/ServiceControl/issues/501

Concurrent Bulk Retries consumes too many resources on the machine. Specifically the creation of so many retry message documents floods the IO and causes a backlog in the Esent Version Store. This change will push them all into a single background thread. 

Note that while this helps, it is not sufficient to resolve the issue in all cases. If the user has large messages then they need to set [Raven/Esent/MaxVerPages](http://ravendb.net/docs/article-page/2.5/csharp/server/administration/configuration) to something higher than 512MB which is the default. If we're doing batches of 1000 messages at 2MB each then a setting of at least 2048 should be used.

### DIFF
--- a/src/ServiceControl/Recoverability/Retries/RetriesGateway.cs
+++ b/src/ServiceControl/Recoverability/Retries/RetriesGateway.cs
@@ -1,10 +1,10 @@
 namespace ServiceControl.Recoverability
 {
     using System;
+    using System.Collections.Concurrent;
     using System.Collections.Generic;
     using System.Linq;
     using System.Linq.Expressions;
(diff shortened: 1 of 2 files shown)


danielmarbach:
You could use a `ConcurrentBag` or a `ConcurrentQueue` instead of the `BlockingCollection`. Rest is fine

johnsimons:
@mikeminutillo u happy to ship this one out as part of 1.6.4?

johnsimons:
Part of #566

PR #565, opened 2015-10-16
Retries get stuck on a batch when any of the messages fails to be sent

johnsimons:
## Who's affected
Anyone using ServiceControl version 1.6.0 and up.

## Symptoms
The issue is that if the Send back to endpoint queue fails (it could be because the queue no longer exists because the user has decommission it, or Azure is having a temporary glitch) that causes the batch to never finish and therefore any subsquent batches do not get processed.

--

This fix does a few things:
- It published an error in the appending log  
image

- It replaces the fault manager of the `ReturnToSenderDequeuer` so that failures never go into the error queue of service control since there is no point;
- It updates the status of the  message back to unresolved, so it can be retried again;
- It still uses the FLR to try to send a few times;
- It always counts messages even if they fail all FLR;


johnsimons:
@mikeminutillo @indualagarsamy @andreasohlund please review

johnsimons:
This fix does not introduce a new status (RetryFailed) for the failed message as before discussed.
The reason is that it would require updates to SI.

src/ServiceControl/Recoverability/Retries/Infrastructure/ReturnToSenderDequeuer.cs
```diff
@@ -1,25 +1,120 @@
...
+                        {
+                            new PatchRequest
+                            {
+                                Type = PatchCommandType.Set,
+                                Name = "Status",
+                                Value = (int) FailedMessageStatus.Unresolved,
+                            }
+                        });
+
+                    store.DatabaseCommands.Delete(FailedMessageRetry.MakeDocumentId(messageUniqueId), null);
```
mikeminutillo: Do these two database commands respect transaction scope? Ideally we want both or neither of them to work.

src/ServiceControl/Recoverability/Retries/Infrastructure/ReturnToSenderDequeuer.cs
```diff
@@ -1,25 +1,120 @@
...
+                        {
+                            m.Reason = "Failed to retrieve reason!";
+                        }
+
+                    });
+                }
+                catch (Exception ex)
+                {
+                    // If something goes wrong here we just ignore, not the end of the world!
+                    Log.Error("A failure occurred when trying to handle a retry failure.", ex);
```
mikeminutillo: It doesn't block future retries from happening but it may leave an individual error message in an invalid state.

johnsimons: So what do you propose ?

src/ServiceControl/Recoverability/Retries/Infrastructure/ReturnToSenderDequeuer.cs
johnsimons: yeah I thought about that but not sure how to achieve that without loading the documents ?

mikeminutillo: Maybe in this instance that's OK. This is a failure scenario.

src/ServiceControl/Recoverability/Retries/Infrastructure/ReturnToSenderDequeuer.cs
mikeminutillo: Not sure. If we can get a transaction around the database bits that may be OK.

tools/ResetConcurrencyFlag/README.md
```diff
@@ -0,0 +1,38 @@
...
+```
+
+This should give the following output:
+```txt
+Resetting message...
+Done
+```
+
+**Note:**
+The tool assumes that RavenDB instance is exposed at the default url, i.e. `http://localhost:33333/storage`. However, if you customized the ServiceControl configuration to alter the URL, you can pass the full URL in on the command line after the id, like this:
```
petersgiles: +The tool assumes that RavenDB instance is exposed at the default url, i.e. `http://localhost:33333/storage`. If you have altered the URL in the ServiceControl configuration, you can pass the full URL to the command like this:

tools/ResetConcurrencyFlag/README.md
```diff
@@ -0,0 +1,38 @@
...
+```
+
+This should give the following output:
+```txt
+Resetting message...
+Done
+```
+
+**Note:**
+The tool assumes that RavenDB instance is exposed at the default url, i.e. http://localhost:33333/storage. If you have altered the URL in the ServiceControl configuration, you can pass the full URL to the executable like this:
```
mikeminutillo: Can we link this to the doco explaining how to expose raven (if it isn't already) http://docs.particular.net/servicecontrol/use-ravendb-studio

johnsimons: Or do we just tell them to run in maint mode ?

tools/ResetConcurrencyFlag/README.md
```diff
@@ -0,0 +1,38 @@
+# Retries concurrency flag reset
+
+An [issue in ServiceControl 1.6.0 to 1.6.3](https://github.com/Particular/ServiceControl/pull/565) was identified that causes the retries feature of ServiceControl to stall and not process any subsequent retries. This issue is fixed in [ServiceControl 1.6.4](https://github.com/Particular/ServiceControl/releases/tag/1.6.4), however in order to reset stalled retries a manual step is needs to be performed by the user, this is where this tool will help.
+
```
mikeminutillo: When a message is retried in ServiceControl 1.6.x, a lock is placed on the message to ensure it cannot be retried again. If we did not do this then a message could be send to the receiving endpoint more than once and processed multiple times. If the message fails again the lock is removed allowing subsequent retries.

A situation can occur where ServiceControl creates the lock on the failed message but the message cannot be sent to the original destination. When this happens the message being retried gets placed on the ServiceControl error queue but the lock remains.

This tool allows you to remove the lock from a failed message allowing it to be retried again.

johnsimons:
Will resubmit against 1.7 instread

PR #567, opened 2015-10-26
revert expiry index

SimonCropp:
so as to not create startup load


SimonCropp:
@mikeminutillo @johnsimons based on both your input i have reverted back the message expiry index so as to not cause load

johnsimons:
So are we doing a standalone expiry for saga histories ?

SimonCropp:
@johnsimons since u said u didnt want the load caused by changing the expiry message index

johnsimons:
No worries @SimonCropp 
So is the new expiration for saga histories to go out as par of 1.7 release ?

johnsimons:
:+1: to merge

mikeminutillo:
Looks good to me as well :+1:

PR #568, opened 2015-10-26
Corrected sorting of failed messages inside an error group 

petersgiles:
## Who's affected
Anyone using ServiceControl version 1.6.0 and up.

## Symptoms
When browsing failed message inside an Error Group the messages were not sorted properly.

Here is an example of messages sorted by Time Sent:
untitled

Relates to https://github.com/Particular/ServicePulse/issues/195


petersgiles:
@johnsimons @mikeminutillo please review

johnsimons:
This looks good to me @petersgiles but the test can be simplified a lot more if u use the TM directly, see https://github.com/Particular/ServiceControl/blob/develop/src/ServiceControl.AcceptanceTests/MessageFailures/When_a_retry_fails_to_be_sent.cs#L150-L160, so use this code to send raw messages to the error queue and fill in the appropriate headers with the data u need, if u need help ping me

Issue #590, opened 2015-11-12
SignalR component of ServiceControl not publishing events


particularbot:
### Who's affected

All users that are running ServiceControl 1.7 or higher and ServicePulse

### Symptoms

Neither the indicatiors or the evenlog updates automatically as new failed messages comes in or custom checks failes. Pretty sure the same applies to heartbeats as well but haven't tested.
Navigating to other pages or hitting browser refresh updates the UI properly.
Looking in the developer tools the consoles says SignalR connected but no network traffic is registered.

-- 

### Versions

ServiceControl Version 1.7.4
ServicePulse Version 1.4.0

### To repro

* Make a custom check fail
* Notice that the indicator stays green and no entry is shown in the event log

particularbot:
**Comment by [johnsimons](https://github.com/johnsimons)** _https://github.com/Particular/ServicePulse/issues/219#issuecomment-155735315_

----

@andreasohlund have you validated that the event is raised by SC ? In other words have you disprove that is not a SC bug ?

johnsimons:
Fixed in https://github.com/Particular/ServiceControl/commit/c91a1577a2e7579fa330976d8b885528443fa9ec

PR #591, opened 2015-11-12
Event broadcast not working

mikeminutillo:
Connects to #590

SignalR was resolving a separate connection when broadcasting messages. This meant that successfully connected clients were not receiving messages.

### DIFF
--- a/src/ServiceControl.AcceptanceTests/AcceptanceTest.cs
+++ b/src/ServiceControl.AcceptanceTests/AcceptanceTest.cs
@@ -44,7 +44,7 @@
 
         protected Dictionary<string, string> AppConfigurationSettings = new Dictionary<string, string>();
         string pathToAppConfig;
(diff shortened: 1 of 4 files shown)


johnsimons:
@mikeminutillo this looks good to me :+1:

johnsimons:
@mikeminutillo have added test coverage, can u review ?

@mikeminutillo @andreasohlund 
I have only tested one single signalrr event, do you think that is enough or do we test them all ?

mikeminutillo:
@johnsimons The test looks good for testing the infrastructure. We should raise a separate issue to test the "public api" of the SignalR integration point. That is, every event that Pulse (the only consumer) relies on, we should check to ensure that it gets raised and that it contains the expected data in the expected format.

johnsimons:
Merged as part of hotfix-1.7.5

Issue #596, opened 2015-11-18
Write an Acceptance Test for Retries mixing with Replies


mikeminutillo:
Connects to #583 

When a handler fails to process a message, that message is replied and the successful retry causes a Reply, the response should go back to the originating endpoint, not ServiceControl.

This was not working with the RabbitMQ transport until #583 

### Outline of test

A sends MSG to B
B fails to process MSG and forwards to SC
SC retries MSG
B successfully processes MSG and sends REPLY to A
If A gets REPLY then success

PR #597, opened 2015-11-19
Acceptance Test for Retries with Replies

mikeminutillo:
Connects to #596

Works on MSMQ, Not on RabbitMQ


mikeminutillo:
@SzymonPobiega @johnsimons 

The test seems to work for all of the transports except for Rabbit.

* `OriginatingEndpoint` sends `OriginalMessage` to `ReceivingEndpoint`
* `ReceivingEndpoint` fails processing on `OriginalMessage` and forwards it to the error queue
* `ManagementEndpoint` ingests `OriginalMessage` and it appears in the API
* A retry request is sent to `ManagementEndpoint` for `OriginalMessage`
* `OriginalMessage` is routed back to `ReceivingEndpoint`
* In the `OriginalMessage` handler within `ReceivingEndpoint`, `bus.Reply(new ReplyMessage())` is called
* `ReplyMessage` is routed to `OriginatingEndpoint` which handles it and marks the test as completed.
  * This step last step is where it fails. Instead of being routed to `OriginatingEndpoint` the `ReplyMessage` is being routed to `ReceivingEndpoint`

johnsimons:
Related to https://github.com/Particular/NServiceBus.RabbitMQ/issues/105

johnsimons:
Merged in https://github.com/Particular/ServiceControl/commit/21852e2c5e65c09b60b3066527bb682d5d6b3ed2

PR #606, opened 2015-12-09
Handling failures when publishing to external integrations

johnsimons:
If the publishing of an event to external integrations fails continuosly, SC is shutdown.

This PR catches exceptions as part of the publishing, logs them in the log file and also adds them to the append log that SC keeps and is displayed in SP.

### DIFF
--- a/src/ServiceControl.AcceptanceTests/CustomChecks/Custom_check_should_only_trigger_events_on_transition.cs
+++ b/src/ServiceControl.AcceptanceTests/CustomChecks/Custom_check_should_only_trigger_events_on_transition.cs
@@ -32,7 +32,6 @@
 
         public class MyContext : ScenarioContext
         {
-            public string CustomCheckId { get; set; }
         }
 
         public class EndpointWithFailingCustomCheck : EndpointConfigurationBuilder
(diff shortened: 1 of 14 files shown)


johnsimons:
@mikeminutillo @DavidBoike thoughts ?

mikeminutillo:
Does the exception thrown indicate which endpoint was the problem?

johnsimons:
@mikeminutillo nope, we don't know, we don't have that info, hopefully the exception message will say

DavidBoike:
Not a huge fan. Unlikely they'll subscribe to the new event. There's a lot of samples and doco out there (not all controlled by us) about how to subscribe to SC events and this isn't mentioned. So SC is not falling over, which is good, but *much* higher likelihood of complete and total message loss, because a Publish to 0 subscribers leaves absolutely no trace.

But at least SC doesn't fall over. :smile:

johnsimons:
@DavidBoike in this proposal I am logging the exception into two different locations.
- the log fie, which we tend to tell the users to monitor, but being an app maybe not the right place;
- logging to the append only log in SC, which gets displayed in SP for the user to take action. We also use this same technique for other areas of SC where things fail and we need to notify the user.

Issue #612, opened 2015-12-18
Messages with TTBR set cannot be forwarded to Audit / Error Log


mikeminutillo:
## Who's affected
This bug affects you if you are using Audit/Error Forwarding in ServiceControl and are setting Time To Be Received (TTBR) on messages in the Audit/Error Queue. 

### Audits

You may be explicitly setting the TTBR flag by setting the [OverrideTimeToBeReceived](http://docs.particular.net/nservicebus/operations/auditing#audit-configuration-options-overridetimetobereceived) setting on the audit configuration. If you are using ServiceControl, this setting should not be necessary. 

It may be copied from the incoming message. This can happen if:

- The receiving endpoint (that forwards to the audit queue) does not contain the fix for https://github.com/Particular/NServiceBus/issues/2886 OR
- The receiving endpoint (that forwards to the audit queue) is not transactional

### Errors

Messages in the error queue will only have a TTBR set if it has been copied from the input queue of the processing endpoint. This can only happen if the processing endpoint does not have the fix for https://github.com/Particular/NServiceBus/issues/2886
If the endpoint was not transactional then FLR would not forward the message to the error queue.

## Symptoms
When ServiceControl tries to ingest these messages, they are not fowarded to the audit/error log queues and the logs contain the message `Sending messages with a custom TimeToBeReceived is not supported on transactional MSMQ.`

If enough of these messages are encountered in a short timeframe, ServiceControl will shut down.

### Affected Versions

ServiceControl
1.8.0
1.8.1
1.8.2

### Plan of attack

Modify the ServiceControl ingestion to strip out TTBR headers before forwarding to Audit / Error Log queues AFTER the message has been persisted to the database.

PR #613, opened 2015-12-21
Strip TimeToBeReceived before forwarding to audit log

mikeminutillo:
Relates to #612 

@Particular/servicecontrol please review

### DIFF
--- a/src/ServiceControl.AcceptanceTests/Audit/When_an_audit_message_has_ttbr_set.cs
+++ b/src/ServiceControl.AcceptanceTests/Audit/When_an_audit_message_has_ttbr_set.cs
@@ -77,7 +77,8 @@
                 {
                     context.TtbrStipped = true;
                     // MSMQ gives incoming messages a magic value so we can't compare against MaxValue
-                    // context.TtbrStipped = transportMessage.TimeToBeReceived == TimeSpan.MaxValue;
+                    // Ensure that the TTBR given is greater than the 10:00:00 configured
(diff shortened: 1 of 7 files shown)


gbiellem:
@mikeminutillo  in your POA in #612  you say

> strip out TTBR headers before forwarding to Audit/Error queue

But the code only deals with audit forwarding unless I'm missing something.  Is there a reason to skip error forwarding?

mikeminutillo:
@gbiellem Missed that one. I forgot they were handled separately.

mikeminutillo:
@gbiellem We're not actually able to get into that state.

In order for the error/audit forwarding to work with TTBR you need to turn transactions off. If you do that then the error forwarding doesn't occur because it is currently handled by the FLR and that is disabled for non-transactional endpoints. 

This raises another interesting issue with https://github.com/Particular/NServiceBus/issues/2886

If you send a message with TTBR to a transactional endpoint, and it fails, forwarding to the error queue fails.

mikeminutillo:
@gbiellem I was wrong. It IS still possible, it's just hard to test. If the processing endpoint does not contain the fix https://github.com/Particular/NServiceBus/issues/2886, then it can and will forward error messages with TTBR to ServiceControl. I wrote a test targeting this scenario. Can you have another look for me?

gbiellem:
@mikeminutillo The AT for this is failing against SQLServer - http://builds.particular.net/viewLog.html?buildTypeId=ServiceControl_32RunAcceptanceTestsSqlServer&buildId=138854

gbiellem:
:+1:  - Merged into hotfix 1.8.3

Issue #614, opened 2015-12-21
Validation of Service account fails on non-english versions of Windows


johnfake:
### Symptom

When adding an instance of the ServiceControl service using a builtin account as the service account returns  "Not a valid account" on non-English version of Windows.

### Who's affected

Potentially anyone adding a ServiceControl instance via the Management UI or PowerShell module on an non-English version of Windows,

---

Hi,

cannot add an instance of service controll (msmq) on windows 10 pro.
![image](https://cloud.githubusercontent.com/assets/7387234/11940579/de73f414-a829-11e5-9ea2-faeb5ed667b2.png)

![image](https://cloud.githubusercontent.com/assets/7387234/11940591/ef96a5f2-a829-11e5-9202-fcc0c3799d0a.png)

   bei ServiceControlInstaller.Engine.Accounts.UserAccount.ParseAccountName(String accountName) in C:\BuildAgent\work\7189a56f9f44affc\src\ServiceControlInstaller.Engine\Accounts\UserAccount.cs:Zeile 113.
   bei ServiceControlInstaller.Engine.Validation.ServiceAccountValidation.Validate(IServiceAccount instance) in C:\BuildAgent\work\7189a56f9f44affc\src\ServiceControlInstaller.Engine\Validation\ServiceAccountValidation.cs:Zeile 14.
   bei ServiceControlInstaller.Engine.Instances.ServiceControlInstanceMetadata.Validate() in C:\BuildAgent\work\7189a56f9f44affc\src\ServiceControlInstaller.Engine\Instances\ServiceControlInstanceMetadata.cs:Zeile 213.
   bei ServiceControl.Config.Framework.Modules.Installer.Add(ServiceControlInstanceMetadata details, IProgress`1 progress) in C:\BuildAgent\work\7189a56f9f44affc\src\ServiceControl.Config\Framework\Modules\InstallerModule.cs:Zeile 40.
   bei ServiceControl.Config.UI.InstanceAdd.InstanceAddAttachment.<>c__DisplayClass6_1.<Add>b__0() in C:\BuildAgent\work\7189a56f9f44affc\src\ServiceControl.Config\UI\InstanceAdd\InstanceAddAttachment.cs:Zeile 80.
   bei System.Threading.Tasks.Task`1.InnerInvoke()
   bei System.Threading.Tasks.Task.Execute()
--- Ende der Stapelüberwachung vom vorhergehenden Ort, an dem die Ausnahme ausgelöst wurde ---
[...]

PR #615, opened 2015-12-22
Fix Mgmt UI localization issue

gbiellem:
Relates to  #614

Mgmt Util fails on non-english OSes because it does not cater for the localized equivalent of LocalSystem and NTAuthority when doing account validation.

### DIFF
--- a/src/ServiceControlInstaller.Engine/Accounts/UserAccount.cs
+++ b/src/ServiceControlInstaller.Engine/Accounts/UserAccount.cs
@@ -27,7 +27,12 @@
             }
         }
 
-        
+        // ReSharper disable once InconsistentNaming
+        static string LocalizedNTAuthority()
+        {
+            var localSystem = new SecurityIdentifier(WellKnownSidType.LocalSystemSid, null);
+            return localSystem.Translate(typeof(NTAccount)).Value.Split('\\')[0];
+        }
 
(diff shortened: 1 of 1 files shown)


gbiellem:
@mikeminutillo quick review.  Simple fix for #614.

src/ServiceControlInstaller.Engine/Accounts/UserAccount.cs
```diff
@@ -27,7 +27,12 @@ public string QualifiedName
             }
         }
 
-        
+        // ReSharper disable once InconsistentNaming
+        static string LocalizedNTAuthority()
+        {
+            var localSystem = new SecurityIdentifier(WellKnownSidType.LocalSystemSid, null);
+            return localSystem.Translate(typeof(NTAccount)).Value.Split('\\')[0];
```
mikeminutillo: Are we OK to assume that the current user has the ControlPrincipal security permission? https://msdn.microsoft.com/en-us/library/system.security.principal.securityidentifier.translate(v=vs.110).aspx#Anchor_2

gbiellem:
@mikeminutillo  - Yes we force elevation

mikeminutillo:
The :+1:

gbiellem:
Merged into hotfix-1.8.3 branch

Issue #623, opened 2016-01-25
V6 upgrade


danielmarbach:
@mikeminutillo @johnsimons @gbiellem 

I would like to take a stab at this with your help. I'd be OK to start earlier in the morning to create more overlapp with you so that we can do a few pair programming sessions or whatever is required to get feedback from you. Any ideas, suggestions to pull this off?

gbiellem:
@danielmarbach sounds good

I don't think we should tag it as part of V6 though as I don't think it's a blocker for the V6 release.
If this isn't the case it would be good if you can detail why you think it is.

johnsimons:
:+1: on what @gbiellem said

danielmarbach:
I've expressed my opinion about this several times. Let's jump on a call

gbiellem:
> Let's jump on a call

@danielmarbach  sure. It'll have to be next week. I've quit for the day,  

> I've expressed my opinion about this several times.

Great but i was hoping to capture your reasoning here so we don't have to go through this again when we are queried about why it's part of V6 and why we are spending time on it.  I've not been in any conversation about it so I can't speak for you

danielmarbach:
I even stated it in GH issue (which I can't find now...) that in my opinion
v6 is not done before we dogfooded it with SC.

mikeminutillo:
I think we could spike this to ensure that the seam is correct. There is a non-trivial amount of effort involved in doing releasing a version of ServiceControl on NSB v6:

- Updating to v6 requires us to update the RavenDB Persistence as well which will require an update to Raven DB version 3. This will require a data conversion.
- ServiceControl (rightly or wrongly) subscribes to it's own events. We did this as a hack in v4 and v5. We do not know if it will continue to work in v6.
- The ingestion process is also based on hacking the v5 pipeline. This infrastructure is different in v6 and will need to be tested. 
- The transport seams (which the retry mechanism relies upon) have changed in v6.
- The update to v5 was specifically to support a feature in the SQL Transport (multiple databases) which is being dropped from v6. As we only maintain a single version of ServiceControl, moving forward would also mean dropping this feature.

If the decision has been made to hold the release of NSB v6 until all affected downstream repos have been updated then ServiceControl is one of those repos and needs to be updated. If we don't do that, then we run the risk that the seam is not right and updating ServiceControl is difficult/impossible. This could lead to a situation where v6 releases new features that are not compatible with ServiceControl and we can't quickly respond. That is undesirable.

[...]

PR #627, opened 2016-02-01
[WIP] Spike Upgrade to v6

danielmarbach:
Connects to #623

## Summary

An upgrade to Version 6 is possible when those issues have been taken into account

* [Redesign FailedAuditImport and FailedErrorImport](https://github.com/Particular/ServiceControl/issues/6330)

### DIFF
--- a/src/ServiceControl.AcceptanceTests/Contexts/ManagementEndpointSetup.cs
+++ b/src/ServiceControl.AcceptanceTests/Contexts/ManagementEndpointSetup.cs
@@ -40,7 +40,7 @@
 
             configurationBuilderCustomization(builder);
 
-            var startableBus = new Bootstrapper(configuration: builder).Bus;
+            var startableBus = new Bootstrapper(configuration: builder).EndpointInstance;
 
             LogManager.Configuration = SetupLogging(endpointConfiguration);
 
--- a/src/ServiceControl.IntegrationDemo/App.config
+++ b/src/ServiceControl.IntegrationDemo/App.config
@@ -12,4 +12,13 @@
     </MessageEndpointMappings>
   </UnicastBusConfig>
   <AuditConfig QueueName="audit" />
+<startup><supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5.2" /></startup>
+  <runtime>
+    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
+      <dependentAssembly>
+        <assemblyIdentity name="NServiceBus.Core" publicKeyToken="9fc386479f8a226c" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-6.0.0.0" newVersion="6.0.0.0" />
+      </dependentAssembly>
+    </assemblyBinding>
+  </runtime>
 </configuration>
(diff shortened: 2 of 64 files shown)


danielmarbach:
Currently the biggest blockers are

* Direct usage of `TransportMessage` on the storage model, i.e. `FailedErrorImport`, so we need to copy paste the old TransportMessage definition including all related data types to ServiceControl if we don't want to patch the indexes.
* Usage of IAdvancedSatellites. We can workround that by using the satellite pipeline but we have to decide how we do the error handlling and whether we do need FLR for the satellite pipelines. We used a few custom failure managers to store failed messages inside Raven. This needs to be done directly in the behaviors registered in the corresponding satellite pipeline.

danielmarbach:
@Particular/servicecontrol-maintainers @mat-mcloughlin continues working tomorrow on the upgrade spike and pings you in his morning. Feel free to take the PR over in the meantime

mat-mcloughlin:
Just working on the `FeatureStartupTask`s. I'm about halfway through them.

johnsimons:
So I was looking at converting the existing `IAdvancedSatellite` to use the new `Feature.AddSatellitePipeline` but it seems that api only caters for subqueues.
So I was thinkning of instead we host multi endpoints in the same process, so in total we would start (maybe more later on, eg heartbeats, custom checks, ...) with `particular.servicecontrol`, `audit` and 'error', thoughts ?

danielmarbach:
Multi hosting is the better and more future proof approach.

johnsimons:
@mat-mcloughlin  yes just split the address on https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/Contracts/Operations/ImportFailedMessage.cs#L14

mat-mcloughlin:
Thanks @johnsimons :+1:

johnsimons:
@mat-mcloughlin @danielmarbach I can't see us proceeding with this till https://github.com/Particular/NServiceBus/issues/3389 is done, what are your thoughts ?

johnsimons:
Guys, after spending a day looking at this, I'm coming to the conclusion that upgrading SC to NSB v6 is plausible.
Most challenges have already been dealt with and the ones that have not (eg the satellites/multi host) we are aware of what are the limitations (see comment above).

So I think we can put this one to rest and tick it as done.

danielmarbach:
@johnsimons @mikeminutillo @mat-mcloughlin 

I raised 

https://github.com/Particular/ServiceControl/issues/633

as an autocome of this investigation. @johnsimons could you summarize the satellite pipeline hacks and the micro endpoint discussions in an issue as well and link it here?

I think we can close this now.

johnsimons:
@danielmarbach the summary is that in order for us to complete the upgrade to v6, we would need for core to expose either:

1. a `AddSatellitePipeline` that accepts a queue name (it currently only accepts a qualifier, mostly used for subqueues);
1. a way to create an endpoint that does not contain all the baggage eg no behaviours, no features, ..., pretty much the same as a satelite but has an endpoint;

Obviously the preferred option would be 2 because of the advantage that we would get from being able to host these new endpoints separatelly.
An issue has been raised for 2, see https://github.com/Particular/NServiceBus/issues/3389

PR #635, opened 2016-02-03
DO NOT MERGE - Added additional logging to RetryAllInGroup handler

WilliamBZA:
This is a hotfix for LightSail to investigate why their RetryAllInGroup messages fail silently.


WilliamBZA:
Connects to https://github.com/Particular/ServiceControl/issues/636

johnsimons:
@WilliamBZA I have asked the user to send us their database and that way we can debug it locally, hence no need for this PR

Issue #639, opened 2016-02-09
Under high load heartbeats can have concurrency issues


mikeminutillo:
When ServiceControl receives a heartbeat message it loads a document from it's persistence, updates it and saves it back again. Under high load two incoming heartbeats from the same endpoint can be worked on concurrently causing a concurrency exception. Even when one of these gets a concurrency exception and rolls back, if it is earlier than the one successfully processed then ServiceControl will discard the older one when it tries to re-process it. This further clogs up the already congested ServiceControl input queue.

I would propose the following solution:

1. Keep all heartbeat information in an in-memory concurrent dictionary keyed on endpoint id.
2. Serve all heartbeat queries from this dictionary.
3. Persist the dictionary to Raven on a predictable schedule (say every 2 minutes) rather than on every heartbeat.
4. When a user toggles heartbeats on / off in Pulse this will need to persist immediately. The same when endpoints transition from Dead to Heartbeating and back again.

Pros: 
- Saving heartbeats will (almost) never cause concurrency exceptions. Toggling heartbeats on/off and state transitions may still cause this.

Cons: 
- As the data is in memory, we are restricted to a single instance. ServiceControl is already limited to a single instance by the failed message recovery mechanism. 
- If ServiceControl shuts down unexpectedly between modifying and saving the data we may lose the latest heartbeats. If the endpoint is still heartbeating when ServiceControl comes back online we will update anyway. If the endpoint is not heartbeating when ServiceControl comes back online then the difference of a few heartbeats as to when it went down is unlikely to be very relevant. 

NOTES:
- We already DO keep an in-memory cache of heartbeats and we use it to determine when an endpoint has stopped heartbeating. 
- When we detect an endpoint that's not heartbeating we send a message to ourselves. This message can be enqueued behind a heartbeat message from that endpoint so that when we get to processing it, the heartbeat has already been restored. If heartbeats are meant to be low latency we should probably handle this synchronously as soon as it is discovered if possible.

mikeminutillo:
@MarcinHoppe this is the issue that @andreasohlund was talking about.

@mauroservienti @sfarmar @ramonsmits as you've had recent dealings with customers about SC I'd appreciate your input here 

Heartbeats have another issue as well which is that they can bank up in the ServiceControl queue if the transport does not handle TTBR well. https://github.com/Particular/NServiceBus.SqlServer/issues/158

mauroservienti:
[...]

PR #640, opened 2016-02-10
Using an inmemory heartbeats cache to avoid so much contention with saving docs into raven

johnsimons:
This addresses the issue raised in #639

### DIFF
--- a/src/ServiceControl/HeartbeatMonitoring/HeartbeatsMonitor.cs
+++ b/src/ServiceControl/HeartbeatMonitoring/HeartbeatsMonitor.cs
@@ -10,6 +10,8 @@ namespace ServiceControl.HeartbeatMonitoring
     using NServiceBus.Features;
     using Raven.Client;
     using ServiceBus.Management.Infrastructure.Settings;
+    using ServiceControl.Contracts.HeartbeatMonitoring;
+    using ServiceControl.Contracts.Operations;
     using ServiceControl.HeartbeatMonitoring.InternalMessages;
     using ServiceControl.Infrastructure;
 
@@ -20,6 +22,7 @@ namespace ServiceControl.HeartbeatMonitoring
             EnableByDefault();
             RegisterStartupTask<StatusInitialiser>();
             RegisterStartupTask<HeartbeatMonitor>();
+            RegisterStartupTask<HeartbeatsWriter>();
         }
 
         protected override void Setup(FeatureConfigurationContext context)
@@ -33,7 +36,6 @@ namespace ServiceControl.HeartbeatMonitoring
             public IBus Bus { get; set; }
 
             public HeartbeatStatusProvider HeartbeatStatusProvider { get; set; }
-
 
             protected override void OnStart()
             {
@@ -56,7 +58,7 @@ namespace ServiceControl.HeartbeatMonitoring
 
                 try
                 {
-                    timer.Change((int)TimeSpan.FromSeconds(5).TotalMilliseconds, -1);
+                    timer.Change(TimeSpan.FromSeconds(5), Timeout.InfiniteTimeSpan);
                 }
                 catch (ObjectDisposedException)
                 {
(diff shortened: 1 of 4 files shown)


johnsimons:
@mikeminutillo is this what u have in mind ?

johnsimons:
Also I wonder if it would be better to use the patch api to do it all ?

johnsimons:
After reviewing the changes, we cannot use the patch api because based on certain state changes events get published

johnsimons:
@mikeminutillo this one is done, can u review

johnsimons:
So I have hardcoded 2 minutes for the sync timer, this means it can take up to 2 minutes for an endpoint to be seen as dead or alive, Is this ok ?

src/ServiceControl/HeartbeatMonitoring/HeartbeatsMonitor.cs
```diff
@@ -147,5 +149,114 @@ void Initialise()
...
+                            };
+
+                            Bus.Publish(new HeartbeatingEndpointDetected
+                            {
+                                Endpoint = heartbeat.EndpointDetails,
+                                DetectedAt = heartbeat.LastReportAt,
+                            });
+                        }
+
+                        session.Store(heartbeat);
```
mikeminutillo: Should only need to do this for new heartbeats. If this is an existing heartbeat, it is already tracked as a part of the session.

src/ServiceControl/HeartbeatMonitoring/HeartbeatsMonitor.cs
```diff
@@ -147,5 +149,114 @@ void Initialise()
...
+
+                        HeartbeatStatusProvider.RegisterHeartbeatingEndpoint(heartbeat.EndpointDetails, heartbeat.LastReportAt);
```
mikeminutillo: Shouldn't need to do this. We just pulled the info from the heartbeat stats provider

src/ServiceControl/HeartbeatMonitoring/HeartbeatsMonitor.cs
```diff
@@ -147,5 +149,114 @@ void Initialise()
...
+                }
+            }
+
+            void Refresh(object _)
+            {
+                Sync();
+
+                try
+                {
+                    timer.Change(TimeSpan.FromMinutes(2), Timeout.InfiniteTimeSpan);
```
mikeminutillo: Does this mean when an endpoint starts beating again that we won't notify anyone for 2 minutes?

src/ServiceControl/HeartbeatMonitoring/HeartbeatsMonitor.cs
```diff
@@ -56,7 +58,7 @@ void Refresh(object _)
 
                 try
                 {
-                    timer.Change((int)TimeSpan.FromSeconds(5).TotalMilliseconds, -1);
+                    timer.Change(TimeSpan.FromSeconds(5), Timeout.InfiniteTimeSpan);
```
mikeminutillo: Should we use [`PeriodicExecutor`](https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/Infrastructure/PeriodicExecutor.cs) in these startup tasks?

src/ServiceControl/HeartbeatMonitoring/HeartbeatsMonitor.cs
mikeminutillo: Same for new heartbeats

src/ServiceControl/HeartbeatMonitoring/HeartbeatsMonitor.cs
```diff
@@ -147,5 +149,114 @@ void Initialise()
...
+
+                    foreach (var memHeartbeat in inmemoryHeartbeats)
+                    {
+                        Heartbeat heartbeat;
+                        if (storedHeartbeats.TryGetValue(memHeartbeat.Key, out heartbeat))
+                        {
+                            if (heartbeat.Disabled)
+                            {
+                                continue;
+                            }
```
mikeminutillo: This means that we don't update the details of heartbeats when we aren't monitoring them. Were we before? Should we?

mikeminutillo:
> So I have hardcoded 2 minutes for the sync timer, this means it can take up to 2 minutes for an endpoint to be seen as dead or alive, Is this ok ?

I think it is not OK. One of the discussions we had around heartbeats is that they need to be low latency.
@Particular/servicecontrol-maintainers please chime in.

src/ServiceControl/HeartbeatMonitoring/HeartbeatsMonitor.cs
johnsimons: That is how we were doing it before, not sure if it makes sense updating the last heartbeat timestamp.
If we look at the whole reason to stop monitoring an endpoint, most times has to do with decomissioning an endpoint and hence that endpoint will not be sending heartbeats anyway, so I don't think we need to worry.

src/ServiceControl/HeartbeatMonitoring/HeartbeatsMonitor.cs
johnsimons: PeriodicExecuter uses Tasks + Task.Delay to in essence replicate what a Timer gives us for free.
In other words, I am not sure we should be using it at all

johnsimons:
@mikeminutillo did u endup working on this one ?

mikeminutillo:
@johnsimons yeah but I've stashed it at the moment. Let me pull it out and see where I got up to

PR #652, opened 2016-02-15
Improving shutdown

johnsimons:
Added capability to stop unbounded queries at shutdown.


johnsimons:
There is only 2 more locations where we do unbounded queuries, these are:

1. [RetryDocumentManager](https://github.com/Particular/ServiceControl/blob/ef197b0573fa9bf6e65fc27ba7c7b1cfcefc7764/src/ServiceControl/Recoverability/Retries/RetryDocumentManager.cs#L132-L138)
1. [RetriesGateway](https://github.com/Particular/ServiceControl/blob/ef197b0573fa9bf6e65fc27ba7c7b1cfcefc7764/src/ServiceControl/Recoverability/Retries/RetriesGateway.cs#L72-L88)

@mikeminutillo any ideas what are the implications of cancelling these mid flight ?

mikeminutillo:
Retry Document Manager - At worst there are 1,000 messages in any given batch. If we shut this process down partway through then we'll just pick it up again when we restart. Note that we're querying for Ids and only once we have the all do we issue an update statement.

Retries Gateway - This one generates batches of 1000 messages from the larger group. If this is interrupted then the retry request is effectively lost. The end user experience is "Request Retry, Shutdown SC, Nothing Happens". I don't think this is unreasonable.

johnsimons:
@mikeminutillo I think we can abort RetriesGateway, see https://github.com/Particular/ServiceControl/commit/5c00cae22cd95bab2038eb6245bd6e9cc98241df

johnsimons:

src/ServiceControl/Recoverability/Retries/RetryDocumentManager.cs
```diff
@@ -112,7 +120,7 @@ internal bool AdoptOrphanedBatches()
 
                 AdoptBatches(session, orphanedBatchIds);
 
-                var moreToDo = stats.IsStale || orphanedBatchIds.Any();
+                var moreToDo = abort || stats.IsStale || orphanedBatchIds.Any();
```
mikeminutillo: This doesn't read well. There's more to do if we're aborting. 

The bool value returned is "all done". Perhaps we should add a second bool so we can have:

```CSharp
var allDone = abort || !moreToDo;
return allDone;
```

mikeminutillo:
I take it we need a separate notifier because we want to stop certain activities before disposing of the Raven store?

johnsimons:
> I take it we need a separate notifier because we want to stop certain activities before disposing of the Raven store?

yes that is the scenario

mauroservienti:
What happens if we try to shutdown ServiceControl while in `AdoptBatches`?

mauroservienti:
Ignore my previous comment

> What happens if we try to shutdown ServiceControl while in `AdoptBatches`?

I just saw the commit related to that.

johnsimons:
@Particular/servicecontrol-maintainers this one is ready

Issue #651, opened 2016-02-12
Logging detail not sufficient for proper debugging


WilliamBZA:
When dealing with a customer support case it became very clear that while we log a lot of information in a lot of places, we are still missing log messages in plenty other places making it difficult to identify where in the code the process is breaking.

The solution might be to add more log messages in places that have none, or perhaps look at instrumentation.

johnsimons:
@WilliamBZA do you have a list where we need to provide more diagnostic logging ?

mikeminutillo:
I can add some based on my experience with this case. We do not log retry requests. We don't log anything to do with a retry until we start actually staging it. There is a whole batching process before this step that we have no visibility on. Obviously from #650 we know that there is no logging in that area either. I will start a PR to add at least these logging pieces.

ramonsmits:
Another one, log progress and size in unbounded query processing.

gbiellem:
@Particular/servicecontrol-maintainers  How do we address this and get it to  point were it could be closed?  At the moment the issue seems very open ended.

mauroservienti:
This is not related to ServiceControl only, this spans the entire platform from Core to transports, I think we should move this to Platform Dev.

For example one of the issues I faced at a customer while dealing with SC and the SQL Transport is that the transport might fail to create queues, it logs the error but misses the queue that is causing the failure. It's not that SC misses something, something is not logged correctly at the transport level in this case.

johnsimons:
@gbiellem how about we close this one and raise specific smaller issues to deal with, eg better logging to diagnose retries, archival, ...

mikeminutillo:
I'm all for that @johnsimons. The alternative is to do an audit of the entire code-base looking for opportunities to provide better context via logging.

johnsimons:
So I am closing this one, and every time we find an area that needs better logging for diagnostics purpose, we will create specific issues to deal with it

Issue #400, opened 2014-10-24
Add Authentication and Authorization Support to the ServiceControl REST API


fafachd:
#### Is your feature related to a problem? Please describe.

There is currently no way to secure access to ServicePulse / ServiceControl API which exposes the traffic to everyone on the network.

#### Describe the requested feature

Include support for authentication, authorization, and TLS/SSL. Support JWT bearer tokens, OpenID Connect, and OAuth 2.0, with ServicePulse authentication configuration automatically fed from the ServiceControl API. Authentication defaults to disabled for backward compatibility.

Add direct HTTPS hosting via Kestrel, HTTPS redirect, HSTS configuration, and improved security controls for CORS and reverse proxy scenarios. Include comprehensive documentation covering all security configuration options and best practices at https://docs.particular.net.

### Additional Context

These features cannot be used with ServiceInsight. Please note that its core features have been moved to ServicePulse, which is now the recommended tool for monitoring and debugging distributed systems.
New development on ServiceInsight has stopped.

<details>
<summary>Original feature request</summary>
See the following for background:

- https://github.com/Particular/ServicePulse/issues/3 

There needs to be a way to do authentication and authorization on the SC REST API so IT OPS can run ServiceInsight on their desktops without exposing the REST API to everyone on the network.
The alternative is to have IT Ops remote into the server, but in larger orgs this access is very difficult to obtain.
</details>

fafachd:
Any update on this, @johnsimons or @andreasohlund?

johnsimons:
No updates @fafachd, at the moment we are focusing on performance improvements.

@Particular/servicecontrol-maintainers should we raise this in plat dev ?

gbiellem:
@johnsimons  I'd say yes as it concerns multiple products.

johnsimons:
This issue has been raised internally in plat dev.

@fafachd thanks a lot for starting this discussion, the way we manage this kind of suggestions, is by raising them internally in a private repo, where we prioritise it and manage it from now on.
So with that in mind, I'll close this suggestion for now, and once we are ready to act on it we will reopen it. This does not mean we will not be working on it.

fafachd:
Thanks, @johnsimons.

andreasnilsen:
It's been 8 years, will there be any release of authentication/authorization features in ServiceControl in the foreseeable future?

[...]

PR #653, opened 2016-02-15
Improved Logging

mikeminutillo:
Connects to https://github.com/Particular/ServiceControl/issues/651

Looking for ways to improve the overall logging of ServiceControl to improve visibility when debugging issues in client environments. 

- Now logging every API call

### DIFF
--- a/src/ServiceControl/Infrastructure/OWIN/ApiLogger.cs
+++ b/src/ServiceControl/Infrastructure/OWIN/ApiLogger.cs
@@ -1,0 +1,22 @@
+﻿namespace ServiceControl.Infrastructure.OWIN
+{
+    using System.Threading.Tasks;
+    using Microsoft.Owin;
+    using NServiceBus.Logging;
+
+    class LogApiCalls : OwinMiddleware
+    {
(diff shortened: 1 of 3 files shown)


johnsimons:
looks good @mikeminutillo

johnsimons:
@mikeminutillo after looking at this again, I am doubting the value that this really provides.

Should we reevaluate the use case for this one ?

@Particular/servicecontrol-maintainers thougths ?

mikeminutillo:
@johnsimons it means that we can turn on API Logging and see when users actually perform actions. You could probably achieve something similar by running Fiddler on the machine that's running Pulse.

johnsimons:
> it means that we can turn on API Logging and see when users actually perform actions

I get that, but is there a requirement for it ?
Majority of the requests are related to authentication, see #400 

Adding this info to the log file seems wrong to me, we keep on adding more and more stuff to the log file, but is this the right location for such info ?

mikeminutillo:
Will rebase onto https://github.com/Particular/ServiceControl/pull/667 Release 1.10

mikeminutillo:
Recreated here https://github.com/Particular/ServiceControl/pull/668

Issue #650, opened 2016-02-12
ArchiveAllInGroupHandler is doing an unbounded stream update in one operation


ramonsmits:
ArchiveAllInGroupHandler is doing an unbounded stream update in one operation.

If we need to archive 100.000+ messages then this operation will take a very long time at each document is processed sequentially and patched separately.

The logic should to doing this by moving through all items either one by one or in (small) batches.

Possible fix:

- Query should be bounded by using `.Take(100)`
- If there are results, then at the end do a SendLocal of the same message

On a side node, we should add logging about its behavior.

ramonsmits:
Possible we can use a completely different implementation by doing a `UpdateByIndex` containing a patch request. Making roundtrip do RavenDB instead of the number roundtrips based on the number of matching documents.

- http://stackoverflow.com/questions/8595636/conditional-patch-requests-in-ravendb

ramonsmits:
By running a patch via the admin interface interface on an index we could update a large set of matching documents within a second.

ramonsmits:
```
            var store = Session.Advanced.DocumentStore;
            store.DatabaseCommands.UpdateByIndex("FailedMessages/ByGroup",
                new IndexQuery { Query = string.Format(CultureInfo.InvariantCulture, "FailureGroupId: {0} AND Status:{1}", message.GroupId, (int)FailedMessageStatus.Unresolved) },
                new[] {
                            new PatchRequest
                                {
                                    Type = PatchCommandType.Set,
                                    Name = "Status",
                                    Value = (int)FailedMessageStatus.Archived
                                }
                        },
                false);
```

The published `FailedMessageGroupArchived` has GroupName and MessageIds properties which we cannot set based on this patch logic so it requires additional queries but are we really going to publish an event with 100.000+ ID's?

johnsimons:
Being addressed as part of #652

ramonsmits:
@johnsimons #652 talks about breaking out of a stream operation. This issue is talking about replacing the stream operation by making use of UpdateByIndex.

Its a completely different issue.

johnsimons:
@ramonsmits here is the PR that is fixing this https://github.com/Particular/ServiceControl/pull/656

johnsimons:
Fixed in https://github.com/Particular/ServiceControl/issues/656

PR #656, opened 2016-02-15
Archiving by index

johnsimons:
This relates to #650

### DIFF
--- a/src/ServiceControl/EventLog/Definitions/FailedMessageGroupArchivedDefinition.cs
+++ b/src/ServiceControl/EventLog/Definitions/FailedMessageGroupArchivedDefinition.cs
@@ -6,9 +6,8 @@
     {
         public FailedMessageGroupArchivedDefinition()
         {
-            Description(m => string.Format("Archived {0} messages from group: {1}", m.MessageIds.Length, m.GroupName));
+            Description(m => string.Format("Archived {0} messages from group: {1}", m.MessagesCount, m.GroupName));
             RelatesToGroup(m => m.GroupId);
-            RelatesToMessages(m => m.MessageIds);
         }
     }
 }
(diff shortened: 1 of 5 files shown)


johnsimons:
@Particular/servicecontrol-maintainers so now the question is what do we report on ?

```c#
Bus.Publish<FailedMessageGroupArchived>(m =>
            {
                m.GroupId = message.GroupId;
                m.GroupName = message.GroupId;
                m.MessageIds = new []{ result.ToString() };
            });
```

As @ramonsmits and @mauroservienti point out, there is no point inreporting on a massive list of ids, so do we get rid of it ?

mikeminutillo:
We probably don't want to have 100s of thousands of ids in there but it's not that difficult to get.

```CSharp
var result = operation.WaitForCompletion();
	
var updatedDocuments = result.JsonDeserialization<DocumentPatchResult[]>();
var updatedIds = updatedDocuments.Select(x => x.Document).ToArray();

public class DocumentPatchResult
{
	public string Document { get; set; }
	public Results Result { get; set; }
	public string[] Debug { get; set; }

	public class Results
	{
		public string PatchResult { get; set; }
		public string Document { get; set; }
	}
}
```

mikeminutillo:
Now capturing the group name and related failed message Ids correctly. I suspect we want to get rid of the message Ids but we need to discuss the impact that will have first.

mikeminutillo:
Added an explicit cutoff date. Without this cutoff the call to `UpdateByIndex()` will fail under high load. The other alternative is that we can explicitly allow Stale Indexes.

mauroservienti:
Which the reason we report the `message ids` other than the `group id`? We are archiving a group in the end.

> Without this cutoff the call to UpdateByIndex() will fail under high load. The other alternative is that we can explicitly allow Stale Indexes.

Interesting, do you remember the error?
cause when we tried that the load was super heavy and it worked as expected.

johnsimons:
I have modified the UpdateByIndex query to allowStale = true, otherwise we could block for a long time.

mikeminutillo:
> Interesting, do you remember the error?
> cause when we tried that the load was super heavy and it worked as expected.

If the index is stale the update should fail and throw an exception. If the index is keeping up then it shouldn't be a problem.

johnsimons:
I have removed the list of guids and instead we just display the total count.
@Particular/servicecontrol-maintainers anything else that needs to be done here before merging ?

Issue #654, opened 2016-02-15
ServiceControl RavenDB logging in insufficient to pass on to Hibernating Rhinos support


gbiellem:
The embedded RavenDB logging is hard coded to "Error" and the log entries are mixed in with the ServiceControl logs.  The logging level was set to remove the Raven "noise" from the ServiceControl logs however this presents a problem when we want detailed RavenDB logs from a customer to   hand over to Hibernating Rhinos support for analysis.

To address this I suggest we add another logging target specifically for the Raven logs and add a config entry to enable verbose raven logging.

gbiellem:
@johnsimons  as discussed

johnsimons:
> To address this I suggest we add another logging target specifically for the Raven logs and add a config entry to enable verbose raven logging.

:+1:

gbiellem:
Implemented in #657

Issue #641, opened 2016-02-10
Log file should roll over for large log files


ramonsmits:
Log file is not rolling over. Log file becomes too large to open with 'standard' windows tools.

![capture-logfile-size](https://cloud.githubusercontent.com/assets/152998/12946698/ad6ef9a0-cff6-11e5-92fc-0c97d4cae6d5.PNG)

johnsimons:
We have increased the default logging level to Warn in v1.9, so large logging files should be a rare occasion.

@Particular/servicecontrol-maintainers 
So do we still want to do this ?

gbiellem:
Warn may now be the default but when diagnosing an issue the log level will be bumped to info or debug leading to  the same unwieldy file sizes. So I'd be +1 for still doing this

johnsimons:
@gbiellem want to do this one ?

gbiellem:
Sure thing, I'll take it

ramonsmits:
@gbiellem Here is the NLog configuration that I used:

```xml
	<nlog>
		<variable name="LogDir" value="C:\Windows\System32\config\systemprofile\AppData\Local\Particular\ServiceControl\logs"/>
		<variable name="LogDay" value="${date:format=dd}"/>
		<targets>
			<target name="logfile" type="AsyncWrapper" queueLimit="5000" overflowAction="Discard">
				<target name="LogTarget1" type="File" fileName="${LogDir}/sc-${LogDay}.log" encoding="utf-8" maxArchiveFiles="100" archiveNumbering="Sequence" archiveAboveSize="250485760" archiveFileName="${LogDir}/sc-{#######}.log" layout="${longdate}|${threadid}|${level}|${logger}|${message}${onexception:${newline}${exception:format=tostring}}"/>
			</target>
			<target name="logfileRaven" type="AsyncWrapper" queueLimit="5000" overflowAction="Discard">
				<target name="LogTarget1" type="File" fileName="${LogDir}/raven-${LogDay}.log" encoding="utf-8" maxArchiveFiles="100" archiveNumbering="Sequence" archiveAboveSize="250485760" archiveFileName="${LogDir}/raven-{#######}.log" layout="${longdate}|${threadid}|${level}|${logger}|${message}${onexception:${newline}${exception:format=tostring}}"/>
			</target>
		</targets>
		<rules>
			<logger name="Raven.*" minlevel="Trace" writeTo="logfileRaven" final="true"/>
[...]

PR #657, opened 2016-02-15
Change raven logging

gbiellem:
Relates to #654 and #641

- Raven Logging separated out to a different file
- New Setting RavenDBLogLevel to control the level of data written to the separate log.
- Console Target is no longer used when running as a service
- Logs now roll at 30MB
- Log file names and  archived logs names are date based for easier sorting

### DIFF
--- a/src/ServiceControl/Bootstrapper.cs
+++ b/src/ServiceControl/Bootstrapper.cs
@@ -123,6 +123,7 @@ namespace Particular.ServiceControl
 
         static void ConfigureLogging()
         {
+            const long MegaByte = 1073741824;
             if (NLog.LogManager.Configuration != null)
             {
                 return;
@@ -139,6 +140,7 @@ namespace Particular.ServiceControl
                 ArchiveNumbering = ArchiveNumberingMode.Rolling,
                 Layout = simpleLayout,
                 MaxArchiveFiles = 14,
+                ArchiveAboveSize =  30 * MegaByte
             };
 
             var consoleTarget = new ColoredConsoleTarget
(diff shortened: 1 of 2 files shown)


johnsimons:
Looks good @gbiellem.

What are your thoughts on naming log files with dates, to make it simpler for us to sort them when we ask for the logs ?

gbiellem:
> What are your thoughts on naming log files with dates, to make it simpler for us to sort them when we ask for the logs?

That's what I normally do so :+1: from me.

johnsimons:
@gbiellem do it as part of this PR ?

gbiellem:
Sure.  I'll add it

gbiellem:
@johnsimons  One downside If we change the format of the log file name the existing logs won't be archived and aged out.  How do suggest we handle this? Leave them in place or purge them?

johnsimons:
So the aging does not work for date rolling ?

gbiellem:
> So the aging does not work for date rolling ? 

No it works for any new files going forward   But because the pattern has changed the old files from before the pattern change are never touched by the archiving - effectively they are orphaned

gbiellem:
> No it works for any new files going forward But because the pattern has changed the old files from before the pattern change are never touched by the archiving - effectively they are orphaned

Discussed with @johnsimons  we'll leave the orphaned files in place and document it.

mikeminutillo:
@gbiellem does there need to be a doco change for this? Specifically: 

> New Setting RavenDBLogLevel to control the level of data written to the separate log.

gbiellem:
Definitely. I've already started it but not pushed it yet.

mikeminutillo:
Ah cool. Just lining up ducks on the release.

Issue #655, opened 2016-02-15
Modify the Managment Util to add recommended Raven Settings to app.config on upgrade/install


gbiellem:
From @johnsimons ...

Hibernating Rhinos have suggested we should add the following configuration entries into the ServiceControl app.config to boost the performance of the embedded RavenDB. 

```xml
<configuration>
  <appSettings>
    <add key="Raven/Esent/MaxVerPages" value="2048" />
  </appSettings>
  <runtime>
    <gcServer enabled="true"/>
  </runtime>
</configuration>
```
The management util will be updated to set these options if they do not exist when adding or installing an instance of ServiceControl.  If the config already has a any configuration related to these settings then the existing values will be preserved.

gbiellem:
Implemented in #658

PR #658, opened 2016-02-15
Add default app.config settings on update and install

gbiellem:
Relates to #655

- Add RavenDB settings recommend by Hibernating Rhinos to ServiceControl app config.
- This occurs on install and update of an SC instance
- No existing setting is overwritten


johnsimons:
looks good

PR #663, opened 2016-02-17
Improvements to data being served to ServicePulse

johnsimons:
So far the improvements are:
- `/eventlogitems` REST call was unbounded, and ServicePulse really only needs the top 10 documents.  
  I have added paging to the REST api call
- Stop populating related_to field for `EventLogItemAdded`, this data at the moment does not make sense for an end user and can cause a lot of data to sent out to ServicePulse

### DIFF
--- a/src/ServiceControl/EventLog/EventLogApiModule.cs
+++ b/src/ServiceControl/EventLog/EventLogApiModule.cs
@@ -5,6 +5,7 @@
     using Raven.Client;
     using ServiceBus.Management.Infrastructure.Extensions;
     using ServiceBus.Management.Infrastructure.Nancy.Modules;
+    using ServiceControl.Infrastructure.Extensions;
 
     public class EventLogApiModule : BaseModule
     {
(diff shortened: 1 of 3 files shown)


johnsimons:
@petersgiles In SP you should from now on specify the page size for the `/eventlogitems` REST call, eg `/eventlogitems?per_page=20`
If you have any issues let me know

johnsimons:

gbiellem:
Aside from My nitpicking about the comment this looks good

johnsimons:
I will add the comment

petersgiles:
If we are going to change the rest api for list paging can we do it the same way for all lists.

if you are going to make an api change have a look the ajax bits of this https://www.datatables.net/ might give you an idea of the scope of things people want to do with lists and tables

johnsimons:
> If we are going to change the rest api for list paging can we do it the same way for all lists.

we are not planning to do any changes in paging @petersgiles

petersgiles:
>  REST call was unbounded, and ServicePulse really only needs the top 10 documents.
I have added paging to the REST api call

hmm looks as if you are.

All I'm suggesting is that it might be a good thing while we are in there to add a bit of filtering etc. not hard to do but would have made @mauroservienti & @ramonsmits happy during their last fault finding adventure

Feel free to ignore me as this is only a suggestion and not in any way intended to distract you from what you are doing

johnsimons:
@petersgiles all I did was add the existing way we do paging to `/eventlogitems`, that is all.
I haven't seen any need to support filtering/sorting for event log items, maybe I missed it?

PR #664, opened 2016-02-17
Using patch api to update heartbeats

johnsimons:
This PR uses the patch api to update heartbeats.

This saves us from having to do a read.

### DIFF
--- a/src/ServiceControl/HeartbeatMonitoring/SaveHeartbeatHandler.cs
+++ b/src/ServiceControl/HeartbeatMonitoring/SaveHeartbeatHandler.cs
@@ -1,19 +1,28 @@
 ﻿namespace ServiceControl.HeartbeatMonitoring
 {
     using System;
-    using Contracts.HeartbeatMonitoring;
+    using System.Linq;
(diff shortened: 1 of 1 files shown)


johnsimons:
@mikeminutillo this is a small but hopefully significant improvement.
I know is not exactly what you described in #639, but given the difficulty to pull off #639 as described and also given that we know we want to split out heartbeats and rearchitect it from the ground up, I wonder if this PR is a good alternative for now ?

ramonsmits:
:+1: Good improvement, reduced API calls by 50% regarding heartbeat updates. 

Issie with #639 is also that the SQL transport does a lousy job regarding expired messages.

adamralph:
Any functional change here or just a refactoring?

johnsimons:
@adamralph just a change how we update documents

adamralph:
@johnsimons @gbiellem thanks, if no functional change then refactoring label makes sense.

johnsimons:
Actually this is just an improvement, a label that is missing

johnsimons:
Merged into release-1.11

PR #682, opened 2016-02-25
AdoptOrphanedBatches running more then necessary

### LINKED ISSUE #681: AdoptOrphanedBatches running more then necessary

johnsimons:
Fixes #681

### DIFF
--- a/src/ServiceControl/Infrastructure/TimeKeeper.cs
+++ b/src/ServiceControl/Infrastructure/TimeKeeper.cs
@@ -10,21 +10,23 @@
         ConcurrentDictionary<Timer, object> timers = new ConcurrentDictionary<Timer, object>();
         private ILog log = LogManager.GetLogger<TimeKeeper>();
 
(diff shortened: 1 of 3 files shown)


johnsimons:

johnsimons:
@mikeminutillo please review it before monday

mauroservienti:
If I understood everything correctly:

* when messages are retried, group of messages, they are retried in batches
* it may happen, e.g. SC is shut down while retrying, that one or more batch is orphaned 
* this process looks for orphans and if found restarts the retry batch, otherwise stops itself from looking for orphans
* before we were continuously looking for orphans 

If that's the case, then it looks fine and I cannot spot any smell in the code. @mikeminutillo last word to you :smile: , I'm :+1:

mikeminutillo:
I'm usually against `out` params on `void` methods but in this case I think it makes it clearer what's going on. :+1: from me

mikeminutillo:
This will all go away if we implement https://github.com/Particular/PlatformDevelopment/issues/635 ;)

PR #694, opened 2016-03-16
Add a retention policy for error messages

johnsimons:
## Symptoms
ServiceControl currently only has a retention policy for audit messages.
Error messages are currently kept forever.
At worse this causes ServiceControl to eventually run out of disk space and bring down the process.

## Who's affected
All users of ServiceControl that are currently running v1.11.1.

#### Service Control Management per instance 
- [x] New field (`ServiceControl/ErrorRetentionPeriod`) to capture retention for the unresolved error messages, this field range has to be between 10-45 days, so minimum would be 10 days and maximum 45 days;
- [x] New field (`ServiceControl/AuditRetentionPeriod`) to capture retention for the audits, this field range is between minutes-days, so minimum would be 1 hour and maximum 365 days;

#### Service Control  
- Api Changes:
  - [x] Modify /api/errors to include `LastModified` datetime in results and allow to sort and filter on that same new field.  
  Example of usage `
  http://localhost:33333/api/errors/?sort=modified&modified=2016-01-25T15:38:35.6767764Z...2016-01-25T15:38:36.6767764Z&status=archived`.  
  Date formatting for the query parameters must follow the [ISO8601 standard](http://en.wikipedia.org/wiki/ISO_8601), [here is an example on how to do it in javascript](http://www.w3schools.com/jsref/jsref_toisostring.asp).
  - [x] We are also including #703
  - [x] Unarchive by date range and by array of ids.  
  Example of usage by date range:
  ```
  PATCH http://localhost:33333/api/errors/2016-01-25T15:38:35.6767764Z...2016-01-25T15:38:36.6767764Z/unarchive
  ```
  Example of usage by ids:
  ```
  PATCH http://localhost:33333/api/errors/unarchive
  ```
  ```json
  [ 
 "e989a17f-4cf5-79f4-6399-2bddd3084412",
 "5feb1b8c-c171-abfe-afad-f863b7681d09"
  ]
  ```
  - [x] Get current cool-off settings  
  Example of usage:
  ```
  GET http://localhost:33333/api/configuration
  ```
  ```json
{
    "host": {
        "service_name": "Particular.ServiceControl",
        "raven_db_path": "C:\\ProgramData\\Particular\\ServiceControl\\localhost-33333",
        "logging": {
            "log_path": "C:\\Users\\john\\AppData\\Local\\Particular\\ServiceControl\\logs",
            "logging_level": "Warn",
            "raven_db_log_level": "Warn"
        }
    },
    "data_retention": {
        "audit_retention_period": "01:00:00",
        "error_retention_period": "02:00:00"
    },
    "performance_tunning": {
        "maximum_message_throughput_per_second": 350,
        "max_body_size_to_store": 102400,
        "http_default_connection_limit": 100,
        "external_integrations_dispatching_batch_size": 100,
        "expiration_process_batch_size": 32756,
        "expiration_process_timer_in_seconds": 600
    },
    "transport": {
        "transport_type": "NServiceBus.MsmqTransport, NServiceBus.Core, Version=5.0.0.0, Culture=neutral, PublicKeyToken=9fc386479f8a226c",
        "audit_log_queue": "audit.log",
        "audit_queue": "audit",
        "error_log_queue": "error.log",
        "error_queue": "error",
        "forward_audit_messages": false,
        "forward_error_messages": false
    },
    "plugins": {
        "heartbeat_grace_period": "00:00:40"
    }
}
  ```
  - [x] SignalR messages for service pulse, when messages are unarchived
  - [x] Schedule job to do deletion of unresolved error messages that are older than the cool-off period
- [x] Rename existing `ServiceControl/HoursToKeepMessagesBeforeExpiring` to `ServiceControl/AuditRetentionPeriod`

#### Update doco
- [x] http://docs.particular.net/servicecontrol/creating-config-file
- [x] http://docs.particular.net/servicecontrol/how-purge-expired-data
- [x] http://docs.particular.net/servicecontrol/installation-silent

Ongoing PR - https://github.com/Particular/docs.particular.net/pull/1318

Connects to https://github.com/Particular/PlatformDevelopment/issues/667


petersgiles:
Signlr messages should occur when
- a synchronous action finishes
- an asynchronous action starts and when it finishes
- any action's state changes i.e. progress updates, workflow state

So if deleting is done via a back ground job I should get a message 
- When the job has been accepted
- When the job starts
- When the processing of the job has achieved a milestone
- When the job is done

The signlr updates should be consistent with the data returned via the web api. That is, when I refresh the page, the data presented should be consistent with the state of the application as derived from the signlr messages. So if signlr showed processing of a thing at 50% then the data from the web api should show processing of a thing at 50%.

The same would be true for un-archiving

petersgiles:
@johnsimons I need to have the current retention policy in minutes come back with the archived messages. can you stick it in the header

gbiellem:
Bumped this to 1.13 so we can get the SQLTransport patch out now

petersgiles:
@johnsimons  the total on this /api/errors?status=archived is the total returned not the total of all archived messages. you should probably return both in the header of the response

http://localhost:33333/api/errors?status=archived&page=1&sort=modified&direction=desc&modified=2016-03-30T13:03:15...2016-03-30T15:03:15

johnsimons:
@petersgiles I'll try tomorrow to check this out, but it should work exactly like the other queries

petersgiles:
@johnsimons I'm suggesting that all the other queries are unsuitable too

petersgiles:
@john I archived 5 messages in the last 2 hours 

when i query to last 2 hours I get nothing

http://localhost:33333/api/errors?status=archived&page=1&sort=modified&direction=desc&modified=2016-03-31T12:03:24...2016-03-31T14:03:24

when I query the last day I get the five I'm after

http://localhost:33333/api/errors?status=archived&page=1&sort=modified&direction=desc&modified=2016-03-30T14:03:18...2016-03-31T14:03:18

ramonsmits:
I would expect the following:

    /api/errors/unarchive?from=2016-01-25T15:38:35.6767764Z&to=2016-01-25T15:38:36.6767764Z

Instead of:

    /api/errors/2016-01-25T15:38:35.6767764Z...2016-01-25T15:38:36.6767764Z/unarchive

That would be much more consistent with all the other arguments listed in the description.

It is weird that you can also do 

    /api/errors/unarchive

Also, based on the description it seems you cannot do:

    /api/errors/unarchive?id=e989a17f-4cf5-79f4-6399-2bddd3084412

or

    /api/errors/unarchive?id=e989a17f-4cf5-79f4-6399-2bddd3084412,5feb1b8c-c171-abfe-afad-f863b7681d09

ramonsmits:
As an example, its weird to have date ranges in the paths because they can overlap. On a real filesystem you would not have that.

    /api/errors/2016-01-25...2016-01-28/unarchive

    /api/errors/2016-01-27...2016-01-31/unarchive

johnsimons:
@mikeminutillo can I get you to do a code review on this one ?

johnsimons:
@ramonsmits I am sorry that the REST api feels weird to you, maybe raise another issue if you feel strong about it.
For now this is it, @petersgiles has already coded ServicePulse using it, so unfortunately too late now.

src/ServiceControl.AcceptanceTests/CustomChecks/Custom_check_transition_should_trigger_signalr_event.cs
```diff
@@ -36,7 +37,6 @@ public class MyContext : ScenarioContext
             public bool SignalrEventReceived { get; set; }
             public int SCPort { get; set; }
             public string SignalrData { get; set; }
-            public bool SignalRConnected { get; set; }
```
mikeminutillo: Why did this get removed? It is useful in determining the root cause of errors in commonly failing tests on the Build Server

src/ServiceControl.UnitTests/Expiration/ChunkerTests.cs
```diff
@@ -0,0 +1,80 @@
...
+            Assert.AreEqual(499, ends[0]);
+
+            Assert.AreEqual(500, starts[1]);
+            Assert.AreEqual(999, ends[1]);
+
+            Assert.AreEqual(1000, starts[2]);
+            Assert.AreEqual(1499, ends[2]);
+
+            Assert.AreEqual(3, starts.Count);
+            Assert.AreEqual(3, ends.Count);
```
mikeminutillo: Stick a message on these asserts otherwise we won't be able to tell which one fails (not that I envisage them ever failing)

src/ServiceControl/Contracts/Operations/FailureDetails.cs
```diff
@@ -4,6 +4,11 @@ namespace ServiceControl.Contracts.Operations
 
     public class FailureDetails
     {
+        public FailureDetails()
+        {
+            TimeOfFailure = DateTime.UtcNow;
```
mikeminutillo: Is this correct. Is there ever a case where this does not get set? I thought we default this if we don't get a header.

johnsimons: There shouldn't, but just in case I decided to initialise it, because indexes depend on it

src/ServiceControl.UnitTests/Expiration/ChunkerTests.cs
johnsimons: Well, the stacktrace will tell u the line number :wink:, but yes I could add a message to make it simpler

src/ServiceControl.AcceptanceTests/CustomChecks/Custom_check_transition_should_trigger_signalr_event.cs
johnsimons: No reason, it was a field that we don't use, we can put it back for diagnostics if we need it again

src/ServiceControl/Infrastructure/Extensions/QueryableExtensions.cs
```diff
@@ -161,11 +169,52 @@ public static IDocumentQuery<FailedMessageViewIndex.SortAndFilterOptions> Filter
             }
             sb.Append(")");
 
+            source.AndAlso();
```
mikeminutillo: This is a terrible API. What does this do? Should we add a comment here to explain why we need this?

mikeminutillo: The Raven API `AndAlso()` is the thing I am referring to as terrible.

src/ServiceControl/Infrastructure/Nancy/Modules/RootModule.cs
```diff
@@ -27,21 +27,58 @@ public RootModule()
...
+            Func<dynamic, dynamic> configuration = p => Negotiate
+                .WithModel(new
+                {
+                    Host = new
+                    {
+                        Settings.ServiceName,
+                        RavenDBPath = Settings.DbPath,
+                        Logging = new
+                        {
+                            LoggingSettings.LogPath,
```
mikeminutillo: Should we prefer explicit property names here? i.e. `LogPath = LoggingSettings.LogPath`. Otherwise refactoring the settings class will impact the public json API and we won't detect that straight away. The same comment holds for the rest of this method.

src/ServiceControl/Infrastructure/Extensions/QueryableExtensions.cs
johnsimons: This is the api to build a lucene query, the `AndAlso` adds `AND` to the query

src/ServiceControl/Infrastructure/Nancy/Modules/RootModule.cs
johnsimons: We can do that, but then we have R# complaining, so I opted to please the static analysis tool

src/ServiceControl/Infrastructure/RavenDB/Expiration/ErrorMessageCleaner.cs
```diff
@@ -0,0 +1,112 @@
...
+                        });
+                }
+                catch (OperationCanceledException)
+                {
+                    //Ignore
+                }
+
+                logger.Debug("Batching deletion of {0} documents.", items.Count);
+
+                docsToExpire += items.Count;
```
mikeminutillo: docsToExpire will always be 0 before this, won't it? We are not in a loop at this point

johnsimons: this is not the latest code

gbiellem:
Thanks @petersgiles  - I'll fix that

src/ServiceControl/Infrastructure/RavenDB/Expiration/AuditMessageCleaner.cs
```diff
@@ -0,0 +1,125 @@
...
+                //Ignore
+            }
+
+            var deletionCount = 0;
+
+            Chunker.ExecuteInChunks(items.Count, (s, e) =>
+            {
+                logger.InfoFormat("Batching deletion of {0}-{1} audit documents.", s, e);
+                var results = database.Batch(items.GetRange(s, e - s + 1));
+                logger.InfoFormat("Batching deletion of {0}-{1} audit documents completed.", s, e);
```
mikeminutillo: This will look weird:

> Batching deletion of 500-999 audit documents completed.

Sounds like we aren't sure how many documents we deleted.

src/ServiceControl/Infrastructure/RavenDB/Expiration/ExpiryErrorMessageIndex.cs
```diff
@@ -0,0 +1,24 @@
...
+    using Raven.Client.Indexes;
+    using ServiceControl.MessageFailures;
+
+    public class ExpiryErrorMessageIndex : AbstractIndexCreationTask<FailedMessage>
+    {
+        public ExpiryErrorMessageIndex()
+        {
+            Map = messages => from message in messages
+                where message.Status != FailedMessageStatus.Unresolved
+                let last = message.ProcessingAttempts.Last()
```
mikeminutillo: Does this index actually use `last`?

johnsimons: Fixed, good catch :smile:

src/ServiceControl/Infrastructure/RavenDB/Expiration/ExpirySagaAuditIndex.cs
```diff
@@ -15,16 +15,13 @@ public class Result
...
                 select new Result
                 {
-                    LastModified = MetadataFor(sagaHistory)["Last-Modified"],
-                });
+                    LastModified = MetadataFor(sagaHistory).Value<DateTime>("Last-Modified").Ticks
+                };
 
             DisableInMemoryIndexing = true;
-
-            Sort(result => result.LastModified, SortOptions.String);
```
mikeminutillo: Should there still be a sort on this index? When we read the index we apply a sort to the query so it will need to get sorted anyway.

johnsimons: Don't think so because it is a numeric field now

src/ServiceControl/MessageFailures/Api/UnArchiveMessages.cs
```diff
@@ -0,0 +1,60 @@
...
+                return HttpStatusCode.Accepted;
+            };
+
+            Patch["/errors/{from}...{to}/unarchive"] = parameters =>
+            {
+                DateTime from, to;
+
+                try
+                {
+                    from = DateTime.Parse(parameters.from, CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind);
```
mikeminutillo: Do we accept requests where from >= to? And what happens when we execute the query that way around?

johnsimons: it work both ways, because essentially we use a lucene range query

src/ServiceControl/MessageFailures/Handlers/UnArchiveMessagesHandler.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    using System.Linq;
+    using Contracts.MessageFailures;
+    using InternalMessages;
+    using NServiceBus;
+    using Raven.Client;
+
+    public class UnArchiveMessagesHandler : IHandleMessages<UnArchiveMessages>
+    {
+        public IDocumentSession Session { get; set; }
```
mikeminutillo: Should this use doc store and manage it's own session? As we aren't executing in a transaction it's possible for the bulk update to fail if one of the messages gets updated after it's loaded and the publish will go out before the save fails. 

It's not terribly important at this stage because nothing should be happening to errors in the archived state. It may cause problems with really bad luck and retention deletion.

johnsimons: > As we aren't executing in a transaction

we are in a transaction, MSMQ transaction, will roll back and try again.
Or maybe I am misunderstanding your comment ?

mikeminutillo: OK so the outgoing MSMQ message (representing the event) will be rolled back if the Unit of Work fails? Do all of the transports capture outgoing messages in the receive transaction?

mikeminutillo:
Review complete

src/ServiceControl/MessageFailures/Handlers/UnArchiveMessagesHandler.cs
johnsimons: > Do all of the transports capture outgoing messages in the receive transaction?

No,but for those we delay the sending till the end of everything else

petersgiles:
The more I think on it the more I reckon the questions on upgrade should have a previous button etc, make it more like a traditional wizard

petersgiles:
Also it seems to me that a summary of the things I have chosen should appear before the upgrade starts, (not important for now)

petersgiles:
I used the defaults during my upgrade but when i go to edit them the values look like this

o oh

should be 30 days and 15 days respectively

gbiellem:
@petersgiles  Are the values correct in the config? i.e did it fail to set them or read them back on refresh?

petersgiles:
@gbiellem tracing through the way instances are loaded I think we need to go have a look all the way down to WindowsServiceController

gbiellem:
So I think we're done with this one.  @Particular/servicecontrol-maintainers  any object to merging and starting a release

petersgiles:
I say we're ready to go

johnsimons:
May the force be with you!
And good luck stabilising acceptance tests :grimacing:

PR #711, opened 2016-04-19
SagaUpdatedMessage handler can cause lots of concurrency exceptions.

### LINKED ISSUE #403: SagaUpdatedMessage handler can cause lots of concurrency exceptions
## Who's affected
Anyone using the [Saga plugin](http://docs.particular.net/servicecontrol/plugins/saga-audit) for ServiceControl

## Symptoms
When a `SagaUpdatedMessage` message is [handled](https://github.com/Particular/ServiceControl/blob/7db35f37e92c2e949ac111003dc05826387934be/src/ServiceControl/SagaAudit/SagaUpdatedHandler.cs), we attempt to create a new or update an existing raven document.

In an environment that sends lots of (100+) `SagaUpdatedMessage` messages to ServiceControl and the `SagaId` is the same, when we attempt to save the updated saga we are very prone to get a concurrency exception because ServiceControl processes messages using [multiple threads](https://github.com/Particular/ServiceControl/blob/7db35f37e92c2e949ac111003dc05826387934be/src/ServiceControl/ConfigTransportConfig.cs#L13).

Because in ServiceControl we only have FLR enabled ([SLR is disabled](https://github.com/Particular/ServiceControl/blob/7db35f37e92c2e949ac111003dc05826387934be/src/ServiceControl/Bootstrapper.cs#L36)), the chances of these messages ending up in the ServiceControl error queue is quite high.

We have had a report of this exact situation occurring.

--

### Proposal
Change the way we store saga history so that instead of using a single document, we split it into multiple documents.
So we would end-up with a `SagaHistory` document and multiple standalone `SagaStateChange`.
This way the `SagaUpdatedMessage` handler would only do inserts into raven.
We would then create an index that merges all saga info into a single output.

johnsimons:
Fixes #403


src/ServiceControl/SagaAudit/SagaDetailsIndex.cs
```diff
@@ -0,0 +1,58 @@
...
+                            OutgoingMessages = doc.OutgoingMessages,
+                            StartTime = doc.StartTime,
+                            StateAfterChange = doc.StateAfterChange,
+                            Status = doc.Status
+                        }
+                    }
+                });
+
+            //Legacy so we still scan old sagahistories
+            AddMap<SagaHistory>(docs => from doc in docs
```
danielmarbach: So the old index will still be running. Technically we haven't solved anything when this one is not deprecated or do I misunderstand?

johnsimons: @danielmarbach u are misunderstand.
The problem is the inserts, where we used to do a load and update, also we used to have an unbouded collection history on the document.
We now do inserts only, no load or update
And the we do a map reduce on the query side

Hope this makes sense

danielmarbach: Ok thanks!

johnsimons:
@Particular/servicecontrol-maintainers anyone opposed to ship this one out ?

mauroservienti:
No. Let's ship this.

PR #629, opened 2016-04-22
ServiceControl Management and ServiceControl report different license status

gbiellem:
### Sympton

ServiceControl Management Util reports that a valid license is installed but ServiceControl reports that there is no valid license

### Analysis 
The ServiceCotnrol Management  Util scans for licenses in multiple locations.  One of those locations is `HKEY_CURRENT_USER` registry hive.  This isn't appropriate as it is very unlikely the ServiceControl service is running as the installing user.

### DIFF
--- a/src/ServicControlInstaller.Packaging/ServiceControlInstaller.Packaging.csproj
+++ b/src/ServicControlInstaller.Packaging/ServiceControlInstaller.Packaging.csproj
@@ -22,6 +22,7 @@
     <DefineConstants>DEBUG;TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
+    <LangVersion>6</LangVersion>
   </PropertyGroup>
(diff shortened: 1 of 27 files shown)


gbiellem:
@johnsimons  review please and merge.  Just changed over to new particular.licensing source only package that addresses this bug.  Bumped the C# version to 6 as string interpolation is used in that code

gbiellem:
@Particular/servicecontrol-maintainers  review?

mikeminutillo:
@gbiellem it looks good to me (and the code is much cleaner and easier to follow). Do we have any tests in the product that cover finding license files? 

I'd love to go over it and verify that if it worked before, it's going to work still without customers having to update their license file.

gbiellem:
@mikeminutillo  Tests are in the private repo https://github.com/Particular/Operations.Licensing not SC. If you pull that repo down  you can also generate test licenses locally. I can help you with this tomorrow morning your time if you like.

HEskandari:
I'm :+1: on it. No objection on merging it in. Do we have separate task(s) to update downstream repos? If not should we raise one for each impacting repo?

PR #718, opened 2016-05-04
Added a unit test and a fix to use a proper culture when converting p…

### LINKED ISSUE #717: User's culture is used to serialize objects sent through the endpoints
If user are using a culture like Turkish where there are different upper/lower case characters it will affect how the output json is rendered. 

In the example of accessing /messages endpoint, the rendered json will have "message_ıd" instead of "message_id" for that property (notice the different "i). This is not limited to any given endpoint and is applied through all the property names.

HEskandari:
…roperty names. Fixes #717


johnsimons:
looks good

PR #720, opened 2016-05-06
Upgrade to RavenDB v3

johnsimons:
This PR upgrades RavenDB to v3

As part of this upgrade ServiceControl is now using a separate database in RavenDB , before we were co-sharing the <system> database.
In RavenDB v3, the embedded database feature has been changed, it now requires all connections to be http based, so this means that we need to ensure `/storage` url is accessible.
Because RavenDB v3 data storage has changed, we need to migrate existing data to a different location on the disk.
The purging of data is now integrated as part of ServiceControl, we no longer use a RavenDB bundle.

So what is left to do here:
- [x] migration of data to the new location, this needs to be performed as part of the install  
  The Data needs to be moved from eg `C:\ProgramData\Particular\ServiceControl\localhost-33333` to `C:\ProgramData\Particular\ServiceControl\localhost-33333\Databases\ServiceControl`, so the src content needs to be moved to the destination folder.
- [x] `/api` url acl need to be removed  
  We now use OWIN for all urls (/api, /storage, ...), so OWIN is actually listening on `/` and then redirects traffic to the correct handler.
- [x] ensure `/storage` url is accessible  
  This needs to be done all the time, we just need it to be accessible via localhost, but if the customer wants to expose it differently, that should also work
- [ ] ~~Given that RavenDB is releasing v3.5 end of May, should we wait and upgrade to it instead ?~~  
I have upgraded to the unstable to test it and it seems to work fine, we probably would just need to get an updated license
- [ ] We are not running the latest version of NServiceBus.RavenDB because of https://github.com/Particular/NServiceBus.RavenDB/issues/229
- [x] Deal with the schema update of the lucene storage in a nicer way  
see https://github.com/Particular/ServiceControl/blob/e03310739a3ca21cae4608a8fa80998ca77a462c/src/ServiceControl/Infrastructure/RavenDB/RavenBootstrapper.cs#L49-L62, we probably want to do it as port of the migration of the Data
- [x] Figure out what to do with `<add key="Raven/Licensing/AllowAdminAnonymousAccessForCommercialUse" value="true" />`


mauroservienti:
### RavenDB 3.5 new features

* [Data exploration](https://ayende.com/blog/171139/what-is-new-in-ravendb-3-5-exploring-data-in-the-dark): a studio feature that can be usefull foir us when developing new features, totally unrelated to our end users since they do not make any change to the data structures
* [Clusters are now first class citizens](https://ayende.com/blog/173893/ravendb-3-5-whirl-wind-tour-is-this-a-cluster-in-your-pocket-and-you-are-happy-to-see-me): not rerevant to us unless we decide to support RavenDB cluster type as a scale-out option, requires a full redisgn to handle eventual consistency of the cluster itself, master-master locig and merge replication conflits
* [Global and shared configuration options](https://ayende.com/blog/173894/ravendb-3-5-whirl-wind-tour-configuring-once-is-best-done-after-testing-twice) not relevant to us
* [Hot spare support](https://ayende.com/blog/174017/ravendb-3-5-whirl-wind-tour-can-you-spare-me-a-server), again a cluster feature not interesting for ServiceControl as it stands now
* [Admin scripts](https://ayende.com/blog/173896/ravendb-3-5-whirl-wind-tour-im-the-admin-and-i-got-the-power), handle with a lot of care, could be an interesting option to debug/fix on the fly/workaround production issues
* [Better replication control](https://ayende.com/blog/173898/ravendb-3-5-whirl-wind-tour-you-want-all-the-data-you-cant-handle-all-the-data) not interesting right now
* [I/O bandwidth analysis](https://ayende.com/blog/173899/ravendb-3-5-whirl-wind-tour-ill-find-who-is-taking-my-i-o-bandwidth-and-they-shall-pay) can be an interesting addition to better understand what is going on ion a production environment
* [Better patch results, better delete by index API, better stats](https://ayende.com/blog/173901/ravendb-3-5-whirl-wind-tour-ill-have-the-3-1-goodies-to-go-please) are all nice to have
* [Fully custom thread pool](https://ayende.com/blog/171138/what-is-new-in-ravendb-3-5-my-thread-pool-is-smarter) to better handling resources at the database level

mauroservienti:
### Authentication options

When it comes to [authentication](https://ravendb.net/docs/article-page/3.0/csharp/server/configuration/authentication-and-authorization) we have 3 options at the database server level:

1. ignore and set up anonymous access, this works just fine since the database is embedded however if forces us to set `<add key="Raven/Licensing/AllowAdminAnonymousAccessForCommercialUse" value="true" />` in the ServiceControl config file
* use Windows Authentication, IMO, not an option since we could say that by default we allow access to the user running the ServiceControl service, that is well know at setup time but can be manually changed later and we need to find a way to keep things in sync
* setup and `API Key` and use oAuth to talk to RavenDB from ServiceControl

The API Key approach is fairly simple, at setup time we can run the following code:

```
store
	.DatabaseCommands
	.Put(
		"Raven/ApiKeys/sample",
		null,
		RavenJObject.FromObject(new ApiKeyDefinition
		{
			Name = "sample",
			Secret = "ThisIsMySecret",
			Enabled = true,
			Databases = new List<ResourceAccess>
			        {
				        new ResourceAccess {TenantId = "*"},
				        new ResourceAccess {TenantId = Constants.SystemDatabase},
			        }
		}), new RavenJObject());
```

And then setup our own document store(s) in the following way:

```
var store = new DocumentStore
{
	ApiKey = "sample/ThisIsMySecret",
       ....
};
```

I cannot find anything dramatically wrong in using option [1]. Thoughts?

johnsimons:
@mauroservienti I actually think Windows Authentication is the way to go, and we already have a way for user to change username/password via the installer, they actually have to do that otherwise they will have issues with urlacls (those also need to be updated)

johnsimons:
@mauroservienti regarding authentication, what happens when a user also wants to access the studio from a non localhost url ? if we have windows auth on, do they get prompted for credentials ?

johnsimons:
@mikeminutillo another cool think, database compaction is much nicer:
image

mikeminutillo:
## Justifications for Promotion

1. This is a quick win for stability. We have seen situations where ServiceControl loses stability under very high load. Initial testing shows that the upgrade to Raven 3 greatly improves this. 
  - Under Raven 2.5, when the background purge task happens to delete old messages, ingestion rates drop dramatically. In some systems, this drop can lead to a situation where SC is unable to ever catch up even after the purge has completed. Under Raven 3, the deletes take longer to complete but we see very little impact on ingestion rates. Arguably, it is more important to keep ingesting new messages with lower latency then to delete old messages.
  - We see less garbage collector pressure.
  - We see overall lower resource utilization.
2. It lays the groundwork for componentization effort.
  - Raven 3 uses the same interfaces for embedded DBs as external DBs making it easier to externalize the data store
  - Raven 3 allows us to create additional databases in embedded mode so we are not confined to the system database. This would allow us to store data in separate databases with separate indexes as we undergo the componentization effort. This in turn will lead to easier physical separation if and when we decide to go down that route.
3. It's already mostly done and tested. The remaining work is to ensure that the customer has a smooth upgrade experience when moving from Raven 2.5 to Raven 3. This mostly involves moving the data to a second database named ServiceControl and re-indexing it. 
4. Keeping it rebased while we do other work is painful. This is a large PR and every time we do another fix or release we need to rebase the PR on top of the new changes.

johnsimons:
So some news on the 3.5 release - https://groups.google.com/d/msg/ravendb/ggQgP_V3JsI/BOrGPwEiFAAJ, in summary they expect a stable release around Mid Jul / Early Aug. I don't think is worth waiting

johnsimons:
So it seems we have windows auth by default :smile:

johnsimons:
@mauroservienti I need some help setting the windows auth, I am trying to set it for both dbs:
image
but that is not working.

Also I have modified the code to include:
```c#
documentStore.Configuration.AllowLocalAccessWithoutAuthorization = true;
documentStore.Configuration.AnonymousUserAccessMode = AnonymousUserAccessMode.None;
documentStore.Credentials = CredentialCache.DefaultNetworkCredentials;
```
Is this what we should have ?
How about the <system> database ?

johnsimons:
So authentication in SC vnext is as follow:

The user account that SC runs under has Read/Write to ServiceControl database
image

And then I have added the BUILTIN\Administrators to be admins:
image

johnsimons:
So I am now waiting for the lucene schema update:
image

I wonder if we need a cancel ?

johnsimons:
@Particular/servicecontrol-maintainers I wonder if we should start using RavenFS ?
So any new instances would use it and any existing would keep using the obsolete way.
We know if the user is doing an update or a fresh install so it is easy to differentiate.

johnsimons:
Have just tried it out and it seems our license does not include RavenFS!

gbiellem:
@johnsimons  FYI... the SC MSMQ ATs hung on the last run of this branch - https://builds.particular.net/viewLog.html?tab=buildLog&buildTypeId=bt58&buildId=179736

johnsimons:
No worries @gbiellem, still a work in progress, will take me some time to upgrade this branch

johnsimons:
This PR has been superseded by #772

Issue #484, opened 2015-07-08
Circuit breaker settings are hard coded in Service Control and plugins


vellad:
The two circuit breaker settings are hard coded for Service Control to these values:
timeToWaitBeforeTriggering: 2 mins
delayAfterFailure: 20 secs

They are also hardcoded in the heartbeat plugin to these values:
timeToWaitBeforeTriggering: 2 mins
delayAfterFailure: 1 sec (default from circuit breaker class)

This could be the case in other plugins as well.

They need to all be configurable in code and/or a config file.

johnsimons:
`EventDispatcher` still has hardcoded settings, but I wonder why a user needs to change them.
@Particular/servicecontrol-maintainers what do u think, can u see a reason to make thises settings configurable ?

mauroservienti:
The only good reason that comes toy mind is that we have customers using brokers, such as RabbitMQ or SQL, on a not so reliable infrastructure. We added the ability to configure the circuit breaker to overcome to that issue, could this be affected?

johnsimons:
@mauroservienti in that scenario I see us doing them a favor and pointing out issues with their own infrastructure.
The current hardcoded times are IMO quite generous, we currently wait up to 2mins before arming the breaker, this is IMO quite a bit of time to wait for a network to respond, increasing this number even further could have dangerous effects.

yossita:
Just in this subject, when communicating with a sql server cluster, some
environments take time to fail over, 2 minutes might not be enough to
recover.

johnsimons:
@yossita what time do you think it would be safer ? Would 5mins work ?

yossita:
Not sure, this is why we asked it to be configurable :) We've seen
different behaviours with different SQL Server versions, configuration and
environments. The major issue we had was when the circuit breakers are
triggered, a critical exception is thrown which causes the hosting service
to crash.

johnsimons:
> The major issue we had was when the circuit breakers are
> triggered, a critical exception is thrown which causes the hosting service
> to crash.

@yossita that is actually on purpose, we want to raise awarness of it, so you can act on it.
The issue is that if we cannot contact the transport, we are in essence broken and not processing messages, hence the monitoring capability is stopped!
Maybe there is a different way of raising awarness?

yossita:
Like logging an error?

johnsimons:
> Like logging an error?

I'm pretty sure we log errors, but those do not have enough impact and therefore get ignored.
[...]

PR #722, opened 2016-05-10
Circuit-breaker in EventDispatcher is not being reset

johnsimons:
## Symptoms
Once the circuit-break triggers, example an intermittent break in comms to the broker, the circuit-breaker will never be reset and therefore ServiceControl will shutdown at the end of the circuit-breaker timeout.

## Who's Affected
All versions of ServiceControl prior to v1.15


johnsimons:

SzymonPobiega:
Looks good :+1: . BTW, why changing the timeout from 2 to 5 mins?

johnsimons:
> BTW, why changing the timeout from 2 to 5 mins?

@SzymonPobiega https://github.com/Particular/ServiceControl/issues/484

Issue #723, opened 2016-05-11
ServiceControl cannot be used through an ARR reverse proxy in ServicePulse


boblangley:
## Symptoms

ServicePulse does not receive updates from the ServiceControl

## Who is affected

Any user attempting to install ServicePulse using ARR proxy in a non-root directory.

## POA

- [x] Update ServiceControl SignalR version to latest stable. #724 
- [x] Smoke Test ServiceControl / ServicePulse
- [ ] Create new release of ServiceControl
- [ ] Announcement

johnsimons:
@boblangley is there an link to when this was fixed in SignalR ?

johnsimons:
Fixed in https://github.com/Particular/ServiceControl/pull/724

PR #724, opened 2016-05-11
Signalr 2.2.0

boblangley:
## Symptoms

ServicePulse does not receive updates from the ServiceControl

## Who is affected

Any user attempting to install ServicePulse using ARR proxy in a non-root directory.

connects to #723


johnsimons:
looks good

johnsimons:
It seems gitversion is getting confused with the branch name called `signalr-2.2.0` so we need to merge this to a `release-1.16` before merging it to `master` to get the correct version.

johnsimons:
@boblangley should I proceed with the release, Or do u want to handle this one ?

Issue #733, opened 2016-05-19
Nested path check fails when the directories are siblings with common prefix


slahn:
## Symptoms
We have ServiceControl installed under `X:\Particular\ServiceControl`, and the database installed under `X:\Particular\ServiceControlDB`.
We're not allowed to reconfigure this instance using the management tools, because those paths are considered 'nested' by [CheckNoNestedPaths] (https://github.com/Particular/ServiceControl/blob/a5af0081d72e0fa3f88f19c10a5f5036ed45a1c9/src/ServiceControlInstaller.Engine/Validation/PathsValidator.cs#L111).

## Who is affected
Any user attempting to use the management tools to reconfigure an instance with such nesting.

slahn:
I've added a fix and a simple test in my branch (slahn@1d6bfcc800634150163f8014c223dbee5c3af66f).

johnsimons:
I don't think this is a recommended setup hence the reason for not supporting it.
I think you need to manually move the database around to fix this issue. Do you need help with this ?

gbiellem:
@johnsimons - No it's a legitimate bug

johnsimons:
@slahn want to submit a PR ?

johnsimons:
Fixed in https://github.com/Particular/ServiceControl/pull/734

PR #734, opened 2016-05-20
Don't treat sibling paths as nested (#733)

slahn:
Fixes the issue described in #733, and adds a test for it.


johnsimons:
Looks good.
thanks @slahn

johnsimons:
@slahn can you sign our contributions form, see http://particular.net/sourcecode

slahn:
@johnsimons Done.

Issue #715, opened 2016-04-20
Support "NT AUTHORITY\LocalService" in installer


ramonsmits:
I almost always install SC using the *NT AUTHORITY\LocalService* account and have to search for the exact string to paste into the box.

it would help if this account could be added in the list:

![image](https://cloud.githubusercontent.com/assets/152998/14665735/908021d2-06d5-11e6-9700-aad5f0c6d1ae.png)

This makes it possible to install SC for SQL with integrated security.

gbiellem:
If you just enter 'localservice' it works it out so no need to remember the "NT AUTHORITY\Local Service"

gbiellem:
And yeah we should add it to the UI

gbiellem:
@sergioc if we add another option do  you want to change this to a dropdown?

sergioc:
I think the number of options is still low enough to not justify changing to a dropdown.

johnsimons:
I don't think this is as easy as it looks. It seems by default the local service account has no rights to log on as a windows service, so I think we would need to check that ?
@gbiellem thoughts ?

johnsimons:
@gbiellem I am stuck on this one, not sure if we need to modify everything or not, have started a PR in #740

gbiellem:
No problem @johnsimons  I'll take a look

johnsimons:
Fixed in #740

PR #740, opened 2016-06-03
Add support for installing instance as "NT AUTHORITY\LocalService"

johnsimons:
Connects to #715


gbiellem:
@johnsimons  I've fixed up the LocalService option so it works regardless of the locale of the OS.  Tested on English Windows 10 and German Windows 8.1

johnsimons:
So is there anything left to do @gbiellem ?

src/ServiceControlInstaller.Engine/Services/WindowsServiceController.cs
```diff
@@ -107,15 +107,19 @@ public void ChangeAccountDetails(string serviceAccount, string servicePassword)
                     var message = (wmiReturnCode < Win32ChangeErrorMessages.Length)
                         ? $"Failed to change service credentials on service {ServiceName} - {Win32ChangeErrorMessages[wmiReturnCode]}"
                         : "An unknown error occurred";
+
+                    if (wmiReturnCode == 22)
+                    {
+                        message += $"( AccountName {userAccount.QualifiedName} converted to {ConvertAccountNameToServiceAccount(userAccount.QualifiedName)})";
```
johnsimons: @gbiellem is this a debug statement?

johnsimons:
looks good, but one comment added

gbiellem:
I left that in on purpose - on the off chance we do hit a locale issue that this fix cover

PR #741, opened 2016-06-03
Added sql dtc option

johnsimons:
image

### DIFF
--- a/src/ServiceControlInstaller.Engine/Instances/Transports.cs
+++ b/src/ServiceControlInstaller.Engine/Instances/Transports.cs
@@ -18,7 +18,7 @@
                 Name = "AzureServiceBus",
                 TypeName = "NServiceBus.AzureServiceBusTransport, NServiceBus.Azure.Transports.WindowsAzureServiceBus",
                 MatchOn = "NServiceBus.AzureServiceBus",
-                SampleConnectionString = @"Endpoint=sb://[namespace].servicebus.windows.net; SharedSecretIssuer=<owner>;SharedSecretValue=<someSecret>"
+                SampleConnectionString = "Endpoint=sb://[namespace].servicebus.windows.net; SharedSecretIssuer=<owner>;SharedSecretValue=<someSecret>"
 
             },
             new TransportInfo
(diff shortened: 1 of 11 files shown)


johnsimons:
@mikeminutillo this is it

src/ServiceControlInstaller.Engine/Instances/Transports.cs
```diff
@@ -17,37 +17,44 @@ public static TransportInfo FindByName(string name)
             {
                 Name = "AzureServiceBus",
                 TypeName = "NServiceBus.AzureServiceBusTransport, NServiceBus.Azure.Transports.WindowsAzureServiceBus",
-                MatchOn = "NServiceBus.AzureServiceBus",
+                MatchOn = "NServiceBus.AzureServiceBusTransport",
```
gbiellem: @johnsimons Changing the match value will break upgrades from older ServiceControl that are still on NSB v4.  The V4 transports didn't have "Transport" at the end of the string.

src/ServiceControlInstaller.Engine/Instances/Transports.cs
```diff
@@ -17,37 +17,44 @@ public static TransportInfo FindByName(string name)
...
                 Name = "SQLServer",
                 TypeName = "NServiceBus.SqlServerTransport, NServiceBus.Transports.SQLServer",
-                MatchOn = "NServiceBus.SqlServer",
+                MatchOn = "NServiceBus.SqlServerTransport",
+                SampleConnectionString = @"Data Source=<SQLInstance>;Initial Catalog=nservicebus;Integrated Security=True"
+            },
+            new TransportInfo
+            {
+                Name = "SQLServer with DTC",
+                TypeName = "NServiceBus.SqlServerWithDTCTransport, NServiceBus.Transports.SqlServerWithDTC",
```
gbiellem: @johnsimons 

The TypeName is what ends up in the app config in ""ServiceControl/TransportType" so it that what you intents here?

johnsimons: yes

src/ServiceControlInstaller.Engine/Instances/Transports.cs
johnsimons: In that case I may need to rename my custom transport to not conflict

src/ServiceControlInstaller.Engine/Instances/Transports.cs
```diff
@@ -17,30 +17,30 @@ public static TransportInfo FindByName(string name)
             {
                 Name = "AzureServiceBus",
                 TypeName = "NServiceBus.AzureServiceBusTransport, NServiceBus.Azure.Transports.WindowsAzureServiceBus",
-                MatchOn = "NServiceBus.AzureServiceBusTransport",
+                MatchOn = "NServiceBus.AzureServiceBus,",
```
gbiellem: @johnsimons I think you made a typo in all the MatchOn values you put back - each one has a comma at the end

johnsimons:
Replaced by https://github.com/Particular/ServiceControl/pull/757

Issue #676, opened 2016-02-18
License dialog is not refreshed when importing a new license file


HEskandari:
Here are the steps to reproduce the issue. I have the following license entries in my Registry:

HKLM -> Software -> ParticularSoftware -> License (using Particular's internal license file)
HKLM -> Software -> ParticularSoftware -> TrialStart = 2015-09-30
HKCU -> Software -> ParticularSoftware -> License (again, using Particular's internal license file)
HKCU -> Software -> ParticularSoftware -> TrialStart = 2015-09-30

I launch the Management Utility app. Open the license dialog. I can see that the effective license is shown correctly.

I then update the license (using another valid license file) and import it. Here's what happens:

- There's no indication that the license is updated. (Related to #670)
- I checked the registry, the License key is actually updated

But now when I close the dialog and re-open it, although the content of the license registry key has changed, in the dialog the old license information is shown.

gbiellem:
Thanks @HEskandari  I'll address this in the release after 1.10

johnsimons:
@gbiellem is this fixed ?

gbiellem:
No but since you have me messing with licensing code I'll tackle it

johnsimons:
@gbiellem I fixed this, but I am sure you won't be happy with the fix!

PR #744, opened 2016-06-08
Making license logic simpler

### LINKED ISSUE #669: SCMU license dialog doesn't display the edition
SCMU license dialog  displays the "License Type" from the underlying licensing component.  This is either Trail or Standard which causes some confusion since the expectation is this will show the license edition.   

e.g.  I paid for an enterprise license. Why does the license dialog say "Standard"

Relates to https://github.com/Particular/Operations.Licensing/issues/15

johnsimons:
Only checking license in HKLM
Added edition, fixes #669
Better fix for #676
Updated license package


johnsimons:
@gbiellem please review

PR #746, opened 2016-06-13
[WIP] DTC Support

mikeminutillo:
Adds the ability to switch on DTC support for ServiceControl

### DIFF
--- a/src/ServiceControl/Bootstrapper.cs
+++ b/src/ServiceControl/Bootstrapper.cs
@@ -71,10 +71,19 @@ namespace Particular.ServiceControl
 
             configuration.UseSerialization<JsonSerializer>();
 
-            configuration.Transactions()
-                .DisableDistributedTransactions()
-                .DoNotWrapHandlersExecutionInATransactionScope();
-            
+            if (Settings.EnableDtc)
+            {
+                configuration.Transactions()
+                    .EnableDistributedTransactions()
+                    .WrapHandlersExecutionInATransactionScope();
+            }
+            else
+            {
+                configuration.Transactions()
+                    .DisableDistributedTransactions()
+                    .DoNotWrapHandlersExecutionInATransactionScope();
+            }
+
             configuration.ScaleOut().UseSingleBrokerQueue();
             
             var transportType = DetermineTransportType();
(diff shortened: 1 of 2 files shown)


src/ServiceControl/Bootstrapper.cs
```diff
@@ -71,10 +71,19 @@ public Bootstrapper(ServiceBase host = null, HostArguments hostArguments = null,
...
+                    .EnableDistributedTransactions()
+                    .WrapHandlersExecutionInATransactionScope();
+            }
+            else
+            {
+                configuration.Transactions()
+                    .DisableDistributedTransactions()
+                    .DoNotWrapHandlersExecutionInATransactionScope();
+            }
+
```
mauroservienti: Is this the default in Core?

```
configuration.Transactions()
   .EnableDistributedTransactions()
   .WrapHandlersExecutionInATransactionScope();
```

Is it enough to disable the DTC if `Settings.EnableDtc` is `false`?

mikeminutillo: I went for being explicit rather than relying on defaults. Lines 75 and 76 were the prior behavior which is the default.

MarcinHoppe:
The only thing that I would add is to log on `INFO` level if DTC was on or off on startup. It's something that I lacked in testing and I imagine would be useful in production.

johnsimons:
I would much prefer this to be shipped as an extra dll that is only delivered when user selects the sql transport.
@mikeminutillo are you ok with me changing it ?

gbiellem:
@johnsimons I assume you want the extra DLL  because you are worried it would be enabled by mistake for another transport.  If that the case couldn't we just check the transport type prior to enabling DTC.  And obviously log the misconfiguration.

johnsimons:
> I assume you want the extra DLL because you are worried it would be enabled by mistake for another transport.

Yes


> couldn't we just check the transport type prior to enabling DTC

not easily without using magical strings

johnsimons:
I'll modify it

johnsimons:
My take #757

johnsimons:
Replaced by https://github.com/Particular/ServiceControl/pull/757

Issue #735, opened 2016-05-23
Add option to run in maintenance mode to the Management Utility


johnsimons:
![image](https://cloud.githubusercontent.com/assets/122651/15458931/104877b2-20e3-11e6-8547-a20ee694450e.png)

This is ideal for compacting the database that in Raven v3 is part of the studio

johnsimons:
@sergioc  where do you see this option being displayed ?

sergioc:
Can you explain what the option does exactly?

johnsimons:
it would stop the instance if it is running (it probably needs to warn about that), it would then start ServiceControl in maintenance mode (we probably would display the url to raven), this would pop up the console app and then is up to the user to exit the console.

johnsimons:
Majority of time a user would use this mode for compacting the database, see http://docs.particular.net/servicecontrol/db-compaction#using-servicecontrol-in-maintenance-mode

sergioc:
@johnsimons this is small task for me, can you add it to the current SC budget?

johnsimons:
@sergioc we can do that but I actually see this be a complement to #728.

gbiellem:
@johnsimons At the moment we get users to run in maint mode interactively because that's the easiest way to add a command line switch. If we are going to add it to the UI though there is no need to pop up the console - we could just toggle the mode and restart the service.  Where we store the switch we can discuss later.  Thoughts?

johnsimons:
@gbiellem  Yep that works for me, and an easy solution to implement

Issue #638, opened 2016-02-08
RavenDB Silverlight management UI is a blocker


mauroservienti:
In some environments Silverlight is a no go, especially on production servers, and it is preventing customers to directly access the ServiceControl data stored in RavenDB.

johnsimons:
@mauroservienti technically it should be very rare for users to have to access the raw data, not saying that it won't happen but very rare IMO.

Not sure what to do with this @mauroservienti, as you know the only way out of this is by upgrading Raven, which u had a go and decided not to go ahead, so what do you think we should do with this issue ?

mauroservienti:
@johnsimons it's a good question I have no answer for :(

> technically it should be very rare for users to have to access the raw data, not saying that it won't happen but very rare IMO.

I agree, the issue is that in a lockdown environment in those very rare scenarios there is nothing you can do.
And yes the only option we have is to upgrade to RavenDB 3.0, when @weralabaj and I tested it the effort was not worth the result. 

Is the upgrade to 3.0 required if and/or when we move ServiceControl from NServiceBus V5 to NServiceBus V6?

johnsimons:
@Particular/servicecontrol-maintainers not much we can do here till we move either to raven v3 or something else, I think we can close this one ?

gbiellem:
Agreed. No matter what the the vNext DB is this will be solved so no point keeping it around

mauroservienti:
Another Desk case where Silverlight is a no go, is upgrading to RavenDB 3 an option in the short term? It will remove the problem entirely.

gbiellem:
@mauroservienti i'd be :+1: for doing that  - The hard part would be getting it prioritized at the moment :(

johnsimons:
Upgrading to Raven v3 is probably the easy part, the bigger issue is the migration of the data and the massive inconvenience it puts on users, no monitoring.

gbiellem:
@johnsimons  but do we need to migrate the data? This has never been clear to me.
Is there anything stopping us from continuing to use the system DB.   
When we start splitting capabilites we can make use of multiple DBs and not put the new data in system.

johnsimons:
@gbiellem it is my understanding that v3 does not let us write to system db, so yes we have to migrate it

gbiellem:
[...]

PR #747, opened 2016-06-14
Added ability to enter MaintenanceMode

johnsimons:
This PR addresses #735

Here is how this is going to work:
- User stops the endpoint instance;
- They then can click on "Start in Maintenance" button;
- Once they done the "Stop" the instance, and this will revert back to normal mode

Here is what is looking like so far:
image
image


johnsimons:
@sergioc can we get an icon for the button "Start in Maintenance", something like the usual screwdriver cross ?

We also need the panel that says "This instance is in main...." prettied up :smile:

johnsimons:
@gbiellem regarding this PR and #625, this won't address #625, to fix #625 we need to run the instance with the admin account that is running the installer right ?

gbiellem:
I think we should address #625 before this goes out. 
The only reason an admin account is required for #625 is that we don't set a urlacl for /storage.  There is also some weirdness in the SC Raven Init icode that limits the.binding to local host for wild cards but not fully qualified domain names.  I think we need to get the user to be explicit about that binding rather than trying to determine it.  This would allow the use of Raven studio from a 2nd PC - that would address scenarios where silver light is not an option on the server as highlighted in #638

sergioc:
- Is there a confirmation prompt that displays when switching to maintenance mode?

johnsimons:
@sergioc nope

johnsimons:
> The only reason an admin account is required for #625 is that we don't set a urlacl for /storage. There is also some weirdness in the SC Raven Init icode that limits the.binding to local host for wild cards but not fully qualified domain names. I think we need to get the user to be explicit about that binding rather than trying to determine it. This would allow the use of Raven studio from a 2nd PC - that would address scenarios where silver light is not an option on the server as highlighted in #638

@gbiellem can we discuss this a bit more, what exactly do we need to do ?

johnsimons:
@gbiellem is [this](https://github.com/Particular/ServiceControl/blob/dd3db2a0ea66c6b35d1f44c43e7a685da62cbe3c/src/ServiceControl/Infrastructure/RavenDB/RavenBootstrapper.cs#L62-L64) what u talking about ?

sergioc:
@johnsimons since the instance needs to be stopped before entering maintenance mode, shouldn't we have a prompt to allow the user confirm, just like we do for upgrades?

johnsimons:
Sure we can do that

On Fri, 17 Jun 2016 at 18:27 Sergio <notifications@github.com> wrote:

> @johnsimons <https://github.com/johnsimons> since the instance needs to
> be stopped before entering maintenance mode, shouldn't we have a prompt to
> allow the user confirm, just like we do for upgrades?
>
> —
> You are receiving this because you were mentioned.
>
>
> Reply to this email directly, view it on GitHub
> <https://github.com/Particular/ServiceControl/pull/747#issuecomment-226711204>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAHfG69g8_u4piNcad9w9ebKyqh0Haurks5qMlp5gaJpZM4I1BWu>
> .
>
-- 

*John Simons*
*Engineer, Particular Software*

Email*:*  john.simons@particular.net
Web:    www.particular.net

sergioc:
> Sure we can do that

Excellent.

sergioc:
Here's how SCMU could look like while in Maintenance Mode:

screenshot 2016-06-17 12 17 16

johnsimons:
@sergioc can you give us the icon ?

sergioc:
Icons: https://www.dropbox.com/sh/jvg7faw3i9a7274/AAA1QwX_Ldbtpjozjk-yABZ4a?dl=0

johnsimons:
@gbiellem this is now ready, can u review it

Also, I think the url for raven is better in your PR ;)

gbiellem:
Code looks good. Found the maint mode icon being greyed out a little weird initially but once I stopped the service it made sense

johnsimons:
Awesome

gbiellem:
@johnsimons Above you told @sergioc  that you'd add a confirmation dialog.  I think that would make sense on starting maintenance mode - it might be a good place to put some explaination of what this actually is and what it's for

gbiellem:
Actually the confirmation you promised @sergioc was for stopping the instance so it's not exactly the 
same

johnsimons:
@gbiellem 
yeah, the user has to stop the endpoint before they can start in maintenance mode.

but I agree, i think we need a bit more explanation what this is for, we can talk more about it tomorrow

sergioc:
> yeah, the user has to stop the endpoint before they can start in maintenance mode.

That's why a confirmation prompt for starting is also useful. If the button is greyed out, how will the user understand that the instance has to be stopped first in order to be able to click that button?

Adding a confirmation prompt to start maintenance mode allows the button to be enabled. The prompt will then explain the instance has to be stopped first, which will be done automatically before entering maintenance mode once the user confirms.

PR #751, opened 2016-06-16
Added option to reuse existing Paths

### LINKED ISSUE #732: Reinstalling an instance wont work on the same data files
I tried to reinstall an instance. I first removed the instance via the management tool. It asked if I wanted to remove the data files and selected to keep it. I then added an instance again with the same folders and got a message that it could not add it because there were already files present.

Instead of blocking it should ask the question if you are sure that you want to use those folders and reuse existing database files.

johnsimons:
Fixes #732


johnsimons:
@gbiellem I think I am done here.
Do you want to have a look at the powershell

gbiellem:
Will do

gbiellem:
@johnsimons  Done,

PR #625, opened 2016-06-21
Adding ExposeRavenDB can cause service to fail on startup

gbiellem:
### Symptom 

If the service account is not a local administrator or localsystem enabling the ExposeRavenDB setting can cause the server to fail on startup.  This is because A URLACL is required to allow the Raven Management Studio to be exposed on the `/storage` virtual directory.  Raven will create this automatically if the service account has the privileges.

### Who's affected 

Anyone using a low privilege service account that decides to expose the raven studio.  This setup is typically used by client using the SQL transport and integrated authentication so they are more likely to be affected.

### Workaround 

Until this is fixed the workaround is to manually create the required URLACL as covered in [docs ](http://docs.particular.net/servicecontrol/troubleshooting#unable-to-start-service-after-exposing-ravendb)


gbiellem:
@johnsimons  review please

johnsimons:
@gbiellem this one should be good now

PR #755, opened 2016-06-22
Introducing Advance Options

johnsimons:
Have moved maintenance mode and remove under new advance options


johnsimons:
@mikeminutillo @gbiellem 
As discussed, here it is, please have a look.

What is left:
- [ ] The information text needs to be written;
- [ ] The link to doco needs to be specific, @mikeminutillo I know the doco people don't like this, but IMO we really need it to provide a complete solution to our users;
- [ ] The screen design needs to be reviewed by @sergioc

sergioc:
Why do we need this Advanced Options screen?

mikeminutillo:
@sergioc we discussed separating the Maintenance Mode from the other actions for two reasons:
1. We wanted to add some text to it describing what it does so that users actually have some context before they click it. It's not intuitive what "Maintenance Mode" means and it's not clear what will happen if you click it. The tooltip space we provide is not conducive to lengthy explanations. 
2. It should be used far less commonly than Start/Stop and Edit

We tried adding the Maintenance Mode to the bottom of the section for each item and it didn't look right. It also felt strange to have actions for the same instance on the same page but in different locations. So we settled on creating a separate screen for this "advanced option". Once we had that, it became clear that the Delete button satisfied both of the criteria above and so we moved it to the advanced options screen. 

Does that make sense?

sergioc:
@mikeminutillo it doesn't. It doesn't make sense because SCMU as a whole is already infrequently used. You're adding infrequently used functionality inside an infrequently used tool.

This requires all functions to be very easy to discover, so they should be accessible directly in the main screen. Adding the "Start in maintenance mode" button is only one more button and barely adds clutter. But it makes it easier to reach the option in a tool that users won't easily memorize.

I asked @gbiellem about what would lead users to use maintenance mode:

- reached relevant doco page and concluded from there it's needed
- support contact

At this point, users will understand what maintenance mode is for. Also, you can place the explanation text inside the confirmation prompt. By explaining what maintenance mode does and requiring explicit confirmation, the confirmation prompt helps ensure the user won't do anything silly by impulse.

gbiellem:
@sergioc 

If maintenance mode was the only option that we were considering adding I'd agree with you but it isn't.  We are also looking at adding options for DB backup and restore etc.  It would seem logical for those settings/functions to be accessible through the advanced screen. ( https://github.com/Particular/ServiceControl/pull/760. )

I seem to remember In your original designs for SCMU you had a button on the far left with an ellipsis that led off to additional actions.  We dropped it because at the time because we didn't have anything to put under there.  So I think we are actually returning to your original idea except we now have a spanner icon rather than ellipsis

gbiellem:
Also keep in mind that the SCMU needs to support the current version and previous versions prior to them being upgraded.  I think having different buttons available to the versions listed would be very confusing.

gbiellem:
@HEskandari  Thanks for the refresh fix.

gbiellem:
@mikeminutillo this is now ready for a review. Once done I'd say we are go for a release

sergioc:
> We are also looking at adding options for DB backup and restore etc. 

@gbiellem makes more sense then. 

Last remark: I'd only use red for the delete instance button, though. Deleting an instance seems considerably more dramatic then entering maintenance mode.

gbiellem:
@sergioc cool 

I'll change the color of the start in maint button to the standard one

gbiellem:
@mikeminutillo can you review and merge if you are happy

src/ServiceControl.Config/Commands/StartServiceInMaintenanceModeCommand.cs
```diff
@@ -0,0 +1,67 @@
...
+    using ServiceControl.Config.Events;
+    using ServiceControl.Config.Framework;
+    using ServiceControl.Config.Framework.Commands;
+    using ServiceControl.Config.UI.AdvanceOptions;
+    using ServiceControlInstaller.Engine.ReportCard;
+
+    class StartServiceInMaintenanceModeCommand : AwaitableAbstractCommand<AdvanceOptionsViewModel>
+    {
+        private readonly IWindowManagerEx windowManager;
+        private readonly IEventAggregator eventAggregator;
```
mikeminutillo: Does the installer use a different R# settings file than the rest of the Platform? I thought we didn't use `readonly` in most places.

src/ServiceControl.Config/Commands/StartServiceInMaintenanceModeCommand.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        public StartServiceInMaintenanceModeCommand(IWindowManagerEx windowManager, IEventAggregator eventAggregator) : base(model => !(model.IsRunning && model.InMaintenanceMode))
```
mikeminutillo: Does this mean that I can start in Maintenance Mode if I'm already started?

mikeminutillo: Ah. Yes I can. I can see that below we ask for confirmation in that case.

src/ServiceControl.Config/Commands/StartServiceInMaintenanceModeCommand.cs
```diff
@@ -0,0 +1,67 @@
...
+        {
+            this.windowManager = windowManager;
+            this.eventAggregator = eventAggregator;
+        }
+
+        public override async Task ExecuteAsync(AdvanceOptionsViewModel model)
+        {
+            model.ServiceControlInstance.Service.Refresh();
```
mikeminutillo: Why is this needed? Just in case someone has manually started the service since opening this screen?

mikeminutillo: This feels like a violation of https://en.wikipedia.org/wiki/Law_of_Demeter but I'm not sure how else you'd do it.

src/ServiceControl.Config/Commands/StartServiceInMaintenanceModeCommand.cs
```diff
@@ -0,0 +1,67 @@
...
+                    if (!started)
+                    {
+                        var reportCard = new ReportCard();
+                        reportCard.Warnings.Add("Failed to start the service");
+                        reportCard.SetStatus();
+                        windowManager.ShowActionReport(reportCard, "ISSUES STARTING INSTANCE IN MAINTENANCE MODE", "There were some warnings when attempting to start instance in Maintenance Mode:", String.Empty);
+                    }
+                }
+            }
+            eventAggregator.PublishOnUIThread(new RefreshInstances());
```
mikeminutillo: This gets run even if we don't start in Maintenance Mode. As we didn't change anything, is that necessary?

src/ServiceControl.Config/Commands/StartServiceInMaintenanceModeCommand.cs
```diff
@@ -0,0 +1,67 @@
...
+        public override async Task ExecuteAsync(AdvanceOptionsViewModel model)
+        {
+            model.ServiceControlInstance.Service.Refresh();
+
+            var confirm = model.IsStopped ||
+                          windowManager.ShowYesNoDialog("STOP INSTANCE AND START IN MAINTENANCE MODE", $"{model.Name} needs to be stopped in order to start in Maintenance Mode.", "Do you want to proceed?", "Yes I want to proceed", "No");
+
+            if (confirm)
+            {
+                using (var progress = model.GetProgressObject(String.Empty))
```
mikeminutillo: Why String.Empty? Should this be an optional parameter?

src/ServiceControl.Config/Commands/StartServiceInMaintenanceModeCommand.cs
```diff
@@ -0,0 +1,67 @@
...
+                {
+                    var stopped = await model.StopService(progress);
+
+                    if (!stopped)
+                    {
+                        eventAggregator.PublishOnUIThread(new RefreshInstances());
+                        var reportCard = new ReportCard();
+                        reportCard.Errors.Add("Failed to stop the service");
+                        reportCard.SetStatus();
+                        windowManager.ShowActionReport(reportCard, "ISSUES STARTING INSTANCE IN MAINTENANCE MODE", "There were some errors when attempting to start instance in Maintenance Mode:", String.Empty);
```
mikeminutillo: Could this String.Empty be an optional parameter?

src/ServiceControl.Config/Commands/StartServiceInMaintenanceModeCommand.cs
```diff
@@ -0,0 +1,67 @@
...
+
+                        return;
+                    }
+
+                    progress.Report(new ProgressDetails("Starting Service in Maintenance Mode"));
+                    var started = await Task.Run(() =>
```
mikeminutillo: the model instance has a `StopService()` method but not a `StartInMaintenanceMode()` method? These two things feel like they should be at the same level of abstraction.

src/ServiceControl.Config/Commands/UpgradeInstanceCommand.cs
```diff
@@ -113,25 +113,40 @@ public UpgradeInstanceCommand(IWindowManagerEx windowManager, IEventAggregator e
...
                     if (reportCard.HasErrors || reportCard.HasWarnings)
                     {
                         windowManager.ShowActionReport(reportCard, "ISSUES UPGRADING INSTANCE", "Could not upgrade instance because of the following errors:", "There were some warnings while upgrading  the instance:");
                     }
                     else
                     {
-                        if (isRunning) await instanceViewModel.StartService();
+                        if (restartAgain)
+                        {
+                            await model.StartService(progress);
```
mikeminutillo: Could this step fail? And if so is that handled within the `StartService` call?

src/ServiceControl.Config/UI/AdvanceOptions/AdvanceOptionsView.xaml
```diff
@@ -0,0 +1,174 @@
...
+                >
+                Enter database maintenance mode to access to RavenDB Management Studio.  While in this mode all message processing is disabled and the REST API is unavailable. This will prevent ServicePulse and ServiceInsight connecting to this instance.
+                </TextBlock>
+                
+                <TextBlock FontSize="12px" 
+                           TextWrapping="Wrap" 
+                           Visibility="{Binding InMaintenanceMode, Converter={StaticResource boolToVis}}"
+                           Margin="0,0,0,20"
+                >
+                This instance is in database maintenance mode. All message processing is disabled and the REST API is unavailable.<LineBreak />Launch <Hyperlink Command="{Binding OpenUrl}" CommandParameter="{Binding StorageUrl}">
```
mikeminutillo: Is it worth mentioning again that ServicePulse and ServiceInsight cannot connect to this instance in this state?

gbiellem: I'll add that back in

src/ServiceControl.Config/Commands/StartServiceInMaintenanceModeCommand.cs
gbiellem: I believe it does.  We can standardize but lets do that as a separate PR / refactoring exercise

src/ServiceControl.Config/Commands/StartServiceInMaintenanceModeCommand.cs
gbiellem: Agreed but as you've noted there isn't another way. (AFAIK) to handle this.

src/ServiceControl.Config/Commands/StartServiceInMaintenanceModeCommand.cs
gbiellem: Agreed - I'll move it.

src/ServiceControl.Config/Commands/StartServiceInMaintenanceModeCommand.cs
gbiellem: It should say "Stopping Service" not String.Empty.

src/ServiceControl.Config/Commands/StartServiceInMaintenanceModeCommand.cs
gbiellem: Agreed

gbiellem:
@mikeminutillo I've acted on your feedback. Can you give this another review.

PR #758, opened 2016-06-23
[WIP] Message redirects

boblangley:
Connects to https://github.com/Particular/V6Launch/issues/77

Provides a means to return failed messages to a queue other than the original processing queue.
Adds Nancy API module to create, delete, and query existing redirects.
Applies redirect when failed queue address has a redirect created.


boblangley:
@Particular/servicecontrol-maintainers Would you mind doing an interim code review while I finish up test cases?

src/ServiceControl/MessageRedirects/Api/MessageRedirectsModule.cs
```diff
@@ -0,0 +1,144 @@
...
+                if (string.IsNullOrWhiteSpace(message.FromPhysicalAddress) || string.IsNullOrWhiteSpace(message.ToPhysicalAddress))
+                {
+                    return HttpStatusCode.BadRequest;
+                }
+
+                message.MessageRedirectId = DeterministicGuid.MakeId(message.FromPhysicalAddress, message.ToPhysicalAddress);
+
+                using (var session = Store.OpenSession())
+                {
+                    var redirect = session.Load<MessageRedirect>(MessageRedirect.GetDocumentIdFromMessageRedirectId(message.MessageRedirectId));
```
mikeminutillo: Is this check strictly necessary? If you try to create a redirect and one already exists that does exactly the same thing and they have no identity beyond the two endpoints they connect, what does it matter whether we actually created a new one or just return you the existing one?

src/ServiceControl/MessageRedirects/Api/MessageRedirectsModule.cs
```diff
@@ -0,0 +1,144 @@
...
+
+                    if (redirect != null)
+                    {
+                        return Negotiate.WithReasonPhrase("Duplicate").WithModel(redirect).WithStatusCode(HttpStatusCode.Conflict);
+                    }
+
+                    var dependents = session.Query<MessageRedirect>().Where(r => r.ToPhysicalAddress == message.FromPhysicalAddress).ToList();
+
+                    if (dependents.Any())
```
mikeminutillo: What if there is a dependency the other way? We are creating a new redirect that goes to a queue that is itself redirected?

src/ServiceControl/MessageRedirects/Api/MessageRedirectsModule.cs
```diff
@@ -0,0 +1,144 @@
...
+                }
+
+                message.MessageRedirectId = DeterministicGuid.MakeId(message.FromPhysicalAddress, message.ToPhysicalAddress);
+
+                Bus.SendLocal(new RemoveMessageRedirect
+                {
+                    MessageRedirectId = oldMessageRedirectId
+                });
+
+                Bus.SendLocal(message);
```
mikeminutillo: Will this work OK if these two messages are processed out of order?

What happens if I try to update a redirect to cause a dependency chain. That check is missing from here.

src/ServiceControl/MessageRedirects/Api/MessageRedirectsModule.cs
```diff
@@ -0,0 +1,144 @@
...
+
+                return Negotiate.WithStatusCode(HttpStatusCode.Accepted);
+            };
+
+            Delete["/redirects/{messageredirectid}"] = parameters =>
+            {
+
+                Guid messageRedirectId = parameters.messageredirectid;
```
mikeminutillo: We don't check if the input param is null here as we do above? I guess if someone sends us garbage to delete and it's not there then that's OK but we're (Pulse) likely to be the ones that suffer.

src/ServiceControl/MessageRedirects/Api/MessageRedirectsModule.cs
boblangley: Good catch

boblangley: Actually I thought more about this and still think this is the only check that is needed. The opposite scenario is excluded from this validation is where multiple queues are redirected to the same queue, which should be supported.

src/ServiceControl/MessageRedirects/Api/MessageRedirectsModule.cs
boblangley: The original plan was not to support Update, I think this wasn't wise. We need there to be 1 state for redirects, not 2 different ones. So we will support updates to prevent 2 redirects from the same from queue being available at the same time.

I pulled out the validation routine so it can be used for the update as well.

src/ServiceControl/MessageRedirects/Api/MessageRedirectsModule.cs
boblangley: This should be caught at develop time, I do not think it is unreasonable to assume we will test ServicePulse to make sure the correct URL is being called.

src/ServiceControl/Infrastructure/Extensions/NegotiatorExtensions.cs
```diff
@@ -123,18 +123,18 @@ public static Negotiator WithEtagAndLastModified(this Negotiator negotiator, Que
 
         public static Negotiator WithEtagAndLastModified(this Negotiator negotiator, RavenQueryStatistics stats)
         {
-            var etag = stats.IndexEtag;
-            var responseLastModified = stats.IndexTimestamp;
+            var etag = stats?.IndexEtag;
```
johnsimons: I can't figure out why we need to make this nullable?

src/ServiceControl/Infrastructure/Extensions/NegotiatorExtensions.cs
```diff
@@ -123,18 +123,18 @@ public static Negotiator WithEtagAndLastModified(this Negotiator negotiator, Que
...
+            var responseLastModified = stats?.IndexTimestamp;
```
johnsimons: I can't figure out why we need to make this nullable?

src/ServiceControl/Infrastructure/Extensions/NegotiatorExtensions.cs
```diff
@@ -123,18 +123,18 @@ public static Negotiator WithEtagAndLastModified(this Negotiator negotiator, Que
...
 
             return WithEtagAndLastModified(negotiator, etag, responseLastModified);
         }
 
-        public static Negotiator WithEtagAndLastModified(this Negotiator negotiator, Etag etag, DateTime responseLastModified)
+        public static Negotiator WithEtagAndLastModified(this Negotiator negotiator, Etag etag, DateTime? responseLastModified)
         {
-            var currentEtag = etag.ToString();
+            var currentEtag = etag?.ToString();
```
johnsimons: I can't figure out why we need to make this nullable?

src/ServiceControl/Infrastructure/Extensions/NegotiatorExtensions.cs
```diff
@@ -123,18 +123,18 @@ public static Negotiator WithEtagAndLastModified(this Negotiator negotiator, Que
...
             return negotiator
                 .WithHeader("ETag", currentEtag)
-                .WithHeader("Last-Modified", responseLastModified.ToString("R"));
+                .WithHeader("Last-Modified", responseLastModified?.ToString("R"));
```
johnsimons: I can't figure out why we need to make this nullable?

src/ServiceControl/MessageRedirects/Api/MessageRedirectsModule.cs
```diff
@@ -0,0 +1,169 @@
...
+                }
+
+                Bus.SendLocal(new CreateMessageRedirect
+                {
+                    FromPhysicalAddress = request.fromphysicaladdress,
+                    ToPhysicalAddress = request.tophysicaladdress,
+                    MessageRedirectId = DeterministicGuid.MakeId(request.fromphysicaladdress)
+                });
+
+                return HttpStatusCode.Created;
```
johnsimons: This should be `HttpStatusCode.Accepted`

src/ServiceControl/Contracts/MessageRedirects/MessageRedirectCreated.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace ServiceControl.Contracts.MessageRedirects
+{
+    using System;
+
+    public class MessageRedirectCreated
```
johnsimons: missing `IEvent`

src/ServiceControl/Contracts/MessageRedirects/MessageRedirectRemoved.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace ServiceControl.Contracts.MessageRedirects
+{
+    using System;
+
+    public class MessageRedirectRemoved
```
johnsimons: missing `IEvent`

src/ServiceControl/MessageRedirects/Api/MessageRedirectsModule.cs
```diff
@@ -0,0 +1,169 @@
...
+
+                    return Negotiate
+                        .WithModel(queryResult)
+                        .WithPagingLinksAndTotalCount(stats, Request)
+                        .WithEtagAndLastModified(stats);
+                }
+            };
+        }
+
+        private Negotiator ValidateRedirectRequest(MessageRedirectRequest request, bool checkForExisting = true)
```
johnsimons: I wonder if there is much point doing this validation here.
The data could have changed once we actually doing the update in Raven so maybe just accept the data and report an error later on via a `EventLogMappingDefinition<MessageRedirectCreated>` and `EventLogMappingDefinition<MessageRedirectChanged>`

src/ServiceControl/MessageRedirects/MessageRedirect.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+
+namespace ServiceControl.MessageRedirects
+{
+    public class MessageRedirect
+    {
+        private const string DocumentIdNamespace = "redirects/";
+
+        public string Id { get; set; }
```
johnsimons: Why is this not a `Guid` ?

johnsimons:
We are missing `EventLogMappingDefinition`s for `MessageRedirectCreated`, `MessageRedirectChanged`, `MessageRedirectDeleted`

johnsimons:
I wonder if we should store redirect documents as:
`redirects/[FromPhysicalAddressHash]/[ToPhysicalAddressHash]`
These would allow us to use the `LoadStartsWith` raven api, see https://ravendb.net/docs/article-page/2.5/csharp/client-api/basic-operations/loading-editing-existing-document#load-by-prefix-and-pattern

The advantage is that we have a multi key (FromPhysicalAddress and ToPhysicalAddress) id, which is what I think we need, thoughts ?

src/ServiceControl/MessageRedirects/Handlers/RemoveMessageRedirectHandler.cs
```diff
@@ -0,0 +1,28 @@
...
+
+    public class RemoveMessageRedirectHandler : IHandleMessages<RemoveMessageRedirect>
+    {
+        public IDocumentSession Session { get; set; }
+
+        public IBus Bus { get; set; }
+
+        public void Handle(RemoveMessageRedirect message)
+        {
+            var messageRedirect = Session.Load<MessageRedirect>(MessageRedirect.GetDocumentIdFromMessageRedirectId(message.MessageRedirectId));
```
johnsimons: You can avoid the `Load` by using:
`Store.DatabaseCommands.Delete((MessageRedirect.GetDocumentIdFromMessageRedirectId(message.MessageRedirectId), typeof(MessageRedirect), false), null)`.
See DeleteCustomCheckHandler class

boblangley:
Replaced with #762

Issue #759, opened 2016-06-24
Backup raven database as part of updates


johnsimons:
As part of a user selecting to upgrade, we should prompt the user to backup the current database before starting the new version

johnsimons:
@gbiellem is this now done ?

boblangley:
@gbiellem this looks completed by #776, is that correct?

gbiellem:
@boblangley  That code is part of the failed Raven 3 update so it never got merged back to master.

The backup code produced worked two ways 

- For version running Raven 3 it called the standard Raven 3  backup API.
- For older SC instances that used Raven 2.5 embedded we had no api to work with so came up with a workaround that involve a one off replacement of the servicecontrol.exe with a cut down exe who's only purpose was to backup the system db.  We tested the drop in exe with version 1.0 through to the latest and what we thought would be the last Raven 2.5 version of SC at the time and it  work for all of them.

Of course now that we haven't moved off Raven 2.5 this approach is a little more fragile as there is a change we may produce a version with a later patch of Raven 2.5.  Then what happens? Do we maintain multiple versions of the  drop in replacement exe?

PR #760, opened 2016-06-24
Enabling incremental backups

johnsimons:
A quick spike to enable incremental backups.

This is quite easy to get going.

The next steps would be to as part of a new instance install ask the user the location for the backups and frequency.

This is related to #759

### DIFF
--- a/src/ServiceControl/Infrastructure/RavenDB/RavenBootstrapper.cs
+++ b/src/ServiceControl/Infrastructure/RavenDB/RavenBootstrapper.cs
@@ -10,10 +10,12 @@
     using NServiceBus.Persistence;
     using NServiceBus.Pipeline;
     using Particular.ServiceControl.Licensing;
+    using Raven.Abstractions.Data;
     using Raven.Abstractions.Extensions;
     using Raven.Client;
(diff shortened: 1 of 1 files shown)


johnsimons:
Adding a couple of extra questions to the install ui shouldn't take a long time, we then store those settings in the config file and we read them out of there.

Regarding turning this on for an existing database, I tested it and it does add a bit of pressure because I assume it is doing a full backup to start with.

johnsimons:

mauroservienti:
well enabling incremental backups is absolutely a nice addition, allows customers to save space and increase data security running backups more frequently. We should test a disaster recovery as well before implementing this.

gbiellem:
@mauroservienti 

>  allows customers to save space 

How does it save space?  Surely it's using more space.  

@johnsimons  

Are you proposing this as an alternative to the idea of offering the user the option to do a complete backup on upgrades?

mauroservienti:
If the only option is a full backup and you need to retain deletions you must preserve multiple full backups leading to more wasted space compared to the incremental solution. Isn't it?

johnsimons:
Will re-evaluate this when we go Raven v3

PR #761, opened 2016-07-07
Speed up our acceptance tests

johnsimons:
The purpose of this PR is to speed up our acceptance tests.

## Main changes
- Made NServiceBus a Owin middleware (This enables to test SC using http://dhickey.ie/2015/04/testing-owin-applications-with-httpclient-and-owinhttpmessagehandler/ which completely bypasses the tcp stack);
- All acceptance tests run now in a single AppDomain;
- Is now possible to configure ServiceControl instance entirely in code;
- Tunning all aspects of the acceptance tests framework to make them faster;
- Fixed issue with acceptance test framework, where the Done was being called before the Whens were done;
- Added concept of SessionId to the running of individual acceptance tests, so with this in place we drop any messages where the message sessionid does not match the current sessionid, this means we do not need to do purge of queues/topics at startup any longer;
- Tunned ASB transport to make it go fast :zap: 
- Disabling all plugins by default in Acceptance Tests, they are now explicitly enabled on a per test;
- Parallelized the internal subscriptions in ServiceControl;


johnsimons:
@gbiellem I think the next steps are to change the `AcceptanceTest` base class to include:
```c#
var app = new AppBuilder();
bootstrapper.startup.Configuration(app);
var appFunc = app.Build();
var handler = new OwinHttpMessageHandler(appFunc);

client = new HttpClient(handler);
```

johnsimons:
@Particular/servicecontrol-maintainers this is now ready to be merged, finally, can someone start reviewing ;)
I have documented in the PR description the bulk of the changes.
The speed improvements can be seen by going to https://builds.particular.net/project.html?projectId=ServiceControl&branch_ServiceControl=761

johnsimons:
I have squashed the commits into more distinct pieces of work, to hopefully make it simpler to review

johnsimons:
One very important thing left to do, fix the urlacls, before the acl was on `/api`, it now needs to be moved to `/`

src/ServiceControl.AcceptanceTesting/Support/EndpointBehavior.cs
```diff
@@ -2,8 +2,6 @@
 {
     using System;
     using System.Collections.Generic;
-    using System.Diagnostics;
-    using System.Threading;
 
     [Serializable]
     public class EndpointBehavior : MarshalByRefObject
```
mikeminutillo: Does this still need to be MarshalByRef?

src/ServiceControl.AcceptanceTesting/Support/ScenarioConfigSource.cs
```diff
@@ -24,9 +24,22 @@ public T GetConfiguration<T>() where T : class, new()
...
+
                 return configurationSection as T;
             }
 
+            if (type == typeof(TransportConfig))
+            {
+                return new TransportConfig
+                {
+                    MaxRetries = 1,
+                    MaximumConcurrencyLevel = 2
```
mikeminutillo: Why these numbers specifically? Should we have a comment or something here explaining?

src/ServiceControl.AcceptanceTests/AcceptanceTest.cs
```diff
@@ -306,110 +158,127 @@ static void Delete(string path)
...
+        }
 
-            if (response == null)
+        private async Task<T> Get<T>(string url) where T : class
+        {
+            if (!url.StartsWith("http://"))
             {
-                Thread.Sleep(1000);
-                return null;
+                url = $"http://localhost:{port}{url}";
```
mikeminutillo: What if url doesn't start with a `/`?

src/ServiceControl.AcceptanceTesting/Support/ScenarioConfigSource.cs
johnsimons: those are the defaults based on our scenarios

src/ServiceControl.AcceptanceTests/AcceptanceTest.cs
johnsimons: I guess the test will fail

src/ServiceControl.AcceptanceTesting/Support/EndpointBehavior.cs
johnsimons: no, I'll remove it

src/ServiceControl.AcceptanceTests/AcceptanceTest.cs
```diff
@@ -306,110 +158,127 @@ static void Delete(string path)
...
+            using (var stream = await response.Content.ReadAsStreamAsync())
             {
                 var serializer = JsonSerializer.Create(serializerSettings);
 
                 return serializer.Deserialize<T>(new JsonTextReader(new StreamReader(stream)));
             }
         }
 
-        protected bool TryGet<T>(string url, out T response, Predicate<T> condition = null) where T : class
+        protected bool TryGetMany<T>(string url, out List<T> response, Predicate<T> condition = null) where T : class
```
mikeminutillo: Should this one be async as well (to match the above)

mikeminutillo: Wait...probably not. await and out params don't mix well.

src/ServiceControl.AcceptanceTests/AcceptanceTest.cs
```diff
@@ -419,113 +288,358 @@ protected byte[] DownloadData(string url)
...
+                throw new InvalidOperationException($"Call failed: {(int) response.StatusCode} - {response.ReasonPhrase} - {body}");
             }
+        }
 
-            return true;
+        protected byte[] DownloadData(string url, HttpStatusCode successCode = HttpStatusCode.OK)
+        {
+            if (!url.StartsWith("http://"))
+            {
+                url = $"http://localhost:{port}/api{url}";
```
mikeminutillo: This one is inconsistent with the logic of all the other methods (it adds the /api). Should these be consistent? Or even pulled out as a private helper method?

src/ServiceControl.AcceptanceTests/AcceptanceTest.cs
```diff
@@ -419,113 +288,358 @@ protected byte[] DownloadData(string url)
...
+            }
+
+            protected override void Setup(FeatureConfigurationContext context)
+            {
+                context.Container.ConfigureComponent<IMessageMapper>(builder => new MessageMapperWrapper(), DependencyLifecycle.SingleInstance);
+            }
         }
 
-        public IScenarioWithEndpointBehavior<T> Define<T>(Func<T> contextFactory) where T : ScenarioContext, new()
+        private class MessageMapperWrapper : IMessageMapper
```
mikeminutillo: If we were to change the default message mapper in ServiceControl this would hide that change. I don't know that there's anything we can do about that :(

johnsimons: Yeah, this is just a hack because we are now in a single appdomain

PR #762, opened 2016-07-07
Pending retries message redirects

boblangley:
Connects to https://github.com/Particular/V6Launch/issues/77

### Improvements
This release adds support for creating failed message redirects to override the source failed message queue in the case of permanent decommissioning. Additionally this release adds support for listing failed messages that have been been returned for retry, but have not yet been seen again by ServiceControl (either failed again or successful via auditing). 

### Should You Upgrade Immediately
Users will only be required to update to support the version of ServicePulse that supports message redirects and pending retries as described above.


boblangley:
@WojcikMike I re-combined the 2 PRs into 1 to make it easier. Check the nancy module for message redirects, there are few changes to the status codes (mostly Accepted to OK) and perhaps some metadata (e-tags are no longer relevant).

boblangley:
I pushed updates making sure they do not cause dependencies. I also re-based, squashing the earlier changes into a single commit.

src/ServiceControl/MessageFailures/Api/QueueAddressModule.cs
```diff
@@ -0,0 +1,52 @@
...
+                        .Take(take)
+                        .OrderBy(q => q.PhysicalAddress)
+                        .ToArray());
+
+                    var queueAddresses = failedMessageQueues.Select(q => q.PhysicalAddress).ToArray();
+
+                    if (failedMessageQueues.Count < take)
+                    {
+                        failedMessageQueues.AddRange(session.Query<QueueAddress, QueueAddressIndex>()
+                            .Where(q => !queueAddresses.Contains(q.PhysicalAddress) && q.PhysicalAddress.Contains(search))
```
WojcikMike: This is throwing an exception while trying to call Contains. RavenDB is not supporting that method.

src/ServiceControl/MessageFailures/Api/FailedMessageViewIndex.cs
```diff
@@ -30,6 +31,7 @@ select new
                     message.Status,
                     TimeSent = (DateTime) processingAttemptsLast.MessageMetadata["TimeSent"],
                     ReceivingEndpointName = ((EndpointDetails) processingAttemptsLast.MessageMetadata["ReceivingEndpoint"]).Name,
+                    QueueAddress = processingAttemptsLast.FailureDetails.AddressOfFailingEndpoint.ToLowerInvariant(),
```
johnsimons: @boblangley why are we `ToLowerInvariant()` in an index ?
We should do that as part of the display not the index

src/ServiceControl/MessageRedirects/Api/MessageRedirectsModule.cs
```diff
@@ -0,0 +1,201 @@
...
+
+                    var toMessageRedirectId = DeterministicGuid.MakeId(request.tophysicaladdress.ToLowerInvariant());
+
+                    if (redirects[toMessageRedirectId] != null)
+                    {
+                        return HttpStatusCode.Conflict;
+                    }
+
+                    messageRedirect.ToPhysicalAddress = request.tophysicaladdress.ToLowerInvariant();
+                    messageRedirect.LastModifiedTicks = DateTime.UtcNow.Ticks;
```
johnsimons: Why do we need this ?

boblangley:
@Particular/servicecontrol-maintainers @WojcikMike  Please review PR:

To support pending retries listing added failed message queue to filtering.
- Failed messages now include the physical queue address property as `QueueAddress`
- **NEW** GET `/errors/queues/addresses` List failed message queue physical addresses. Supports paging.
- **NEW** GET `/errors/queues/addresses/search/{search}` List failed message queue physical addresses filtered by startswith search term. Supports paging.
- **UPDATED** GET `/errors` Add support for `queueaddress` List failed messages. Added support for query parameter to filter by failed message queue physical address.
- **NEW** POST `/errors/queues/{queueaddress}/retry` Retry failed messages by physical queue address by status. Default status when omitted is `RetryIssued`. `Archived` and `Resolved` status will be rejected.

To support message redirects:
- Redirects are a From and a To. From (toLower) is the unique identifier which is sometimes represented using a deterministic GUID.
- Redirects are stored in a single document. This prevents query index issues. Document Id is `messageredirects`. Initial estimate was ~10k redirects per MB document size.
- Failed message retry will replace failed queue address with redirect To when failed queue address matches from, case insensitive.
- The redirects document is re-loaded for each batch of retries.
- **NEW** POST `/redirects` Create a new redirect. 
   - Validate from/to are populated. 
   - Validate From is unique (ignore if To is the same). 
   - Validate no existing To point to new From.
   - Retry by failed queue address can be called when redirect is created (for pending retries) by including `retryexisting` query parameter
- **NEW** PUT `/redirects/{messageredirectid:guid}/` Update existing redirect. Validate To is populated. Validate From exists. Validate To does not point to any existing From.
- **NEW** HEAD `/redirects` Get stats for redirects.
- **NEW** GET `/redirects` List existing redirects. Supports paging.

src/ServiceControl/MessageRedirects/Api/MessageRedirectsModule.cs
```diff
@@ -0,0 +1,200 @@
...
+                        ToPhysicalAddress = messageRedirect.ToPhysicalAddress = request.tophysicaladdress.ToLowerInvariant(),
+                    };
+
+                    messageRedirect.LastModifiedTicks = DateTime.UtcNow.Ticks;
+
+                    redirects.Save(session);
+
+                    Bus.Publish(messageRedirectChanged);
+
+                    return HttpStatusCode.OK;
```
johnsimons: I wonder if we should return a 204 ?

src/ServiceControl/MessageRedirects/Api/MessageRedirectsModule.cs
```diff
@@ -0,0 +1,200 @@
...
+
+                    Bus.Publish<MessageRedirectRemoved>(evt =>
+                    {
+                        evt.MessageRedirectId = messageRedirectId;
+                        evt.FromPhysicalAddress = messageRedirect.FromPhysicalAddress;
+                        evt.ToPhysicalAddress = messageRedirect.ToPhysicalAddress;
+                    });
+                }
+
+                return HttpStatusCode.OK;
```
johnsimons: I wonder if we should return a 204 ?

src/ServiceControl/MessageFailures/Api/FailedMessageViewIndex.cs
boblangley: Will we need it when we filter by QueryAddress?

https://github.com/Particular/ServiceControl/pull/762/files#diff-59f9f660f25fbc5f2a406e858443f133R233

src/ServiceControl/MessageRedirects/Api/MessageRedirectsModule.cs
```diff
@@ -0,0 +1,202 @@
...
+                        {
+                            r.MessageRedirectId,
+                            r.FromPhysicalAddress,
+                            r.ToPhysicalAddress,
+                            LastModified = new DateTime(r.LastModifiedTicks)
+                        });
+
+                    return Negotiate
+                        .WithModel(queryResult)
+                        .WithEtagAndLastModified(redirects.ETag, queryResult.Max(r => r.LastModified))
```
WojcikMike: When the collection is empty the Max throws Invalid Operation Expcetion:  Sequence contains no elements

src/ServiceControl/MessageRedirects/Api/MessageRedirectsModule.cs
```diff
@@ -0,0 +1,202 @@
...
+                {
+                    var redirects = MessageRedirectsCollection.GetOrCreate(session);
+
+                    return Negotiate
+                        .WithEtagAndLastModified(redirects.ETag, redirects.Redirects.Max(r => new DateTime(r.LastModifiedTicks)))
+                        .WithTotalCount(redirects.Redirects.Count);
+                }
+            };
+
+            Get["/redirects"] = _ =>
```
WojcikMike: I have a problem with that call. I have 2 redirects, I remove 1 using delete. Now when I call this method again, even that when I debug in code I see that there is only 1 redirect, after return browser shows me that response has a status 304 and 2 redirects. Something is off with Nancy recognizing that results are not changed

boblangley: Addressed in https://github.com/Particular/ServiceControl/pull/762/commits/9cbcb86519b7107bf92a5e7daf2fdb6f3f6c144c Used document last modified metadata instead of internal data which may not be complete or change when the document is updated

boblangley:
Fixed issue with `queue_address` not being populated with https://github.com/Particular/ServiceControl/pull/762/commits/e10a0bee98935d73c88d29284f634f2855662fd9. Updated comment https://github.com/Particular/ServiceControl/pull/762#issuecomment-234110999 to reflect the correct name of the property when calling the api

src/ServiceControl/MessageRedirects/Api/MessageRedirectsModule.cs
boblangley: As you stated in slack both are acceptable, but I think 204 is the more appropriate.

src/ServiceControl/MessageRedirects/Api/MessageRedirectsModule.cs
boblangley: As you stated in slack both are acceptable, but I think 204 is the more appropriate.

boblangley:
@WojcikMike I updated Message Redirects PUT/DELETE to return 204 on success instead of 200.

boblangley:
@Particular/servicecontrol-maintainers @WojcikMike SC is ready. ServicePulse will require some adjustments. The Pending Retries actions have been changed (parameters moved from URI to POST body) and we need to test long queue names (ala Azure Storage Queues) with the ServicePulse UI.

boblangley:
Merged #762

PR #768, opened 2016-08-11
Support GMSA accounts

gbiellem:
Customer request:

> Description: I want the Windows Service that NServiceBus.Host.exe creates
> to be executed as a Group Managed Service Account. In Windows Server 2012,
> you can create a special account in Active Directory called a Group Managed
> Service Account (gMSA). You can use this account to host Windows Services,
> IIS App Pools, Scheduled Tasks, and more. The advantage of using a gMSA is
> that you don't have to know the password. In fact, no one knows the
> password. When an administrator of a server creates a Windows Service
> using services.msc, he (or she) can click on the login tab, type the gMSA's
> UPN and leave the password blank. The server (if it is permitted to do so)
> will retrieve the managed password from Active Directory. The password is
> automatically changed and refreshed on a scheduled basis.
> 
> 
> 
> We run as many services as we can using gMSA's.
> 
> 
> 
> You can read more about gMSA's here:
> https://blogs.technet.microsoft.com/askpfeplat/2012/12/16/windows-server-2012-group-managed-service-accounts/
> 

See also https://github.com/Particular/NServiceBus.Host/issues/93

### DIFF
--- a/src/ServiceControl.Config/UI/InstanceAdd/InstanceAddView.xaml
+++ b/src/ServiceControl.Config/UI/InstanceAdd/InstanceAddView.xaml
@@ -42,17 +42,16 @@
                             </Grid.RowDefinitions>
                             <controls:FormTextBox Grid.Row="1"
                                                   Header="SERVICE ACCOUNT"
-                                                  IsEnabled="{Binding UseProvidedAccount}"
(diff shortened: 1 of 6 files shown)


johnsimons:
looks good @gbiellem

PR #771, opened 2016-08-16
Testing new deps

johnsimons:


src/ServiceControl/Infrastructure/OWIN/Startup.cs
```diff
@@ -38,6 +38,9 @@ public Startup(IContainer container, ServiceBase host, Settings settings, Embedd
 
         public void Configuration(IAppBuilder app)
         {
+            var signalrIsReady = new SignalrIsReady();
+
+            configuration.RegisterComponents(components => components.RegisterSingleton(signalrIsReady));
```
WojcikMike: This addition cause Service Control to throw null ref when running from visual studio. As configuration object is null.

PR #772, opened 2016-08-17
Upgrade to Raven v3

johnsimons:
This PR upgrades RavenDB to v3

As part of this upgrade ServiceControl is now using a separate database in RavenDB , before we were co-sharing the <system> database.

In RavenDB v3, the embedded database feature does not exist anymore, it now requires all connections to be http/https based, so this means we are going ahead with using an external RavenDB server, this means that at install time we will give the option to install a RavenDB server.

The installer will create the database and indexes.

Because RavenDB v3 data storage has changed, we need to migrate existing data to the new standalone RavenDB server, to accomplish this we are backing up the existing embedded database and then importing it in.

The purging of data is now integrated as part of ServiceControl, we no longer use a RavenDB bundle.


johnsimons:
@gbiellem I say we bring in #774 before this one ?

gbiellem:
@johnsimons  I'm ok with that - Feel free to start reviewing #774

src/ServiceControl/Operations/Audit/ProcessAudits.cs
```diff
@@ -117,14 +132,35 @@ public void Stop()
                 }
 
                 meter.Mark(processedFiles.Count);
+            }
+            catch (Exception ex)
+            {
+                logger.Error("Something went wrong ingesting a batch", ex);
```
johnsimons: This will never trigger the breaker!

mikeminutillo: True. Should each one trigger the breaker or should the breaker be triggered once if there were any failures?

johnsimons:
We are backlogging this upgrade and re-evaluate the ServiceControl future direction

Issue #766, opened 2016-07-15
Users should be warned when they turn on audit and error forwarding


indualagarsamy:
When installing SC, users need to be aware that they should only choose audit and error forwarding if they already have a process that consumes messages from the forwarded queue. 

If users aren't aware, they might choose it, not knowing that messages will start piling up in the forwarded queues and eventually run out of disk space causing bad things to happen.

gbiellem:
@indualagarsamy  I knew there was a related issue.

Do you see this as just explanatory text in line with the options or via some sort of help tip - See @sergioc's suggestion here - https://github.com/Particular/ServiceControl/issues/534#issuecomment-185142743

indualagarsamy:
@gbiellem @sergioc - I don't think tooltip is the way to go. It can be easily missed. I prefer always readable text. 
Something like: Select this option only if you have a service consuming the forwarded messages
or like it.

sergioc:
I agree with @indualagarsamy in principle, but that comes at the cost of added clutter.

How about showing the in your face text only if the user starts typing something in that input?

indualagarsamy:
How about the default set to No for forwarding, with a line of text that's inside the group box option for forwarding that says:
"Turn on forwarding only if the forwarded messages are being consumed by another service"
I don't think 1 line of text will be adding clutter. In this case its more meaningful to the user.

sergioc:
@indualagarsamy why would you need to give that information in advance, before the user would turn the option ON?

indualagarsamy:
@sergioc - because if users know that in advance, they might not even find the need to change the default from No to Yes. It helps them decide correct in first place. Much more intuitive.

johnsimons:
We currently already display a warning text when upgrading, here it is:
"Error messages can be forwarded to a secondary error queue known as the Error Forwarding Queue. This queue exists to allow external tools to receive error messages. If you do not have a tool processing messages from the Error Forwarding Queue this setting should be disabled."

So I suggest we use about the same ?

sergioc:
@indualagarsamy I don't think that's worth the cost of the additional clutter. The first time you turn that option ON, you'll get that message and learn about the consequences of having the option ON. After you learn about that it is unlikely you'll need the default message clutter again.

johnsimons:
@gbiellem this one would be an easy one to work on ?

PR #773, opened 2016-08-18
Added warning about forwarding queues

johnsimons:
This addresses #766

### DIFF
--- a/src/ServiceControl.Config/Themes/Generic.xaml
+++ b/src/ServiceControl.Config/Themes/Generic.xaml
@@ -55,7 +55,7 @@
                                 <ScrollViewer x:Name="PART_ContentHost" />
                             </Border>
                             <StackPanel Grid.Row="2">
-                                <StackPanel Orientation="Horizontal" Visibility="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=(Validation.HasError), Converter={StaticResource boolToVis}}">
+                                <StackPanel Orientation="Horizontal" Visibility="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=(Validation.HasError), Converter={StaticResource boolToVis}}" Margin="0,3,0,0">
                                     <Control Width="16"
                                              Height="16"
                                              Margin="5,0,0,0"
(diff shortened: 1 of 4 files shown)


johnsimons:
@gbiellem want to review this one ?

gbiellem:
@johnsimons LGTM. I made the warnings contextual so they only show when the options are on.  I believe that what @sergioc wanted.

sergioc:
Behavior is 👍 . However, the warning text is a bit cramped and the yellow text makes it a bit harder to read:

screenshot 2016-08-22 09 21 22

I've mocked up a version with black text, more white space, better alignment and a shortened message:

screenshot 2016-08-22 09 36 55

gbiellem:
Thanks @sergioc I'll adjust it

sergioc:
@gbiellem if we have documentation about why should Audit Forwarding remain OFF if that queue is not being processed, I suggest adding a "Learn more" link at the end of the text linking to the documentation page.

gbiellem:
@sergioc 

Not sure about that, I believe we usually try to steer clear of deep linking to docs within the products

sergioc:
> Not sure about that, I believe we usually try to steer clear of deep linking to docs within the products

Why? It would provide more context to the warning. Otherwise you either leave the user wondering about why or you have to resort to lengthy messages in the UI.

gbiellem:
@sergioc tweaks made. 

capture

I won't add a link because
-  the message is pretty self explanatory -
- as I've stated we don't deep link to docs.  

I believe the reason we don't deep link is that by anchoring this to a version of a page in docs we are then bound to maintain that URL and the relevance of the content for the life of this version of the product.  So effectivley we'd expect it to be a permalink

gbiellem:
@johnsimons  review and merge if you are happy

sergioc:
Makes sense why we don't deep link. However, that doesn't mean we should leave the user with questions unanswered and no clue about where to find the answer.

> the message is pretty self explanatory

@gbiellem as a user I'd ask "why is it telling me to enable in those circumstances?". How about creating a modal with the full explanation, accessible via a "Learn more" or "More information" link?

gbiellem:
@sergioc I'm struggling to see what is vague in the message.  Anyone who is installing this is setting it up to monitor queues so they'll have a basic understanding on message queuing. It makes perfect sense to not send messages to a queue if nothing is processing them. The end result is at best a waste of disk space and at worst out of disk space/queue resources.

gbiellem:
If you do think it is still vague I'd suggest we actually re work the form and make a section called something like External Integration and put all the Forwarding configuration under that with an overarching blurb on what the forwarding queues are

gbiellem:
@indualagarsamy  Any thoughts since you raised the original issue this PR is trying to addresses

sergioc:
> If you do think it is still vague I'd suggest we actually re work the form and make a section called something like External Integration and put all the Forwarding configuration under that with an overarching blurb on what the forwarding queues are

I'd definitely not hold up merging this in order to do what you suggest since what's done already constitutes an improvement. The UI should not be cluttered with too much explanation text either.

I'm just concerned that users might not understand why we're issuing that warning, because the cause may not seem explicit. We'd only be able to figure this out during a user test, though. So I'd release this as-is, then later assess in a test.

gbiellem:
@sergioc  :+1:  - I'd still like to get @indualagarsamy to review before we merge though since she raised the original issue

indualagarsamy:
@gbiellem - I like the changes you made. 👍 for merging.

PR #774, opened 2016-08-22
[WIP} SC HTTPS Support

gbiellem:
This a pre-requisite to moving to Raven3 and also to adding access and auth to SC.


gbiellem:
This is pretty much done. Parking it for now as there is no reason to pull it in yet.

gbiellem:
@johnsimons  you can start reviewing but I need to retest the powershell parts before its merged and tweet the ps doco

johnsimons:
@gbiellem the code looks good, but I am a bit confused on the user interactions, can you expand on them, eg:
- A fresh new instance install, how do I make it https ? I couldn't see any modifications to the new instance screens shouldn't the `InstanceAddView.xaml` be modified ?
- An upgrade? I was expecting us to ask whether they want to turn https on as part of the upgrade ?

What is the default, on or off ? Should it be ON ?

gbiellem:
As discussed in zoom.  SSL is not the normal use case, it's something that should be on when the SC is on an untrusted network and creds are being passed which is why this was thought of as part of either the Raven3 update or adding access control which is one of the other issue in the pipeline.

For any install where the instance is set to localhost it would make no sense to turn on SSL,
I like the idea you proposed in zoom of prompting or warning the user if they set the hostname to something other than localhost that SSL should be considered.  I'll add that to the UI

johnsimons:
@gbiellem do we close this ?

PR #775, opened 2016-09-01
Extra performance tweaks for raven v3 upgrade

johnsimons:
A few extras:
- bulk inserts
- file storage
- cachable subscriptions
- less chatty EventDispatcher
- Metrics.net
- Voron storage by default


johnsimons:
@mikeminutillo I think this is it

PR #777, opened 2016-09-09
[WIP] Remove Raven UoW

johnsimons:
@mikeminutillo we spoke about this before.

I have got rid of the injectable `IDocumentSession` and instead I have changed it to explicit OpenSession at the handler so we can decide if concurrency check should be enabled or not.

I am not 100% sure I am covering all the places that we need to disallow concurrent updates.

Currently I have only turned on concurrency checks for:
- RegisterEndpointHandler
- MessageFailureResolvedHandler
- UnArchiveMessagesHandler

Custom checks, heartbeats, I don't think we should support concurrent updates, this means we could be out of sync but i think that is only momentarily, eventually we good again.

What are your thoughts ?


src/ServiceControl/CompositeViews/Endpoints/KnownEndpointUpdateHandler.cs
```diff
@@ -8,24 +8,34 @@ namespace ServiceControl.CompositeViews.Endpoints
...
-            }
+                knownEndpoint = session.Load<KnownEndpoint>(message.EndpointId);
+
+                if (knownEndpoint == null)
+                {
+                    throw new Exception($"No endpoint with found with id: {message.EndpointId}");
+                }
 
-            knownEndpoint.Monitored = true;
+                knownEndpoint.Monitored = true;
```
mikeminutillo: I wonder if this shouldn't just be a patch request. It seems silly to do a load-update-save cycle to update a single boolean flag

src/ServiceControl/CompositeViews/Endpoints/KnownEndpointUpdateHandler.cs
```diff
@@ -8,24 +8,34 @@ namespace ServiceControl.CompositeViews.Endpoints
...
 
-            Session.Store(knownEndpoint);
+                session.Store(knownEndpoint);
+            }
 
-            Bus.Publish(new MonitoringEnabledForEndpoint
+            bus.Publish(new MonitoringEnabledForEndpoint
             {
                 EndpointInstanceId = message.EndpointId,
                 Endpoint = knownEndpoint.EndpointDetails
```
mikeminutillo: Oh wait, we need details. Never mind

johnsimons:
anything else @mikeminutillo ?

PR #795, opened 2016-10-03
Modify `GetErrorById` to return data in similar format as `GetAllErrors`

WojcikMike:
Modify `GetErrorById` to return data in similar format as `GetAllErrors`

### DIFF
--- a/src/ServiceControl/MessageFailures/Api/GetErrorById.cs
+++ b/src/ServiceControl/MessageFailures/Api/GetErrorById.cs
@@ -1,8 +1,11 @@
 ﻿namespace ServiceControl.MessageFailures.Api
 {
     using System;
+    using System.Linq;
     using Nancy;
+    using Raven.Client;
     using ServiceBus.Management.Infrastructure.Nancy.Modules;
+    using ServiceControl.Contracts.Operations;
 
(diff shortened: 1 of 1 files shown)


WojcikMike:
@Particular/servicecontrol-maintainers can someone have a look at this?

johnsimons:
I have no issues with this change, just confirming with @Particular/serviceinsight-maintainers to make sure it does not affect it.

HEskandari:
From ServiceInsight perspective, at the moment we don't use the response from the 'retry' endpoint so changing the return type would not be an issue. There's also no strict 'contract' or proxy objects that'd need regenerating, so it LGTM.

With all said, I'd like to test ServiceInsight with the updated version of ServiceControl to absolutely make sure it works fine.

HEskandari:
I stand corrected, we don't use this specific API at all, we use the Retry one at /errors/{id}/retry. Feel free to change this as it has no impact on ServiceInsight.

WojcikMike:
In that case please merge @Particular/servicecontrol-maintainers if you are satisfied with the code

WojcikMike:
@johnsimons I assume that this PR will be released with any next feature that will be released right?

PR #799, opened 2016-10-11
Update to Autofac 4.1.1 and NServiceBus.Autofac 5.1.1

danielmarbach:
Relates to https://github.com/Particular/docs.particular.net/pull/2024

To be honest this update is not strictly necessary. I figured I'll send it in nonetheless to see how the latest container changes would influence SC

### DIFF
--- a/src/ServiceControl.AcceptanceTests/ServiceControl.AcceptanceTests.csproj
+++ b/src/ServiceControl.AcceptanceTests/ServiceControl.AcceptanceTests.csproj
@@ -37,8 +37,8 @@
     <LangVersion>6</LangVersion>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="Autofac, Version=3.5.0.0, Culture=neutral, PublicKeyToken=17863af14b0044da, processorArchitecture=MSIL">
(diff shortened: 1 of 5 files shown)


johnsimons:
So @danielmarbach you mean that if we decide to update to the latest version of the NServiceBus.Autofac then we also need to dispose of the container ?

danielmarbach:
@johnsimons correct

SimonCropp:
> So @danielmarbach you mean that if we decide to update to the latest version of the NServiceBus.Autofac then we also need to dispose of the container

then how is that a safe behaviour change to put in a patch?

danielmarbach:
@SimonCropp discussed this with @andreasohlund @timbussmann and @Particular/container-maintainers and we treated it as a bug. Let's face it:

If a user owns a container and passed it into the bus configuration then that code becomes the composition root of the container. It is and always was the responsibility of the composition root to dispose disposables. Arguably if a user code doesn't dispose the created container the bug was always there in the user code and our wrong container adapter behavior made it sorta work but it never should have.

johnsimons:
Fixed in https://github.com/Particular/ServiceControl/commit/b8995d60ccfb1bae400b9d544344cbc3732321a2

PR #802, opened 2016-10-11
Revert "Modify `GetErrorById` to return data in similar format as `GetAllErrors`"

johnsimons:
Reverts Particular/ServiceControl#795

### DIFF
--- a/src/ServiceControl/MessageFailures/Api/GetErrorById.cs
+++ b/src/ServiceControl/MessageFailures/Api/GetErrorById.cs
@@ -1,11 +1,8 @@
 ﻿namespace ServiceControl.MessageFailures.Api
 {
     using System;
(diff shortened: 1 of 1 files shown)


johnsimons:
@WojcikMike heads up, I have reverted #795, since it broke a lot of stuff on our acceptance tests, Instead I will create a new dedicated API

johnsimons:
@WojcikMike see https://github.com/Particular/ServiceControl/commit/40099c0e77783829b682cf06755d08a6495d430b#diff-6fe7def7f518c7a5832858f293f5e95dR31

WojcikMike:
@johnsimons ok, thanks for info. BTW it seems that original GerErrorById method (the one that I modified) is not used neither by SP nor by SI. FYI

Issue #789, opened 2016-09-22
License reporting in ServiceControl does not refresh


johnsimons:
When a user updates the license via management utility, the SC instances do not pick up the new license. So they keep reporting on the old license.

gbiellem:
It's also broken even without a license change.because the license status isn't periodically checked  
So a long running SC instance won't report the expiry date has passed until the next service restart

gbiellem:
Related to #783

Issue #783, opened 2016-09-20
Improve license reporting to ServicePulse


gbiellem:
Related to #782 

License status reported to ServicePulse has only two states - either the license is good or expired.
No indication of expiration dates.  upgrade protection dates or type of license is forwarded to ServicePulse.  This can lead to confusion from the customer perspective as to why the license is being reported as expired

johnsimons:
So what can we improve exactly ?

gbiellem:
@johnsimons  Make the license details available for display (Licensee,  expiration date and upprade protection date, and type of license)  

We could also include a warning status if the license is out of maintenance or almost of out of maintenance or that can be left to SP to derive from the data we make available.

Issue #782, opened 2016-09-20
Management Utility does not handle upgrade protection correctly.


gbiellem:
The management utility does not give any indication that the maintenance period for a license has expired.  It should also prevent an instance from being upgraded if the release date of the version being upgraded falls after the upgrade protection expiration

gbiellem:
@Particular/servicecontrol-maintainers  thoughts on this?

If the new version of ServiceControl falls outside the maintenance period of the currently installed license should the installation of the new version be blocked or should it just result in a warning prior to upgrade indicating that after the upgrade has completed ServiceControl will be unlicensed?

gbiellem:
@Particular/servicecontrol-maintainers no thoughts on this?

I propose we go with a warning on upgrade and not block

gbiellem:
Raised a corresponding discussion issue in platdev - https://github.com/Particular/PlatformDevelopment/issues/950

gbiellem:
Result of plat dev discussion was to move ahead with blocking upgrades and installation if an expired license is found.  This will prevent users upgrading and then needing to downgrade because the license is out of date.

PR #803, opened 2016-10-11
License Fixes

gbiellem:
Related to #782 #783 and #789

Added a feature to periodically poll for license changes and expose the license details via rest api 

Changes the Management utility to block install and upgrading instances if the license has expired or the upgrade version falls outside the maintenance period


src/ServiceControl/Licensing/LicenseModule.cs
```diff
@@ -0,0 +1,36 @@
...
+        public ActiveLicense ActiveLicense { get; set; }
+
+        public LicenseModule()
+        {
+            Get["/license"] = parameters =>
+            {
+                var licenseInfo = new LicenseInfo
+                {
+                    TrialLicense = ActiveLicense.Details.IsTrialLicense,
+                    Edition = ActiveLicense.Details.Edition ?? "" ,
```
johnsimons: `String.Empty` ?

src/ServiceControl/Infrastructure/Nancy/Modules/RootModule.cs
```diff
@@ -30,6 +30,7 @@ public RootModule()
                     Name = SettingsReader<string>.Read("Name", "ServiceControl"),
                     Description = SettingsReader<string>.Read("Description", "The management backend for the Particular Service Platform"),
                     LicenseStatus = License.IsValid ? "valid" : "invalid",
```
johnsimons: Do we move the status to `LicenseDetails` ?

gbiellem: Wouldn;t that break older versions of pulse?

src/ServiceControl/Licensing/LicenseModule.cs
gbiellem: fixed

gbiellem:
@johnsimons  are you happy with the rest of this?

src/ServiceControl/Licensing/LicenseCheckFeature.cs
```diff
@@ -0,0 +1,58 @@
...
+                manualResetEvent.WaitOne();
+            }
+        }
+
+        void Refresh(object _)
+        {
+            activeLicense.Refresh();
+            try
+            {
+                timer.Change((int)TimeSpan.FromMinutes(10).TotalMilliseconds, -1);
```
johnsimons: Does it need to be this frequent ?

johnsimons:
i thought we also wanted to warn about license about to expire ?

src/ServiceControl/Licensing/LicenseCheckFeature.cs
gbiellem: How frequent do you suggest? Having it too long would be a weird experience for anyone updating the license..  i.e.  I updated the license but ServicePulse still says it's expired.  We already get those support cases

johnsimons: 1 hour

gbiellem:
> i thought we also wanted to warn about license about to expire ?

I'm not sure I see a benefit in adding a warning x days before expiry.  Pulse will tell them when the license has expired and there is no degradation of service once it has expired.  I'm been trying to think of another product that does a impending license expiry and I can't come up with one.   

Given we are now exposing the license expiration and maintenance we could also do that from within pulse without using a notification

src/ServiceControl/Licensing/LicenseCheckFeature.cs
johnsimons: BTW, you can change that to use `TimeKeeper`, see https://github.com/Particular/ServiceControl/blob/40607aacd062404253797a1f3221718d2177598c/src/ServiceControl/Recoverability/Retries/FailedMessageRetries.cs#L124

gbiellem: I just had a look at the Raven code - it checks for a license every 15 minutes.  An hour seems like a long time.  If we're going to set it to be that long then I'll change the mgmt UI to indicate that when updating/adding a license

gbiellem: Thanks for the TimeKeeper tip - Will update

gbiellem:
@johnsimons Changed to Timekeeper and set the license refresh to 8 hours.   Also added a query string option to force the license to reload.  Tweaked mgmt util to try to call the license refresh on each instance after license is imported so we don;t need a high refresh rate

johnsimons:
looks good @gbiellem

gbiellem:
One last thing to do.  The license expirty only blocks if installs and upgrades  are done through the mgmt UI.   Once I add the checks to the Powershell cmdlets this should be good to go

gbiellem:
@johnsimons When I tested the artifact from the build service I'm getting some binding errors in the management UI that I don't get when debugging.  Only difference i can see is this branch is 4.5 but after merging with master it's 4.5.2.  So don't merge yet til I resolve this

gbiellem:
@johnsimons  Think I'm done with this. Appreciate if you could take another look and merge if you are happy with it.

src/ServiceControl.Config/Commands/AddInstanceCommand.cs
```diff
@@ -5,21 +5,31 @@
...
             this.addInstance = addInstance;
+            this.installer = installer;
         }
 
         public override void Execute(object obj)
         {
+            var licenseCheckResult = installer.CheckLicenseIsValid();
+            if (!licenseCheckResult.Valid)
+            {
+                windowManager.ShowMessage(licenseCheckResult.Message, "Installation could not continue due to an issue with the current license. Contact sales@particular.net", hideCancel: true);
```
johnsimons: @gbiellem, it seems the error message `licenseCheckResult.Message` is actually displayed as the title of the popup dialog, shouldn't the `licenseCheckResult.Message` be part of the body ?

src/ServiceControl.Config/Commands/UpgradeInstanceCommand.cs
```diff
@@ -28,13 +28,19 @@ public UpgradeInstanceCommand(Func<InstanceDetailsViewModel, bool> canExecuteMet
...
             this.eventAggregator = eventAggregator;
             this.installer = installer;
         }
 
         public override async Task ExecuteAsync(InstanceDetailsViewModel model)
         {
+            var licenseCheckResult = installer.CheckLicenseIsValid();
+            if (!licenseCheckResult.Valid)
+            {
+                windowManager.ShowMessage(licenseCheckResult.Message, "Upgrade could not continue due to an issue with the current license. Contact sales@particular.net", hideCancel: true);
```
johnsimons: @gbiellem, it seems the error message `licenseCheckResult.Message` is actually displayed as the title of the popup dialog, shouldn't the `licenseCheckResult.Message` be part of the body ?

src/ServiceControlInstaller.Engine/FileSystem/DeploymentZipInfo.cs
```diff
@@ -55,5 +57,58 @@ public void ValidateZip()
...
+        public bool TryReadServiceControlReleaseDate(out DateTime releaseDate)
+        {
+            releaseDate = DateTime.MinValue;
+            var tempDomain = AppDomain.CreateDomain("TemporaryAppDomain");
+            var tempFile = Path.Combine(Path.GetTempPath(), @"ServiceControl\ServiceControl.exe");
+            try
+            {
+                using (var zip = ZipFile.Read(FilePath))
+                {
+                    var entry = zip.Entries.FirstOrDefault(p => p.FileName == "ServiceControl/ServiceControl.exe");
```
johnsimons: is the slash right ?

src/ServiceControlInstaller.Engine/Unattended/UnattendInstaller.cs
```diff
@@ -228,6 +243,45 @@ public bool Delete(string InstanceName, bool removeDB, bool removeLogs)
...
+            if (!license.Details.ValidForServiceControl)
+            {
+                return new CheckLicenseResult(false, "This license edition does not include ServiceControl");
+            }
+
+            if (ZipInfo.TryReadServiceControlReleaseDate(out releaseDate))
+            {
+                if (license.Details.ReleaseNotCoveredByMaintenance(releaseDate))
+                {
+                    return new CheckLicenseResult(false, "This version was released afer the upgrade protection expiration date specified in the license");
```
johnsimons: spelling "afer"

src/ServiceControlInstaller.Engine/FileSystem/DeploymentZipInfo.cs
gbiellem: Yep - Slash is correct. Entries in a zip use that slash

src/ServiceControl.Config/Commands/AddInstanceCommand.cs
gbiellem: that was intentional.

src/ServiceControl.Config/Commands/UpgradeInstanceCommand.cs
gbiellem: that was intentional.

src/ServiceControlInstaller.Engine/Unattended/UnattendInstaller.cs
gbiellem: oops - will fix

src/ServiceControl.Config/Commands/UpgradeInstanceCommand.cs
johnsimons: not following ? Doesn't it look strange that way ?

gbiellem:
@johnsimons I've addressed your comments above. Anything else or we good to merge?

PR #806, opened 2016-10-13
Update NServiceBus framework

johnsimons:
Removed usages of pipeline from importers


johnsimons:
@mikeminutillo As part of updating to NServiceBus v5.2.19, it seems we also need to remove the pipeline from the importers.
So I have backported those changes to this PR.
Can you please review ?

We also need to backport the removal of the raven UoW, not sure if we bundle it here ?

mikeminutillo:
LGTM.

We should remove `AuditMessageHandler` as it is no longer in use.

PR #811, opened 2016-10-19
Record a possibly incomplete batch

WilliamBZA:


src/ServiceControl/Recoverability/Grouping/Retries/RetryOperationManager.cs
```diff
@@ -83,6 +95,7 @@ public class RetryOperationSummary
     {
         public int? MessagesRemaining { get; internal set; }
 
+        public bool IsPossiblyIncomplete { get; set; } // Need a better name for this
```
janovesk: How about `WasInterrupted`?

src/ServiceControl/Recoverability/Grouping/FailureGroupsApi.cs
```diff
@@ -84,7 +84,8 @@ dynamic GetAllGroups(string classifier)
                             First = failureGroup.First,
                             Last = failureGroup.Last,
                             RetryStatus = summary != null ? "In Progress" : null,
-                            NumberOfRetryMessagesRemaining = summary?.MessagesRemaining
+                            NumberOfRetryMessagesRemaining = summary?.MessagesRemaining,
+                            IsPossiblyAnIncompleteRetry = summary.IsPossiblyIncomplete
```
janovesk: Don't we know for a fact that it WAS interrupted?

src/ServiceControl/Recoverability/Retries/RetryDocumentManager.cs
```diff
@@ -114,26 +114,33 @@ internal void AdoptOrphanedBatches(IDocumentSession session, out bool hasMoreWor
         {
             RavenQueryStatistics stats;
 
-            var orphanedBatchIds = session.Query<RetryBatch, RetryBatches_ByStatusAndSession>()
+            var orphanedBatches = session.Query<RetryBatch, RetryBatches_ByStatusAndSession>()
```
janovesk: Not sure how smart Raven is for this. Will this change what's going over the wire from a list of id's to a list of full documents?

src/ServiceControl/Recoverability/Retries/RetryDocumentManager.cs
```diff
@@ -114,26 +114,33 @@ internal void AdoptOrphanedBatches(IDocumentSession session, out bool hasMoreWor
...
 
-            log.InfoFormat("Found {0} orphaned retry batches from previous sessions", orphanedBatchIds.Length);
+            log.InfoFormat("Found {0} orphaned retry batches from previous sessions", orphanedBatches.Length);
 
-            AdoptBatches(session, orphanedBatchIds);
+            AdoptBatches(session, orphanedBatches.Select(b => b.Id));
+
+            foreach (var batch in orphanedBatches)
+            {
+                if (batch.RetryType != RetryType.MultipleMessages)
```
janovesk: I'd drop this if check if that means we have to load the full document compared to just an Id previously. No harm in having status for `RetryType.MultipleMessages` in the RetryOperationManager in-memory cache. They won't be queried, but will be cleared out when the adopted batch is forwarded.

src/ServiceControl/Recoverability/Grouping/FailureGroupsApi.cs
WilliamBZA: No, all the documents could have been marked and SC could have crashed before updating the batch status

src/ServiceControl/Recoverability/Retries/RetryDocumentManager.cs
WilliamBZA: We still need to load the retry type and the request id to mark the correct batch

src/ServiceControl/Recoverability/Grouping/FailureGroupsApi.cs
janovesk: So you're defining updating of the batch status as not part of the retry? :D

PR #815, opened 2016-10-24
Added progress indicator for retry operations

janovesk:


janovesk:
Ping @WilliamBZA @WojcikMike

src/ServiceControl/Recoverability/Grouping/Retries/RetryOperationProgressionCalculator.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    public static class RetryOperationProgressionCalculator
+    {
+        public static double CalculateProgression(RetryOperationSummary retry)
+        {
+            const double waitingWeight = 0.05;
+            const double prepairedWeight = 0.475;
+            const double forwardedWeight = 0.475;
+
+            if (retry.RetryState == RetryState.Waiting)
```
WilliamBZA: Are these checks necessary? Are they not covered by the formula at the end?

janovesk: If the state of the the Retry Operation is correct, then yes. :)  Good point, I'll clean it up.

Issue #790, opened 2016-09-23
Handle Duplicate messages in Error queue


boblangley:
With our current level of concurrency and the RavenDB PUT mechanics, duplicate error messages can trip the circuit breaker and cause ServiceControl to crash.

/cc @bording @SeanFeldman

boblangley:
In talking with @johnsimons SC 2.0 (which is currently in development) addresses this issue. I am going to write a failing test for this in the 1.x branch (ignored for builds) and then port it to the 2.0 branch (if necessary) to confirm this has been resolved there.

johnsimons:
@boblangley given v2 upgrade is cancelled, I think we need to repro this and address it in the current version.
We can pair on this and try to figure it out ?

boblangley:
Yes, I will contact you shortly to coordinate.

johnsimons:
Fixed in https://github.com/Particular/ServiceControl/pull/818

PR #818, opened 2016-10-25
Using patch api in error importer

johnsimons:
This is trying to address #790


johnsimons:
@boblangley can you review this and make sure it addressed the issue raised.

PR #829, opened 2016-10-27
Add acknowledgements of completed retry operations

WilliamBZA:
Ping @janovesk @WojcikMike @sergioc

### DIFF
--- a/src/ServiceControl/Recoverability/Grouping/Retries/RetryOperationCompletedHandler.cs
+++ b/src/ServiceControl/Recoverability/Grouping/Retries/RetryOperationCompletedHandler.cs
@@ -1,0 +1,34 @@
+﻿using System;
+using NServiceBus;
+using Raven.Client;
+
+namespace ServiceControl.Recoverability
+{
+    public class RetryOperationCompletedHandler : IHandleMessages<RetryOperationCompleted>,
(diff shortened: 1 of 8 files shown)


src/ServiceControl/Recoverability/Grouping/Retries/RetryOperationCompletedHandler.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System;
+using NServiceBus;
+using Raven.Client;
+
+namespace ServiceControl.Recoverability
+{
+    public class RetryOperationCompletedHandler : IHandleMessages<RetryOperationCompleted>,
+        IHandleMessages<AcknowledgeRetryOperationCompleted>
+    {
+        public IDocumentSession Session { get; set; }
```
johnsimons: Sorry @WilliamBZA forgot to also mention that we removed the raven UoW! We now opted to be explicit, instead of having unnecessary raven sessions that may or may not be used.
The way to go about it now, is to inject `IDocumentStore`, and then do your own sessions, but in this scenario u don't even need a session, u could just `store.DatabaseCommands.Delete(id)`

janovesk:
LGTM, but let's land the "should we let the user acknowledge" discussion before merging?

PR #831, opened 2016-11-01
Add the concept of retry operations history

WilliamBZA:


src/ServiceControl/Recoverability/Grouping/Retries/RetryOperationCompletedHandler.cs
```diff
@@ -0,0 +1,37 @@
...
+        public RetryOperationManager RetryOperationManager { get; set; }
+
+        public void Handle(RetryOperationCompleted message)
+        {
+            var completedOperation = new CompletedRetryOperation
+            {
+                RequestId = message.RequestId,
+                RetryType = message.RetryType,
+                CompletionDate = message.CompletionDate
+            };
```
janovesk: We don't store or act on this being a failed operation or not. Is this what we want if we are showing this in the UI as "LAST 10 SUCCESSFUL GROUP RETRIES"? I would expect either to only see successfully completed retries here or a different header like "LAST 10 GROUP RETRIES".

WilliamBZA: What failures could there be?

janovesk: SC restart that cut the retry operation short for example. From the end user, that would not be a full success.

PR #837, opened 2016-11-07
Disabling monitoring of endpoints does not persist changes

SzymonPobiega:
### Symptoms
When disabling monitoring of endpoints, the command does not work because the changes are not persisted to RavenDB. 

### Who's Affected
Users that are running ServiceControl v1.25.0.


SzymonPobiega:

PR #838, opened 2016-11-08
Fixes the problem of showing endpoints with no HB plugin as "inactive"

### LINKED ISSUE #726: Timeouts and timeoutsdispatcher are registered as endpoints, but they don't generate heartbeats so always stay "in red"
This was originally raised in https://github.com/Particular/ServicePulse/issues/340

This happens because when a timeout fails is sent to the error queue and then ServiceControl picks it up and registers that as an endpoint.
And obviously we then do not have heartbeats for satellites and hence these are seen in error all the time.

SzymonPobiega:
This includes the original problem in #726 which was related to satellite receivers (which by definition don't send heartbeats) showing as "inactive" if there was a failure in processing that resulted in moving a message to the error queue.

There are two changes:
 * Remove the behavior which caused endpoints to be added to the monitored group immediately after they were detected (regardless of the detection source: HB, audit, error)
 * Enabling adding of endpoints to the monitored group when a HB is first detected or is restored

Fixes #726

### DIFF
--- a/src/ServiceControl.AcceptanceTests/HeartbeatMonitoring/When_an_unmonitored_endpoint_is_marked_as_monitored.cs
+++ b/src/ServiceControl.AcceptanceTests/HeartbeatMonitoring/When_an_unmonitored_endpoint_is_marked_as_monitored.cs
@@ -1,0 +1,105 @@
+﻿namespace ServiceBus.Management.AcceptanceTests.HeartbeatMonitoring
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using Contexts;
+    using NServiceBus;
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.Features;
+    using NUnit.Framework;
+    using ServiceControl.CompositeViews.Endpoints;
+    using Conventions = NServiceBus.AcceptanceTesting.Customization.Conventions;
+
+    public class When_an_unmonitored_endpoint_is_marked_as_monitored : AcceptanceTest
+    {
+        enum State
+        {
+            WaitingForEndpointDetection,
+            WaitingForHeartbeatFailure
+        }
+
+        static string EndpointName => Conventions.EndpointNamingConvention(typeof(MyEndpoint));
+
+        [Test]
+        public void It_is_shown_as_inactive_if_it_does_not_send_heartbeats()
+        {
+            var context = new MyContext();
+            List<EndpointsView> endpoints = null;
+            var state = State.WaitingForEndpointDetection;
+            Define(context)
+                .WithEndpoint<MyEndpoint>()
+                .Done(c =>
+                {
+                    if (state == State.WaitingForEndpointDetection)
+                    {
+                        var found = TryGetMany("/api/endpoints/", out endpoints, e => e.Name == EndpointName && !e.Monitored);
+                        if (found)
+                        {
+                            var endpointId = endpoints.First(e => e.Name == EndpointName).Id;
+                            Patch($"/api/endpoints/{endpointId}",new EndpointUpdateModel
+                            {
+                                MonitorHeartbeat = true
+                            });
+                            state = State.WaitingForHeartbeatFailure;
+                            Console.WriteLine("Patch successful");
+                        }
+                        return false;
+                    }
+                    return state == State.WaitingForHeartbeatFailure && TryGetMany("/api/endpoints/", out endpoints, e => e.Name == EndpointName && e.MonitorHeartbeat && e.Monitored && !e.IsSendingHeartbeats);
+                })
+                .Run();
+
+            var myEndpoint = endpoints.FirstOrDefault(e => e.Name == EndpointName);
+            Assert.NotNull(myEndpoint);
+            Assert.IsTrue(myEndpoint.Monitored);
+            Assert.IsTrue(myEndpoint.MonitorHeartbeat);
+            Assert.IsFalse(myEndpoint.IsSendingHeartbeats);
+        }
+
+        public class MyContext : ScenarioContext
+        {
+        }
+
+        public class MyEndpoint : EndpointConfigurationBuilder
+        {
+            public MyEndpoint()
+            {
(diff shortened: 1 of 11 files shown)


SzymonPobiega:

johnsimons:
Should we have an acceptance test for this ?

SzymonPobiega:
@johnsimons how about this:
 * `When_an_endpoint_without_heartbeat_plugin_starts` `It_is_not_monitored`
 * `When_not_monitored_endpoint_starts_sending_heartbeats` `It_shows_as_active`
 * `When_monitored_endpoint_stops_sending_heartbeat` `It_shows_as_inactive`
 * `When_an_endpoint_with_heartbeat_starts` `It_shows_as_active`

The problems I see here is that we are heavily time-dependent, especially with the second test. Do you think adding these test is worth the time on the build server? So far there is only one test for HB that ensures that when heartbeating endpoint starts, it is detected.

SzymonPobiega:
@johnsimons I added a couple of acceptance tests that cover most of the HB logic. I had to do a small change in the acceptance test base class to allow multiple rounds of `Define-Run-Assert` for the case when the endpoint first does not have HB and then starts sending it. 

I also added a property to the `RegisterEndpoint` message. It defaults to `false` so there should be no problem with (de)serialization.

src/ServiceControl.AcceptanceTests/HeartbeatMonitoring/When_an_unmonitored_endpoint_is_marked_as_monitored.cs
```diff
@@ -0,0 +1,105 @@
...
+                {
+                    if (state == State.WaitingForEndpointDetection)
+                    {
+                        var found = TryGetMany("/api/endpoints/", out endpoints, e => e.Name == EndpointName && !e.Monitored);
+                        if (found)
+                        {
+                            var endpointId = endpoints.First(e => e.Name == EndpointName).Id;
+                            Patch($"/api/endpoints/{endpointId}",new EndpointUpdateModel
+                            {
+                                MonitorHeartbeat = true
```
SzymonPobiega: @johnsimons well, an alternative would be to not allow to enable monitoring of an endpoint if we haven't received any HBs. I think I prefer this approach. Here, by force-enabling monitoring we say "I expect this endpoint is going to have HB plugin installed and working" and it will show in read until somebody actually installs and configures it.

johnsimons: not sure @SzymonPobiega 

SzymonPobiega: @johnsimons @Particular/servicecontrol-maintainers I guess this is the last question that blocks this PR.

johnsimons: well, I think we just merge it as it is.
What are the implications of this change has on SP ?

SzymonPobiega: None. The UI in SP stays the same. The only difference is that the endpoints that don't have HB won't show in red by default and they will be shown as `OFF` in the configuration tab.

johnsimons: And if they send this patch request, then they would be red again ?

SzymonPobiega: Yes. If you manually switch them to `ON` they will start showing up as red.

johnsimons:
👍 to merge.

@mikeminutillo do u want to also review this ?

src/ServiceControl.AcceptanceTests/HeartbeatMonitoring/When_an_unmonitored_endpoint_starts_sending_heartbeats.cs
```diff
@@ -0,0 +1,96 @@
...
+            var myEndpoint = endpoints.FirstOrDefault(e => e.Name == EndpointName);
+            Assert.NotNull(myEndpoint);
+            Assert.IsFalse(myEndpoint.Monitored);
+
+            Define(context)
+                .WithEndpoint<MyEndpointWithHeartbeat>()
+                .Done(c => TryGetMany("/api/endpoints/", out endpoints, e => e.Name == EndpointName && e.Monitored && e.MonitorHeartbeat && e.IsSendingHeartbeats))
+                .Run();
+
+            myEndpoint = endpoints.FirstOrDefault(e => e.Name == EndpointName);
```
mikeminutillo: Should this be SingleOrDefault? Just in case the two runs generate two endpoint records with the same name.

mikeminutillo:
Probably not something to fix in the current PR but `When_an_unmonitored_endpoint_starts_sending_heartbeats` relies on the Endpoint detection feature. Should we bypass this and just put Endpoint data into the DB at test startup?

src/ServiceControl.AcceptanceTests/HeartbeatMonitoring/When_an_unmonitored_endpoint_is_marked_as_monitored.cs
mikeminutillo: I think this is fine. If you force an endpoint to be monitored even though it doesn't have the heartbeat plugins installed, it shows as failing. I think that's the correct behavior.

mikeminutillo:
Review completed. I am :+1: for this. 

Should we hold off for a bit on releasing as we are going to send an email telling users to upgrade to 1.27 next week around the critical fix.

SzymonPobiega:
@mikeminutillo I pushed the fix.

PR #840, opened 2016-11-11
Improvements to retries

johnsimons:
- Evicting documents early
- Not sending bodies of message at staging time, instead we retrieve the body at forwarding time, this will save on bandwidth


johnsimons:
Based on a chat with @janovesk, here are a bunch of improvements to the retries pipeline.
@mikeminutillo thoughts ?

mikeminutillo:
I like it. Less I/O is always better.

PR #843, opened 2016-11-11
Add queue slot to waiting state, both api and event published

janovesk:


janovesk:
Add "Slot" to both the group API and the waiting event. It can be used to show the queue position of the retry operation. 0 is lowest. 

Have a look @WilliamBZA, @WojcikMike and @sergioc

janovesk:
I'll rebase this.

Issue #839, opened 2016-11-10
"Please select" instead of blank in Audit Forwarding dropdown


adamralph:
When I create a new instance via Management, the Audit Forwarding dropdown is pre-selected with a blank entry, which co-exists alongside the "On" and "Off" entries. ~~I see that the Error Forwarding dropdown only has "On" and "Off" with the latter pre-selected. This leads me to believe that the blank entry in the Audit Forwarding should be removed.~~ The reason for this is that there is no sensible default, so we force the user to select an entry. It would be better if this entry read "(Please select)" than be blank.

gbiellem:
@adamralph what've described is by design.  Error forwarding has a default of off as it's rarely a good idea to have it one.  So it's a sensible default. With audit forwarding though there is no sensible default.

adamralph:
@gbiellem ah, I see, thanks. I just tried to submit without selecting and indeed it tells me "Must select audit forwarding".

Would it be better for the entry to say "(Please select)" rather than be blank?

gbiellem:
Sure we could do that

adamralph:
Thanks @gbiellem, I've re-titled and altered the description to make this an improvement. Will send a PR.

adamralph:
I can't work out where the blank entry is coming from. I can only see the On and off entries in the ItemsSource: https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl.Config/UI/SharedInstanceEditor/SharedInstanceEditorViewModel.cs#L29

Can anyone help?

gbiellem:
@adamralph I also had a look. It's not straight forward.  Basically the blank entry is because we bind to a list but don't select an entry.  The hacky way to add the "Please Select" would be be to add another entry to the list with a null value but I think I've got a better way.  Leave it with me and I'll try and get my idea into a PR over the weekend

adamralph:
👍 thanks @gbiellem

gbiellem:
@adamralph take a look at the PR and see what you think.

PR #844, opened 2016-11-14
Fix Blank entries in dropdowns

gbiellem:
Related to https://github.com/Particular/ServiceControl/issues/839

Adjusted the control template for the FormComboBox so the DropDowns in the Add Instance screen show "Please Select.." instead of a blank entry


johnsimons:
@gbiellem is this ready ?

gbiellem:
@johnsimons Just waiting on feedback from @adamralph before merging- See https://github.com/Particular/ServiceControl/issues/839#issuecomment-260244824

PR #846, opened 2016-11-14
[SPIKE] Using websockets to transmit messages

johnsimons:


gbiellem:
@mikeminutillo  Is there any plan to continue with this or should it be cleaned up?

PR #851, opened 2016-11-15
[WIP] Migrations

mikeminutillo:
- Add the concept of migrations
- Add a migration to detect problematic failed messages and split them up

### DIFF
--- a/src/ServiceControl/DbMigrations/1.27/SplitFailedMessageDocumentsMigration.cs
+++ b/src/ServiceControl/DbMigrations/1.27/SplitFailedMessageDocumentsMigration.cs
@@ -85,6 +85,11 @@
                     failedMessage.ProcessingAttempts.Remove(attempt);
                 }
             }
+
+            if (failedMessage.ProcessingAttempts.Count == 0)
+            {
+                session.Delete(failedMessage);
+            }
         }
 
         class ProcessingAttemptRecord
--- a/src/ServiceControl/DbMigrations/1.27/SplitFailedMessageDocumentsMigration.cs
+++ b/src/ServiceControl/DbMigrations/1.27/SplitFailedMessageDocumentsMigration.cs
@@ -1,17 +1,115 @@
 ﻿namespace Particular.ServiceControl.DbMigrations
 {
+    using System.Linq;
+    using global::ServiceControl;
+    using global::ServiceControl.MessageFailures;
+    using global::ServiceControl.Operations;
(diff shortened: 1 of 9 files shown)


mikeminutillo:
@johnsimons thoughts?

I need to write some tests for these

src/ServiceControl/DbMigrations/1.27/SplitFailedMessageDocumentsMigration.cs
```diff
@@ -0,0 +1,115 @@
...
+                {
+                    Check(failedMessage, session);
+                }
+
+            } while (retrievedResults == PageSize);
+        }
+
+        private void Check(FailedMessage failedMessage, IDocumentSession session)
+        {
+            var originalRecords = failedMessage.ProcessingAttempts.Select(x => new ProcessingAttemptRecord(x)).ToArray();
```
johnsimons: don't we need to filer based on `Unresolved` only ?

src/ServiceControl/DbMigrations/1.27/SplitFailedMessageDocumentsMigration.cs
```diff
@@ -0,0 +1,115 @@
...
+                {
+                    // We've retried to this endpoint so it's the primary
+                    continue;
+                }
+
+                var newFailedMessage = new FailedMessage
+                {
+                    Id = FailedMessage.MakeDocumentId(group.Key),
+                    UniqueMessageId = group.Key,
+                    ProcessingAttempts = group.Select(x => x.Attempt).OrderBy(x => x.FailureDetails.TimeOfFailure).ToList(),
```
johnsimons: I don't get this? When would there be more then one attempt ?

src/ServiceControl/DbMigrations/1.27/SplitFailedMessageDocumentsMigration.cs
mikeminutillo: What if we have a FailedMessage that we retried but it has one of these duplicates in it? Shouldn't we pull the duplicate out?

src/ServiceControl/DbMigrations/1.27/SplitFailedMessageDocumentsMigration.cs
mikeminutillo: When we've performed at least one retry there will be multiple processing attempts

Issue #849, opened 2016-11-14
Error and audit queue names are enforced when not required


adamralph:
When creating a new instance, I'm forced to enter queue names for error forwarding and audit forwarding even though I'm leaving those options off.

Normally, I wouldn't care, but when I was adding a second SC instance (for testing a change in SI) I had to manually edit it them because they clashed with the existing instance.

Would it be possible to make to so that those queue names can be omitted/ignored if error forwarding and audit forwarding are switched off?

gbiellem:
The historical reason for this working this way is that if the queue isn't created as setup time then toggling the setting from off to on directly via an app.config will result in an error and the service won't start and
we don't create the queues on demand at service startup as there is no guarantee the service account has the rights to do so.

If these setting were not exposed via app config I'd be 👍 for changing this.  Otherwise I'm 👎

johnsimons:
@gbiellem given that we want everyone to not update the config file manually but instead go via  management tool, should we rethink this and actually let it fail ?

gbiellem:
@johnsimons If you're suggesting pulling those settings out of app.config and storing them elsewhere (maybe in the DB) then I'd be all for changing this to the behavior @adamralph has suggested

johnsimons:
> If you're suggesting pulling those settings out of app.config and storing them elsewhere

no I am not suggesting that.

All I am saying is that even thought users can manually update the config file and make changes that would break ServiceControl, we are not really gaining much for enforcing Error and audit queue names even when we are turning them off, eg, if user decides to change the forwarding queues to something else manually, they would break it since the queues do not exist.

So I am for removing this enforcement and let the system fail to start if they decide to manually turn it on, instead of using the management tool.

gbiellem:
@johnsimons  and I got on a zoom call and discussed this further.  He has won me over so now I'm 👍 for modifying this to the behavior outlined by @adamralph and leaving the setting in the app.config for the time being. Removing those should be a separate exercise.  I'll put together a PR

gbiellem:
Implemented in #852

PR #852, opened 2016-11-15
Don't create forwarding queues if option is disabled

gbiellem:
Related to #849

When adding a new instance and forwarding is turned off don't create the queues.  

### Mgmt UI Changes

The Add Instance UI will only ask and validate Error forwarding queue name or Audit Forwarding queue name if that related combo box is "on"

The Edit Instance screen will feature switch because of the different behavior in old versions.  
 - For version prior to 1.12 the forwarding queues are always displayed and the ErrorForwarding combo is not shown
 - For version 1.12 to the version prior to this release the ErrorForwarding combo is always shown and the  forwarding queues are always displayed and created
- For this version and above forwarding queues are only displayed 

### Mgmt PowerShell Changes.  

Cmdlets for adding ServiceControl instances no longer require the forwarding queue name if the forwarding option is off

### Config changes

For new ServiceControl instances or ServiceControl instances upgraded by this release:

- the 'ServiceBus/ErrorLogQueue' is omitted if 'ServiceControl/ForwardErrorMessages' is set to 'false'
- the 'ServiceBus/AuditLogQueue' is omitted if 'ServiceControl/ForwardAuditMessages' is set to 'false'

### ServiceControl service changes

- The queue creation for forwarding queues durign setup is only executed if the forwarding is enabled
- The empty test message sent to the forwarding queues at startup has been removed

### DIFF
--- a/src/ServiceControl.Config/ServiceControl.Config.csproj
+++ b/src/ServiceControl.Config/ServiceControl.Config.csproj
@@ -254,6 +254,7 @@
     <Compile Include="Xaml\Controls\AppHeader.xaml.cs">
       <DependentUpon>AppHeader.xaml</DependentUpon>
     </Compile>
+    <Compile Include="Xaml\Controls\FormWarningTextBlock.cs" />
     <Compile Include="Xaml\Controls\FormSlider.cs" />
     <Compile Include="Xaml\Controls\FormPasswordBox.cs" />
     <Compile Include="Xaml\Controls\FormPathTextBox.cs" />
--- a/src/ServiceControl.Config/ServiceControl.Config.csproj
+++ b/src/ServiceControl.Config/ServiceControl.Config.csproj
@@ -196,6 +196,7 @@
     </Compile>
     <Compile Include="UI\MessageBox\YesNoCancelDialogViewModel.cs" />
     <Compile Include="UI\MessageBox\ReportCardViewModel.cs" />
+    <Compile Include="UI\SharedInstanceEditor\ForwardingOption.cs" />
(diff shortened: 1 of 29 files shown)


gbiellem:
@johnsimons  review?

johnsimons:
so small 😲

gbiellem:
Yeah I know... It sounded so simple in #849 when first proposed and it looks more complicated than it is.
Most of the changes are due to the edit and add screens sharing a base class. I've moved large chunks out of the common model to accommodate the change.  Which means I also had to move chunks of validation etc...

johnsimons:
@gbiellem I don't think this is related to this change but if I shrink the width of the form the text gets chopped:
image

johnsimons:
Also, I wonder if it would be possible to make the gap between the fields and the warning smaller ?
image

gbiellem:
@johnsimons  good pickup on both counts - I'll take a look

johnsimons:
fields are not lined up
image

johnsimons:
The tab order stops on the warning icon 
image

gbiellem:
@johnsimons  Can't repro the alignment issue.  A little help please.

alignment

johnsimons:
@gbiellem how come we still need UI\SharedInstanceEditor\SharedInstanceEditorView.xaml.cs ?

gbiellem:
@johnsimons I was trying to only change what was necessary for this fix.  As you already noted it was already getting big.

johnsimons:
Everything else looks ok to me

gbiellem:
Cool. OK to merge then?

PR #857, opened 2016-11-18
Fix collapsed subscribers noretention

tmasternak:


src/ServiceControl/DbMigrations/1.27/SplitFailedMessageDocumentsMigration.cs
```diff
@@ -1,156 +1,179 @@
...
+                        failedMessage.FailureGroups = failedMessageFactory.GetGroups((string)messageType, lastAttempt.FailureDetails);
+                    }
+                    failedMessage.FailureGroups.Add(CreateSplitFailureGroup(lastAttempt, messageType, originalStatus));
+                }
 
-            writer.WriteLine();
+                if (failedMessage.UniqueMessageId == originalFailedMessage.UniqueMessageId) return;
+
+                session.Store(failedMessage);
+                stats.Created++;
```
mikeminutillo: This counts the original document as well if it is still around. I'm not opposed to that, just noting it.

src/ServiceControl/MessageFailures/FailedMessagesFeature.cs
```diff
@@ -54,12 +54,14 @@ public override void Enrich(IReadOnlyDictionary<string, string> headers, IDictio
                 if (isOldRetry)
                 {
                     bus.Publish<MessageFailureResolvedByRetry>(m => m.FailedMessageId = headers.UniqueId());
+                    return;
                 }
 
-                if (isNewRetry)
+                bus.CurrentMessageContext.Headers.Add("ServiceControl.Retry.UniqueMessageId", headers.UniqueId());
```
mikeminutillo: Does this work? I believe that `CurrentMessageContext.Headers` refers to the headers on the message being processed, not the headers on outgoing messages.

This type is OK to add a property to because although it is in the Contracts namespace it is not actually exposed externally. The only types we don't want to make changes to are the ones found in the [`ServiceControl.Contracts` assembly](https://github.com/Particular/ServiceControl.Contracts/tree/master/src/ServiceControl.Contracts).

tmasternak:
The two failing test `It_should_mark_failedmessage_resolved_when_it_succeeds_with_old_header` and `It_should_mark_failedmessage_resolved_when_it_succeeds_with_new_header` are failing for the same reason (ASB sets ReplyToAddress differently on audit messages than other transports).
I think this is good enough even not working for ASB and would vote for disabling it.

johnsimons:
> I think this is good enough even not working for ASB and would vote for disabling it.

So is this fix working for ASB or not ?

src/ServiceControl/MessageFailures/Handlers/MessageFailureResolvedHandler.cs
```diff
@@ -59,7 +62,7 @@ public void Handle(MarkPendingRetriesAsResolved message)
             }
         }
 
-        private void MarkMessageAsResolved(string failedMessageId)
+        private MarkMessageAsResolvedStatus MarkMessageAsResolved(string failedMessageId)
```
johnsimons: `bool TryMarkMessageAsResolved` ?

mikeminutillo:
> So is this fix working for ASB or not ?

No. It is not. The problem is that ASB **doesn't** override the `ReplyToAddress` on Audited messages. It looks like the transport is doing something different to preserve that value.

src/ServiceControl/MessageFailures/FailedMessagesFeature.cs
```diff
@@ -51,15 +53,25 @@ public override void Enrich(IReadOnlyDictionary<string, string> headers, IDictio
                     return;
                 }
 
-                if (isOldRetry)
-                {
-                    bus.Publish<MessageFailureResolvedByRetry>(m => m.FailedMessageId = headers.UniqueId());
-                }
+                var altHeaders = AuditHeadersToFailedMessageHeaders(headers);
```
mikeminutillo: Agreed. I'll update it to just hard-code these Ids

- MessageId, ProcessingEndpoint
- MessageId, ReplyToAddress

johnsimons:
> No. It is not. The problem is that ASB doesn't override the ReplyToAddress on Audited messages. It looks like the transport is doing something different to preserve that value.

So what does this mean ? Are we going to put a fix out for ASB ?

mikeminutillo:
I'm investigating. 

The issue is error messages have a Unique Message Id that is based on Message Id and Failed Q. Failed Q is the physical address of the processing endpoint. We don't have the physical address of the processing endpoint for audited messages. For most transports, that's OK because the v5 Audit pipeline overrides the ReplyToAddress of audited messages to be the physical address of the processing endpoint. This allows us to fall back to using using Message Id and ReplyToAddress to correlate an incoming Audit message with the Error message that triggered it. 

In ASB, the audited message has the correct ReplyToAddress which points to the physical address of the sender. There is no header with a physical address that points to the processing endpoint.

PR #860, opened 2016-11-23
Use existing classifiers

mikeminutillo:
This PR adds a prefix onto all groups for split documents. This prevents us from having to have a different failure group classifier type and allows customers to proceed without updating ServicePulse.


mikeminutillo:
@johnsimons is this what you meant?

PR #863, opened 2016-11-25
Manually subscribe to events to stop racing conditions

### LINKED ISSUE #859: Native Pub/Sub Acceptance Tests assume Sub too early
When the SC Acceptance Tests run under a transport with native pub/sub, they occasionally fail because they publish the event before all subscriptions are in place. Typically there is code in place to prevent this by setting a boolean flag in the Scenario Context when the endpoint starts and then checking this flag before publishing the event.

Here is an example taken from [one of the tests](https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl.AcceptanceTests/MessageFailures/When_an_event_with_multiple_subscribers_fails.cs#L25)

```CSharp
Define(context)
	.WithEndpoint<FailingSubscriber1>(behavior => behavior.Given((bus, ctx) =>
	{
		if (ctx.HasNativePubSubSupport)
		{
			ctx.Subscriber1Subscribed = true;
		}
	}))
	.WithEndpoint<FailingSubscriber2>(behavior => behavior.Given((bus, ctx) =>
	{
		if (ctx.HasNativePubSubSupport)
		{
			ctx.Subscriber2Subscribed = true;
		}
	}))
	.WithEndpoint<Publisher>(behavior => behavior
			.CustomConfig(cfg => cfg.OnEndpointSubscribed(sub =>
				{
					if (sub.SubscriberReturnAddress.Queue.Contains("Subscriber1"))
					{
						context.Subscriber1Subscribed = true;
					}
					else if (sub.SubscriberReturnAddress.Queue.Contains("Subscriber2"))
					{
						context.Subscriber2Subscribed = true;
					}
					else
					{
						throw new Exception($"Unknown subscriber subscribed to publisher: {sub.SubscriberReturnAddress}");
					}
				})
			).When(
				ctx => ctx.Subscriber1Subscribed && ctx.Subscriber2Subscribed,
				bus => bus.Publish<SampleEvent>()
			)
```

The problem is that the flag is set in a `Given` and these are executed [before the endpoint starts](https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl.AcceptanceTesting/Support/EndpointRunner.cs#L123). This is before the subscription has been put into place.

We can move these to a `When` and use the scenario context flag `EndpointsStarted` to set the subscriber flag. The `EndpointsStarted` flag gets set [after the endpoints have started](https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl.AcceptanceTesting/Support/ScenarioRunner.cs#L217). But the subscriptions may STILL not have been set up at this point. 

Subscriptions are configured in a [feature startup task](https://github.com/Particular/NServiceBus/blob/support-5.2/src/NServiceBus.Core/AutomaticSubscriptions/AutoSubscribe.cs#L59) and the startup [does not wait for those to finish](https://github.com/Particular/NServiceBus/blob/support-5.2/src/NServiceBus.Core/Unicast/UnicastBus.cs#L680) before setting the flag!

johnsimons:
Fixes #859

### DIFF
--- a/src/ServiceControl.AcceptanceTests/Contexts/DefaultServerWithAudit.cs
+++ b/src/ServiceControl.AcceptanceTests/Contexts/DefaultServerWithAudit.cs
@@ -15,6 +15,7 @@
     using NServiceBus.AcceptanceTesting.Support;
     using NServiceBus.Config.ConfigurationSource;
     using NServiceBus.Configuration.AdvanceExtensibility;
+    using NServiceBus.Features;
     using NServiceBus.Hosting.Helpers;
     using ServiceBus.Management.AcceptanceTests.Contexts.TransportIntegration;
 
@@ -43,6 +44,7 @@
             });
             var builder = new BusConfiguration();
 
+            builder.DisableFeature<AutoSubscribe>();
             builder.UsePersistence<InMemoryPersistence>();
             builder.EndpointName(endpointConfiguration.EndpointName);
             builder.TypesToScan(types);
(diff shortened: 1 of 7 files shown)


johnsimons:
Is this it @mikeminutillo ?

mikeminutillo:
Looks good to me

PR #864, opened 2016-11-28
Use Opacity to mean disable

### LINKED ISSUE #788: We use grey for both an action button and a disable button
So which one is enabled?

image

johnsimons:
Fixed #788


johnsimons:
@HEskandari can you please review this change, it is to implement - Reducing opacity to 30%

HEskandari:
@johnsimons Seems simple enough. LGTM.

PR #865, opened 2016-11-29
Do deletes outside paging

johnsimons:
Otherwise the paging numbers change and it becomes much harder to recalculate

### DIFF
--- a/src/ServiceControl.UnitTests/Migrations/1.27/SplitFailedMessageDocumentsMigrationTests.cs
+++ b/src/ServiceControl.UnitTests/Migrations/1.27/SplitFailedMessageDocumentsMigrationTests.cs
@@ -98,7 +98,6 @@
         public void Should_handle_larger_than_pagesize_number_of_failedmessages()
         {
             const int multiplier = 2;
-
             using (var session = documentStore.OpenSession())
             {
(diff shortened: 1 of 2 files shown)


johnsimons:
@mikeminutillo I think this is it ?

tmasternak:
@johnsimons is this fixing statistics or is it a fix to the logic?

mikeminutillo:
It fixes a bug in the pagination. It should go in before sending the email.

The code looks right to me @johnsimons 

My only concern here is that by not deleting the old document in the same session as we create the new ones we might fail part way through a migration and have a bunch of FailedMessage documents lying around with no ProcessingAttempts on them. 

Having said that, if we fail partway through a migration, SC won't finish the update process.

mikeminutillo:
@johnsimons Are we going to make this 1.28 or do we release 1.27.1?

johnsimons:
> or do we release 1.27.1?

what is that ? You know we don't do those 😉

mikeminutillo:
Well in this case I thought we might make an exception as the only thing going in to a release a bug fix to a fix that was added to the previous release.

johnsimons:
@mikeminutillo 
no is not https://github.com/Particular/ServiceControl/milestone/69

johnsimons:
the train always rolls forward

PR #870, opened 2016-12-08
[WIP] Prepare SC for transport adapters

SzymonPobiega:
This PR prepares SC to add transport adapter layer in order to redesign the internal workings of SC later on. It should not change any observable behaviors for existing users, both normal endpoints as well as integration event consumers.

### Integration events dispatching

Removes the integration event dispatcher logic which was about storing integration events in RavenDB before publishing them. The original purpose of this logic was to make sure we don't lose any outgoing integration evens and decouple their publishing from other stuff in SC. In the meantime the code has been change so that a failed publishing of integration event is never re-tried. Instead a notification is generated and displayed on SP front page. This change removes the whole thing and replaces it with a plain old send/publish in a `try...catch` clause with standard `error` logging

### Integration events sending

Adds a config section which can define "adapter" addresses to which the integration events are being sent (in addition to being published to all subscribers). This change allows to deploy adapters and, in future, remove the feature of publishing the integration events in SC and move it to the adapter layer. In future this config section will be configured via the SC admin UI.

### RetryTo header

Adds handling of `ServiceControl.RetryTo` header when retrying a failed message. If a failed message contains that header, it is used instead of `ServiceControl.TargetEndpointAddress` as a destination. In that case the `ServiceControl.TargetEndpointAddress` is preserved on the message so that the adapter can actually forward the message to the ultimate destination


SzymonPobiega:
I've re-raised a [bigger issue](https://github.com/Particular/PlatformDevelopment/issues/399) of decoupling the SC from the transports in PlatDev. I consider this one as a technical first step that allows it but also allows us to move away from it. I don't expect us announcing this one in public before the parent PlatDev TF does not figure out the details.

SzymonPobiega:
Should wait until https://github.com/Particular/ServiceControl/pull/877 is merged.

src/ServiceControl/EventLog/Definitions/ExternalIntegrationEventFailedToBePublishedDefinition.cs
```diff
@@ -1,13 +0,0 @@
-namespace ServiceControl.EventLog.Definitions
-{
-    using ServiceControl.ExternalIntegrations;
-
-    class ExternalIntegrationEventFailedToBePublishedDefinition : EventLogMappingDefinition<ExternalIntegrationEventFailedToBePublished>
```
mikeminutillo: This has just been replaced with a log message right? Which means we don't send it to Pulse and customers won't necessarily know it has happened.

src/ServiceControl/ExternalIntegrations/Config/AdapterElement.cs
```diff
@@ -0,0 +1,14 @@
...
+{
+    using System.Configuration;
+
+    public class AdapterElement : ConfigurationElement
+    {
+        [ConfigurationProperty("Name", IsRequired = true)]
+        public string Name
+        {
+            get { return (string)this["Name"]; }
+            set { this["Name"] = value; }
```
mikeminutillo: Are we assuming that these adapters will only ever be NServiceBus addresses? Is it too early to consider other types of adapters?

src/ServiceControl/ExternalIntegrations/CustomCheckFailedPublisher.cs
```diff
@@ -1,51 +1,21 @@
...
-    using System.Linq;
-    using Raven.Client;
     using CustomCheckFailed = ServiceControl.Contracts.CustomCheckFailed;
 
-
-    public class CustomCheckFailedPublisher : EventPublisher<Contracts.CustomChecks.CustomCheckFailed, CustomCheckFailedPublisher.DispatchContext>
+    public class CustomCheckFailedPublisher : EventPublisher<Contracts.CustomChecks.CustomCheckFailed, CustomCheckFailed>
     {
-        protected override DispatchContext CreateDispatchRequest(Contracts.CustomChecks.CustomCheckFailed @event)
+        protected override CustomCheckFailed Convert(Contracts.CustomChecks.CustomCheckFailed message)
```
mikeminutillo: This is good because it will catch existing instances that are in transit during an update. Would it make sense to go back a step and just publish the original external integration event at the point that we published the event being published here? Or should that happen in a future update?

src/ServiceControl/ExternalIntegrations/EventMappingHandler.cs
```diff
@@ -1,49 +0,0 @@
...
-    {
-        private readonly IDocumentStore store;
-        private readonly IEnumerable<IEventPublisher> eventPublishers;
-
-        public EventMappingHandler(IDocumentStore store, IEnumerable<IEventPublisher> eventPublishers)
-        {
-            this.store = store;
-            this.eventPublishers = eventPublishers;
-        }
-        public void Handle(IEvent message)
```
mikeminutillo: YAY! Getting rid of the handlers for IEvent makes me eternally happy.

src/ServiceControl/ExternalIntegrations/ExternalIntegrationsFeature.cs
```diff
@@ -1,24 +1,26 @@
...
-            var eventPublisherTypes = context.Settings.GetAvailableTypes().Implementing<IEventPublisher>();
+            var adaptersConfig = AdapterConfigurationSection.GetAdapters();
 
-            foreach (var eventPublisherType in eventPublisherTypes)
-            {
-                context.Container.ConfigureComponent(eventPublisherType, DependencyLifecycle.SingleInstance);
-            }
+            var adapters = adaptersConfig.Adapters.Cast<AdapterElement>().Select(e => e.Name);
+
+            var integrationEndpoints = adapters.Select(a => $"{a}.integration").ToArray();
```
mikeminutillo: Are we adding a suffix on to the address that our customers provided? Why? What if they provided an address with a machine name? Shouldn't we use `Address.Parse(a).SubScope("integration")` if this is something we want to do?

src/ServiceControl/ExternalIntegrations/IntegrationEventSender.cs
```diff
@@ -0,0 +1,44 @@
...
+            this.bus = bus;
+        }
+
+        public void Send(object evnt)
+        {
+            foreach (var queue in integrationForwardQueues)
+            {
+                try
+                {
+                    bus.Send(Address.Parse(queue), evnt);
```
mikeminutillo: Is there an issue with sending an event (rather than publishing it)? Do we need to disable the best practices analyzer (which I wouldn't be opposed to in SC anyway).

src/ServiceControl/ExternalIntegrations/IntegrationEventSender.cs
```diff
@@ -0,0 +1,44 @@
...
+                }
+                catch (Exception e)
+                {
+                    log.Error($"Failed dispatching external integration event to {queue}.", e);
```
mikeminutillo: So customers need to check the log file to see that there is an integration event issue?

src/ServiceControl/Infrastructure/NServiceBusFactory.cs
```diff
@@ -42,7 +42,7 @@ public static IStartableBus Create(Settings.Settings settings, IContainer contai
 
             var transportType = DetermineTransportType(settings);
 
-            configuration.Conventions().DefiningEventsAs(t => typeof(IEvent).IsAssignableFrom(t) || IsExternalContract(t));
+            configuration.Conventions().DefiningMessagesAs(t => ImplementsMessageInterface(t) || IsExternalContract(t));
```
mikeminutillo: Is to get around the issue with sending events? I'm not sure why this changed.

src/ServiceControl/Recoverability/Retries/Infrastructure/ReturnToSenderDequeuer.cs
```diff
@@ -104,10 +101,17 @@ void HandleMessage(TransportMessage message)
                 }
                 message.Headers.Remove("ServiceControl.Retry.Attempt.MessageId");
             }
-
+            var destination = message.Headers["ServiceControl.TargetEndpointAddress"];
```
mikeminutillo: Is this change pulled in from #877 ? If so I'll take a look at it there.

WilliamBZA:
@SzymonPobiega #877 has been merged, what do we need to push this forward?

SzymonPobiega:
@WilliamBZA I'll close this one but keep the branch for some time for reference.

PR #877, opened 2016-12-19
Allows overriding retry address in failed messages

SzymonPobiega:
Adds handling of `ServiceControl.RetryTo` header when retrying a failed message. If a failed message contains that header, it is used instead of `ServiceControl.TargetEndpointAddress` as a destination. In that case the `ServiceControl.TargetEndpointAddress` is preserved on the message so that the adapter can actually forward the message to the ultimate destination

### DIFF
--- a/src/ServiceControl.UnitTests/Recoverability/ReturnToSenderDequeuerTests.cs
+++ b/src/ServiceControl.UnitTests/Recoverability/ReturnToSenderDequeuerTests.cs
@@ -1,0 +1,148 @@
+﻿namespace ServiceControl.UnitTests.Operations
+{
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Text;
+    using NServiceBus;
+    using NServiceBus.Transports;
+    using NServiceBus.Unicast;
+    using NUnit.Framework;
+    using ServiceControl.Operations.BodyStorage;
+    using ServiceControl.Recoverability;
+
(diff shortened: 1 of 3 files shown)


SzymonPobiega:

SzymonPobiega:
@mikeminutillo in order for the adapters to work, the original `FailedQ` address needs to be preserved so I thought this will be the least invasive method to do it.

When the message contains `ReplyTo` then the target header should be preserved. Otherwise, it should be removed. If the `ReplyTo` logic should go to the staging we would have to:
 * Set `ServiceControl.TargetEndpointAddress` to value from `ReplyTo`
 * Set some other header to `addressOfFailingEndpoint`

Also, the override should be applied before the replyto because you still want to override the input queues of individual endpoints, even if the retries are going to be handled via an adapter.

SzymonPobiega:
@mikeminutillo ping

SzymonPobiega:
@mikeminutillo well, it overrides where the message is sent on the wire but it does not override the logical destination. If `RetryTo` is present, the actual message is sent to the queue where `RetryTo` points but the destination (FailedQ or outcome of redirection) is passed as a header.

For now it does not have to be documented because if you don't pass that header, everything works as expected. I plan to add documentation gradually.

mikeminutillo:
Removed from milestone as this is an internal refactoring intended to support transport adapters later. Not to be included in the release notes

Issue #878, opened 2016-12-20
ServiceControl crashes because of TimeoutException


SzymonPobiega:
### Symptoms

ServiceControl process is killed due to an unhandled exception in the transport (RabbitMQ)

### Affected users

Users who run ServiceControl on broker transports that can experience connectivity problems (MSMQ, RabbitMQ, ASB, ASQ)

### Root cause

The `ServiceControl.HeartbeatMonitoring.HeartbeatsMonitor+HeartbeatMonitor.UpdateStatuses()` method sends messages without proper defensive code which causes the exception to be escalated to the timer and effectively kill the application.

### Logs

```
Application: ServiceControl.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: System.TimeoutException
   at RabbitMQ.Util.BlockingCell.GetValue(System.TimeSpan)
   at RabbitMQ.Client.Impl.SimpleBlockingRpcContinuation.GetReply(System.TimeSpan)
   at RabbitMQ.Client.Impl.ModelBase.ModelRpc(RabbitMQ.Client.Impl.MethodBase, RabbitMQ.Client.Impl.ContentHeaderBase, Byte[])
   at RabbitMQ.Client.Framing.Impl.Model._Private_ChannelOpen(System.String)
   at RabbitMQ.Client.Framing.Impl.Connection.CreateModel()
   at NServiceBus.Transports.RabbitMQ.ConfirmsAwareChannel..ctor(RabbitMQ.Client.IConnection, Boolean, System.TimeSpan)
   at NServiceBus.Transports.RabbitMQ.ChannelProvider.GetNewPublishChannel()
   at NServiceBus.Transports.RabbitMQ.RabbitMqMessageSender.Send(NServiceBus.TransportMessage, NServiceBus.Unicast.SendOptions)
   at NServiceBus.DispatchMessageToTransportBehavior.SendOrDefer(NServiceBus.TransportMessage, NServiceBus.Unicast.SendOptions)
   at NServiceBus.DispatchMessageToTransportBehavior.InvokeNative(NServiceBus.Unicast.DeliveryOptions, NServiceBus.TransportMessage)
   at NServiceBus.DispatchMessageToTransportBehavior.Invoke(NServiceBus.Pipeline.Contexts.OutgoingContext, System.Action)
[...]

PR #879, opened 2016-12-20
Fix missing heartbeat detection

SzymonPobiega:
Added `try...catch` around sending potential missing heartbeat messages. Because looking for potential missing heartbeats is retried every 5 seconds, there is not much we can/should do in case of send failure. Bringing down the whole endpoint via `CriticalError` seems to not be the best idea here.

Fixes https://github.com/Particular/ServiceControl/issues/878


SzymonPobiega:
@Particular/servicecontrol-maintainers please review. Also, I've went through other usages of `TimeKeeper` and noticed they action passed to `New` or `NewTimer` does not have a `try...catch` around the logic which means that whenever an error occurs, the action is never ever scheduled again. Could that cause any problems? I've added that kind of top level `try...catch` here.

SzymonPobiega:
@Particular/servicecontrol-maintainers ping

PR #880, opened 2016-12-21
Halt return to sender on shutdown

mikeminutillo:


SzymonPobiega:
I've verified this on my local machine and this version does have a problem with the deadlock when shutting down.

SzymonPobiega:
@mikeminutillo what's the status of this one?

mikeminutillo:
Just waiting on the acceptance tests and then this one is good to go

PR #882, opened 2016-12-22
Prevent dispatch getting stuck if indexes are stale

mikeminutillo:
The event dispatcher an trigger the `signal` before the indexes are up to date and miss the external integration event. This change will check for this condition and keep trying until the index catches up.

### DIFF
--- a/src/ServiceControl/ExternalIntegrations/EventDispatcher.cs
+++ b/src/ServiceControl/ExternalIntegrations/EventDispatcher.cs
@@ -28,6 +28,7 @@
         private IDisposable subscription;
         private Task task;
         private CancellationTokenSource tokenSource;
+        private Etag latestEtag = Etag.Empty;
 
         public EventDispatcher(IDocumentStore store, IBus bus, CriticalError criticalError, Settings settings, IEnumerable<IEventPublisher> eventPublishers)
         {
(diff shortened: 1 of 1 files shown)


SzymonPobiega:
@mikeminutillo the current code waits for a second if there are no results. The new one will hammer RavenDB until the index is up to date. Will this solve the problem of stuck integration events? I mean, retrying every second should cover the index stale problems.

mikeminutillo:
This version should wait for a second even if the index is stale. The only change should be 

> if the index is empty *and is stale* then there is more data

Previously we had

> if the index is empty then there's no more data

SzymonPobiega:
@mikeminutillo but that means that if the index is stale we will continue polling it in a tight loop, right?

mikeminutillo:
@SzymonPobiega We'll continue polling but with a 1 second pause between queries. 

When the "changes" subject fires, it should unlock the `DispatcherTask` which calls `DispatchEvents`. This method will call `TryDispatchEventBatch` which returns a boolean indicating whether or not there are more events. If there are more events (and we aren't shutting down) then the we pause for one second and then call `TryDispatchEventBatch` again. If there aren't more events then fall all the way back to the `DispatcherTask` and block on the signal again.

All this change does is affect the question "are there more events?". When the "changes" subject fires we record the etag for the change that triggered it. If, when we query, if the index is not at least up to this point, then we assume there are more events. That should lead to the loop in `DispatchEvents` to pause for a second and then query again. Because we are waiting for a specific etag to pass, we should eventually catch up before we relatch on the blocking signal.

SzymonPobiega:
@mikeminutillo sorry, I misread the surroundings of line 137 :-( Now it is all clear.

Issue #883, opened 2016-12-24
Validation on add instance throws NRE if Audit Forwarding is not set


gbiellem:
A Null  Reference Exception is being thrown by the validation on the Add Instance screen of version 1.31 of the Management Utility. This only occurs in the Audit Forwarding dropdown is not set to on or off.

PR #884, opened 2016-12-24
Fix NRE in Mgmt Util Add Instance screen

gbiellem:
Related to #883 

Validation on Add Instance screen throws NRE in Audit Forwarding in not set to On or Off when Submitted
Corrected to show the validation message when not selected.

Initial report of the issue was through the unhandled exception screen logged by an end user through Raygun


gbiellem:
@mikeminutillo  please review.  Think this should be released straight away

Issue #872, opened 2016-12-09
Remove integration events dispatcher


SzymonPobiega:
Remove the complexity associated with publishing integration events via RavenDB collection. Instead of publishing the events, send them to well-known (configurable via `app.settings`, UI in future) list of queues.

Connects to https://github.com/Particular/ServiceControl/issues/872

### Customer communication

As part of release `{1.x}` we are changing the [integration event](https://docs.particular.net/servicecontrol/contracts) delivery method. Instead of relying on Publish/Subscribe this version uses a list of recipient addresses that can be configured in `ServiceControl` config file:

```
<configSections>
  <section name="Adapters" type="ServiceControl.ExternalIntegrations.Config.AdapterConfigurationSection, ServiceControl"/>
</configSections>
<Adapters>
  <add Name="MyIntegrationEventConsumer@SomeMachine"/>
</Adapters>
```

Currently the addresses in the `Adapters` collection should point to your existing consuming endpoints. This is, however, the very first step on the way to introduce ServiceControl transport adapters which, among other things, will allow using multiple transports with a single instance of ServiceControl.

PR #885, opened 2016-12-30
Allow an external ServiceControl.Contracts.MessageFailed event with a Status of MessageStatus.ArchivedFailure to be published per the documentation

brdmllr:
We are unable to receive a ServiceControl.Contracts.MessageFailed event with a Status of MessageStatus.ArchivedFailure per the documentation linked below. We engaged support to review this issue (in conjunction with another issue) and uncovered the root cause when reviewing the ServiceControl code during our troubleshooting. A fix is contained in this pull request.

https://docs.particular.net/servicepulse/event-types#messagefailed

### DIFF
--- a/src/ServiceControl/ExternalIntegrations/FailedMessageArchivedPublisher.cs
+++ b/src/ServiceControl/ExternalIntegrations/FailedMessageArchivedPublisher.cs
@@ -1,0 +1,29 @@
+﻿namespace ServiceControl.ExternalIntegrations
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using Contracts.MessageFailures;
+    using Raven.Client;
+
+    public class FailedMessageArchivedPublisher : EventPublisher<FailedMessageArchived, MessageFailedPublisher.DispatchContext>
+    {
+        protected override MessageFailedPublisher.DispatchContext CreateDispatchRequest(FailedMessageArchived @event)
+        {
+            return new MessageFailedPublisher.DispatchContext
+            {
+                FailedMessageId = new Guid(@event.FailedMessageId)
+            };
+        }
+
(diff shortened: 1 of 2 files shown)


mikeminutillo:
Thanks @brdmllr. We're reviewing this and we'll get back to you. 

One thing to note is that ServiceControl allows bulk operations on groups of messages. In those cases, the message archived event is not published. Instead, a different event is published (`FailedMessageGroupArchived`). This event does not contain the ids of all of the messages that have been archived as this frequently exceeds the allowed message size. 

As I mentioned on the support case, we never intended for archiving a message to trigger an external integration event. The only trigger that should fire these events is ingestion from the error queue.

SzymonPobiega:
HI @brdmllr 

Unfortunately we're not able to merge this. There are two reasons for this. The first is the fact that we want to remove the integration event dispatcher infrastructure ([issue](https://github.com/Particular/ServiceControl/issues/872)) soon as we concluded it does not add enough value over simple `bus.Send`. The second is, we won't be able to ever support publishing archived notifications for group archives due to reasons related to message size and send throughput.

If you need to know about messages being archived, I'd suggest a different way. Because ServiceControl is built on top of NServiceBus
 * it's is based on a system of messages and handlers
 * it uses assembly scanning to discover handlers
That means that you can create a custom project that contains handlers for `FailedMessageArchived` and  `FailedMessageGroupArchived` and, in these handlers, publish (or send to a known queue) any message you like. These events are public types in the SC assembly so all you need is to reference the SC dll. This way you code will be not affected by the upcoming changes in the ServiceControl in near future.

We can't guarantee, however, that such a code will always work with SC. Looking further into future we plan to decouple the SC from NServiceBus. This will be a long journey and I am sure the final solution we arrive at will somehow support also your scenario.

brdmllr:
Thanks for your consideration. We were successfully able to follow the approach outlined above to publish the desired messages.

PR #902, opened 2017-01-20
[WIP] Draft of requirements

kbaley:
PR is meant to initiate discussions on the requirements of ServiceControl and ServicePulse. So far, we've captured what ServiceControl currently does as well as the business rationale for the major features.

### DIFF
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -10,16 +10,27 @@ ServiceControl should provide the following features:
 - process and store the contents of the error queue
 - replay messages from the error queue
 - process and store the contents of the audit queue
-- generate and collect heartbeat messages
+- monitor the availability of endpoints
 - generate and collect saga state changes
 - provide an API for monitoring user-defined conditions
-- provide operational events for consumption by other applications
+- provide integration hooks for developers to consume
+- search for messages
+- export failed message data for external analysis
+- report on the contents of the timeout manager
+- allow user to provide metadata about messages for operational support
+- allow user to easily provide feedback and usage data and diagnostics for the platform
+- monitor the health of SC itself
 
-Implicit in these requirements is that ServiceControl should:
+Implicit in these requirements is that the Service Platform should:
 
 - scale to an entire system based on the Particular Service Platform
 - be durable (i.e. no message loss)
 - be secure (i.e. calls to ServiceControl should be authenticated)
+- provide a smooth installation and upgrade experience
+- be performant
+- provide a clear and intuitive UI
+- be responsive with UI components (not in the HTML/mobile sense); i.e. it should always indicate when it is "doing something"
+- support all versions of all transports and all transport features
 
 ### Process and store the contents of the error queue
 
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -19,6 +19,7 @@ Implicit in these requirements is that ServiceControl should:
 
 - scale to an entire system based on the Particular Service Platform
 - be durable (i.e. no message loss)
+- be secure (i.e. calls to ServiceControl should be authenticated)
 
 ### Process and store the contents of the error queue
 
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -1,0 +1,49 @@
+At its heart ServiceControl and ServicePulse provide operational support for systems built on the Particular Service Platform. They do this by:
+
+- monitoring system health
+- diagnosing and treating errors
+
+In this context, system refers to NServiceBus endpoints and communication with third-party services.
+
+ServiceControl should provide the following features:
+
+- process and store the contents of the error queue
+- replay messages from the error queue
+- process and store the contents of the audit queue
+- generate and collect heartbeat messages
+- generate and collect saga state changes
+- provide an API for monitoring user-defined conditions
+- provide operational events for consumption by other applications
+
+Implicit in these requirements is that ServiceControl should:
+
+- scale to an entire system based on the Particular Service Platform
+- be durable (i.e. no message loss)
+
+### Process and store the contents of the error queue
+
+Rationale: To view the contents of the error queue in order to triage potential bugs or infrastructure failures in a system.
+
+### Replay messages from the error queue
+
+Rationale: To recover from bugs or infrastructure failures in order to maintain stability of a system.
+
+### Process and store the contents of the audit queue
+
+Rationale: To verify that a replayed message completed successfully. To understand the emergent design of a system through correllated messages. To identify design flaws in a system.
+
+### Generate and collect heartbeat messages
+
+Rationale: Verify that endpoints are up and able to send messages. Facilitates notification in the event an endpoint is no longer operational.
+
+### Generate and collect saga state changes
+
+Rationale: To understand the operational behavior of a long-running process and to troubleshoot design flaws.
+
+### Provide an API for monitoring user-defined conditions
+
+Rationale: To monitor endpoint dependencies. Allows team to more quickly identify and react to infrastructure failures.
+
+### Provide operational events for consumption by other applications
+
+Rationale: To provide extensibility for developers to react to events in ServiceControl. Allows users to provide better diagnostics for their systems.


docs/requirements.md
```diff
@@ -0,0 +1,49 @@
...
+
+Rationale: To view the contents of the error queue in order to triage potential bugs or infrastructure failures in a system.
+
+### Replay messages from the error queue
+
+Rationale: To recover from bugs or infrastructure failures in order to maintain stability of a system.
+
+### Process and store the contents of the audit queue
+
+Rationale: To verify that a replayed message completed successfully. To understand the emergent design of a system through correllated messages. To identify design flaws in a system.
```
SzymonPobiega: I'd question the "store" part because it implies persistence (writing to disk). We might be able to fulfill the requirements listed in the *rationale* without storing messages on disk.

docs/requirements.md
```diff
@@ -0,0 +1,49 @@
+At its heart ServiceControl and ServicePulse provide operational support for systems built on the Particular Service Platform. They do this by:
+
+- monitoring system health
+- diagnosing and treating errors
+
+In this context, system refers to NServiceBus endpoints and communication with third-party services.
+
+ServiceControl should provide the following features:
```
SzymonPobiega: For the final version I'd like to get rid of "ServiceControl" term because it implies monolithic design whereas I believe the features listed below can be fulfilled by a set of fairly independent components.

docs/requirements.md
```diff
@@ -0,0 +1,49 @@
...
+
+### Replay messages from the error queue
+
+Rationale: To recover from bugs or infrastructure failures in order to maintain stability of a system.
+
+### Process and store the contents of the audit queue
+
+Rationale: To verify that a replayed message completed successfully. To understand the emergent design of a system through correllated messages. To identify design flaws in a system.
+
+### Generate and collect heartbeat messages
```
SzymonPobiega: Does "generate" imply SC will actively "ping" endpoints to check if they are able to receive messages?

kbaley: @SzymonPobiega As I understand it, the heartbeats plugin is technically part of ServiceControl. That's why we said SC generates the messages.

docs/requirements.md
```diff
@@ -0,0 +1,49 @@
...
+
+Rationale: Verify that endpoints are up and able to send messages. Facilitates notification in the event an endpoint is no longer operational.
+
+### Generate and collect saga state changes
```
SzymonPobiega: I think long term this should be built into the saga storage rather than SC i.e. the saga storage should have a "diagnostic" mode where it stores each state change object. Then SI should be able to "somehow" load this diagnostic info.

docs/requirements.md
```diff
@@ -0,0 +1,49 @@
...
+
+Rationale: To understand the operational behavior of a long-running process and to troubleshoot design flaws.
+
+### Provide an API for monitoring user-defined conditions
```
SzymonPobiega: I would question if we really need this. Proper monitoring software will do this better than we will ever be able to do.

SzymonPobiega:
❤️ Good job! I'd add one more thing: proper user-facing HTTP API that would allow users to automate stuff such as retries.

docs/requirements.md
SzymonPobiega: 👍 makes sense.

gbiellem:
> Implicit in these requirements is that ServiceControl should: 
>
> - scale to an entire system based on the Particular Service Platform
>- be durable (i.e. no message loss)

I' like to add something about security in here.  One of the glaring holes with SC and SP from day 1 has been the lack of any authorization layer on the REST API.  We've worked around it to a degree by documenting how to reverse proxy it via IIS but this is far from elegant and very unhelpful for cloud deployments.

kbaley:
@Particular/servicecontrol-maintainers Added some more requirements after @boblangley and I went through the backlog.

docs/requirements.md
```diff
@@ -0,0 +1,61 @@
...
+- process and store the contents of the error queue
+- replay messages from the error queue
+- process and store the contents of the audit queue
+- monitor the availability of endpoints
+- generate and collect saga state changes
+- provide an API for monitoring user-defined conditions
+- provide integration hooks for developers to consume
+- search for messages
+- export failed message data for external analysis
+- report on the contents of the timeout manager
```
SzymonPobiega: Not sure if this requirement belongs here. Especially that we want to move away from the timeout manager towards the native implementations. These native implementations would be transport-specific and would require transport-aware management tooling.

docs/requirements.md
```diff
@@ -0,0 +1,61 @@
...
+- allow user to provide metadata about messages for operational support
```
SzymonPobiega: All messages or failed messages?

docs/requirements.md
```diff
@@ -0,0 +1,61 @@
...
+- allow user to easily provide feedback and usage data and diagnostics for the platform
+- monitor the health of SC itself
+
+Implicit in these requirements is that the Service Platform should:
+
+- scale to an entire system based on the Particular Service Platform
+- be durable (i.e. no message loss)
+- be secure (i.e. calls to ServiceControl should be authenticated)
+- provide a smooth installation and upgrade experience
+- be performant
```
SzymonPobiega: Pretty sure it is not an English word ;-)

docs/requirements.md
sergioc: "Should" here means it's supposed to already deliver or it's our vision?

docs/requirements.md
sergioc: http://wordnik.com/words/performant

docs/requirements.md
```diff
@@ -0,0 +1,61 @@
...
+- provide a clear and intuitive UI
```
sergioc: Should we separate UI concerns in a different document? I'd say perhaps just indicate how can SC help whatever UI (not just SP) work well.

docs/requirements.md
```diff
@@ -0,0 +1,61 @@
...
+- be responsive with UI components (not in the HTML/mobile sense); i.e. it should always indicate when it is "doing something"
+- support all versions of all transports and all transport features
+
+### Process and store the contents of the error queue
+
+Rationale: To view the contents of the error queue in order to triage potential bugs or infrastructure failures in a system.
```
sergioc: Given the short rationale descriptions, how about just appending them to the bullet list above?

docs/requirements.md
sergioc: Can the features be linked to whatever available related doco?

docs/requirements.md
```diff
@@ -0,0 +1,61 @@
...
+- provide an API for monitoring user-defined conditions
+- provide integration hooks for developers to consume
+- search for messages
+- export failed message data for external analysis
+- report on the contents of the timeout manager
+- allow user to provide metadata about messages for operational support
+- allow user to easily provide feedback and usage data and diagnostics for the platform
+- monitor the health of SC itself
+
+Implicit in these requirements is that the Service Platform should:
```
sergioc: We mentioned before that the list above are features, not requirements.

Could we treat the list below as additional features instead of requirements?

docs/requirements.md
SzymonPobiega: https://weblogs.asp.net/jongalloway/performant-isn-t-a-word :trollface:

docs/requirements.md
SzymonPobiega: @sergioc my understanding is that "should" is used to define the vision boundaries for future SC.

docs/requirements.md
SzymonPobiega: 👍 for treating these as features which will make them more explicit and force us to define what "acceptable performance" means (as a feature).

docs/requirements.md
mikeminutillo: In what way do we need to be performant? 

Low latency between API request/response?
Low latency between event occurring in system and being available via the API/notification events?
Resource utilization?

mikeminutillo: Should there be a non-functional requirement about being suitable for deployment to the cloud?

docs/requirements.md
mikeminutillo: I'd also question whether reading from the audit queue is the most efficient way to detect successful retries. Without that requirement, the decision to audit or not is completely decoupled from the recoverability/retries capability.

docs/requirements.md
mikeminutillo: Well, in theory there is a difference between "the Sales DB is down" vs "the Sales endpoint cannot connect to the Sales DB" and Custom Checks lets you develop the latter trivially. If we ever put the control channel in I would love to see endpoints expose Custom Checks to SC but SC control when/how they are run so we could have an interface in SP to manage it. i.e. Click an endpoint, get a list of checks to run, click the "Check Connection to Sales DB" button, get back a response saying "Success!". Then the whole concept of periodic checks goes away and you put the scheduling part in SC as it's an operational automation concern. Developers just expose whatever checks they feel are appropriate.

docs/requirements.md
```diff
@@ -0,0 +1,61 @@
...
+
+Rationale: To understand the operational behavior of a long-running process and to troubleshoot design flaws.
+
+### Provide an API for monitoring user-defined conditions
+
+Rationale: To monitor endpoint dependencies. Allows team to more quickly identify and react to infrastructure failures.
+
+### Provide operational events for consumption by other applications
+
+Rationale: To provide extensibility for developers to react to events in ServiceControl. Allows users to provide better diagnostics for their systems.
```
mikeminutillo: I really want to explore this further. I hate the current set of Failed Message events and would rather see them based on Groups. I also think a well defined and public API for SC would help us to reduce the number of events published here.

mikeminutillo:
@Particular/servicecontrol-maintainers do we need to keep this around? Is it good enough as it is to merge and we can make changes to it later?

sergioc:
I'd merge this after indicating inside the document that it's WIP.

boblangley:
How about we close the PR, but leave the branch for now.

I will do the first part now, we can delete the branch or re-open and mark as WIP if there are objections.

PR #912, opened 2017-02-02
Preserve ReplyToAddress for ingested errors

### LINKED ISSUE #911: Upon retrying messages, ServiceControl is changing the ReplyTo address of messages incorrectly
User reported:
https://groups.google.com/forum/?pli=1#!topic/particularsoftware/CnObyrIkuK0

Steps to Repro:
- Setup 2 machines PC1 and PC2
- Install SC on PC1 and run the Full Duplex Sample's Server on PC1 (tweak it to throw an exception)
- Run the Full Duplex Sample's Client on PC2 and send a message 
- Close the Server App on PC2
- Retry the failed message either using SP or SI
- Inspect the message that's on the Server's queue and look at the ReplyTo Address and notice that it is using the SC's machine name which is PC1 and not PC2.

mikeminutillo:
Fixes #911

### DIFF
--- a/src/ServiceControl/Operations/ErrorQueueImport.cs
+++ b/src/ServiceControl/Operations/ErrorQueueImport.cs
@@ -137,6 +137,8 @@
 
             var failureDetails = failedMessageFactory.ParseFailureDetails(message.Headers);
 
+            var replyToAddress = message.ReplyToAddress == null ? null : message.ReplyToAddress.ToString();
+
             var processingAttempt = failedMessageFactory.CreateProcessingAttempt(
                 message.Headers,
                 metadata,
@@ -144,7 +146,7 @@
                 message.MessageIntent,
                 message.Recoverable,
                 message.CorrelationId,
-                message.ReplyToAddress?.Queue);
+                replyToAddress);
 
             var groups = failedMessageFactory.GetGroups((string) metadata["MessageType"], failureDetails);
 
--- a/src/ServiceControl.AcceptanceTests/Recoverability/When_a_message_is_retried_with_a_replyTo_header.cs
+++ b/src/ServiceControl.AcceptanceTests/Recoverability/When_a_message_is_retried_with_a_replyTo_header.cs
@@ -1,0 +1,119 @@
+﻿namespace ServiceBus.Management.AcceptanceTests.Recoverability
+{
+    using System;
+    using System.Collections.Generic;
+    using NServiceBus;
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.MessageMutator;
+    using NServiceBus.Settings;
+    using NServiceBus.Transports;
+    using NServiceBus.Unicast;
+    using NUnit.Framework;
+    using ServiceBus.Management.AcceptanceTests.Contexts;
+    using ServiceControl.MessageFailures.Api;
+
+    public class When_a_message_is_retried_with_a_replyTo_header : AcceptanceTest
+    {
+        [Test]
+        public void The_header_should_not_be_changed()
+        {
+            var context = new ReplyToContext
+            {
+                ReplyToAddress = "ReplyToAddress@SOMEMACHINE"
+            };
+
+            List<FailedMessageView> failedMessages;
+
+            Define(context)
+                .WithEndpoint<VerifyHeaderEndpoint>()
+                // Verify that the wait for failed message and issue retry
+
+                .Done(x =>
+                {
+                    if (!x.RetryIssued && TryGetMany("/api/errors", out failedMessages))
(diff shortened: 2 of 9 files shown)


mikeminutillo:
@Particular/servicecontrol-maintainers please review and keep an eye on the build status for this one.

indualagarsamy:
@mikeminutillo - what about here? 
https://github.com/Particular/ServiceControl/blob/preserve-replytoaddress/src/ServiceControl/Infrastructure/TransportMessageExtensions.cs#L31

Is this relevant here as well?

indualagarsamy:
@mikeminutillo - do we need any changes to this test here, based on the bug and now?
https://github.com/Particular/ServiceControl/blob/preserve-replytoaddress/src/ServiceControl.UnitTests/Infrastructure/TransportMessageExtensionsTests.cs#L51

mikeminutillo:
This won't fix data on disk. It will only work for newly ingested Failed Messages. 

I don't know why we are overwriting the ReplyToAddress header that was ingested with the metadata one at all. Removing [the code that does this](https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/Recoverability/Retries/RetryProcessor.cs#L199) will fix it but I am uncertain why this code was present in the first place. Investigating further.

mikeminutillo:
This appears to be a holdout from when SC ran on NSB4. In that version, ReplyToAddress was a property of TransportMessage. The reason to pull it out as a separate piece of metadata was because in V4, `ReplyToAddress` was [just a property and was not linked to the header](https://github.com/Particular/NServiceBus/blob/4.6.10/src/NServiceBus.Core/TransportMessage.cs#L111).

mikeminutillo:
The test referenced is a different path of code entirely. It uses the headers directly and is not supposed to preserve the machine name. It is looking for a Logical Endpoint name so it explicitly drops the machine name from the physical address. 

Having said that, the test should include a machine name to prove that it's getting dropped.

mikeminutillo:
This fix has a green build. 

Should we remove the metadata `ReplyToAddress` on Processing Attempts? I don't believe anyone is using it. It doesn't have to be a part of this fix. We can add a refactoring issue to the backlog.

Also should I revert the change to the Ingestion pipe to put it back the way it was for now? As I said, I'm pretty sure there is no code using it.

mikeminutillo:
This symptom is still present when running on the SQL Server transport but it is a [different issue](https://github.com/Particular/NServiceBus.SqlServer/issues/345)

src/ServiceControl/Operations/ErrorQueueImport.cs
```diff
@@ -137,14 +137,16 @@ private void InnerHandle(TransportMessage message)
 
             var failureDetails = failedMessageFactory.ParseFailureDetails(message.Headers);
 
+            var replyToAddress = message.ReplyToAddress == null ? null : message.ReplyToAddress.ToString();
```
SzymonPobiega: Given the fact that `ReplyToAddress` is implemented as

```
get
      {
        string destination;
        if (this.Headers.TryGetValue("NServiceBus.ReplyToAddress", out destination))
          return Address.Parse(destination);
        return (Address) null;
      }
```
we should just plain get the header value here and not pass it through the `Address` class. This way we'll preserve whatever the transport-defined value was in the first place.

SzymonPobiega:
@mikeminutillo I believe that if we change this to what I proposed, we won't need the SQL server fix.

src/ServiceControl/Operations/ErrorQueueImport.cs
mikeminutillo: You can safely ignore this value. It used to be used to preserve the value when the underlying transport core did not do so. It should eventually be removed.

The main issue is [here](https://github.com/Particular/ServiceControl/blob/d78e169da41f38e54645b254465a40dcd0ad6d09/src/ServiceControl/Recoverability/Retries/RetryProcessor.cs#L211). This transport message contains the unadulterated ReplyToAddress header but the transport will run it through `Address.Parse().ToString()` before writing it to the table as outlined in https://github.com/Particular/NServiceBus.SqlServer/issues/345

mikeminutillo:
If a message has already been through a retry attempt (via ServiceControl), and the ReplyToAddress has been corrupted, if the processing fails again the the newly imported ReplyToAddress header will be recorded in it's corrupt state. 

I just added code to detect and fix this case (at least for MSMQ) on outgoing retries. I extracted the behavior into a new class and unit tested it. In order to acceptance test it, we need the sending endpoint to run on a different machine than ServiceControl runs on :s

boblangley:
Merged #912

PR #914, opened 2017-02-10
[WIP] Send errors from .Retries to endpoint queue

### LINKED ISSUE #913: Errors imported from .Retries queues have required FailedQ header stripped when retrying
The secondary retries processors require a `FailedQ` header that points to the normal input queue for the endpoint. If a `FailedQ` header is not present the message is rejected and forwarded to the system error queue.

Additionally, when a message fails in the secondary retries processor, it is forwarded to the system error queue with the `.Retries` address set to the `FailedQ` header value.

Service Control strips the `FailedQ` header when sending a message back for retry. 

The result is, for a message that originally fails in the `.Retries` queue, and is then retried from Service Control, the message will never able to be retried being immediately returned to the system error queue. The message is now stuck in a loop:

image

boblangley:
Fixes #913 

Updates:

- Looks for queue names ending in `.Retries` and strips `.Retries` when forwarding for retry.
- Added test case

### DIFF
--- a/src/ServiceControl.AcceptanceTests/Recoverability/When_a_message_is_retried_from_the_retries_queue.cs
+++ b/src/ServiceControl.AcceptanceTests/Recoverability/When_a_message_is_retried_from_the_retries_queue.cs
@@ -1,0 +1,151 @@
+﻿namespace ServiceBus.Management.AcceptanceTests.Recoverability
+{
+    using System;
+    using System.Threading;
+    using NServiceBus;
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.Config;
+    using NServiceBus.Settings;
+    using NServiceBus.Transports;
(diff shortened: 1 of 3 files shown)


boblangley:
The test is timing out. When I test locally with MSMQ, with journaling turned on, the failing endpoint receives the message (it appears in the queue journal) and appears to process it, but there is no evidence the handler is called.

Turning journaling on for the `error` queue also only shows 1 message, which is expected of a passing test.

SzymonPobiega:
What if I have `ExamPolicy.Retries` endpoint in my system?

boblangley:
We have talked about this and it is a valid concern. The question we come to is this likely, or is this an edge case that we should handle later?

Is there another way to handle this scenario?

mikeminutillo:
@boblangley can we close this?

PR #915, opened 2017-02-13
Update to latest ASQ transport

### LINKED ISSUE #910: Unable to ingest messages from ASQ CoreV6 SendOnly endpoints
There is an issue between the v5 and v6 versions of the Azure Storage Queues that prevents v5 receivers from reading messages sent by v6 SendOnly endpoints. ServiceControl runs on the v5 version of the ASQ transport (6.2.1) which means it is unable to ingest messages from v6 SendOnly endpoints.

Related issue https://github.com/Particular/NServiceBus.AzureStorageQueues/issues/181 V6 SendOnly cannot send message to V5

_Edit by @boblangley_
Raised from customer case 00023775

mikeminutillo:
Fixes #910

### DIFF
--- a/src/ServiceControl.UnitTests/app.config
+++ b/src/ServiceControl.UnitTests/app.config
@@ -1,0 +1,43 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <runtime>
+    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
+      <dependentAssembly>
+        <assemblyIdentity name="Microsoft.Data.Edm" publicKeyToken="31bf3856ad364e35" culture="neutral" />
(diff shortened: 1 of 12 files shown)


src/ServiceControl.UnitTests/app.config
```diff
@@ -0,0 +1,43 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
SzymonPobiega: Why do we need this?

mikeminutillo: It came with the upgrade to NSB

mikeminutillo:
Merging this to put a release out. If we need to, we can revisit https://github.com/Particular/ServiceControl/pull/915/files/6b33905c18c2abb3d62e2a4a06515249750a3c1f#diff-80ee757a291baf1ce278e196a549baebR1 later but as it's on a test project I'm not too concerned about it.

PR #916, opened 2017-02-15
Update SQL Transport

mikeminutillo:
Waiting on release of 2.2.5 to fix SQL version of #911

### DIFF
--- a/src/ServiceControl.AcceptanceTests/AcceptanceTest.cs
+++ b/src/ServiceControl.AcceptanceTests/AcceptanceTest.cs
@@ -111,7 +111,7 @@ namespace ServiceBus.Management.AcceptanceTests
             var key = (object) "IgnoreTransports";
             if (TestContext.CurrentContext.Test.Properties.Contains(key))
             {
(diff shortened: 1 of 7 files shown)


SzymonPobiega:
@mikeminutillo 2.2.5 is out on NuGet

src/ServiceControl.AcceptanceTests/AcceptanceTest.cs
```diff
@@ -103,9 +106,26 @@ public void Setup()
             ravenPath = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
         }
 
+        private void AssertTransportNotExplicitlyIgnored()
+        {
+            var key = (object) "IgnoreTransports";
```
SzymonPobiega: why do we need a cast?

SzymonPobiega:
@mikeminutillo is it still WIP?

mikeminutillo:
Not still WIP. Happy to remove the cast before merging. It used to be a different key and for some reason the dictionary is (object) -> (object)

Issue #919, opened 2017-02-17
StackTrace Parser consumes 100% CPU with some stack traces


WilliamBZA:
## Who's Affected

Anyone using ServiceControl to that receives messages with strange EOL encodings

## Symptoms 

ServiceControl does not ingest messages and CPU spikes to 100%. Removing the messages from front of the queue and restarting Service Control allows processing to continue.

## Details

The stack trace parsing code did not handle certain line encodings correclty which resulting in a lot of backtracing within a regular expression. This resulting in a large spike of CPU usage and the prevention of processing messages.  This release resolves that problem so no messages are prevented from processing.

## Original Bug Report
Raised from customer support incident, internal case# 00023775

This came in via a support case where the customer was experiencing SC hitting 100% CPU and no messages were being ingested.

Upon inspection it appears that the Regex we use to parse the stackframe from the headers was locking during matching and consuming 100% CPU.

PR #920, opened 2017-02-17
Update stack trace regex parsing to support failed case

WilliamBZA:
Connects to https://github.com/Particular/ServiceControl/issues/919

* Updates the stack trace parser regex to work with the failing case
* Adds an additional test to validate the fix

Ping @Particular/servicecontrol-maintainers


WilliamBZA:
Another option is to remove the regex completely as we only ever use the first stackframe and not all of them. If we can validate that assumption then perhaps we can come up with a non-regex based solution that doesn't look like black magic.. @mikeminutillo Do you know enough about this piece of code to confirm the assumption?

WojcikMike:
I've spent some time reviewing the changes the botom line is that it replaces:
 - \s with [\x20\t] //which is space (ascii 32) or tab
 - . with [^\x20\t] //which is everything beside above

In theory \s represents: [\r\n\t\f\v ] (I found it on some site) so it include more than only space and tab. yet the change seems to work correctly. Looking at the [StackTraceParser repo](https://github.com/atifaziz/StackTraceParser)
They found this issue: https://github.com/atifaziz/StackTraceParser/issues/2 which resulted in the same change being made. However the issue found there is incorrect parsing of the stacktrace rather than 100 % CPU utilization. 

cc: @mikeminutillo @janovesk

janovesk:
Looks good to me, but then again.. I'm no expert at regex! Replacing with custom parsing is a nightmare though, so I would keep this.

 Maybe we should consider pulling in the Nuget source package instead, https://github.com/atifaziz/StackTraceParser?

WilliamBZA:
@janovesk @WojcikMike I've updated to use the nuget package directly.

If you're happy please approve and I'll merge and release.

WojcikMike:
@WilliamBZA approved.

Issue #438, opened 2015-01-20
MSMQ Messages from 3rd party integration are not available via the REST API  


gbiellem:
Failed MSMQ Messages from third party integrations may not have a message id.  These messages are imported to RavenDB as failed messages but then when ServiceControl is queried these messages throw 404.  Since the message id is missing the native MSMQ message is is used.  This message id has a "\" in it which is incorrectly handled by the Nancy self host routing.

PR #922, opened 2017-02-22
remove erroneous manipulation of search keyword

### LINKED ISSUE #921: Search strings with a forward slash followed by a numeric do not work via the path based search API
Due to [this string replacement](https://github.com/Particular/ServiceControl/blob/10f82ebab164c62833fd3c71f9ead21ff346955e/src/ServiceControl/CompositeViews/Messages/GetMessagesByQuery.cs#L25):

- `/messages/search/foo/bar` ✔️ works
- `/messages/search/foo/123` ❌ does not work!

I stepped through the code and it appears that this is some weird quirk of RavenDB. When searching for "foo\bar", RavenDB matches "foo/bar", but when searching for "foo\1", RavenDB does not match "foo/1". 😕 

Note that there are [four variants of the search API](https://github.com/Particular/ServiceControl/blob/10f82ebab164c62833fd3c71f9ead21ff346955e/src/ServiceControl/CompositeViews/Messages/GetMessagesByQuery.cs#L15-L44) and only one of them does this replacement in the keyword. Something definitely seems off with this.

Ultimately, if you arbitrarily replace a character in a search string, then obviously no client will ever be able to search using that character. And to only do it for one variant of the search is even worse:

1. `/messages/search?q=foo/123` - ✔️ works
2. `/messages/search/foo/123` - ❌ does not work!
3. `/endpoints/myendpoint/messages/search?q=foo/123` - ✔️ works
4. `/endpoints/myendpoint/messages/search/foo/123` - ✔️ works

To fix the immediate problem, we can switch ServiceInsight to using the query string variant (1) which doesn't do the replacement, but still, it doesn't seem right to leave variant (2) in this broken state.

<details>
<summary>Previous description</summary>
The search API appears to support two formats:

- `http://localhost:33333/api/messages/search/foo` (used by ServiceInsight)
- `http://localhost:33333/api/messages/search?q=foo`

However, strings containing forward slashes. For example, when searching for the string "foo/123", the correct, encoded URLs would be:

- `http://localhost:33333/api/messages/search/foo%2f123`
- `http://localhost:33333/api/messages/search?q=foo%2f123`

But neither of these return any results. The only way to search for "foo/123" is to use the following un-encoded URL:

- `http://localhost:33333/api/messages/search?q=foo/123`

(There is no equivalent of the other URL format that works.)

ServiceInsight correctly encodes the URL's. Until this is fixed, there is no way to search for strings which contain characters which need to be URL encoded from ServiceInsight.
</details>

adamralph:
fixes #921

### DIFF
--- a/src/ServiceControl/CompositeViews/Messages/GetMessagesByQuery.cs
+++ b/src/ServiceControl/CompositeViews/Messages/GetMessagesByQuery.cs
@@ -22,7 +22,6 @@ namespace ServiceControl.CompositeViews.Messages
             Get["/messages/search/{keyword*}"] = parameters =>
             {
                 string keyword = parameters.keyword;
-                keyword = keyword?.Replace("/", @"\");
                 return SearchByKeyword(keyword);
             };
 


src/ServiceControl/CompositeViews/Messages/GetMessagesByQuery.cs
```diff
@@ -22,7 +22,6 @@ public GetMessagesByQuery()
             Get["/messages/search/{keyword*}"] = parameters =>
             {
                 string keyword = parameters.keyword;
-                keyword = keyword?.Replace("/", @"\");
```
adamralph: Note how the equivalent case for searching within a specific endpoint [does not do this replacement](https://github.com/Particular/ServiceControl/blob/10f82ebab164c62833fd3c71f9ead21ff346955e/src/ServiceControl/CompositeViews/Messages/GetMessagesByQuery.cs#L37-L43).

mikeminutillo:
The original fix for #438 also performs this replacement in a second API `/messages/{id}/body` https://github.com/adamralph/ServiceControl/commit/e7e626d3041a4e83a883819266c61402a1c9922b

It looks as if native MSMQ Message Ids contain a slash (not sure if `/` or `\`) and that's why this fix was put in place. I suspect that if SI has been switched to the query string API then we can remove this API instead of trying to fix it.

PR #926, opened 2017-03-03
WIP - Not use callbacks when using SqlServer transport.

indualagarsamy:
Relates to a customer case. 

This will work for Sql according to:
https://github.com/Particular/NServiceBus.sqlserver/issues/2#issuecomment-24493875

TODO:
Find a way where we can do this in an agnostic way for RabbitMQ as well.

### DIFF
--- a/src/ServiceControl/Infrastructure/NServiceBusFactory.cs
+++ b/src/ServiceControl/Infrastructure/NServiceBusFactory.cs
@@ -32,6 +32,9 @@ namespace ServiceBus.Management.Infrastructure
             configuration.DisableFeature<TimeoutManager>();
             configuration.DisableFeature<Outbox>();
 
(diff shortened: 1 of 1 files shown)


SzymonPobiega:
@indualagarsamy I don't think this is the proper API. I think it worked for SQL 1.x targeting NSB 4.x. In SQL 2.x you need to call `DisableCallbackReceiver`.

indualagarsamy:
@SzymonPobiega - Thanks. The documentation also confirms what you said.

SzymonPobiega:
@indualagarsamy In case of SQL Server transport we have a dedicated configuration assembly we drop for DTC configuration: https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl.SqlServerWithDTC/ServiceControlDTCSupport.cs. Probably we can reuse it to disable the callback queue.

PR #932, opened 2017-03-14
Keep original UniqueMessageId for Retried Messages

mikeminutillo:
Fixes https://github.com/Particular/ServicePulse/issues/425

> When a failed message is ingested by ServiceControl it is stored in the embedded database. If the record was already in the database it's status is updated (usually from Pending to Unresolved). In order to accommodate events, the unique identifier for each failed message is a combination of the Message Id (usually a GUID) and the Processing Endpoint.
> 
> With a redirect, you are effectively changing the Processing Endpoint. When the message fails again, the resulting failed message is not matched back to the original one.
> 
> The result is two failed message records. One for the original address and one for the new address. The original one remains pending because we haven't linked it back to the incoming failed message. The new one is listed as Unresolved with only a single processing attempt.

This PR fixes the issue by retaining the original Unique Messaging Id (which was already applied as a header to all outgoing retry messages). This Id will be used for all ingested messages. If a message fails and then is retried with a redirect, when the new failure is ingested it will be correlated back to the original id and added to that document.

### DIFF
--- a/src/ServiceControl.AcceptanceTests/MessageRedirects/When_a_message_fails_a_retry_with_a_redirect.cs
+++ b/src/ServiceControl.AcceptanceTests/MessageRedirects/When_a_message_fails_a_retry_with_a_redirect.cs
@@ -1,0 +1,138 @@
+﻿namespace ServiceBus.Management.AcceptanceTests.MessageRedirects
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Net;
+    using NServiceBus;
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.Config;
+    using NServiceBus.Features;
+    using NServiceBus.Settings;
+    using NUnit.Framework;
+    using ServiceBus.Management.AcceptanceTests.Contexts;
+    using ServiceControl.Infrastructure;
+    using ServiceControl.MessageFailures;
+    using ServiceControl.MessageFailures.Api;
+
+    class When_a_message_fails_a_retry_with_a_redirect : AcceptanceTest
+    {
+        [Test]
+        public void The_original_failed_message_record_is_updated()
+        {
+            FailedMessage failedMessage;
+            List<FailedMessageView> failedMessages = null;
+
+            Define<Context>()
+                .WithEndpoint<OriginalEndpoint>(b =>
+                        b.Given(bus => bus.SendLocal(new MessageToRetry()))
+                            .When( // Failed Message Received
+                                ctx => ctx.UniqueMessageId != null && TryGet($"/api/errors/{ctx.UniqueMessageId}", out failedMessage),
+                                (bus, ctx) =>
+                                {
+                                    // Create Redirect
+                                    Post("/api/redirects", new RedirectRequest
+                                    {
+                                        fromphysicaladdress = ctx.FromAddress,
+                                        tophysicaladdress = ctx.ToAddress
+                                    }, status => status != HttpStatusCode.Created);
+
+                                    // Retry Failed Message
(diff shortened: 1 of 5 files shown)


mikeminutillo:

SzymonPobiega:
one test failed

SzymonPobiega:
re-running to confirm.

SzymonPobiega:
OK, I see that there is some problem in this test. Investigating...

SzymonPobiega:
@mikeminutillo I fixed the test as I believe the problem was not with the code but with test but I would feel better if you could check after me.

mikeminutillo:
Looking at this with fresh eyes I suspect this will legitimately break the data migration code.

The data migration was pulling apart the processing attempts, recalculating their UniqueMessageId and grouping them that way. This worked because the UniqueMessageId was a combination of Message Id (which is more or less immutable) and some identifier for the processing endpoint. In cases where things got tangled up, this would pull processing attempts apart that were handled by different endpoints. 

This change means that if you've retried a message and it failed again then the data migration will no longer be able to pull them apart because the UniqueMessageId will just short-circuit back to what it was when you issued the retry. 

I can adjust the migration code to use a copy of the original UniqueMessageId algorithm which will fix it. Are there any problems with that approach?

SzymonPobiega:
@mikeminutillo not sure I understand. My understanding is that the migration thing splits the failure records by generating the new unique ID in a correct way and grouping the failure attempts with matching ID together. Is that fair assumption?

No assuming we have the ID generation code fixed and no longer using `ReplyTo`, all the new failure should have proper IDs. When we now fixed the code for generating failure attempt IDs to take into account the header, how does this affect the migration?

mikeminutillo:
Before this PR:

- A publishes event E
- B handles E and fails, forwarding E to the error Q
- C handles E and fails, forwarding E to the error Q
- SC picks up the error Q and incorrectly combines BE and CE into a single failed message FM which has an incorrect Unique Message Id.
- User retries E and the response only goes back to C. This retry attempt fails and a new RetriedCE goes to the error Q. RetriedCE has gets added to FM
- The Data Migration runs
  - It should regenerate new Unique Message Ids for all the processing attempts in FM (BE, CE, and RetriedCE).
  - The processing attempts are regrouped by their new Unique Message Ids
  - New Failed Messages are created for each grouping

After this PR, he data migration will not create the right Unique Message Id for RetriedCE. This is because the code to generate the Unique Message Id will use the retry header id if it finds one. RetriedCE will continue to have the old Unique Message Id and will not be matched against CE.

WilliamBZA:
The migrations are only run once, correct? So this is only a problem if a user has a failed message with the incorrect Id and then upgrades to the >= version that this will be included in.

If that is the case, why can we not tweak the migration code for 1.27 to use custom Id generator in whatever version this will be released as?  That migration doesn't have to rerun for users who have already run it, but for future users that might be at risk here, the correct Ids can be generated.

TL;DR: I see no problem with adjusting the migration code to use a copy of the Id.

SzymonPobiega:
@mikeminutillo suppose we implemented the fix you proposed to ignore the header in the migration script, the scenario you described would work fine. What would happen if, when E is retried user redirects the retry to D? Then we would end up with BE, CE and Retried DE in the same FM, right? (because `ReplyTo` address is used to generate the ID). Splitting should then behave fine so I can't poke any holes in this process.

SzymonPobiega:
@mikeminutillo @WilliamBZA I've pushed the change Mike proposed and added a test for scenario Mike described. What Mike said! (TM)`

SzymonPobiega:
@mikeminutillo I'll leave it to you for final check and merge.

PR #937, opened 2017-03-23
Add failing endpoint address classifier

WilliamBZA:
## Who's affected

* Any user using Service Control to manage errors

## Symptoms

Adds an additional classifier to group errors based on the address of the endpoint.

## Original request 

https://github.com/Particular/ServicePulse/issues/427


WilliamBZA:
Ping @Particular/servicecontrol-maintainers. There shouldn't be a practical issues here, but I wonder how far down the rabbit hole of "what classifiers should we add" can we go before it becomes an extensability problem.

Perhaps we should split these extensions out to a plugin-type model? Or allow for enabling/disabling classifiers?

SzymonPobiega:
@WilliamBZA does this require an automated test or are we sure it will be automatically discovered and "just work"?

WilliamBZA:
Good catch @SzymonPobiega. I didn't update the test

SzymonPobiega:
@WilliamBZA I don't like these "magic" functions where you just need to add a class and it will be automatically picked up. Hard to tell if the code is working just by looking at the text.

WilliamBZA:
It's actually not only missing the test, it's also missing the container configuration - which would have made it more explicit that we're registering a grouper.

I've fixed both.

SzymonPobiega:
LGTM @WilliamBZA . Do you want it merged?

WilliamBZA:
Thanks @SzymonPobiega. I'll merge and release with 1.39.0

Issue #944, opened 2017-04-04
Can't acknowledge retried groups after refactoring


WilliamBZA:
There is no code that acknowledges retries anymore, only archiving.

If a user retries a group they are no longer able to dismiss the notification that the retry completed.

WilliamBZA:
Fixed

PR #945, opened 2017-04-04
Acknowledge Retries as well as Archive operations

WilliamBZA:
Connects to https://github.com/Particular/ServiceControl/issues/944

Changes the acknowledgement operation to try and acknowledge archive operations first, followed by retry operations. There can only be one of those per group at a time.

Returns `404` if a group is provided that is either not archived or retried.


WojcikMike:
A bit post factum but LGTM. I will probably make a PR to SP to handle NotFound for acknowledge

PR #963, opened 2017-05-15
In-Memory Heartbeats

mikeminutillo:
- Tracks heartbeats in memory only
- Checking for state changes happens on a background thread instead of when ingesting heartbeats
- Persists KnownEndpoints only on state changes

Known issues:
- Loses status when restarted. This is OK. If the endpoint is not monitored then we don't care. If the endpoint is monitored but not sending heartbeats, it shows as failed (but no external integration event is raised). If the endpoint is monitored and sending heartbeats, it shows as active.


src/ServiceControl/HeartbeatMonitoring/LegacyHandler.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace ServiceControl.HeartbeatMonitoring
```
SzymonPobiega: Is this one only to consume messages that might still be in the queue when upgrading?

mikeminutillo: Yeah. Otherwise they'll go through FLR/SLR. Every event in SC is handled anyway by the global handlers, it's only for commands. These commands don't really make a lot of sense after a restart and can cause problems when trying to ingest them in parallel anyway leading to more congestion.

src/ServiceControl/Monitoring/MonitoringDataPersister.cs
```diff
@@ -0,0 +1,101 @@
...
+{
+    using System.Linq;
+    using Raven.Client;
+    using ServiceControl.Contracts.HeartbeatMonitoring;
+    using ServiceControl.EndpointControl;
+    using ServiceControl.EndpointControl.Contracts;
+    using ServiceControl.Infrastructure;
+    using ServiceControl.Infrastructure.DomainEvents;
+
+    public class MonitoringDataPersister : IDomainHandler<HeartbeatingEndpointDetected>,
```
mikeminutillo: These persistence operations should probably all be done with a patch. Only Monitored and HostName are really variable. Endpoint Name and Host Id are invariant as they're used to calculate the Id.

src/ServiceControl/Monitoring/EndpointMonitoringStats.cs
```diff
@@ -0,0 +1,16 @@
...
+
+    public class EndpointMonitoringStats
+    {
+        private int active;
+        private int failing;
+
+        public int Active => active;
+        public int Failing => failing;
+
+        public void RecordActive() => Interlocked.Increment(ref active);
```
SzymonPobiega: If I read correctly `RecordActive` and `RecordFailing` are only called by  `EndpointInstanceMonitor::AddTo` which seems to be invoked in sync by `GetStats` so I think the `Interlocked` is redundant here.

src/ServiceControl/Monitoring/HeartbeatMonitor.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            var done = false;
+
+            while (!done)
+            {
+                var priorReading = heartbeat;
+                if (priorReading.Timestamp.GetValueOrDefault() < cutoff)
+                {
+                    var newReading = new RecordedHeartbeat(HeartbeatStatus.Dead, priorReading.Timestamp);
+                    done = Interlocked.CompareExchange(ref heartbeat, newReading, priorReading) == priorReading;
```
SzymonPobiega: Why do we need to do `CompareExchange`? Is it to ensure that we are the last one to update the value? Is this guarantee necessary?

src/ServiceControl/Monitoring/EndpointMonitoringStats.cs
mikeminutillo: GetStats can be called from a web-api call as well (although it will be on a different instance so the interlocked is again redundant). 

Having said that, the status of the endpoints won't update between refreshes and we already have a cached copy of the stats so we could easily just return the cached version and make the update fully sync.

src/ServiceControl/Monitoring/HeartbeatMonitor.cs
mikeminutillo: It's to ensure consistency. If I read the timestamp and see that my timestamp is from further into the future then I can overwrite that record. If, in the time between the read and the write another heartbeat is written in, I can't overwrite that one without checking it. BUT that new timestamp might still be from before the cutoff. 

If we don't do the check then we run the risk of going slightly out of sync when the queue is backed up. We declare the endpoint dead moments after the latest heartbeat is received.

mikeminutillo: We can get around all of that by just locking the heartbeat when we update it but it means a lot of locking and synchronization, especially when SC first starts and has a backlog of heartbeats it's trying to process concurrently.

SzymonPobiega: Ah, ok. Now it makes sense. I see it is mainly for concurrent marking as alive. At first I only thought about the race conditions between marking alive and dead. I ❤️ it.

src/ServiceControl/Monitoring/EndpointInstanceId.cs
```diff
@@ -0,0 +1,47 @@
...
+            HostName = hostName;
+            HostGuid = hostGuid;
+        }
+
+        public override bool Equals(object obj)
+        {
+            if (ReferenceEquals(null, obj)) return false;
+            if (ReferenceEquals(this, obj)) return true;
+            if (obj.GetType() != this.GetType()) return false;
+            return Equals((EndpointInstanceId)obj);
```
WilliamBZA: If you use the `as` operator instead, you'd be able to remove all the checks above and instead only use the ones in the strongly typed Equals method.

```CSharp
public override bool Equals(object obj)
{
      return Equals(obj as EndpointInstanceId);
}
```

src/ServiceControl/Monitoring/EndpointInstanceMonitor.cs
```diff
@@ -0,0 +1,122 @@
...
+        private void RaiseStateChangeEvents(HeartbeatStatus newStatus, DateTime? latestTimestamp)
+        {
+            if (newStatus == HeartbeatStatus.Alive)
+            {
+                if (status == HeartbeatStatus.Unknown)
+                {
+                    // NOTE: If an endpoint starts randomly sending heartbeats we monitor it by default
+                    // NOTE: This means we'll start monitoring endpoints sending heartbeats after a restart
+                    Monitored = true;
+                    DomainEvents.Raise(new HeartbeatingEndpointDetected
```
WilliamBZA: So after a restart all endpoints will be monitored again once they start sending messages? Or are the "don't monitor" flags persisted ?

src/ServiceControl/Monitoring/EndpointInstanceMonitor.cs
```diff
@@ -0,0 +1,122 @@
...
+        public void AddTo(EndpointMonitoringStats stats)
+        {
+            if (Monitored)
+            {
+                if (status == HeartbeatStatus.Alive)
+                    stats.RecordActive();
+                else
+                    stats.RecordFailing();
+            }
+
```
WilliamBZA: Pedantic: there's a blank line here between the 2 closing braces

src/ServiceControl/Monitoring/MonitoringDataPersister.cs
```diff
@@ -0,0 +1,101 @@
...
+        IDomainHandler<MonitoringDisabledForEndpoint>
+    {
+
+        public void Handle(HeartbeatingEndpointDetected domainEvent)
+        {
+            var id = DeterministicGuid.MakeId(domainEvent.Endpoint.Name, domainEvent.Endpoint.HostId.ToString());
+
+            using (var session = store.OpenSession())
+            {
+                var knownEndpoint = new KnownEndpoint
```
WilliamBZA: Should we set the session here to use optimistic concurrency so that we don't overwrite previously stored values? I guess it doesn't really make much of a difference though

src/ServiceControl/Monitoring/MonitoringDataPersister.cs
```diff
@@ -0,0 +1,101 @@
...
+                knownEndpoint.Monitored = false;
+                session.SaveChanges();
+            }
+        }
+
+        public void WarmupMonitoringFromPersistence()
+        {
+            using (var session = store.OpenSession())
+            {
+                var endpoints = session.Query<KnownEndpoint>().ToArray();
```
WilliamBZA: This will break if there are more endpoints than the default page size in raven.  Will have to remove the ToArray and switch to an advanced query and stream the results.

```csharp
using (var endpointsEnumerator = session.Advanced.Stream(session.Query<KnownEndpoints>()))
{
    while (endpointsEnumerator.MoveNext())
    {
        var endpoint = endpointsEnumerator.Current.Document;
        monitor.Register(new EndpointInstanceId(endpoint.EndpointDetails.Name, endpoint.EndpointDetails.Host, endpoint.EndpointDetails.HostId), endpoint.Monitored);
    }
}
```

WilliamBZA:
Mostly happy, the only item that really has t o be changed is the raven query. The rest of the comments can be safely ignored.

SzymonPobiega:
@mikeminutillo do you need any help in pushing this forward?

mikeminutillo:
@SzymonPobiega I wonder if we shouldn't move this code (minus the persistence part) over to the Monitoring Instance and then once that's released, deprecate the heartbeat plugin and all that goes with it (including the original code here).

SzymonPobiega:
That would probably requires SP to be able to query the heartbeat-related API from two sources and merge the results but it should not be very difficult.

mikeminutillo:
OK. So that still means we may want to release this version in the meantime.

WilliamBZA:
I think there's value in releasing this as a stand alone too.

SzymonPobiega:
@WilliamBZA stand alone?

WilliamBZA:
@SzymonPobiega Releasing it without waiting for the monitoring instance. ServiceControl will still benefit and there will be almost no change in current user behavior.

src/ServiceControl/Monitoring/EndpointInstanceId.cs
mikeminutillo: @WilliamBZA I didn't write this implementation. I just used the R# generated one. This is much simpler. I'll update it

src/ServiceControl/Monitoring/EndpointInstanceMonitor.cs
mikeminutillo: The "don't monitor" flags are persisted but there is a major assumption built into this new heartbeats code: If you are sending me heartbeats then I should monitor you.

So, if an endpoint is sending heartbeats and the user marks it as "don't monitor", when you restart SC, we'll start monitoring the endpoint again. Additionally, if an endpoint is sending heartbeats and the user marks it as "don't monitor", when you restart the endpoint we'll start monitoring it again. 

I think this is the correct way to handle the situation. If you really don't want to monitor it, why is it sending heartbeats? 

If we always respected the "don't monitor" flag then you get the risk that a new (or renamed endpoint) happens to cache hit one of the "don't monitor" records and fails to show up.

WilliamBZA: 💯

src/ServiceControl/Monitoring/EndpointMonitoringStats.cs
mikeminutillo: Actually my statement above is not quite true. `GetStats` can't be cached between updates. The endpoint status won't change but the monitored status might.

src/ServiceControl/Monitoring/HeartbeatMonitor.cs
```diff
@@ -0,0 +1,53 @@
+namespace ServiceControl.Monitoring
+{
+    using System;
+    using System.Threading;
+
+    class HeartbeatMonitor
+    {
+        volatile RecordedHeartbeat heartbeat = new RecordedHeartbeat(HeartbeatStatus.Unknown, null);
```
mikeminutillo: R# complains that this usage of `volatile` with the two cases of `ref` below don't play nicely together. I've done some research on this but I'd appreciate more eyes.

SzymonPobiega: @mikeminutillo according to https://stackoverflow.com/questions/425132/a-reference-to-a-volatile-field-will-not-be-treated-as-volatile-implications it is fine.

src/ServiceControl/Monitoring/HeartbeatMonitor.cs
```diff
@@ -0,0 +1,53 @@
...
+                    done = Interlocked.CompareExchange(ref heartbeat, newReading, priorReading) == priorReading;
+                }
+                else
+                {
+                    done = true;
+                }
+            }
+        }
+
+        public RecordedHeartbeat Age(DateTime cutoff)
```
SzymonPobiega: How about calling this `MarkDeadIfNoRecentHeartbeat` ? Every time I come back to this class I need to figure out what does the `Age` method do ;-)

src/ServiceControl/Monitoring/EndpointInstanceMonitoring.cs
```diff
@@ -0,0 +1,88 @@
...
+    using System.Collections.Concurrent;
+    using System.Collections.Generic;
+    using System.Linq;
+    using ServiceControl.CompositeViews.Endpoints;
+    using ServiceControl.Contracts.HeartbeatMonitoring;
+    using ServiceControl.Infrastructure.DomainEvents;
+
+    public class EndpointInstanceMonitoring
+    {
+        const bool MONITOR_NEW_ENDPOINTS_BY_DEFAULT = true;
```
SzymonPobiega: Why Is it screaming at me? And why do we need this const anyway?

src/ServiceControl/Monitoring/EndpointMonitoringStats.cs
SzymonPobiega: @mikeminutillo hmmm I am a bit confused now. Can we remove the `Interlocked`?

SzymonPobiega:
I've rebased this on top of the measter and approved.

SzymonPobiega:
Do we need anything more here? I'll go fix the inspections.

mikeminutillo:
@SzymonPobiega I think it's OK. The only external difference is what happens when an endpoint goes offline while SC is down. SC will show it as down, but it won't raise an external integration event. As long as that's OK, I think we can merge this.

SzymonPobiega:
I think it is fine @mikeminutillo . @WilliamBZA ?

WilliamBZA:
:shipit:

SzymonPobiega:
I squashed the commits. Will merge when green.

SzymonPobiega:
Great job @mikeminutillo ❤️

Issue #964, opened 2017-05-15
ServiceControl Manager doesn't reflect new port when it is changed


kbaley:
## Steps to reproduce

1. Install an SC instance
1. Edit the instance and change the port

After saving the changes, the port displayed in the manager remains as the old one. Restarting the manager will show the new (correct) port.

adamralph:
It's not just the port. I see the same behaviour with other fields such as the name of the error queue.

It's probably worth checking through all the fields to see if it's a general problem across all of them.

WilliamBZA:
This has been resolved, and will reflect in the next release of ServiceControl (1.4.1)

DavidBoike:
@WilliamBZA You mean 1.41.0 (now patched up to 1.41.2) right?

WilliamBZA:
@DavidBoike Correct, I do

adamralph:
Or, rather, 1.41.1?

PR #965, opened 2017-05-15
Instance refresh fails 

gbiellem:
Connects to #964

When changing values in the SC Management app like port the update is sometimes not reflected in the UI until restart. Simplified the refresh to clear and reload the data completely


WilliamBZA:
@gbiellem @SzymonPobiega happy for this to be merged?

SzymonPobiega:
@WilliamBZA go for it! :shipit:

Issue #908, opened 2017-01-31
Mgmt util not catching "The service cannot accept control messages at this time"


gbiellem:
When starting and stopping instances the underlying Windows Service Controller can refuse the request with "The service cannot accept control messages at this time".  The mgmt utility is not handling this state and  instead the error is bubbling up to the unhandled exception management

TraGicCode:
This is a really frustrating bug.  Any expectation on when this will be worked on?

SzymonPobiega:
@TraGicCode the fix has been merged. I will be released in the next SC release.

PR #967, opened 2017-05-30
Fix handling of "The service cannot accept control messages at this time"

gbiellem:
Connects to #908 

Correct the handling of start and stop  services throwing "The service cannot accept control messages at this time" when the service is in a pending state.  Prior to this fix this bubbled to the unhandled exception manager and was often sent through to Raygun

### DIFF
--- a/src/ServiceControlInstaller.Engine/Instances/ServiceControlInstance.cs
+++ b/src/ServiceControlInstaller.Engine/Instances/ServiceControlInstance.cs
@@ -344,8 +344,17 @@ namespace ServiceControlInstaller.Engine.Instances
                 return true;
             }
 
-            Service.Stop();
+            try
+            {
 
+                // Will throw with "The service cannot accept control messages at this time" message if the service is in a pending state 
+                Service.Stop();
+            }
+            catch (InvalidOperationException)
+            {
+                return false;
+            }
+            
             try
             {
                 Service.WaitForStatus(ServiceControllerStatus.Stopped, TimeSpan.FromSeconds(60));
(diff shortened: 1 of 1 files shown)


gbiellem:
@mikeminutillo  as discussed, I think we should release this and #965

src/ServiceControlInstaller.Engine/Instances/ServiceControlInstance.cs
```diff
@@ -344,8 +344,17 @@ public bool TryStopService()
...
             }
 
-            Service.Stop();
+            try
+            {
 
+                // Will throw with "The service cannot accept control messages at this time" message if the service is in a pending state 
+                Service.Stop();
+            }
+            catch (InvalidOperationException)
```
SzymonPobiega: Can any other than "The service cannot accept control messages" InvalidOperationException happen here? i.e. do we risk swallowing something we should not?

gbiellem: No it should be safe.  "Stop" only requests the stop operation. It does not wait for the service to stop or report a failure if the stop operation fails.  This exception will occur if the Windows ServiceControl manager rejects the "Stop" request and that should only occur if it thinks the service it's in a pending or bad state.

gbiellem:
@mikeminutillo the fact that users are getting this message at all to me means there may be an issue with SC shutting down. -  To get into this state the service needs to take longer than 30 seconds to shutdown.  From SCMU perspective it's hard to know if this is just slowness or an actual problem.   In testing I assimilated this with a dummy service that always took 60 seconds to shutdown.  I got no issue with us adding a more detailed message to cover this case and indicate there was a timeout waiting for the service to shutdown but I'm 50/50 as too how much value  it would add

gbiellem:
FYI whatever the final fix here is also needs to be applied to baseservice.cs in the add monitoring to scmu  branch

PR #971, opened 2017-06-06
[WIP] Add update banner

boblangley:
Add an update banner that notifies the users that instances have upgrades, and allows them to filter the instance list to only those that have upgrades.

image

Still some issues with the checkbox and awaiting a final determination if this is a viable MVP for the feature.

### DIFF
--- a/src/ServiceControl.Config/Styles/Icons.xaml
+++ b/src/ServiceControl.Config/Styles/Icons.xaml
@@ -338,4 +338,12 @@
             </Canvas>
         </Viewbox>
     </ControlTemplate>
+
+    <ControlTemplate x:Key="UpgradeIcon">
+        <Viewbox Stretch="Uniform">
+            <Canvas Width="12" Height="15">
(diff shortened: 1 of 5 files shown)


SimonCropp:
do we need a "hide banner until next updates"?

boblangley:
I think it is OK for us to be naggy here, since this isn't an app that the end user works for any significant length of time in, and it's purpose is to install, configure, and upgrade. 

Having said that I would not object strongly to adding the ignore option if others want it.

Besides this, should SCMU take into account the license with regards to upgrades? For customers who have a perpetual license, but are outside upgrade protection this process could be annoying. Perhaps a different banner for them that tells them there are new versions and encourage them to update their license.

Another suggestion would be to actually count the updates in some way and lead them to release notes. This would be good to give them context and encourage them to stay up to date.

mikeminutillo:
I wouldn't spend too much time making this feature overly complex or smart. As you said, people are likely to open this app to start/stop/create/delete instances. They aren't hanging out in here.

Whether or not instances can be upgraded depends on which version of SCMU you have installed, not which versions have been released. People will only be presented with a "nag-screen" for something they cannot update if they have installed a version of ServiceControl that their license doesn't cover.

boblangley:
@mikeminutillo: After reflecting on it, and consulting with @sergioc , I am not convinced this PR adds value. Most people will only have a few instances, and the banner doesn't tell them anything the instances doesn't already. If we want to make the upgrade link stand out that would probably be a better effort. I am going to close this for now.

Issue #977, opened 2017-06-21
SagaAudit Duplication Key Fatal Exception, but with no information on what key


danielHalan:
Hi, 

When trying to start Service Control a Fatal Exception is thrown causing Service Control to Stop, the error message informs that there is a duplicate Saga key, but not which ID. (Using Log Level = Trace)

Also how would one solve this issue? That is in which Raven collection or ASB queue would one find this duplicate record, that needs to be removed. 

``
2017-06-21 09:25:21.9356|50|Info|NServiceBus.Unicast.Transport.TransportReceiver|Failed to process message
System.ArgumentException: An item with the same key has already been added.
   at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at ServiceControl.SagaAudit.SagaAuditing.SagaRelationshipsEnricher.Enrich(IReadOnlyDictionary`2 headers, IDictionary`2 metadata) in C:\BuildAgent\work\7189a56f9f44affc\src\ServiceControl\SagaAudit\SagaAuditing.cs:line 36
   at ServiceControl.Operations.AuditQueueImport.ConvertToSaveMessage(TransportMessage message) in C:\BuildAgent\work\7189a56f9f44affc\src\ServiceControl\Operations\AuditQueueImport.cs:line 118
   at ServiceControl.Operations.AuditQueueImport.InnerHandle(TransportMessage message) in C:\BuildAgent\work\7189a56f9f44affc\src\ServiceControl\Operations\AuditQueueImport.cs:line 93
   at ServiceControl.Operations.AuditQueueImport.Handle(TransportMessage message) in C:\BuildAgent\work\7189a56f9f44affc\src\ServiceControl\Operations\AuditQueueImport.cs:line 52
   at NServiceBus.Satellites.SatelliteLauncher.HandleMessageReceived(Object sender, TransportMessageReceivedEventArgs e, ISatellite satellite) in C:\BuildAgent\work\3206e2123f54fce4\src\NServiceBus.Core\Satellites\SatelliteLauncher.cs:line 89
   at NServiceBus.Unicast.Transport.TransportReceiver.OnTransportMessageReceived(TransportMessage msg) in C:\BuildAgent\work\3206e2123f54fce4\src\NServiceBus.Core\Unicast\Transport\TransportReceiver.cs:line 413
[...]

PR #978, opened 2017-06-21
Fix saga audit import

SzymonPobiega:
Connects to #977

* Deduplicates entries in the `NServiceBus.InvokedSagas` header to ensure there is one state change record created for each message even if that message caused invocation of two saga methods
* Handles various state change overrides well in case there were multiple methods invoked i.e. Completed overrides New overrides Updated

### Rationale

I've decided to go with deduplication instead of retaining multiple state changes in the list because I am not sure how the downstream consumers would react if they get a message that caused two state changes of a single saga instance. Would order matter etc.?

### DIFF
--- a/src/ServiceControl/SagaAudit/SagaAuditing.cs
+++ b/src/ServiceControl/SagaAudit/SagaAuditing.cs
@@ -53,7 +53,9 @@
                         }
                     }
 
-                    var sagas = sagasInvokedRaw.Split(';')
+                    var invokedSagas = SplitInvokedSagas(sagasInvokedRaw);
+                    
+                    var sagas = invokedSagas
                         .Distinct()
                         .Select(saga =>
                         {
@@ -126,6 +128,33 @@
                     });
                 }
 
+            }
+
+            static IEnumerable<string> SplitInvokedSagas(string sagasInvokedRaw)
+            {
+                var semicolonCount = sagasInvokedRaw.Count(c => c == ';');
+                var colonCount = sagasInvokedRaw.Count(c => c == ':');
+                if (colonCount != semicolonCount + 1) //Malformed data coming from old version of saga audit plugin
+                {
+                    var tailSemicolon = sagasInvokedRaw.LastIndexOf(";", StringComparison.Ordinal);
+                    var tail = sagasInvokedRaw.Substring(tailSemicolon + 1);
+                    var head = sagasInvokedRaw.Substring(0, tailSemicolon);
+
+                    var headDeduplicated = head.Substring(0, head.Length / 2);
+
+                    foreach (var part in SplitInvokedSagas(headDeduplicated))
+                    {
+                        yield return part;
+                    }
+                    yield return tail;
+                }
+                else
+                {
+                    foreach (var part in sagasInvokedRaw.Split(';'))
+                    {
+                        yield return part;
+                    }
+                }
             }
         }
     }
--- a/src/ServiceControl/SagaAudit/SagaAuditing.cs
+++ b/src/ServiceControl/SagaAudit/SagaAuditing.cs
@@ -19,7 +19,7 @@
             context.Container.ConfigureComponent<SagaRelationshipsEnricher>(DependencyLifecycle.SingleInstance);
         }
 
-        class SagaRelationshipsEnricher : ImportEnricher
+        public class SagaRelationshipsEnricher : ImportEnricher
         {
             public override void Enrich(IReadOnlyDictionary<string, string> headers, IDictionary<string, object> metadata)
             {
--- a/src/ServiceControl/SagaAudit/SagaAuditing.cs
+++ b/src/ServiceControl/SagaAudit/SagaAuditing.cs
@@ -42,10 +42,10 @@
                             {
                                 sagasChanges[id] = thisChange;
                             }
-                            else //Completed overrides New overrides Updated
+                            else
                             {
-                                if (string.Equals(thisChange, "Completed", StringComparison.OrdinalIgnoreCase)
-                                    || string.Equals(thisChange, "New", StringComparison.OrdinalIgnoreCase) && string.Equals(previousChange, "Updated", StringComparison.OrdinalIgnoreCase))
+                                if (thisChange == "Completed"   //Completed overrides everything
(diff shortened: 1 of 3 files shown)


src/ServiceControl/SagaAudit/SagaAuditing.cs
```diff
@@ -35,11 +35,26 @@ public override void Enrich(IReadOnlyDictionary<string, string> headers, IDictio
...
+                            var id = part[0];
+                            var thisChange = part[1];
+                            string previousChange;
+                            if (!sagasChanges.TryGetValue(id, out previousChange))
+                            {
+                                sagasChanges[id] = thisChange;
+                            }
+                            else //Completed overrides New overrides Updated
+                            {
+                                if (string.Equals(thisChange, "Completed", StringComparison.OrdinalIgnoreCase)
```
mikeminutillo: The plugin only ever generates these values with this case. Do we need the ignore case?

Some alternatives:
- Weirdly the values we care about are in alphabetical order. We could just collect them all and sort alphabetically and then return the first one
- We could parse the value into an enum and then sort them numerically

- If we go with a collect and sort later approach this whole thing could be replaced with an ILookup

```csharp
var values = Enumerable.Empty<string[]>();
if...
  values = multiSagaChanges.Select(s => s.Split(':'));
var lookup = values.ToLookup(v => v.FirstOrDefault(), v => v.Skip(1).FirstOrDefault());

// ...
.Select(saga => 
  change = lookup[sagaInvoked[1]].OrderBy(x => x).FirstOrDefault();
```

src/ServiceControl/SagaAudit/SagaAuditing.cs
```diff
@@ -35,11 +35,26 @@ public override void Enrich(IReadOnlyDictionary<string, string> headers, IDictio
...
+                                    || string.Equals(thisChange, "New", StringComparison.OrdinalIgnoreCase) && string.Equals(previousChange, "Updated", StringComparison.OrdinalIgnoreCase))
+                                {
+                                    sagasChanges[id] = thisChange;
+                                }
+                            }
                         }
                     }
 
                     var sagas = sagasInvokedRaw.Split(';')
```
mikeminutillo: Does the previous SagaAudit plugin send us broken data in the form `Type:IdType:Id;Type:Id`? If so, should we handle that? We can detect that because we'll have 2 colons.

SzymonPobiega: Well, we might get more than two. These are possible forms:
 * `Type1:Id1`
 * `Type1:Id1Type1:Id1;Type2:Id`
 * `Type1:Id1Type1:Id1;Type2:IdType1:Id1Type1:Id1;Type2:Id;Type3:Id3`

src/ServiceControl/SagaAudit/SagaAuditing.cs
SzymonPobiega: 👍 for switching to case-sensitive comparison. Not sure I like the "automatic" ordering/comparisons. They look less explicit.

src/ServiceControl/SagaAudit/SagaAuditing.cs
mikeminutillo: Urgh. We might need to create a separate Parser to pull those apart. Are the Ids always going to be Guids? At least then we know the length of them.

SzymonPobiega:
@mikeminutillo I've pushed some changes. It should now handle the malformed headers.

src/ServiceControl/SagaAudit/SagaAuditing.cs
```diff
@@ -127,6 +129,33 @@ public override void Enrich(IReadOnlyDictionary<string, string> headers, IDictio
...
+            {
+                var semicolonCount = sagasInvokedRaw.Count(c => c == ';');
+                var colonCount = sagasInvokedRaw.Count(c => c == ':');
+                if (colonCount != semicolonCount + 1) //Malformed data coming from old version of saga audit plugin
+                {
+                    var tailSemicolon = sagasInvokedRaw.LastIndexOf(";", StringComparison.Ordinal);
+                    var tail = sagasInvokedRaw.Substring(tailSemicolon + 1);
+                    var head = sagasInvokedRaw.Substring(0, tailSemicolon);
+
+                    var headDeduplicated = head.Substring(0, head.Length / 2);
```
mikeminutillo: Doesn't this rely on the two sagas being invoked being the same type. What if the incoming message invokes two different sagas?

src/ServiceControl/SagaAudit/SagaAuditing.cs
```diff
@@ -112,6 +129,33 @@ public override void Enrich(IReadOnlyDictionary<string, string> headers, IDictio
                 }
 
             }
+
+            static IEnumerable<string> SplitInvokedSagas(string sagasInvokedRaw)
```
WilliamBZA: Looking at the plugin code, the format will ALWAYS include at least 1 semicolon if there is >= 1 duplicate, correct?

src/ServiceControl/SagaAudit/SagaAuditing.cs
SzymonPobiega: Nope. The head is always duplicated because of the `+=`. But I'll add a proper test for this case to make sure.

src/ServiceControl/SagaAudit/SagaAuditing.cs
SzymonPobiega: yeah, if there is more than one entry (which means at least one duplicated) there will always be a semicolon.

src/ServiceControl/SagaAudit/SagaAuditing.cs
```diff
@@ -19,7 +19,7 @@ protected override void Setup(FeatureConfigurationContext context)
             context.Container.ConfigureComponent<SagaRelationshipsEnricher>(DependencyLifecycle.SingleInstance);
         }
 
-        class SagaRelationshipsEnricher : ImportEnricher
+        public class SagaRelationshipsEnricher : ImportEnricher
```
SzymonPobiega: It actually is `internal` because it is a nested class. Previously it was `private`.

SzymonPobiega:
@WilliamBZA @mikeminutillo ready for another round of reviews.

PR #981, opened 2017-07-03
Move logging configuration earlier

### LINKED ISSUE #966: ServiceControl logs before logfile settings are loaded
### Problem 
The `Bootstrapper`  class creates a new instance of the `Settings` class before running the `Bootstrapper.Initialize` method.

The `Settings` class writes to log files in some methods for various reasons. Some of these methods that cause log entries are run as part of the ctor of the `Settings` class. This means that we are writing log entries before the `Bootstrapper` can configure the logging correctly.

### Who is affected
Everyone using ServiceControl

### Proposed fix:

Move all of the code from the `Settings` class's ctor to an `Initialize` method and ensure that the new method is called explicitly in the pipeline when it makes sense to do so. Any necessary settings can remain as auto-loaded in the ctor, but all non-required ones should be loaded as part of an explicit method call.

mikeminutillo:
Fixes #966 

- Log settings cannot come from registry (which settings should at this point?), only from config file
- Configure Logging as early as possible (in the Main method)
- Replace file header with a startup message

### DIFF
--- a/src/ServiceControl/Bootstrapper.cs
+++ b/src/ServiceControl/Bootstrapper.cs
@@ -2,7 +2,6 @@ namespace Particular.ServiceControl
 {
     using System;
     using System.IO;
-    using System.Linq;
     using System.Net;
     using System.ServiceProcess;
     using Autofac;
@@ -10,17 +9,12 @@ namespace Particular.ServiceControl
     using global::ServiceControl.Infrastructure.DomainEvents;
     using global::ServiceControl.Infrastructure.SignalR;
     using Microsoft.Owin.Hosting;
-    using NLog;
-    using NLog.Config;
-    using NLog.Layouts;
-    using NLog.Targets;
     using NServiceBus;
     using Raven.Client;
     using Raven.Client.Embedded;
     using ServiceBus.Management.Infrastructure;
     using ServiceBus.Management.Infrastructure.OWIN;
     using ServiceBus.Management.Infrastructure.Settings;
-    using LogLevel = NLog.LogLevel;
     using LogManager = NServiceBus.Logging.LogManager;
 
     public class Bootstrapper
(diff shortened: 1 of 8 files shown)


src/ServiceControl/Bootstrapper.cs
```diff
@@ -56,7 +50,7 @@ public Bootstrapper(Settings settings, BusConfiguration configuration)
         private void Initialize()
         {
             var loggingSettings = new LoggingSettings(settings.ServiceName);
-            ConfigureLogging(loggingSettings);
+            RecordStartup(loggingSettings);
```
WilliamBZA: The settings class at this stage could already written entries to a log file, specifically when loading the [audit and error log queues](https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/Infrastructure/Settings/Settings.cs#L35). 

The only way to prevent that is to actually instantiate the `settings` variable after this initialize method has been called.

What about:


```CSharp
var loggingSettings = new LoggingSettings(host.ServiceName);
settings = settings ?? new Settings(host.ServiceName);
RecordStartup(loggingSettings);
```

Which also means the settings instantiation in the ctor then has to be removed.

mikeminutillo: I made it so the `LoggingSettings` ctor does not log anything. The initialization of the logging framework now happens in the entry point for the console application. This is actually the 2nd time we create a `LoggingSettings` instance in the lifecycle of the application. It was easier to create a new instance than to pass the instance from the entry point all the way to the bootstrapper.

WilliamBZA: `LoggingSettings` is not the problem, the issue is the `Settings` class which could log as part of it's ctor

mikeminutillo: Yes but we now configure the logging long before referencing the `Settings` class [here](https://github.com/Particular/ServiceControl/pull/981/files#diff-a98df16b3668f42c86d0ab2da8be83b2R24)

WilliamBZA: Got it, missed that.

PR #988, opened 2017-07-27
Add additional classifiers for retry group

WilliamBZA:
## Who's affected

* Any users of ServicePulse and ServiceControl

## Description

Adds 2 additional retry group classifications. This allows the user to select all failed messages for a logical endpoint as well as a specific endpoint instance.

## Original Issue

This is the work ongoing for the launchwave.

I'd like to merge this now as the changes don't have a negative impact on the current platform tools and would ideally like to avoid a long-running branch or feature toggles where possible.

If this is released, it will add additional classifiers which will appear in ServicePulse, but will have no other ill effect.


src/ServiceControl/DbMigrations/1.41.3/RerunClassifiersMigration.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        public string Apply(IDocumentStore store)
+        {
+            using (var session = store.OpenSession())
+            {
+                var reclassifySettings = session.Load<ReclassifyErrorSettings>(ReclassifyErrorSettings.IdentifierCase);
+
+                if (reclassifySettings != null)
+                {
+                    reclassifySettings.ReclassificationDone = false;
```
mikeminutillo: Should we do an in-place reclassification here? This will mean that the system will start up and then this data will slowly appear

src/ServiceControl/Recoverability/API/FailureGroupsApi.cs
```diff
@@ -73,16 +76,36 @@ dynamic GetRetryHistory()
             }
         }
 
-        dynamic GetAllGroups(string classifier)
+        dynamic GetAllGroups(string classifier, string classifierFilter = null)
```
mikeminutillo: Is the default needed? don't we always pass in null or a value?

src/ServiceControl/Recoverability/API/GroupFetcher.cs
```diff
@@ -62,14 +62,19 @@ private static UnacknowledgedRetryOperation[] MapAcksToOpenGroups(FailureGroupVi
...
+        private static FailureGroupView[] GetDBGroups(IDocumentSession session, string classifier, string classifierFilter)
         {
             var groups = session.Query<FailureGroupView, FailureGroupsViewIndex>()
-                .Where(v => v.Type == classifier)
-                .OrderByDescending(x => x.Last)
+                .Where(v => v.Type == classifier);
+
+            if (!string.IsNullOrWhiteSpace(classifierFilter))
+            {
+                groups = groups.Where(v => v.Title == classifierFilter);
```
mikeminutillo: Is equality the only type of filter? Should these parameters be `type` and `title`?

Does case matter here?

src/ServiceControl/Recoverability/Grouping/Groupers/ClassifiableMessageDetails.cs
```diff
@@ -3,24 +3,33 @@ namespace ServiceControl.Recoverability
...
-            var last = message.ProcessingAttempts.Last();
+            ProcessingAttempt = message.ProcessingAttempts.Last();
 
-            Details = last.FailureDetails;
-            MessageType = (string)last.MessageMetadata["MessageType"];
+            Details = ProcessingAttempt.FailureDetails;
+            MessageType = (string)ProcessingAttempt.MessageMetadata["MessageType"];
         }
 
         public ClassifiableMessageDetails(string messageType, FailureDetails failureDetails)
```
mikeminutillo: Can we get rid of this constructor? It's only used for tests and for the data migration. What if the user is updating from a version that requires the data migration? They will not receive the new classifications currently.

src/ServiceControl/DbMigrations/1.41.3/RerunClassifiersMigration.cs
WilliamBZA: This is only a problem if ServiceControl is not updated via the installer, correct? Doesn't the upgrade process run the migrations meaning by the time SC starts up the data will be updated already?

I'll verify first though...

It might still be worth changing the reclassification process to append instead of re-classifying _all_ of the existing classifications.

src/ServiceControl/Recoverability/API/GroupFetcher.cs
WilliamBZA: The `Type` is filtered already above.

This is a lucene query, so case is important. The filter value comes from the values that the metrics plugin (and core) provides though, so the case should match between them.

WilliamBZA:
Thanks @mikeminutillo. I've addressed some of your comments. I'll try and come up with a better migration story too.

src/ServiceControl/DbMigrations/1.41.3/RerunClassifiersMigration.cs
WilliamBZA: Verified, you are correct. I've submitted another PR into this one to run the migration synchronously as part of the setup. https://github.com/Particular/ServiceControl/pull/989

PR #989, opened 2017-07-28
Run the migrations synchronously as part of the setup command

WilliamBZA:
This should be a bit better for what you had in mind. The migration now runs synchronously and the handler merely invokes the class that actually does the migration.


WilliamBZA:
The message is still used when receiving a POST to the reclassify messages endpoint, but we can definitely drop the startup message now.

WilliamBZA:
@mikeminutillo happy for me to merge this, and then merge the tf branch PR too?

mikeminutillo:
@WilliamBZA go for it

PR #991, opened 2017-08-07
Fix Retry Operations

### LINKED ISSUE #990: Operations for non-bulk retries marked as complete too early
ServiceControl keeps track of retry operations and their state in memory. For non-bulk (single message, set of user-selected messages) operations, this in-memory state machine is not created until the batch has already been staged, when it is [lazily accessed for the first time](https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/Recoverability/Retrying/RetryProcessor.cs#L59). This leads to an operation containing 0 messages, where 0 messages have been forwarded and 0 messages have been skipped. The operation appears to be complete, even though the batch hasn't actually been forwarded to it's final destination anywhere.

mikeminutillo:
Fixes #990 

Creates the retry operation for single and user-selected sets of messages. Ensures operation exists later when messages are being sent.


mikeminutillo:
I'm going to improve the logging in the retries section as well as a part of fixing this.

mikeminutillo:
Rebased on master

mikeminutillo:
Curse. Null bodies

PR #992, opened 2017-08-07
Make url for specific group id more unique

WilliamBZA:
Makes the url for loading a specific group more unique so that it doesn't clash with the classifier url


SzymonPobiega:
Does SP need to be updated?

WilliamBZA:
This URL is currently only used in the LW code for service pulse, so does not require a release there, but the PR for that is here: https://github.com/Particular/ServicePulse/pull/488

Also, this change was not released to customers so no need to release this

WilliamBZA:
To clarify: The ServicePulse code is in the `tf-wave-metrics` branch, so hasn't been released either

PR #993, opened 2017-08-07
[SPIKE] Give raven a chance to catch up

WilliamBZA:


src/ServiceControl.AcceptanceTests/AcceptanceTest.cs
```diff
@@ -87,7 +87,7 @@ protected AcceptanceTest()
         [SetUp]
         public void Setup()
         {
-            port = FindAvailablePort(33333);
+            port = FindAvailablePort(33335);
```
mikeminutillo: Why change the default port? Just so it doesn't conflict with your existing SC instance?

Do we know how much time we spend in this method per run? I wonder if we couldn't be smarter by building a circular buffer and always moving forwards.

WilliamBZA: I'm trying everything at this point. This was to eliminate caching. This PR will never be merged.

So far it seems like the Nancy routing is behaving differently locally vs on TC. I have no idea why. The routes that match are different.

WilliamBZA:
Found a working fix. Will submit a proper PR shortly

PR #994, opened 2017-08-08
Use querystring parameter for filter value

WilliamBZA:
Fixes the tests on the build server

### DIFF
--- a/src/ServiceControl/Recoverability/API/FailureGroupsApi.cs
+++ b/src/ServiceControl/Recoverability/API/FailureGroupsApi.cs
@@ -27,8 +27,13 @@ namespace ServiceControl.Recoverability
             Post["/recoverability/groups/reclassify"] =
                 _ => ReclassifyErrors();
 
-            Get["/recoverability/groups/{classifier?Exception Type and Stack Trace}/{classifierFilter?}"] =
-                parameters => GetAllGroups(parameters.Classifier, parameters.classifierFilter == "undefined" ? null : parameters.classifierFilter);
+            Get["/recoverability/groups/{classifier?Exception Type and Stack Trace}"] =
+               parameters =>
+               {
+                   var classifierFilter = Request.Query["classifierFilter"] != "undefined" ? Request.Query["classifierFilter"] : null;
+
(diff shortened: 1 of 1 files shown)


WilliamBZA:
I think using a querystring parameter for the filter condition is appropriate and represents the intent that it is, in fact, used in a query. The filter is not an identifier portion of a resource.

src/ServiceControl/Recoverability/API/FailureGroupsApi.cs
```diff
@@ -27,8 +27,13 @@ public FailureGroupsApi()
             Post["/recoverability/groups/reclassify"] =
                 _ => ReclassifyErrors();
 
-            Get["/recoverability/groups/{classifier?Exception Type and Stack Trace}/{classifierFilter?}"] =
-                parameters => GetAllGroups(parameters.Classifier, parameters.classifierFilter == "undefined" ? null : parameters.classifierFilter);
+            Get["/recoverability/groups/{classifier?Exception Type and Stack Trace}"] =
+               parameters =>
+               {
+                   var classifierFilter = Request.Query["classifierFilter"] == null || Request.Query["classifierFilter"] == "undefined" ? null : Request.Query["classifierFilter"];
```
SzymonPobiega: Do we need the first `Request.Query["classifierFilter"] == null` check? 

If I understand correctly this
```
var classifierFilter = Request.Query["classifierFilter"] == null || Request.Query["classifierFilter"] == "undefined" ? null : Request.Query["classifierFilter"];
```
means `classifierFilter` will be either some value of `null` if `classifierFilter` query string param is not present or its value is `"undefined"`.

This should be equivalent:

```
var classifierFilter = Request.Query["classifierFilter"] == "undefined" ? null : Request.Query["classifierFilter"];
```

or even better

```
var classifierFilter = Request.Query["classifierFilter"] != "undefined" ? Request.Query["classifierFilter"] : null;
```

WilliamBZA: Thanks @SzymonPobiega, that's great. I've updated

PR #996, opened 2017-08-10
Add additional archive group log statements

WilliamBZA:
Adds extra logging for the archive group operation


WilliamBZA:
Looking at the platdev discussion it seems to revolve around structured loggers. We don't support other loggers within ServiceControl do we?

mikeminutillo:
Agreed. Happy to leave it for now but we should reach consensus sometime. 

I broke the build. Pushed a fix. Once it passes, this can be merged.

PR #997, opened 2017-08-15
Add Feature Toggle for Monitoring Instances

mikeminutillo:
Adds a feature toggle for monitoring instances. 

By default, Management Utility will not show Monitoring Instances. If you set the app-config setting `Enable-Feature:MonitoringInstances` to `true` then you will get UI to start monitoring instances.

### DIFF
--- a/src/ServiceControl.Config/Features.cs
+++ b/src/ServiceControl.Config/Features.cs
@@ -2,7 +2,7 @@
 
 namespace ServiceControl.Config
 {
-    using System.Collections.Concurrent;
+    using System.Collections.Generic;
     using System.Collections.Specialized;
     using System.Configuration;
     using Autofac;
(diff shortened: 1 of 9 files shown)


src/ServiceControl.Config/Features.cs
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    public static class Feature
+    {
+        public const string MonitoringInstances = "MonitoringInstances";
+    }
+
+    public class FeatureToggles
+    {
+        private ConcurrentDictionary<string, bool> features = new ConcurrentDictionary<string, bool>(StringComparer.InvariantCultureIgnoreCase);
```
WilliamBZA: Doesn't seem like the dictionary part nor the concurrent part is necessary. Maybe a HashSet?

PR #1002, opened 2017-09-11
ASB duplicate events

mikeminutillo:
Due to the way we subscribe to our own events, we are setting up duplicate subscriptions for our events in Azure Service Bus (using the default topology). This means when we publish a message with a parent class, it matches the filters of two subscriptions, each of which sends a copy to our input queue. ServiceControl gets two copies of these messages.

To solve this we have flattened out the event hierarchy. 

## Subclasses of `MessageFailureResolved`
  - `FailedMessageArchived`
  - `MessageFailureResolvedByRetry`
  - `MessageFailureResolvedManually`

There was no handlers for `MessageFailureResolved`. This class has been removed and the corresponding subclasses now all inherit directly from `IEvent`

## Subclasses of `HeartbeatStatusChanged`
  - `HeartbeatingEndpointDetected`
  - `HeartbeatMonitoringEnabled`
  - `HeartbeatMonitoringDisabled`
  - `EndpointHeartbeatRestored`
  - `EndpoingFailedToHeartbeat`

There was one handler for `HearbeatStatusChanged` that simply publishes a heartbeats updated message. This handler has been modified to handle each of the subclasses explicitly and the base class has been removed. All of the corresponding subclasses now inherit directly from `IEvent`.

## `MessageFailedRepeatedly : MessageFailed`

This hierarchy has been split and all properties of `MessageFailed` lifted up to `MessageFailedRepeatedly`. A new property has been added to both message types to indicate repeated message failure. The system no longer publishes `MessageFailedRepeatedly`, preferring to publish `MessageFailed` with the flag set to true. All handlers for both classes have been duplicated (to allow for in-flight messages to be handled properly). In the future, `MessageFailedRepeatedly` and all associated handlers can be removed (and this has been noted as a comment on the message type).

## Test

There is a unit test added to detect deep hierarchies of events and fail the build if these are introduced in the future. The test will be updated to point back to this issue.

### DIFF
--- a/src/ServiceControl.Config/Extensions/VisualTreeExtensions.cs
+++ b/src/ServiceControl.Config/Extensions/VisualTreeExtensions.cs
@@ -1,4 +1,5 @@
-﻿namespace ServiceControl.Config.Extensions
+﻿// ReSharper disable HeuristicUnreachableCode
+namespace ServiceControl.Config.Extensions
 {
     using System.Windows;
     using System.Windows.Controls;
@@ -22,6 +23,7 @@
                 var child = VisualTreeHelper.GetChild(parent, i);
                 // If the child is not of the request child type child
                 var childType = child;
+                // ReSharper disable once ConditionIsAlwaysTrueOrFalse
                 if (childType == null)
                 {
                     // recursively drill down the tree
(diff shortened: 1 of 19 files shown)


SzymonPobiega:
@Particular/servicecontrol-maintainers the acceptance tests pass on most of the transports (including ASB). Please do a final review and merge :-D

PR #1012, opened 2017-09-28
Make ForwaredAuditMessages switch optional (1011)

jmolohon:


SzymonPobiega:
@jmolohon This parametr is intentionally mandatory (while `ForwardErrorMessages` is intentionally optional). This mimics how the UI is design. The rationale is that for the `ForwardErrorMessages` option there is a sensible default (don't) because SC manages the retry flow.

For the `ForwaredAuditMessages` there is no sensible default. For scenarios where you install SC for an existing and running system, users might have some kind of audit processing tool so they would need to check `yes`. For new systems it does not make sense. We want users to think about it before they install SC, either via UI or via the script.

jmolohon:
I see, then my issue is with the type, SwitchParameter, which does not accept a true or false variable as input. I will submit a new pull request that reverts the ForwardAuditMessage type to bool.

PR #1015, opened 2017-10-05
Restore missing header

### LINKED ISSUE #1010: "Deployed instances" header missing in SCMU empty state
"Deployed instances" header should show in SCMU empty state, currently missing:

2017-09-26 at 09 54

mikeminutillo:
Fixes #1010 

Connects to #1010 

With no instances:
withnoitems

With instances:
withitem

### DIFF
--- a/src/ServiceControl.Config/UI/ListInstances/ListInstancesViewModel.cs
+++ b/src/ServiceControl.Config/UI/ListInstances/ListInstancesViewModel.cs
@@ -17,7 +17,7 @@
         public ListInstancesViewModel(Func<BaseService, InstanceDetailsViewModel> instanceDetailsFunc)
         {
             this.instanceDetailsFunc = instanceDetailsFunc;
-            DisplayName = "";
+            DisplayName = "DEPLOYED INSTANCES";
 
             Instances = new BindableCollection<InstanceDetailsViewModel>();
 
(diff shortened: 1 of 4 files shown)


SzymonPobiega:
@mikeminutillo is this ready to be merged?

sergioc:
Needs more spacing between “Deployed instances” and the refresh icon. See the following screenshot for reference:

pasted image at 2017_11_24 09_50 am

Also the distance between “deployed instances” + “refresh” from the separator line below should be equal to the distance between “New instance” and the separator line.

The refresh button shouldn't be omited either.

mikeminutillo:
image

> The refresh button shouldn't be omitted either.

@sergioc can you elaborate on what that means?

mikeminutillo:
Currently the logic to show the refresh button looks like this

```csharp
public bool ShowRefresh => HasInstances && !ShowOverlay && !IsModal;
```

I think we should drop the requirement for instances. In theory, if you used powershell or some other mechanism to create your first instance while you had the UI open, you'd have no way to refresh the UI.

sergioc:
@mikeminutillo The update looks good.

>> The refresh button shouldn't be omitted either.

> can you elaborate on what that means?

When there are no instances configured, the "Refresh" button should still show at the top. In general you don't want to "magically" hide any kind of UI elements without the reason being obvious.

PR #1017, opened 2017-10-05
Remaining gitversiontask

boblangley:
After updating GitVersion task in SC main other projects needed to be updated as well to work. This only leaves some unit test projects

This PR updates Packaging, PowerShell, Config, Engine, Setup, and SqlServerWithDTC to latest version. It removes it from the SC unit tests project.


WilliamBZA:
Oh man, what a disaster. Sorry guys.

PR #1018, opened 2017-10-05
Nsb 5.2.24

boblangley:
Updates ServiceControl to NSB 5.2.24

### DIFF
--- a/src/ServiceControl/ServiceControl.csproj
+++ b/src/ServiceControl/ServiceControl.csproj
@@ -156,8 +156,7 @@
       <Private>True</Private>
     </Reference>
     <Reference Include="NServiceBus.Core, Version=5.0.0.0, Culture=neutral, PublicKeyToken=9fc386479f8a226c, processorArchitecture=MSIL">
-      <HintPath>..\packages\NServiceBus.5.2.21\lib\net45\NServiceBus.Core.dll</HintPath>
-      <Private>True</Private>
+      <HintPath>..\packages\NServiceBus.5.2.24\lib\net45\NServiceBus.Core.dll</HintPath>
     </Reference>
     <Reference Include="NServiceBus.NLog, Version=1.1.0.0, Culture=neutral, PublicKeyToken=9fc386479f8a226c, processorArchitecture=MSIL">
       <HintPath>..\packages\NServiceBus.NLog.1.1.0\lib\net45\NServiceBus.NLog.dll</HintPath>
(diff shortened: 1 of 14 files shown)


boblangley:
This PR relies on #1017 and should be applied afterwards. This is only WIP as long as #1017 has not been merged.

src/ServiceControl/App.config
```diff
@@ -14,5 +14,43 @@ These settings are only here so that we can debug ServiceControl while developin
...
+      </dependentAssembly>
+      <dependentAssembly>
+        <assemblyIdentity name="Microsoft.Data.OData" publicKeyToken="31bf3856ad364e35" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-5.6.4.0" newVersion="5.6.4.0" />
+      </dependentAssembly>
+      <dependentAssembly>
+        <assemblyIdentity name="Microsoft.WindowsAzure.Storage" publicKeyToken="31bf3856ad364e35" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-4.3.0.0" newVersion="4.3.0.0" />
+      </dependentAssembly>
+    </assemblyBinding>
```
mikeminutillo: I don't know that these changes are necessary as SC has some tricksy code to always load the version on local disk.

src/ServiceControl/ServiceControl.csproj
```diff
@@ -656,7 +655,12 @@
       <Install>false</Install>
     </BootstrapperPackage>
   </ItemGroup>
-  <ItemGroup />
+  <ItemGroup>
+    <ProjectReference Include="..\ServiceControl.LearningTransport\ServiceControl.LearningTransport.csproj">
+      <Project>{12f4b317-b885-4d43-a152-fdb5b0ea2042}</Project>
+      <Name>ServiceControl.LearningTransport</Name>
+    </ProjectReference>
+  </ItemGroup>
```
mikeminutillo: What's this?

boblangley: That's not supposed to be in there. I will update the PR when I get home.

src/ServiceControl/App.config
boblangley: Ah, OK. I will drop this from the PR.

src/ServiceControl/ServiceControl.csproj
mikeminutillo: ;)

PR #1020, opened 2017-10-07
[SPIKE]] Learning transport

boblangley:
I marked this as a spike since it was not a planned PR. I have been working on this in my spare time, being mostly frustrated that SC does not support learning transport when doing product demos with docs samples that do. I have no built up some obsolete knowledge on V5 transports 😄 

I based this on the V6 transport, making some changes to match V5 contracts / initialization style.  It is introduced as a separate project that the Service Control project references. One change made from the V6 Core version of the transport is the switch to supporting and relying solely on a connection string,. which is simply a path. It no longer scans for a root solution directory, and I dropped support for configuring the transport directory via code.

It also brings in an acceptance test project using the 5.2.24 acceptance tests. Some of those tests have been removed or changed as (I thought) appropriate. All of the tests pass (on my machine). The acceptance tests require a environment variable, like the others, to set the connection string.

I added the transport to the Config project so it can be selected when an instance is added. I have no problems adding a new instance.

I smoke tested by using the V6 fault tolerance sample, adding audit and error configuration and providing a mechanism to fail in the handler, but succeed when it retries later. I was able to successfully add a new SC instance configured with the learning transport using the V6 Fault Tolerance sample `.learningtransport` directory for the connection string. This limited smoke test passed.

This is submitted as embedded and only for SC, but it could be done as a separate repository with a nuget, with only a change to the Packages / Config project needed for SC to support it.

### DIFF
--- a/src/ServiceControl.AcceptanceTesting/ServiceControl.AcceptanceTesting.csproj
+++ b/src/ServiceControl.AcceptanceTesting/ServiceControl.AcceptanceTesting.csproj
@@ -40,8 +40,7 @@
   <ItemGroup>
     <Reference Include="Microsoft.CSharp" />
     <Reference Include="NServiceBus.Core, Version=5.0.0.0, Culture=neutral, PublicKeyToken=9fc386479f8a226c, processorArchitecture=MSIL">
-      <HintPath>..\packages\NServiceBus.5.2.21\lib\net45\NServiceBus.Core.dll</HintPath>
-      <Private>True</Private>
+      <HintPath>..\packages\NServiceBus.5.2.24\lib\net45\NServiceBus.Core.dll</HintPath>
     </Reference>
     <Reference Include="System" />
     <Reference Include="System.Configuration" />
--- a/src/ServiceControl.AcceptanceTesting/packages.config
+++ b/src/ServiceControl.AcceptanceTesting/packages.config
@@ -1,4 +1,4 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="NServiceBus" version="5.2.21" targetFramework="net452" />
+  <package id="NServiceBus" version="5.2.24" targetFramework="net452" />
 </packages>
--- a/src/ServiceControl.AcceptanceTests/ServiceControl.AcceptanceTests.csproj
+++ b/src/ServiceControl.AcceptanceTests/ServiceControl.AcceptanceTests.csproj
@@ -103,8 +103,7 @@
       <Private>True</Private>
     </Reference>
     <Reference Include="NServiceBus.Core, Version=5.0.0.0, Culture=neutral, PublicKeyToken=9fc386479f8a226c, processorArchitecture=MSIL">
-      <HintPath>..\packages\NServiceBus.5.2.21\lib\net45\NServiceBus.Core.dll</HintPath>
-      <Private>True</Private>
+      <HintPath>..\packages\NServiceBus.5.2.24\lib\net45\NServiceBus.Core.dll</HintPath>
     </Reference>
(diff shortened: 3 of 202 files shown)


SzymonPobiega:
Awesome work @boblangley . I think we can get it even smaller by removing native pub/sub and timeout support.

WilliamBZA:
@boblangley this is great!

mikeminutillo:
@boblangley I haven't had a chance to look at this yet but I had a couple of questions.

a) How hard would it be to push this over to core v5? This would enable v5 endpoints to run on the LearningTransport as well which simplifies some of the doco things like samples.

b) We should have a discussion about relaxing license rules when running on the Learning Transport. This would make it much easier for customers to try out SC.

SzymonPobiega:
@mikeminutillo do we have learning transport for V6? If so than maybe adding it to V6 makes sense but adding it should not block this issue. I agree, we can basically allow users to run on learning transport for free.

SzymonPobiega:
Also, given how @udidahan expressed the goal of the Lunch Wave (show people the value in Advanced License) would it make sense to add this to SC so that we can show the whole Advanced stuff on the learning transport and hence give it for free (on the learning transport of course)

andreasohlund:
Core 6.4 has the learning transport

SzymonPobiega:
@andreasohlund then the impact of having the learning transport for V5 seems to be quite big. We could run the whole platform on the learning transport.

mikeminutillo:
I am working now on getting everything running on Learning Transport using this PR so see if we can package everything into a single zip file. So far, so good

andreasohlund:
>then the impact of having the learning transport for V5 seems to be quite big

Just to make sure I understand: The impact comes from us not having to upgrade SC to v6 right?

Because users "learning" on v5 doesn't seem to have much impact?

mauroservienti:
How would one is expected to run/use this?

1. do I need to setup a SC on learning transport for every project/solution I'm experimenting with?
1. can I double click on `ServiceControl.exe` in the same directory as the solution I'm playing with and get it working with zero friction?

I currently have probably more than 100 NServiceBus samples, spikes and demos on my machine that could be converted to use the learning transport, do I need to setup a SC instance for everyone? Or can I configure each of them, and SC, to use for example `c:\temp\.learningtransport` as a base path?

SzymonPobiega:
@andreasohlund you're right. There is not that big value in having learning transport fully supported for V5.

@mauroservienti If I remember correctly @boblangley used connection string to pass the path so you can set-up a SC instance and then re-configure it for any sample by changing the connection string.

WilliamBZA:
> Because users "learning" on v5 doesn't seem to have much impact?

Users wouldn't learn on V5, but they would be able to learn the platform, which is arguably a large impact

boblangley:
@mauroservienti With SC integration you have to specify the path, we can't really do autodiscovery. So you are correct, however creating new instances is fairly quick.

I see most people using this with the Microservices Workshop and the Showcase examples, or anything custom they have done for pre-sales. Not with every sample a user might be interested in, since many of them don't demonstrate retries anyhow.

That is why I limited this to only support within the current version of SC, it is a stop-gap until we convert to V6/V7.

boblangley:
@mikeminutillo I think the idea of relaxing licensing rules is good. The v5 discussion I agree, the value is small. Learn on V6, but be able to support demonstrating  SC / SP with specific samples/demonstrations would be awesome.

boblangley:
@SzymonPobiega I attempted to pull out the pub/sub and timeout stuff, but ran into snags. Considering all the tests pass, both the transport tests and the SC tests, and this was meant to be a quick win, I suggest we consider keeping them in.

WilliamBZA:
Should this be closed?

PR #1022, opened 2017-10-11
Extend trial link added to licence error messages

tmasternak:
Solves https://github.com/Particular/LaunchWave.DevOpsAdvancedMonitoring/issues/266

### DIFF
--- a/src/ServiceControl.Config/Xaml/Controls/ExtendTrialHyperlink.cs
+++ b/src/ServiceControl.Config/Xaml/Controls/ExtendTrialHyperlink.cs
@@ -1,0 +1,36 @@
+﻿namespace ServiceControl.Config.Xaml.Controls
+{
+    using System;
+    using System.Diagnostics;
+    using System.Windows;
+    using System.Windows.Documents;
+
+    public class ExtendTrialHyperlink : Hyperlink
+    {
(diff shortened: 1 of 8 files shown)


src/ServiceControl.Config/Xaml/Controls/ExtendTrialHyperlink.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        public ExtendTrialHyperlink()
+        {
+            NavigateUri = new Uri(ExtendTrialUrl);
+            Inlines.Add(ExtendTrialUrl);
+
+            Style = new Style(GetType(), FindResource(typeof(Hyperlink)) as Style);
+        }
+
+        protected override void OnClick()
```
HEskandari: Could use existing command instead of custom OnClick. The command 'OpenURLCommand' does this already, all you need to do is to pass in the URL and bind it to the user action (Command property?)

HEskandari:
Not sure if asking them to extend trial every time is a good thing? We have a similar case in SI and there is an [open issue ](https://github.com/Particular/CustomerSuccess/issues/524) for it in CustomerSuccess which suggest the exact opposite.

tmasternak:
We decided to go with #1026 and disable checks temporarily.

PR #1025, opened 2017-10-17
New popup overlay

HEskandari:
Resolves https://github.com/Particular/LaunchWave.DevOpsAdvancedMonitoring/issues/300


src/ServiceControl.Config/UI/Shell/NewInstancePopup.xaml.cs
```diff
@@ -0,0 +1,23 @@
...
+	public partial class NewInstanceOverlay
+	{
+		public NewInstanceOverlay()
+		{
+			InitializeComponent();
+		}
+
+		private void ViewPreviewMouseDown(object sender, MouseButtonEventArgs e)
+		{
+			Console.WriteLine(sender.GetType().Name);
```
mikeminutillo: Should this still be here?

HEskandari: Definitely not :) Removed.

PR #1027, opened 2017-10-18
Always show monitoring instances

mikeminutillo:
Relies on #1026 

Always show monitoring instances.

Wait for release to beta.

### DIFF
--- a/src/ServiceControl.Config/Features.cs
+++ b/src/ServiceControl.Config/Features.cs
@@ -50,6 +50,7 @@ namespace ServiceControl.Config
 
         public void Start()
         {
+            featureToggles.Enable(Feature.MonitoringInstances);
             if (DateTime.Today > new DateTime(2018, 2, 15, 0, 0, 0, DateTimeKind.Utc))
             {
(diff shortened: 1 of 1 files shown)


src/ServiceControl.Config/Features.cs
```diff
@@ -50,6 +50,7 @@ public FeatureToggleDefaults(FeatureToggles featureToggles)
 
         public void Start()
         {
+            featureToggles.Enable(Feature.MonitoringInstances);
```
WilliamBZA: Why not simply remove the `ShowMonitoringInstances` checks that we have, assuming that we will always show monitoring instances in perpetuity?

mikeminutillo: If we leave it in, it's easy to reverse it if something goes wrong. Once everything is stable we can remove the toggle.

WilliamBZA: We can achieve the same with a revert commit.

We could also publish a default app.config that has the config flag enabled. In fact, I might prefer this option instead of hard-coding it on or off if we're not sure how long until it's stable

PR #1031, opened 2017-11-17
Split transports by instance type

mikeminutillo:
Enable monitoring instances and servicecontrol instances to have their own transport types.

Required to pull in https://github.com/Particular/ServiceControl.Monitoring/pull/74


boblangley:
History fail, closing because #1034 was merged already.

PR #1034, opened 2017-11-17
Split v5 and v6 transports

### LINKED ISSUE #1033: SCMU fails to install SC.Monitoring for SQL Server
SCMU fails to install SC.Monitoring for SQL Server because it has a typo in the configuration:

        <add key="Monitoring/TransportType" value="NServiceBus.SqlServerTransport, NServiceBus.Transports.SQLServer" />

This should be:

        <add key="Monitoring/TransportType" value="NServiceBus.SqlServerTransport, NServiceBus.Transport.SQLServer" />

Probably the following logic needs to be fixed:

- https://github.com/Particular/ServiceControl/blob/master/src/ServiceControlInstaller.Engine/Instances/Transports.cs#L42

boblangley:
Fixes #1033 

The dll portion of the TransportType string on https://github.com/Particular/ServiceControl/blob/master/src/ServiceControlInstaller.Engine/Instances/Transports.cs#L42 references the v5 assembly name of the transport. For v6 version of the transport the assembly name was changed, breaking Monitoring installs.

### DIFF
--- a/src/ServiceControl.Config/UI/SharedInstanceEditor/SharedMonitoringEditorViewModel.cs
+++ b/src/ServiceControl.Config/UI/SharedInstanceEditor/SharedMonitoringEditorViewModel.cs
@@ -16,7 +16,7 @@
 
         public SharedMonitoringEditorViewModel()
         {
-            Transports = ServiceControlInstaller.Engine.Instances.Transports.All;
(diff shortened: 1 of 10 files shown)


WilliamBZA:
Used the generated installer to install a SQL instance, seemed to install fine on my machine

PR #1035, opened 2017-11-17
Bump monitoring to 1.0.0-alpha157

boblangley:
Updates to alpha157 which selects a default topology for Azure Service Bus transport.


boblangley:
Withdrawn for #1032

PR #1045, opened 2017-12-05
Beta tag removed from add monitoring instance dialog

tmasternak:


tmasternak:
@Particular/servicecontrol-maintainers can you please review this?

PR #1048, opened 2017-12-12
Restore license checks

mikeminutillo:
Put license checks back in place for the RC of SC Monitoring


SzymonPobiega:
@mikeminutillo is it ready to merge now to be released 1.44.0 ?

mikeminutillo:
@SzymonPobiega will need to check with @tmasternak as it should go with the RC version of SC Monitoring

PR #1055, opened 2017-12-18
Reimport failed audits command line option

SzymonPobiega:
Connects to https://github.com/Particular/PlatformDevelopment/issues/1593

Main changes implemented:

### Code

 * Allow passing `LoggingSettings` to `Bootstrapper` so that command line processes can default to different log levels (e.g. Info)
 * Allow passing `Settings` to `Bootstrapper` to be able to customize the way SC behaves e.g. disable audit and error ingestion when running command-line tooling
 * Extract the audit message conversion code to `ProcessedMessage` to a separate class so that it can be reused outside of the `AuditQueueImport` satellite
 * Allow disabling audit and error ingestion through `Settings`
 * Convert stored failed audits to `FailedTransportMessage` instead of `FailedAuditImport` to correctly deserialize headers
 * Add a command line option `--import-failed-audits` that attempts to re-import all failed audits

### Tests

 * Add `FailedAuditsModule` REST API to be able to test re-importing
 * Add a test that verifies that re-importing failed audits works correctly


src/ServiceControl/Infrastructure/Settings/Settings.cs
```diff
@@ -106,6 +106,9 @@ public TimeSpan HeartbeatGracePeriod
         public bool ForwardAuditMessages { get; set; }
         public bool ForwardErrorMessages { get; set; }
 
+        public bool IngestAuditMessages { get; set; } = true;
+        public bool IngestErrorMessages { get; set; } = true;
```
mikeminutillo: I cannot wait to take further advantage of these

src/ServiceControl/Operations/ErrorQueueImport.cs
```diff
@@ -98,7 +98,7 @@ public void Stop()
 
         public Address InputAddress => settings.ErrorQueue;
 
-        public bool Disabled => InputAddress == Address.Undefined;
+        public bool Disabled => InputAddress == Address.Undefined || !settings.IngestErrorMessages;
```
mikeminutillo: *Can* you leave the error address undefined?

src/ServiceControl/Operations/ImportFailedAudits.cs
```diff
@@ -0,0 +1,86 @@
...
+            {
+                var query = session.Query<T, I>();
+                using (var ie = session.Advanced.Stream(query))
+                {
+                    while (!token.IsCancellationRequested && ie.MoveNext())
+                    {
+                        FailedTransportMessage dto = ((dynamic)ie.Current.Document).Message;
+                        try
+                        {
+                            var transportMessage = new TransportMessage(dto.Id, dto.Headers)
```
mikeminutillo: In v5 there are some properties of TransportMessage that are not linked to corresponding headers. ReplyToAddress is one. Recoverable is one as well. In most cases, if the message has a corresponding header then it gets read from there, otherwise it has some internal state variable. If these are not persisted carefully, we run the risk of losing data

mikeminutillo: Check https://github.com/Particular/NServiceBus/blob/support-5.2/src/NServiceBus.Core/TransportMessage.cs for a complete implementation of TransportMessage. It looks like TimeToBeReceived is one (and I was wrong about ReplyToAddress).

SzymonPobiega: @mikeminutillo what should we do with this property then? If set, assign it to a header?

src/ServiceControl/Operations/ErrorQueueImport.cs
SzymonPobiega: Yeah. In `Settings.GetErrorQueue` there is conditional.

src/ServiceControl/Operations/AuditImporter.cs
```diff
@@ -0,0 +1,63 @@
...
+
+            bodyStorageEnricher.StoreAuditMessageBody(
+                message.Body,
+                message.Headers,
+                metadata);
+
+            var auditMessage = new ProcessedMessage(message.Headers, metadata)
+            {
+                // We do this so Raven does not spend time assigning a hilo key
+                Id = $"ProcessedMessages/{Guid.NewGuid()}"
```
WilliamBZA: I wonder if a deterministic ID is better here to prevent duplicate `ProcessedMessage`s in the event of restarts at just the right time (see below)

src/ServiceControl/Operations/ImportFailedAudits.cs
```diff
@@ -0,0 +1,86 @@
...
+                                Body = dto.Body
+                            };
+
+                            var entity = auditImporter.ConvertToSaveMessage(transportMessage);
+                            using (var storeSession = store.OpenSession())
+                            {
+                                storeSession.Store(entity);
+                                storeSession.SaveChanges();
+                            }
+                            store.DatabaseCommands.Delete(ie.Current.Key, null);
```
WilliamBZA: The `ProcessedMessage` could be persisted to raven and before the key is deleted, we could fail. Are we ok with duplicate `ProcessedMessage`s?

src/ServiceControl/Operations/AuditImporter.cs
SzymonPobiega: That's a very interesting question. I've just copied the code. I would assume that the indexes on the query side deduplicate the results based on the message ID but I need to check it.

WilliamBZA: If it's copy-paste then I don't think this scope should include changing that behaviour.

SzymonPobiega: I did a test by sending the same audit message twice and it seems like we are not deduplicating it in the current version so this does not make it worse. I'll try to create a separate fix for the duplication problem.

SzymonPobiega: @WilliamBZA I've raised a separate issue to deal (or not deal) with the duplication problem https://github.com/Particular/ServiceControl/issues/1056

WilliamBZA: :+1:

src/ServiceControl/Operations/ImportFailedAudits.cs
mikeminutillo: Can do, but then you need to set it on TransportMessage on the other end too. Recoverable is literally just a boolean on the class. I guess all I'm saying is that in V5, a transport message is slightly more than (id, body, headers) and we need to be sure we don't lose those extra bits

SzymonPobiega: Looking at the code only `Recoverable` and `TTBR` are not stored as headers. All the other stuff is. And for audits we don't really need these two as `Recoverable` is always set to `true` (for some legacy reason) and `TTBR` is not used.

Why do you think it is broken for recoverability?

WilliamBZA: If the concern is around recoverability, it shouldn't actually be a concern since this is auditing? Also, since it's copy-paste behaviour from existing import pipeline it should be ok.

These are read-only once imported, right? ServiceInsight can't re-send audit messages, so even if we get the TTBR or Recoverable flags wrong, it's not a deal-breaker surely.

mikeminutillo: OK, you have convinced me

PR #1057, opened 2017-12-19
[SPIKE] Add audit message deduplicatoin when indexing

SzymonPobiega:


mikeminutillo:
Currently ServiceInsight is the only (official) consumer of this API. How does it handle cases where we have duplicate Audit messages?

danielmarbach:
Should we extract ideas from here and close it?

PR #1062, opened 2018-01-10
Amazon SQS added as a supported transport to the monitoring v6 transports

danielmarbach:
Related to https://github.com/Particular/ServiceControl.Monitoring/pull/98

It looks like the V6 transports are currently only used for monitoring purposes. So we went ahead and added SQS to that type. Since we figured it is best to stick to the original design. Note as soon as we start using the v6 transports elsewhere it might be beneficial to separate the transports supported by SC from the ones supported by SC.Monitoring.

The packaging has not yet been pushed to staging so this build might fail
  

image

### DIFF
--- a/src/ServiceControlInstaller.Engine/Instances/V6Transports.cs
+++ b/src/ServiceControlInstaller.Engine/Instances/V6Transports.cs
@@ -50,6 +50,13 @@
                 TypeName = "NServiceBus.RabbitMQTransport, NServiceBus.Transports.RabbitMQ",
                 MatchOn = "NServiceBus.RabbitMQ",
                 SampleConnectionString = "host=<HOSTNAME>;username=<USERNAME>;password=<PASSWORD>"
+            },
+            new TransportInfo
+            {
+                Name = "AmazonSQS",
+                TypeName = "NServiceBus.SqsTransport, NServiceBus.AmazonSQS",
+                MatchOn = "NServiceBus.AmazonSQS",
+                SampleConnectionString = "Region=<REGION>"
             }
         };
     }
--- a/src/ServiceControlInstaller.Engine/Instances/V6Transports.cs
+++ b/src/ServiceControlInstaller.Engine/Instances/V6Transports.cs
@@ -55,7 +55,7 @@
             {
                 Name = "AmazonSQS",
                 TypeName = "NServiceBus.SqsTransport, NServiceBus.AmazonSQS",
-                MatchOn = "NServiceBus.AmazonSQS",
+                MatchOn = "NServiceBus.SqsTransport",
                 SampleConnectionString = "AccessKeyId=<ACCESSKEYID>;SecretAccessKey=<SECRETACCESSKEY>;Region=<REGION>",
                 Help = "AccessKeyId will be promoted to AWS_ACCESS_KEY_ID, SecretAccessKey to AWS_SECRET_ACCESS_KEY and Region to AWS_REGION environment variable."
             }
--- a/src/ServiceControlInstaller.Engine/Instances/V6Transports.cs
+++ b/src/ServiceControlInstaller.Engine/Instances/V6Transports.cs
@@ -34,8 +34,7 @@
                 TypeName = "NServiceBus.MsmqTransport, NServiceBus.Core",
                 MatchOn = "NServiceBus.Msmq",
                 SampleConnectionString = string.Empty,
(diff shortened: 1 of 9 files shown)


SzymonPobiega:
LGTM

WilliamBZA:
Should we mark this as WIP until the package is pushed?

Scooletz:
I wipped it @WilliamBZA

src/ServiceControlInstaller.Engine/Instances/V6Transports.cs
```diff
@@ -50,6 +50,13 @@ public static TransportInfo FindByName(string name)
                 TypeName = "NServiceBus.RabbitMQTransport, NServiceBus.Transports.RabbitMQ",
                 MatchOn = "NServiceBus.RabbitMQ",
                 SampleConnectionString = "host=<HOSTNAME>;username=<USERNAME>;password=<PASSWORD>"
+            },
+            new TransportInfo
+            {
+                Name = "AmazonSQS",
+                TypeName = "NServiceBus.SqsTransport, NServiceBus.AmazonSQS",
+                MatchOn = "NServiceBus.AmazonSQS",
+                SampleConnectionString = "Region=<REGION>"
```
mikeminutillo: Should we include `Help` text about the expected environment variables?

danielmarbach:
We came up with a hybrid approach. When you specify the connection string
it takes precedence but if you leave empty it tries to read the env
variable. I'll add this to the help text

Am 11.01.2018 03:55 schrieb "Mike Minutillo" <notifications@github.com>:

danielmarbach:
@mikeminutillo addressed your comment

danielmarbach:

src/ServiceControlInstaller.Engine/Instances/V6Transports.cs
```diff
@@ -50,6 +50,14 @@ public static TransportInfo FindByName(string name)
...
                 MatchOn = "NServiceBus.RabbitMQ",
                 SampleConnectionString = "host=<HOSTNAME>;username=<USERNAME>;password=<PASSWORD>"
+            },
+            new TransportInfo
+            {
+                Name = "AmazonSQS",
+                TypeName = "NServiceBus.SqsTransport, NServiceBus.AmazonSQS",
+                MatchOn = "NServiceBus.AmazonSQS",
+                SampleConnectionString = "AccessKeyId=<ACCESSKEYID>;SecretAccessKey=<SECRETACCESSKEY>;Region=<REGION>",
+                Help = "AccessKeyId will be promoted to AWS_ACCESS_KEY_ID, SecretAccessKey to AWS_SECRET_ACCESS_KEY and Region to AWS_REGION environment variable."
```
mikeminutillo: Only for the current execution of the process. Rather than saying they will be promoted, can we say:

> If AccessKeyId is omitted then ServiceControl will use the value of the AWS_ACCESS_KEY_ID environment variable.
> If SecretAccessKey is omitted then ServiceControl will use the value of the AWS_SECRET_ACCESS_KEY environment variable.

mikeminutillo: Oh. It doesn't fall back any more. So even if I have these values in environment variables SC won't use them?

danielmarbach:
Correct. We couldn't get it working with they way the process runs. We
could only supporr it if we enforce system wide env variables which seems a
bit hard. That's why we take the mandatory input and promote it to process
variables

Am 15.01.2018 06:14 schrieb "Mike Minutillo" <notifications@github.com>:

src/ServiceControlInstaller.Engine/Instances/V6Transports.cs
danielmarbach: Correct. We couldn't get it working with they way the process runs. We
could only supporr it if we enforce system wide env variables which seems a
bit hard. That's why we take the mandatory input and promote it to process
variables

danielmarbach:
@Particular/servicecontrol-maintainers removed WIP since the SC.Monitoring 1.1.0 version has been released to myget. This is ready for a final review and then we can start releasing it

Scooletz:
@Particular/servicecontrol-maintainers , could you provide the number version for this release as well?
I've crafted this https://github.com/Particular/Announcements/pull/224 assuming 1.46.0 will be the next one. Same goes for the milestone and all the release related activities.

PR #1067, opened 2018-01-17
WIP [SPIKE] Dense Audit

Scooletz:
This PR is a spike of a more dense way of storing audits. My assumption was, that once we want to move away from Raven and move into a new storage, we could use a dense format and some append-only structures (for instance, AppenBlob in Azure) to make audits work. This is just a sketch to get some buy-in. If you want me to elaborate more, I can make the description more concrete.

### Azure deployment

The general premise would be to use two sets of AppendBlobs (if a blob fills up, one of the worker creates a new one with a name incremented by 1). The first set would be responsible for storying metadata, the second - bodies. To audit a message following actions would occur:
- serialize metadata with the new dense serialization
- append body to the second set (returns the file id + position + length => `long/int/int`)
- write serialized metadata with the `long/int/int` body id to the first set

This would not address the indexing. If needed, we could run a function that periodically reads the metadata stream and writes secondary index entries to a table, in a similar way to the azure storage persistence:
- PartitionKey is `PropertyName_value`
- RowKey is `messageId`

The scavenging would be not so hard: 
- read the oldest metadata file
- reindex data, but now, use values to remove them from the table
- remove bodies accordingly (I know, what if there was out of order append etc. we can keep 1 blob more just in case)

### Other deployments

AppendOnly blob can be replaced by any store. A date prefix could be used to get "some ordering" allowing prunning old data. 
Indexing can be done with any Key-Value as it's similar to RAMP transactions, just pointing to a document

### DIFF
--- a/src/ServiceControl/Data/HeaderId.cs
+++ b/src/ServiceControl/Data/HeaderId.cs
@@ -1,0 +1,120 @@
+﻿namespace Particular.ServiceControl.Data
+{
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Linq;
+    using System.Text;
+    using NServiceBus;
+
+    /// <summary>
+    /// Static class containing headers used by NServiceBus.
+    /// </summary>
+    public static class HeaderId
+    {
+        public const byte NotRegistered = byte.MaxValue;
+
+        static readonly Dictionary<string, byte> NameToId;
+        static readonly Dictionary<string, byte[]> NameToIdAsByteArray;
+        static readonly Dictionary<byte, string> IDToName;
+
+        static HeaderId()
+        {
+            NameToId = new Dictionary<string, byte>();
+            byte i = 1;
+
+            void Add(string value, ref byte id)
+            {
+                NameToId.Add(value, id);
+                id += 1;
+            }
+
+            Add(Headers.HttpFrom, ref i);
+            Add(Headers.HttpTo, ref i);
+            Add(Headers.RouteTo, ref i);
+            Add(Headers.DestinationSites, ref i);
+            Add(Headers.OriginatingSite, ref i);
+            Add(Headers.SagaId, ref i);
+            Add(Headers.MessageId, ref i);
+            Add(Headers.CorrelationId, ref i);
+            Add(Headers.ReplyToAddress, ref i);
+            Add(Headers.HeaderName, ref i);
+            Add(Headers.NServiceBusVersion, ref i);
+            Add(Headers.ReturnMessageErrorCodeHeader, ref i);
+            Add(Headers.ControlMessageHeader, ref i);
+            Add(Headers.SagaType, ref i);
+            Add(Headers.OriginatingSagaId, ref i);
+            Add(Headers.OriginatingSagaType, ref i);
+            Add(Headers.Retries, ref i);
+            Add(Headers.FLRetries, ref i);
+            Add(Headers.Retries, ref i);
+            Add(Headers.ProcessingStarted, ref i);
+            Add(Headers.ProcessingEnded, ref i);
+            Add(Headers.TimeSent, ref i);
+            Add(Headers.RelatedTo, ref i);
+            Add(Headers.EnclosedMessageTypes, ref i);
+            Add(Headers.ContentType, ref i);
+            Add(Headers.SubscriptionMessageType, ref i);
+            Add("NServiceBus.SubscriberAddress", ref i);
+            Add("NServiceBus.SubscriberEndpoint", ref i);
+            Add(Headers.IsSagaTimeoutMessage, ref i);
+            Add(Headers.IsDeferredMessage, ref i);
+            Add(Headers.OriginatingEndpoint, ref i);
+            Add(Headers.OriginatingMachine, ref i);
+            Add(Headers.OriginatingHostId, ref i);
+            Add(Headers.ProcessingEndpoint, ref i);
+            Add(Headers.ProcessingMachine, ref i);
+            Add(Headers.HostDisplayName, ref i);
+            Add(Headers.HostId, ref i);
+            Add(Headers.HasLicenseExpired, ref i);
+            Add(Headers.OriginatingAddress, ref i);
+            Add(Headers.ConversationId, ref i);
+            Add(Headers.MessageIntent, ref i);
+            Add("NServiceBus.TimeToBeReceived", ref i);
+            Add("NServiceBus.NonDurableMessage", ref i);
+
+            NameToIdAsByteArray = NameToId.ToDictionary(kvp => kvp.Key, kvp => new[] { kvp.Value });
+            IDToName = NameToId.ToDictionary(kvp => kvp.Value, kvp => kvp.Key);
+        }
+
+        public static bool TryGetId(string headerName, out byte id) => NameToId.TryGetValue(headerName, out id);
+        public static string GetName(byte id) => IDToName[id];
+
+        static readonly UTF8Encoding NoBom = new UTF8Encoding(false);
+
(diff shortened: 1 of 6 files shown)


mikeminutillo:
I've been wondering if we should have (at least) two separate processes for Audits. 

The first would just read from the input queue as quickly as possible and push that data (body and headers) into a blob storage. It could optionally forward to a log queue. This means we can ingest an audit message even if it's missing headers. We can run retention policies directly off the ingestion timestamp.

The second process crawls this data and assembles metadata. Depending on how quickly we can assemble metadata and how big it is (RAM-wise) we could potentially keep it in memory and stick a WebApi over it. Given that metadata is more or less unchanging, we can persist it but it can go into a different storage entirely. One based more on queryability.

SzymonPobiega:
@mikeminutillo we had a short discussion with @tmasternak yesterday about retention policies and after it I am more and more convinced that we don't need it. Instead, we should rotate multiple data stores and aggregate queries. I got it idea from @Scooletz and I believe we got it from the Aeron buffer model ;-). We load audits to one DB/namespace/account/file group/whatever, we have another one as read only and, depending on the need, the third one being removed. We need two instances of API, one over the active part and one over the passive part to server the queries.

Regarding the two-phase approach I am a bit worried that the second phase might be lagging behind the first phase. And before a give message undergoes the second phase it can't be returned by the queries. It just fills up space.

mikeminutillo:
The two phase approach keeps our ingestion separate from metadata extraction. This means we can add additional metadata operations at a later date by re-crawling. We can even do it while ingestion is offline. 

Technically we could have a routing metadata extractor that stores it's own data in it's own database and contributes to queries by plugging into an engine.

It also means that we can use the same infrastructure for audit/error queues. We'd need two things:
1. The ability for a record to be "pinned" (not cleaned up by retention). This might be offered by having a 3rd store that is not cleared up and moving records between them, or it can just be a database flag. All error records start pinned. They can be unpinned when retried (or if we come up with an infallible way of doing so, when we decide the message has been successfully processed). 
2. The ability to mark a record as retryable. Although we may be able to simply do this via metadata extraction if we can tell the difference between a message that has been audited and those that have not. Out of interest, if the retries capability was crawling the entire message store it could find it's own successful retries in the audit store and not rely on the audit ingestion capability to tell it.

NOTE: Pinning allows for more complex retention policies like "keep all of the messages of a conversation until there hasn't been a message for 3 weeks" and "when an error occurs, pin all of the audit messages in the same conversation". These can be created by pinning/unpinning based on certain events and then letting the standard retention policy to roll forward. 

Currently, before a message is ingested it is just taking up space only it's in the broker/input queue. In the case of a sudden spike in audit/error traffic, the two-phase approach ingests at the same pace and the metadata extraction can just catch up. i.e. We still have latency but we don't backfill the transport/broker. If there is a sustained difference, we can warn the user that metadata extraction is falling behind and recommend scaling up the metadata extractor (to do this, we'd need a way for metadata extractors to lay claim to a set of records so that they can scale/work-steal). 

In the case that a message is missing headers we can still ingest it, we can still apply a retention policy to it, we can even show it to the user and ask them to fix it. With that in place we could decide to pin a record when we run into metadata extraction problems.

Scooletz:
Wow! It's like... 😍 ! All your comments! _Stay put Szymon and answer to them_

@mikeminutillo 
> The first would just read from the input queue as quickly as possible and push that data (body and headers) into a blob storage. It could optionally forward to a log queue. This means we can ingest an audit message even if it's missing headers. We can run retention policies directly off the ingestion timestamp.

My assumption was to get big gains by dealing with metadata as soon as possible. This reduces the payload size, lowers the overall costs, increases throughput etc. If there were some missing headers, we could introduce a `BluntEntry` that puts the raw content. I assume this situation would be unlikely to occur.

@SzymonPobiega 

> I believe we got it from the Aeron buffer model ;

I didn't have it in mind. With append-only blobs, we could have up to 1000 blobs being persisted. As we'd use an increasing number for their names (long) we could go on and on forever, deleting the tail as needed. I was not assuming reusing the buffer. It's much easier to just use the infrastructure and delete the tail. Actually, when processing loads of messages with Azure Functions (2 billion a day) I used this approach and was much happier being able to move from "omg, I need to keep the position for this blog" to "just append & cut the tail when needed"

@mikeminutillo 

> Currently, before a message is ingested it is just taking up space only it's in the broker/input queue. In the case of a sudden spike in audit/error traffic, the two-phase approach ingests at the same pace and the metadata extraction can just catch up. i.e. We still have latency but we don't backfill the transport/broker. If there is a sustained difference, we can warn the user that metadata extraction is falling behind and recommend scaling up the metadata extractor (to do this, we'd need a way for metadata extractors to lay claim to a set of records so that they can scale/work-steal).

My assumption "extract and make it dense as early as possible" was meant to address the overall throughput as well. I'd dare to say that it's cheaper to read a message, make it dense and write it, than to read a message and store it in the 1TransportMessage1 form (which is extremely big, with all the headers' names, and enrichment serialized with JSON after all). Less IO, more probable to make it cloud-ready in a cheap manner.

@SzymonPobiega 

> I need to mention that. For the record: EventStore would be an ideal database for this design. The metadata extraction would be just a projection. We get the eviction for free.

That's correct. It's very EventStore-ish. There are a few remarks to mention:
- to make it IO-cheap, we'd still need to drop JSON, which would disallow to run regular v8-based ES projections. We'd need to use subscriptions anyway and run them somewhere
- ES has no indexes over properties (yeah, you can create a stream per value) and we'd need to index it anyway
- ES is like Raven, it's not Azure-ready and you need to run it. With a general Append-Only store, we can run on Azure with services they provide, and on premises, using even dummy file system.

### Note on pinning idea

This idea is approach-independent. No matter what would be chosen (meta+body or two-phase) we could use the same approach for checking for the eviction.

mikeminutillo:
I think manipulation of the headers is a mistake which is why I want to treat them like a binary blob. Headers, like message bodies, are immutable which means you can cache them forever. Do we need to compact each set of headers or would picking a datastore that allows something like gzip compression do the job just as well? I suppose, from a logical perspective, it doesn't matter how the data is stored.

There are always headers we don't necessarily expect and we need to be able to store them too. 

Maybe it makes sense to compact the metadata because it contains only what we expect. Even then I can see a case where we have different services putting data in different places. For example, we might have a routing component that wants to know (sending/receiving endpoint, message type) and a performance component that wants to extract (sent time, processing started, processing ended, processing time, critical time). There's no reason those sets of data need to be stored together or even extracted at the same time. 

Within a UI experience like ServiceInsight we query for a list of messages by conversationId and then ask for the metadata relevant to the visualization we are showing. For a message flow, maybe we just need message type and causation (relatedTo header). For a sequence diagram we need more timing info. 

Once we have basic layout sorted we can layer other info on. Want to show recoverability data? We show a red light indicating that a message has failed and has not been retried. This is less cacheable than message type because it is more volatile.

Scooletz:
@mikeminutillo 

> There are always headers we don't necessarily expect and we need to be able to store them too.

That was not my intention. My intention was to write the well-known headers in a dense format (1 byte for the key) and all the unknown ones (a.k.a external) write with a regular `utf8` name encoding. 

> There's no reason those sets of data need to be stored together or even extracted at the same time.

I see your point and I'm totally ok with having some different extractors. I'd still argue, that the first phase, storing messages, could be done in a better, more dense way, provided by this PR (HeaderId etc).  Just to provide some data, I can refer to my old header-related [PR for the core](https://github.com/Particular/NServiceBus/pull/4802) where I made some measurements. I think that gzip over a byte body won't provide us the gains we could obtain with providing a custom schema that's based on our knowledge (NSB headers are frequest -> better encode them efficiently).

PR #1080, opened 2018-01-25
Api facade saga audit

SzymonPobiega:


src/ServiceControl/SagaAudit/ApiModule.cs
```diff
@@ -1,45 +1,26 @@
...
-                        .WithModel(sagaHistory)
-                        .WithLastModified(lastModified);
-                }
-            };
+            Get["/sagas/{id}", true] = (parameters, token) => GetSagaByIdApi.Execute(this, (Guid) parameters.id);
 
+            Get["/sagas", true] = (_, token) => ListSagasApi.Execute(this, NoInput.Instance);
 
-            Get["/sagas"] = _ =>
+            Get["/sagas2"] = _ =>
```
danielmarbach: This should probably be reverted

src/ServiceControl/CompositeViews/Messages/ScatterGatherApiMessageView.cs
```diff
@@ -1,17 +1,22 @@
 namespace ServiceControl.CompositeViews.Messages
 {
     using System.Collections.Generic;
+    using System.Linq;
     using Nancy;
 
-    public abstract class ScatterGatherApiMessageView<TInput> : ScatterGatherApi<TInput, MessagesView>
+    public abstract class ScatterGatherApiMessageView<TInput> : ScatterGatherApi<TInput, List<MessagesView>>
```
danielmarbach: By using IList here we could avoid the many ToLists but then we'd probably need to cast the IList to list in order to use the sort method which is only available on list :(

src/ServiceControl/SagaAudit/ListSagasApi.cs
```diff
@@ -0,0 +1,32 @@
+namespace ServiceControl.SagaAudit
+{
+    using System.Linq;
+    using System.Threading.Tasks;
+    using Nancy;
+    using Raven.Client;
+    using ServiceControl.CompositeViews.Messages;
+    using ServiceControl.Infrastructure.Extensions;
+
+    public class ListSagasApi : ScatterGatherApi<NoInput, SagaListIndex.Result[]>
```
danielmarbach: Would it make sense to use here IList because that is was is returned from raven but then does newtonsoft support IList?

PR #1081, opened 2018-01-26
Api facade instanceids

boblangley:
- @SeanFeldman and I implemented an `InstanceId` attribute to each `MessageView` returned, which is simply the Base64 encoded string representing the APIUrl for the instance delivering the view. 
- We modified the Retries API to use the InstanceId instead of instance name, and removed that setting.
- We added tests for all including checking for population of `InstanceId`, and added coverage for missing MessageView APIs.

### DIFF
--- a/src/ServiceControl.AcceptanceTests/MultiInstance/When_issuing_retry_on_master.cs
+++ b/src/ServiceControl.AcceptanceTests/MultiInstance/When_issuing_retry_on_master.cs
@@ -69,8 +69,7 @@
                         new RemoteInstanceSetting
                         {
                             ApiUri = addressOfRemote,
-                            QueueAddress = Remote1,
-                            InstanceName = Remote1,
+                            QueueAddress = Remote1
                         }
                     };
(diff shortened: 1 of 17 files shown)


danielmarbach:
@SzymonPobiega and I discussed the following:

When we take the Api URI as an identifier what would happen if the topology changes (like the port)? The implications would be that the returned base64 string is changed. So you'd have to restart SI. An instance ID might be more stable and could be documented to be unique and stable. On the other hand we concluded that changing the topology requires to restart SC master anyway and then it is probably also OK if you have to restart SI.

danielmarbach:
Some additional thoughts which make me currently lean towards the instance id approach. An URI by default has already 17 characters:

http:// or https:// (7-8)
:33335 (port) (4-6)
/api (endpoint) (4) (you could argue this could be removed)

not taking into account the domain part. With base64 encoding, the uri gets at least 25% bigger. Given that we add this information to every message view returned is base64 encoding of the URI a good idea?

PR #1082, opened 2018-01-26
WIP Scatter-gather bypassing when instance_id present

danielmarbach:
This would technically allow bypassing the scatter-gather when the instance id is present. Originally it was intended to be there for the saga list. Unfortunately, the saga list is not used from Service Insight. The question is would we still want this change despite the complexity it introduces into the scatter-gather API?

### DIFF
--- a/src/ServiceControl/SagaAudit/ListSagasApi.cs
+++ b/src/ServiceControl/SagaAudit/ListSagasApi.cs
@@ -4,6 +4,7 @@ namespace ServiceControl.SagaAudit
     using System.Linq;
     using System.Threading.Tasks;
     using Nancy;
+    using Raven.Abstractions.Extensions;
     using Raven.Client;
     using ServiceControl.CompositeViews.Messages;
     using ServiceControl.Infrastructure.Extensions;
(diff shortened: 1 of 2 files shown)


danielmarbach:
@Particular/servicecontrol-maintainers who uses the /sagas API?

src/ServiceControl/SagaAudit/ListSagasApi.cs
```diff
@@ -21,6 +22,8 @@ public class ListSagasApi : ScatterGatherApi<NoInput, List<SagaListIndex.Result>
                     .ToListAsync()
                     .ConfigureAwait(false);
 
+                results.ForEach(r => r.Uri = $"{r.Uri}?instance_id={instanceId}");
```
boblangley: Since ServiceInsight doesn't actually call this API, can we just remove it?

boblangley: ServicePulse does not use the list sagas API either. I vote we remove this API.

boblangley:
Daniel, I rebased this branch, but pushed it up to https://github.com/Particular/ServiceControl/tree/saga-uri-rebased so you can confirm the conflict resolution didn't screw anything up.

src/ServiceControl/SagaAudit/ListSagasApi.cs
mikeminutillo: I agree. If there are no official consumers of this API we should remove it.

mikeminutillo:
I spiked the approach described here https://github.com/Particular/ServiceControl/pull/1085

danielmarbach:
Closing based on https://github.com/Particular/ServiceControl/pull/1085#issuecomment-361564992

PR #1085, opened 2018-01-30
Spike Routed API approach

mikeminutillo:
Based on my review of #1082

### DIFF
--- a/src/ServiceControl/CompositeViews/Messages/RoutedApi.cs
+++ b/src/ServiceControl/CompositeViews/Messages/RoutedApi.cs
@@ -1,0 +1,128 @@
+namespace ServiceControl.CompositeViews.Messages
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Globalization;
+    using System.IO;
+    using System.Linq;
+    using System.Net.Http;
+    using System.Threading.Tasks;
+    using Nancy;
+    using Newtonsoft.Json;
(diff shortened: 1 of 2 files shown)


danielmarbach:
I'm good with that approach since it is also more intention revealing. My original question from the PR remains: Do we need it since we are removing the ListSagasAPI?

SzymonPobiega:
Isn't the main point of this feature to be able to retry messages from SI?

mikeminutillo:
@SzymonPobiega yes. The Retry POST is where we need it.

danielmarbach:
@SzymonPobiega @mikeminutillo not exactly. The base class here assumes querying of either local or remote. As it currently stands the decision was to send messages to remote and not call the API with forwarding, see https://github.com/Particular/ServiceControl/blob/api-facade/src/ServiceControl/MessageFailures/Api/RetryMessages.cs#L20 

@mikeminutillo are you proposing to switch to API forwarding then?

SzymonPobiega:
@danielmarbach yeah. I forgot about that. So in this case I propose to keep the current solution, delete the `ListSagas` API and park this SPIKE until we discover an API that needs it.

SzymonPobiega:
@danielmarbach @mikeminutillo we can close this one also, right?

danielmarbach:
Closing since we currently don't need such an API. We can reopen it if required

PR #1088, opened 2018-01-31
Api facade - GetBody via remote

boblangley:
Queries the remote instance if the request indicates the body is stored on the remote and not local.

Relates to https://github.com/Particular/PlatformDevelopment/issues/1650
Connects to #1069

### DIFF
--- a/src/ServiceControl/CompositeViews/Messages/RoutedApi.cs
+++ b/src/ServiceControl/CompositeViews/Messages/RoutedApi.cs
@@ -97,7 +97,6 @@ namespace ServiceControl.CompositeViews.Messages
 
                 var rawResponse = await httpClient.SendAsync(requestMessage).ConfigureAwait(false);
 
-                // TODO: is it correct to merge headers??
                 var headers = rawResponse.Headers.Union(rawResponse.Content.Headers).ToDictionary(k => k.Key, v => v.Value.FirstOrDefault());
                 var httpStatusCode = (Nancy.HttpStatusCode) Enum.Parse(typeof(HttpStatusCode), rawResponse.StatusCode.ToString(), ignoreCase: true);
 
--- a/src/ServiceControl/CompositeViews/Messages/RoutedApi.cs
+++ b/src/ServiceControl/CompositeViews/Messages/RoutedApi.cs
@@ -1,37 +1,29 @@
 namespace ServiceControl.CompositeViews.Messages
 {
     using System;
-    using System.Collections.Generic;
-    using System.Globalization;
-    using System.IO;
     using System.Linq;
     using System.Net.Http;
     using System.Threading.Tasks;
     using Nancy;
-    using Newtonsoft.Json;
     using NServiceBus.Logging;
     using Raven.Client;
-    using ServiceBus.Management.Infrastructure.Extensions;
-    using ServiceBus.Management.Infrastructure.Nancy;
     using ServiceBus.Management.Infrastructure.Nancy.Modules;
     using ServiceBus.Management.Infrastructure.Settings;
     using ServiceControl.Infrastructure.Settings;
     using HttpStatusCode = System.Net.HttpStatusCode;
 
-    public abstract class RoutedApi<TIn, TOut> : IApi
-        where TOut : class
+    public abstract class RoutedApi<TIn> : IApi
     {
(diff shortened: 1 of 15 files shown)


src/ServiceControl/Operations/BodyStorage/RavenAttachments/GetBodyById.cs
```diff
@@ -2,20 +2,65 @@
...
+                        var rawResponse = await httpClient.SendAsync(new HttpRequestMessage(HttpMethod.Get, instanceUri)).ConfigureAwait(false);
+
+                        if (rawResponse.StatusCode == System.Net.HttpStatusCode.OK)
+                        {
+                            return new Response
+                                {
+                                    Contents = stream => rawResponse.Content.CopyToAsync(stream).GetAwaiter().GetResult()
+                                }
+                                .WithContentType(rawResponse.Content.Headers.ContentType.ToString())
+                                .WithHeader("Expires", DateTime.UtcNow.AddYears(1).ToUniversalTime().ToString("R"))
```
mikeminutillo: This should probably just be copied from the remote as well

src/ServiceControl/Operations/BodyStorage/RavenAttachments/GetBodyById.cs
```diff
@@ -2,20 +2,65 @@
...
+                                .WithHeader("Content-Length", rawResponse.Content.Headers.ContentLength.ToString())
```
mikeminutillo: Can we just iterate through the remote headers and return all of them? If we can generically copy the http response into a nancy response then we can abstract a lot of this away. 

On incoming, if the request is for a remote then call it and pass through it's response.
If the request is not for a remote then execute the local code and return a response.

SeanFeldman: > I'd want to see a lot of the mechanics removed from the Nancy module

@mikeminutillo do you mean moved out from Nancy module into a "helper" class? Yes, that would make sense to keep the module simpler. Keep in mind that we've just added code to delegate to remotes to what was already in place.

SzymonPobiega: 👍 for making a copy of the selected headers we care about.

danielmarbach:
@boblangley @SeanFeldman I was on the fence whether I should push directly to this branch or send in another PR against this one. Since @mikeminutillo and @SzymonPobiega asked for a more bypass approach I figured I'll add it to this branch directly.

I pushed a modified version of @mikeminutillo RouteAPI which allows to directly pass-through http calls to remote instances if required. Have a look. The test passed ;) Especially the part with the header modifications probably needs a thorough review

src/ServiceControl/Operations/BodyStorage/RavenAttachments/GetBodyById.cs
```diff
@@ -1,79 +1,20 @@
...
-                    currentEtag = attachment.Etag;
-                }
-
-                return new Response { Contents = contents }
-                    .WithContentType(contentType)
-                    .WithHeader("Expires", DateTime.UtcNow.AddYears(1).ToUniversalTime().ToString("R"))
-                    .WithHeader("Content-Length", bodySize.ToString())
-                    .WithHeader("ETag", currentEtag)
-                    .WithStatusCode(HttpStatusCode.OK);
+                return await GetBodyByIdApi.Execute(this, messageId);
```
danielmarbach: Pretty sure you can't just return the task here because the API requires Task<object> that's why I think the await is needed

boblangley:
Changes look great.

src/ServiceControl/CompositeViews/Messages/RoutedApi.cs
```diff
@@ -0,0 +1,109 @@
...
+                var method = new HttpMethod(currentRequest.Method);
+                var requestMessage = new HttpRequestMessage(method, instanceUri);
+                foreach (var currentRequestHeader in currentRequest.Headers)
+                {
+                    requestMessage.Headers.Add(currentRequestHeader.Key, currentRequestHeader.Value);
+                }
+
+                var rawResponse = await httpClient.SendAsync(requestMessage).ConfigureAwait(false);
+
+                // TODO: is it correct to merge headers??
```
SeanFeldman: Looks right. `response.Headers` and `response.Content.Headers` should have different headers/keys.

PR #1090, opened 2018-02-02
Skip over empty archive batches instead of terminating the archive process

WilliamBZA:
Resolves https://github.com/Particular/ServicePulse/issues/592

Adds a test to reproduce the behavior and then proposes a fix which is to simply skip over empty archive batches so that the rest of the archive operation continues as it should instead of terminating.

### DIFF
--- a/src/ServiceControl.UnitTests/Archiving/ArchiveGroupTests.cs
+++ b/src/ServiceControl.UnitTests/Archiving/ArchiveGroupTests.cs
@@ -1,0 +1,68 @@
+﻿using NUnit.Framework;
+using Raven.Client;
+using ServiceControl.Contracts.Operations;
+using ServiceControl.MessageFailures;
+using ServiceControl.Recoverability;
+using ServiceControl.UnitTests.Recoverability;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace ServiceControl.UnitTests.Archiving
+{
+    [TestFixture]
+    public class ArchiveGroupTests
+    {
+        [Test]
+        public void ArchiveGroup_skips_over_empty_batches_but_still_completes()
+        {
+            // Arrange
+            using (var documentStore = InMemoryStoreBuilder.GetInMemoryStore())
+            {
+                var groupId = "TestGroup";
+
+                using (var session = documentStore.OpenSession())
+                {
+                    var secondAchiveBatch = new ArchiveBatch { Id = ArchiveBatch.MakeId(groupId, ArchiveType.FailureGroup, 1) };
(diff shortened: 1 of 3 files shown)


src/ServiceControl.UnitTests/Archiving/ArchiveGroupTests.cs
```diff
@@ -0,0 +1,68 @@
...
+        public void ArchiveGroup_skips_over_empty_batches_but_still_completes()
+        {
+            // Arrange
+            using (var documentStore = InMemoryStoreBuilder.GetInMemoryStore())
+            {
+                var groupId = "TestGroup";
+
+                using (var session = documentStore.OpenSession())
+                {
+                    var secondAchiveBatch = new ArchiveBatch { Id = ArchiveBatch.MakeId(groupId, ArchiveType.FailureGroup, 1) };
```
mikeminutillo: Would it make sense to pull this Id out as a `emptyBatchId` or something? It is used below as well.

src/ServiceControl.UnitTests/Archiving/ArchiveGroupTests.cs
```diff
@@ -0,0 +1,68 @@
...
+                    session.Store(secondAchiveBatch);
+
+                    var archiveOperation = new ArchiveOperation
```
mikeminutillo: Why is this needed? It feels like this test knows a lot about the inner workings of the archive subsystem.

WilliamBZA: It does, it's specifically a unit test intended to test a specific scenario. In order to set that scenario up it needs to know what the data looks like.

src/ServiceControl.UnitTests/Archiving/ArchiveGroupTests.cs
WilliamBZA: 👍

WilliamBZA:
By calling the `ArchiveAllInGroup` method before setting this code up, it won't get into the state that the test is trying to validate.

The specific scenario is that an archive group operation was started and stopped but some of the data was not cleaned up which means that when a second call to archive the same group is run, it terminates early because it thinks the archive operation is completed. In order to test this, this test first introduces the data that simulates the "started and stopped but not cleaned up" and then calls the `ArchiveAllInGroup` method again, thereby testing what happens when the user invokes the archive operation on the same group a second time.

mikeminutillo:
@WilliamBZA Would it make sense to rename the variables to `oldArchiveBatch` and `oldArchiveOperation` then? Just to make it more explicit what is going on.

WilliamBZA:
@mikeminutillo I've renamed to `previous`

PR #1091, opened 2018-02-10
Skip custom checks for ATTs

mikeminutillo:


mikeminutillo:
@Particular/servicecontrol-maintainers I hypothesize that the addition of the two new custom checks (failed error/audit imports) to every test slows down everything just enough to nudge passing tests over to failing, especially for high latency transports.

This is the PR that attempts to prove that.

PR #1092, opened 2018-02-15
Disable recoverability if error queue name equals !disable

danielmarbach:
image

The disable entry in the error queue field overrides even the error log forwarding. That is not visible in the installer but SC will automatically take it into account. So in essence `!disable` disables the error queue which also disables error forwarding. We'd argue this makes sense.

image

Leads to no error queue will be created.

We were unable to create a test that checks the absence of the error queue without falling back to transport specific behavior.

### DIFF
--- a/src/ServiceControl/Infrastructure/Settings/Settings.cs
+++ b/src/ServiceControl/Infrastructure/Settings/Settings.cs
@@ -31,8 +31,13 @@
             }
 
             AuditQueue = GetAuditQueue();
-            ErrorQueue = GetErrorQueue();
-            ErrorLogQueue = GetErrorLogQueue();
+
+            {
+                // order matters
+                ErrorQueue = GetErrorQueue();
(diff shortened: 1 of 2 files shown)


WilliamBZA:
In an attempt to do my best to derail everything I touch: would it make sense to extend this to be able to disable the audit queue?

Use case: The master node ingests the error queue only, and has 2 slaves that ingest audit. By disabling audit queue generation on the master, it's a bit neater and that portion doesn't consume cycles.

WilliamBZA:
The comment ☝️ is not intended to prevent merging. Just a thought.

danielmarbach:
@WilliamBZA I think that is a good idea for the multi instance scenario. How about we raise a dedicated PR? I can give it a shot soon

danielmarbach:
@WilliamBZA to make you happy before the weekend https://github.com/Particular/ServiceControl/pull/1093

PR #1093, opened 2018-02-16
Disable auditing if audit queue name equals !disable

danielmarbach:
Based on @WilliamBZA idea mentioned in https://github.com/Particular/ServiceControl/pull/1092#issuecomment-366177427 (he is secretly known as the PR derailer ;) )

image

The disable entry in the audit queue field overrides even the audit log forwarding. That is not visible in the installer but SC will automatically take it into account. So in essence `!disable` disables the audit queue which also disables audit forwarding. We'd argue this makes sense.

image

Leads to no audit queue will be created.

We were unable to create a test that checks the absence of the audit queue without falling back to transport specific behavior.


src/ServiceControl/Infrastructure/Installers/AuditQueueInstaller.cs
```diff
@@ -1,18 +0,0 @@
-namespace ServiceBus.Management.Infrastructure.Installers
-{
-    using NServiceBus;
-    using NServiceBus.Unicast.Queuing;
-    using Settings;
-
-    public class AuditQueueInstaller : IWantQueueCreated
```
danielmarbach: I'm pretty sure this is a leftover which is no longer needed:

- Error queue doesn't have an equivalent counterpart, still, it gets created
- Even when I remove this file the audit queue gets created

src/ServiceControl/Infrastructure/Settings/Settings.cs
```diff
@@ -188,6 +199,14 @@ private Address GetAuditQueue()
...
                 return Address.Undefined;
             }
+
+            if (value.Equals("!disable", StringComparison.OrdinalIgnoreCase))
+            {
+                logger.Info("Audit ingestion disabled.");
+                return null; // needs to be null to not create the queues
+            }
+
+
```
danielmarbach: Happy to burn this spacing here 🔥

danielmarbach:
Rebased

PR #1097, opened 2018-02-22
Raven query statistics to query result

SeanFeldman:
Connects to #1096 (Mike's comments)

Addresses https://github.com/Particular/ServiceControl/pull/1069#discussion_r169548502 to ensure we do not sent empty ETag and Count headers.


SeanFeldman:
Hooray! Can be merged then.

PR #1098, opened 2018-02-22
Known endpoints relative URI

danielmarbach:
Based on the discussion on https://github.com/Particular/ServiceInsight/issues/787 @adamralph and I think the new API endpoint should be relative. SI uses the connection string provided in the API and then can just add the relative part to the URI


adamralph:
FYI I've smoke tested this against the [currently-being-patched SI](https://github.com/Particular/ServiceInsight/pull/789) and it works.

PR #1106, opened 2018-02-28
Use DomainEvents to raise events instead of Bus.Publish

SzymonPobiega:
Connects to the famous "Mike's spreadsheet of events". Tries to limit the number of messages SC sends to itself for the sake of sending them.

 * Use `IDomainEvent` as the interface of the domain events (instead of `IEvent`) to make it explicit
 * Use `IEvent` as a marker interface on selected (two) domain events so that a generic `BusPublisher` knows to re-publish them on the bus
 * Use `IUserInterfaceEvent` as a marker interface on selected domain events so that a generic SignalR gateway knows to re-publish them on the hub


SzymonPobiega:
There is still one test failing on my machine (`When_issuing_retry_on_master`) but I have no idea why :-(

mikeminutillo:
I think I know what that is but I'm not sure how to fix it easily. 

The DomainEvents class is static. When we have multi-instances, they share the same AppDomain, so they share a single instance of DomainEvents. Only one of the containers will end up in the DomainEvents class (probably the master one as it is configured last). 

To fix this, we'd need to make the DomainEvents class non-static and use IoC to make it available to each instance.

SzymonPobiega:
@danielmarbach @tmasternak @boblangley @SeanFeldman @WilliamBZA please review. I hope I didn't miss anybody.

src/ServiceControl.UnitTests/Recoverability/RetryOperationTests.cs
```diff
@@ -132,4 +132,4 @@ public void Skip_and_forward_combination_should_complete_when_done()
             Assert.AreEqual(1000, summary.NumberOfMessagesSkipped);
         }
     }
-}
\ No newline at end of file
+}
```
danielmarbach: Should we have a test that asserts that domaik events are raised?

src/ServiceControl.UnitTests/Recoverability/Retry_State_Tests.cs
```diff
@@ -288,8 +289,8 @@ public override void MoveBatchToStaging(string batchDocumentId, string[] failedM
 
     public class TestReturnToSenderDequeuer : ReturnToSenderDequeuer
     {
-        public TestReturnToSenderDequeuer(IBodyStorage bodyStorage, ISendMessages sender, IDocumentStore store, IBus bus, Configure configure)
-            : base(bodyStorage, sender, store, bus, configure)
+        public TestReturnToSenderDequeuer(IBodyStorage bodyStorage, ISendMessages sender, IDocumentStore store, IDomainEvents domainEvents, Configure configure)
```
danielmarbach: Same question as above. Would mean fake domain events would collect raised events

src/ServiceControl/Bootstrapper.cs
```diff
@@ -107,7 +110,7 @@ public IBus Start(bool isRunningAcceptanceTests = false)
         public void Stop()
         {
             notifier.Dispose();
-            bus?.Dispose();
+            bus?.Bus.Dispose();
```
danielmarbach: Bus.Bus looks like a weird trainwreck. Why not make the instance disposable?

src/ServiceControl/ExternalIntegrations/EventDispatcher.cs
```diff
@@ -14,29 +14,32 @@
...
-        private readonly IDocumentStore store;
-        private RepeatedFailuresOverTimeCircuitBreaker circuitBreaker;
-        private IDisposable subscription;
-        private Task task;
-        private CancellationTokenSource tokenSource;
-        private Etag latestEtag = Etag.Empty;
-
-        public EventDispatcher(IDocumentStore store, IBus bus, CriticalError criticalError, Settings settings, IEnumerable<IEventPublisher> eventPublishers)
+        static ILog Logger = LogManager.GetLogger(typeof(EventDispatcher));
+        IBus bus;
```
danielmarbach: Why is IBus still here?

src/ServiceControl/Infrastructure/BusInstance.cs
```diff
@@ -0,0 +1,17 @@
+namespace ServiceBus.Management.Infrastructure
+{
+    using NServiceBus;
+    using ServiceControl.Infrastructure.DomainEvents;
+
+    public class BusInstance
```
danielmarbach: I'm not sure the name reflects what this does.

src/ServiceControl/Hosting/Commands/ImportFailedAuditsCommand.cs
```diff
@@ -31,7 +31,7 @@ void RunAndWait(HostArguments args)
 
             var loggingSettings = new LoggingSettings(settings.ServiceName, LogLevel.Info, LogLevel.Info);
             var bootstrapper = new Bootstrapper(() => { tokenSource.Cancel();}, settings, busConfiguration, loggingSettings);
-            var bus = (UnicastBus)bootstrapper.Start();
+            var bus = (UnicastBus)bootstrapper.Start().Bus;
```
danielmarbach: This is only needed for resolving. Can we do this in another way without exposing the bus?

src/ServiceControl/Infrastructure/DomainEvents/DomainEventBusPublisher.cs
```diff
@@ -0,0 +1,22 @@
+namespace ServiceControl.Infrastructure.DomainEvents
+{
+    using NServiceBus;
+
+    class DomainEventBusPublisher : IDomainHandler<IDomainEvent>
```
danielmarbach: Not sure if needwd but if polymorphic dispatch would be available we could handle here IEvent

src/ServiceControl/Infrastructure/SignalR/ServicePulseNotifier.cs
```diff
@@ -0,0 +1,26 @@
+namespace ServiceControl.Infrastructure.SignalR
+{
+    using ServiceControl.Infrastructure.DomainEvents;
+
+    /// <summary>
+    /// Forwards all domain events via SignalR
+    /// </summary>
+    public class ServicePulseNotifier : IDomainHandler<IDomainEvent>
```
danielmarbach: Same comment as above

src/ServiceControl/CustomChecks/ReportCustomCheckResultHandler.cs
```diff
@@ -7,16 +7,19 @@
     using NServiceBus;
     using Plugin.CustomChecks.Messages;
     using Raven.Client;
+    using ServiceControl.Infrastructure.DomainEvents;
 
     class ReportCustomCheckResultHandler : IHandleMessages<ReportCustomCheckResult>
     {
-        private readonly IBus bus;
-        private readonly IDocumentStore store;
+        IBus bus;
```
danielmarbach: remove bus usage

src/ServiceControl/EventLog/AuditEventLogWriter.cs
```diff
@@ -3,29 +3,30 @@
...
+        public AuditEventLogWriter(IBus bus, GlobalEventHandler broadcaster, IDocumentStore store, EventLogMappings mappings)
         {
             this.bus = bus;
             this.broadcaster = broadcaster;
             this.store = store;
             this.mappings = mappings;
         }
 
-        public void Handle(IEvent message)
+        public void Handle(IDomainEvent message)
```
danielmarbach: https://github.com/Particular/ServiceControl/blob/9af43ddc2358eee3b74f528f8294c3141e886cdf/src/ServiceControl/EventLog/AuditEventLogWriter.cs#L36 feels leaky now since this is no longer a handler. I guess we need to remove the bus instance and somehow float the message id in over the domain event

src/ServiceControl/EventLog/EventLogMappings.cs
```diff
@@ -13,12 +13,12 @@ public EventLogMappings(Dictionary<Type, Type> mappings)
...
-        public bool HasMapping(IEvent message)
+        public bool HasMapping(IDomainEvent message)
         {
             return mappings.ContainsKey(message.GetType());
         }
 
-        public EventLogItem ApplyMapping(string messageId, IEvent @event)
+        public EventLogItem ApplyMapping(string messageId, IDomainEvent @event)
         {
             var mapping = (IEventLogMappingDefinition)Activator.CreateInstance(mappings[@event.GetType()]);
```
danielmarbach: Not relevant here but we should actually make sure we don't use `Activator.CreateInstance` on a hot path.

danielmarbach: raised https://github.com/Particular/ServiceControl/issues/1109

src/ServiceControl/Infrastructure/SignalR/GlobalEventHandler.cs
```diff
@@ -1,27 +1,22 @@
...
         {
-            var metadata = registry.GetMessageMetadata(@event.GetType());
+            var typeName = @event.GetType().Name;
+            var types = new List<string>
+            {
+                typeName
+            };
             var context = GlobalHost.ConnectionManager.GetConnectionContext<MessageStreamerConnection>();
 
-            context.Connection.Broadcast(new Envelope { Types = metadata.MessageHierarchy.Select(t=>t.Name).ToList(), Message = @event }, emptyArray)
```
danielmarbach: Isn't this a potential breaking change? The message metadata registry might contain multiple types but now we are only ever sending one type. Did you verify that we never ran into the multiple types scenario?

src/ServiceControl/MessageFailures/FailedMessagesFeature.cs
```diff
@@ -27,15 +28,17 @@ protected override void Setup(FeatureConfigurationContext context)
 
         class DetectSuccessfullRetriesEnricher : ImportEnricher
         {
-            public override bool EnrichErrors => false;
-
             IBus bus;
```
danielmarbach: No longer needed

src/ServiceControl/MessageRedirects/Api/MessageRedirectsModule.cs
```diff
@@ -65,7 +67,7 @@ public MessageRedirectsModule()
                     await collection.Save(session).ConfigureAwait(false);
                 }
 
-                Bus.Publish(new MessageRedirectCreated
+                DomainEvents.Raise(new MessageRedirectCreated
```
danielmarbach: What are the potential consequences of this for the callers of the API? Previously this was a publish operation which means the actual handling of the message is decoupled from API call. Now, these two things are coupled together which means a potential failure of handling that message could bubble up to the caller of the API.

src/ServiceControl/EventLog/AuditEventLogWriter.cs
SzymonPobiega: I am going to switch to using guids for ID of event log events

src/ServiceControl/ExternalIntegrations/EventDispatcher.cs
SzymonPobiega: We use it to push the integration events out to the consumers.

src/ServiceControl/Hosting/Commands/ImportFailedAuditsCommand.cs
SzymonPobiega: We can't expose the `IBuilder` but we can expose the underlying Autofac container instead of `IBus`. What do you think?

src/ServiceControl/Infrastructure/DomainEvents/DomainEventBusPublisher.cs
SzymonPobiega: Currently the `DomainEvents` class does not handle polymorphic dispatch. It only matches the exact type and the root type (`IDomainEvent`). I guess we could make it more smart. Want to try? ;-) We can PP

src/ServiceControl/Infrastructure/SignalR/GlobalEventHandler.cs
SzymonPobiega: Yeah. I've looked at the SP codebase and only found usages with concrete leaf type. I also asked the @Particular/servicepulse-maintainers but nobody recalls any usage of base types.

src/ServiceControl/MessageRedirects/Api/MessageRedirectsModule.cs
SzymonPobiega: True. Do you think we should do something about this on SP level? i.e. retry if API call failed or something?

src/ServiceControl/Infrastructure/SignalR/GlobalEventHandler.cs
danielmarbach: Ok then I have no objections

src/ServiceControl/Infrastructure/DomainEvents/DomainEventBusPublisher.cs
danielmarbach: I'm ok with that

src/ServiceControl/Hosting/Commands/ImportFailedAuditsCommand.cs
danielmarbach: I would prefer to expose the container but I still think it is leaky. Normally the bootstrapper should know what to bootstrap. I get that this is basically a non-normal bootstrapping when we want to import the failed audits. That's why we are awaiting here the import audits. Still, it seems weird that we start the bootstrapper and then need to resolve the audit import stuff.

src/ServiceControl/MessageRedirects/Api/MessageRedirectsModule.cs
danielmarbach: Good question. I don't know SC enough to make a judgment call on this one. Maybe @WilliamBZA or @boblangley can join into the discussion

src/ServiceControl/Hosting/Commands/ImportFailedAuditsCommand.cs
SzymonPobiega: Yeah. This is an edge case but I agree it is ugly. The proper way of building this feature would have been an admin UI that allows to start/stop the importing of failed audits. We wanted to cut corners here :-(

src/ServiceControl/MessageRedirects/Api/MessageRedirectsModule.cs
SzymonPobiega: @danielmarbach is this the last thing we need to decide on?

danielmarbach: I think we do

SzymonPobiega: @WilliamBZA @boblangley @tmasternak what do you think?

mikeminutillo: Hmmm. This always had the potential for transactional inconsistency. In other places in the code we accept the user request and turn it into a message so that the handling of the message will go through FLR. Does Pulse use the reason phrases and distinguish all of the error conditions that we detect here?

mikeminutillo: Having said that, the document session is not transactionally consistent with the transport anyway. 

In this case, if there's an error in the Domain Event handlers, then that error will get bubbled all the way to Nancy where it will get turned into a 404. If we successfully created the redirect then a retry will just return a "Duplicate". The rest of the Domain Event handlers and the SendLocal will not get sent. 

How bad is that (in this situation)?

WilliamBZA: Pulse does not. In fact there is almost zero error handling in pulse besides in the monitoring code.

The transport that the publish uses to send messages to pulse (SignalR) is also not safe, and these publish events are at most-once delivery anyway.

WilliamBZA: TL;DR: from the pulse side there is no transactional guarantee around the publish portion anyway. Even if we had full messaging in place at the server side with FLR+SLR, we'd still have to build outbox into the signlar hubs to ensure we don't miss messages during reconnect/negotiation/connection upgrade/etc.

SzymonPobiega: I think that Daniel's original concern was more around the fact that there is more stuff happening now as part of request handling and the likelihood of Pulse showing an error page is higher.

danielmarbach: I'm good with that. It was reasonably discussed. Let's move on

SzymonPobiega:
:heart:

src/ServiceControl/MessageRedirects/Api/MessageRedirectsModule.cs
WilliamBZA: It won't be any more buggy than it currently is, but the value is higher. So 👍 merge it

WilliamBZA: Oops, it already is...

PR #1110, opened 2018-03-14
Storing FailedMessageRetryIds in RetryBatch on creation

tmasternak:
## Overview
This PR changes the way we handle `RetryBatch`es in prepare state of recoverability. Previously preparation would be done in 3 steps:
  1. Create `RetryBatch` document,
  2. Create number of `FailedMessageRetry` documents that link the batch with failed messages,
  3. Update `RetryBatch` with `FailedMessageRetry` ids from step 2 and set status to `Stagging`.

This is invalid in the context of batch adoption process which is performed on startup and is meant to finish preparation step for any batches that when past step 1. When batch is adopted we query for `FailedMessagesRetry` documents by `batchId` using index and continue with step 3. What it means is that we can get stale/incomplete set of ids so some ids might not be assigned to original batch and cause related failed messages to be filterd out if users issues any retry operation for those messages.

## Fix
This fix creates all `FailedMessageRetry` ids and uses them when creating `RetryBatch` in step 1. Step 3 now is only about setting `Stagging` status for given batch. The same goes with adoption operation which now boils down to setting the status as well.

The fact that we could have more ids in batch document than has been acutally created (we restart SC in step 2 e.g.) does not make this approach invalid as we validate the ids one more time during [stage phase](https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/Recoverability/Retrying/RetryProcessor.cs#L152).

/cc: @WilliamBZA @SzymonPobiega

### DIFF
--- a/src/ServiceControl/Recoverability/Retrying/RetriesGateway.cs
+++ b/src/ServiceControl/Recoverability/Retrying/RetriesGateway.cs
@@ -157,21 +157,22 @@ namespace ServiceControl.Recoverability
                 return;
             }
 
-            var batchDocumentId = retryDocumentManager.CreateBatchDocument(requestId, retryType, messageIds.Length, originator, startTime, last, batchName, classifier);
+            var failedMessageRetryIds = messageIds.Select(FailedMessageRetry.MakeDocumentId).ToArray();
+
+            var batchDocumentId = retryDocumentManager.CreateBatchDocument(requestId, retryType, failedMessageRetryIds, originator, startTime, last, batchName, classifier);
 
             log.InfoFormat("Created Batch '{0}' with {1} messages for '{2}'", batchDocumentId, messageIds.Length, batchName);
 
-            var retryIds = new string[messageIds.Length];
             var commands = new ICommandData[messageIds.Length];
             for (var i = 0; i < messageIds.Length; i++)
             {
                 commands[i] = retryDocumentManager.CreateFailedMessageRetryDocument(batchDocumentId, messageIds[i]);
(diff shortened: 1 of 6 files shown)


src/ServiceControl/Recoverability/Retrying/RetriesGateway.cs
```diff
@@ -157,21 +157,22 @@ private void StageRetryByUniqueMessageIds(string requestId, RetryType retryType,
                 return;
             }
 
-            var batchDocumentId = retryDocumentManager.CreateBatchDocument(requestId, retryType, messageIds.Length, originator, startTime, last, batchName, classifier);
+            var failedMessageRetryIds = messageIds.Select(FailedMessageRetry.MakeDocumentId).ToArray();
+
+            var batchDocumentId = retryDocumentManager.CreateBatchDocument(requestId, retryType, failedMessageRetryIds, originator, startTime, last, batchName, classifier);
```
WilliamBZA: If SC goes 💥 after this line, the `RetryBatch` will be persisted but not the `FailedMessageRetry` documents.

Then when SC starts up again and adopts this batch, it will move the batch to staging but the `FailedMessageRetry` will still not be created anywhere. Given that I don't see value to these documents, I don't think this is a problem, but perhaps...?

WilliamBZA: :+1:

SzymonPobiega:
It seems like this one is ready to be released @tmasternak . What do you think?

PR #1112, opened 2018-03-20
Event sourced heartbeat persistence

SzymonPobiega:
* Refactor `EndpointMonitoring` & `EndpointMonitor` to use event-sourced approach where published domain events are also used to restore the state
 * Use simple heartbeat persister which persists last event related to each endpoint instance
 * Use simple component model that does not depend on DI and features:
   * Components consists of parts
   * Components create their parts
   * Components can have dependencies injected when initializing, parts cannot
   * Components are initialized before parts are used
   * Components are torn down after parts are stopped being used
 * Refactor `RavenBootstrapper` to initialize the document store before starting NSB
 * Introduce APIs for processing audits and heartbeats


src/Particular.HealthMonitoring.Uptime/EndpointsView.cs
```diff
@@ -1,4 +1,4 @@
-namespace ServiceControl.CompositeViews.Endpoints
+namespace Particular.HealthMonitoring.Uptime
```
WilliamBZA: Much better name!

src/Particular.HealthMonitoring.Uptime/EventUnitOfWork.cs
```diff
@@ -0,0 +1,38 @@
...
+            {
+                recordedEvents.Add(heartbeatEvent);
+            }
+        }
+
+        public async Task Persist()
+        {
+            foreach (var @event in recordedEvents)
+            {
+                await persister.Store(@event).ConfigureAwait(false);
```
WilliamBZA: I'm struggling to follow this through, but how many of these would be expected at a time? For the most part I see these as single operations except for the timer that would invoke this every 5 seconds. Given that this could be called every 5 seconds for every endpoint, is it worth creating a persistence Unit of Work so that the `SaveChangesAsync` is only called once and the operation is batched?

src/Particular.HealthMonitoring.Uptime/HeartbeatFailureDetector.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        static TimeSpan GetHeartbeatGracePeriod()
+        {
+            try
+            {
+                return TimeSpan.Parse(SettingsReader<string>.Read("HeartbeatGracePeriod", "00:00:40"));
+            }
+            catch (Exception ex)
+            {
+                //logger.Error($"HeartbeatGracePeriod settings invalid - {ex}. Defaulting HeartbeatGracePeriod to '00:00:40'");
```
WilliamBZA: Why is the log statement commented out? Still seems to be valid.

src/Particular.HealthMonitoring.Uptime/HeartbeatFailureDetector.cs
```diff
@@ -0,0 +1,62 @@
...
+                return TimeSpan.FromSeconds(40);
+            }
+        }
+
+        public Task Start(ITimeKeeper timeKeeper)
+        {
+            timer = timer = timeKeeper.New(CheckEndpoints, TimeSpan.Zero, TimeSpan.FromSeconds(5));
```
WilliamBZA: double `timer =`

src/Particular.HealthMonitoring.Uptime/HeartbeatFailureDetector.cs
```diff
@@ -0,0 +1,62 @@
...
+            return Task.FromResult(0);
+        }
+
+        void CheckEndpoints()
+        {
+            try
+            {
+                var now = DateTime.UtcNow;
+                var inactivityThreshold = now - gracePeriod;
+                //log.Debug($"monitoring Endpoint Instances. Inactivity Threshold = {inactivityThreshold}");
```
WilliamBZA: Log statement commented out?

src/Particular.HealthMonitoring.Uptime/HeartbeatFailureDetector.cs
```diff
@@ -0,0 +1,62 @@
...
+                monitoring.CheckEndpoints(inactivityThreshold, now).GetAwaiter().GetResult();
+            }
+            catch (Exception exception)
+            {
+                //log.Error("Exception occurred when monitoring endpoint instances", exception);
```
WilliamBZA: Log statement commented out?

src/Particular.HealthMonitoring.Uptime/UptimeMonitoring.cs
```diff
@@ -1,51 +1,41 @@
...
-            this.timeKeeper = timeKeeper;
-            this.gracePeriod = gracePeriod;
-
-            Monitoring = new EndpointInstanceMonitoring(domainEvents);
+            monitoring = new EndpointInstanceMonitoring();
         }
-
-        public void Start()
+        
+        public IEnumerable<object> CreateParts()
```
WilliamBZA: What is a `Part`? These seem a little like a piece of a `Component`. Is there a better word somewhere between "thing" and "component"?

src/Particular.HealthMonitoring.Uptime/UptimeMonitoring.cs
```diff
@@ -1,51 +1,41 @@
...
-                //log.Error("Exception occurred when monitoring endpoint instances", exception);
-            }
+            return monitoring.Initialize(dependencies.Persister, dependencies.DomainEvents);
         }
 
-        public void Stop()
+        public Task TearDown()
         {
-            timeKeeper.Release(timer);
+            return Task.FromResult(0);
```
WilliamBZA: Shouldn't this chain down to the original timekeeper in `HeartbeatFailureDetector` somehow?

src/ServiceControl/Bootstrapper.cs
```diff
@@ -77,12 +81,25 @@ private void Initialize()
...
 
+            foreach (var component in components)
+            {
+                component.RegisterDependency(containerBuilder);
+
+                var parts = component.CreateParts();
+
+                foreach (var part in parts)
+                {
+                    containerBuilder.RegisterInstance(part).ExternallyOwned().AsImplementedInterfaces().AsSelf();
```
WilliamBZA: Maybe we should invert this so that the creation of the parts has access to the `containerBuilder` in the event that they want to register their own dependencies.

I'm thinking in the case where we want to persist using Sql. Raven will be provided by default, but the component (or piece or part or whatever) could register a SqlConnection factory in the containerBuilder so that it's internal classes will be provided with an instance when they ask for one.


E.g.:


```CSharp
var parts = component.CreateParts(containerBuilder);

...

IEnumerable<objecT>CreateParts(ContainerBuilder builder)
{
    builder.RegisterInstance(sqlPersistenceThing).ExternallyOwned();

    var uptimeModule = new UptimeApiModule(monitoring, builder); // here they can register too
    builder.RegisterInstance(uptimeModule).ExternallyOwned().AsImplementedInterfaces().AsSelf();
    yield return uptimeModule;

    var heartbeatFailureDetected = new HeartbeatFailureDetector(monitoring);
    builder.RegisterInstance(heartbeatFailureDetected).ExternallyOwned().AsImplementedInterfaces().AsSelf();
    yield return heartbeatFailureDetected;
}
```

src/ServiceControl/HeartbeatMonitoring/RegisterEndpointStartupHandler.cs
```diff
@@ -1,26 +1,38 @@
...
             this.domainEvents = domainEvents;
+            this.heartbeatProcessors = heartbeatProcessors;
         }
 
         public void Handle(RegisterEndpointStartup message)
         {
-            monitoring.GetOrCreateMonitor(message.Endpoint, message.Host, message.HostId, true);
+            foreach (var processor in heartbeatProcessors)
+            {
+                processor.Handle(new Particular.Operations.Heartbeats.Api.RegisterEndpointStartup
```
WilliamBZA: Why process them synchronously?

```CSharp

Task.WhenAll(heartbeatProcessers.Select(p => p.Handle(new Particular.Operations.Heartbeats.Api.RegisterEndpointStartup {
    Endpoint = message.Endpoint,
    ...
    StartedAt = message.StartedAt
})).GetAwaiter().GetResult();
```

src/ServiceControl/Operations/ProcessorExecutorFeature.cs
```diff
@@ -0,0 +1,76 @@
...
+            public void Enrich(IReadOnlyDictionary<string, string> headers, IDictionary<string, object> metadata)
+            {
+                var sendingEndpoint = EndpointDetailsParser.SendingEndpoint(headers);
+                var receivingEndpoint = EndpointDetailsParser.ReceivingEndpoint(headers);
+
+                var errorMessage = new ErrorMessage(headers, sendingEndpoint, receivingEndpoint);
+
+                foreach (var processor in errorProcessors)
+                {
+                    processor.Handle(errorMessage).GetAwaiter().GetResult();
```
WilliamBZA: Why do these synchronously?

src/ServiceControl/Operations/ProcessorExecutorFeature.cs
```diff
@@ -0,0 +1,76 @@
...
+            public void Enrich(IReadOnlyDictionary<string, string> headers, IDictionary<string, object> metadata)
+            {
+                var sendingEndpoint = EndpointDetailsParser.SendingEndpoint(headers);
+                var receivingEndpoint = EndpointDetailsParser.ReceivingEndpoint(headers);
+
+                var auditMessage = new AuditMessage(headers, sendingEndpoint, receivingEndpoint);
+
+                foreach (var processor in auditProcessors)
+                {
+                    processor.Handle(auditMessage).GetAwaiter().GetResult();
```
WilliamBZA: Why do these synchronously?

src/ServiceControl/Operations/StartableExecutorFeature.cs
```diff
@@ -0,0 +1,48 @@
...
+            {
+                this.startables = startables;
+                this.timeKeeper = timeKeeper;
+            }
+
+            protected override void OnStart()
+            {
+                foreach (var startable in startables)
+                {
+                    startable.Start(timeKeeper).GetAwaiter().GetResult();
```
WilliamBZA: Why do these synchronously?

src/ServiceControl/Operations/StartableExecutorFeature.cs
```diff
@@ -0,0 +1,48 @@
...
+                }
+            }
+
+            protected override void OnStop()
+            {
+                foreach (var startable in startables)
+                {
+                    startable.Stop(timeKeeper).GetAwaiter().GetResult();
```
WilliamBZA: Why do these synchronously?

src/Particular.HealthMonitoring.Uptime/HeartbeatFailureDetector.cs
SzymonPobiega: I haven't yet figured out what to do with the logging abstraction. I would like to avoid adding a reference to `NServiceBus` in this package. On the other hand I don't like the idea of having an abstraction on top of an abstraction either.

src/Particular.HealthMonitoring.Uptime/HeartbeatFailureDetector.cs
SzymonPobiega: Same reason as above

src/Particular.HealthMonitoring.Uptime/HeartbeatFailureDetector.cs
SzymonPobiega: Same reason as above

src/Particular.HealthMonitoring.Uptime/UptimeMonitoring.cs
SzymonPobiega: Yeah. Part is not the best one. Maybe `component` -> `composite` & `part` -> `component`?

src/ServiceControl/Operations/ProcessorExecutorFeature.cs
SzymonPobiega: Same question as above... Not sure

src/ServiceControl/Operations/StartableExecutorFeature.cs
SzymonPobiega: Yeah, this could be handled async

src/ServiceControl/Operations/StartableExecutorFeature.cs
SzymonPobiega: Yeah, this could be handled async

src/Particular.Operations.Heartbeats.Api/IProcessHeartbeats.cs
```diff
@@ -1,8 +1,10 @@
 ﻿namespace Particular.Operations.Heartbeats.Api
 {
+    using System.Threading.Tasks;
+
     public interface IProcessHeartbeats
     {
-        void Handle(RegisterEndpointStartup endpointStartup);
-        void Handle(EndpointHeartbeat heartbeat);
+        Task Handle(RegisterEndpointStartup endpointStartup);
```
mikeminutillo: Doesn't this mean that every host will have to know how to listen for `EndpointHeartbeat` and `RegisterEndpointStartup` commands? Should these handlers live within this component?

src/Particular.HealthMonitoring.Uptime/UptimeMonitoring.cs
mikeminutillo: The `HeartbeatFailureDetector` needs to be stopped but that's being done by the Startable feature.

src/Particular.HealthMonitoring.Uptime/UptimeMonitoring.cs
mikeminutillo: I would propose that we create a ComponentInitializationContext or something that knows how to handle these things so that we can handle them differently depending on what they are. Envisaging something like this:

```csharp
    async Task Initialize(ComponentInitContext context)
    {
        context.Services.Register<EndpointInstanceMonitoring>();
        context.Services.RegisterDefault<IHeartbeatPersistence, NullHeartbeatPersistence>();

        context.Jobs.Register<MonitorEndpointInstances>();

        context.Handlers.Register<EndpointHeartbeatHandler>();
        context.Handlers.Register<EndpointStartedHandler>();

        context.Api.Get<HeartbeatStatsApi>("/heartbeats/stats");
        context.Api.Get<MonitoringEndpointsApi>("/endpoints");
        context.Api.Patch<ToggleMonitoredEndpointsApi>("/endpoints/{id}");
    }
```

src/Particular.HealthMonitoring.Uptime/HeartbeatFailureDetector.cs
mikeminutillo: What happens if we just rely on native NLog? Or switch to [Microsoft.Extensions.Logging](https://www.nuget.org/packages/Microsoft.Extensions.Logging)?

What is the danger of relying on NSB in this package? Portability between versions?

src/Particular.Operations.Heartbeats.Api/IProcessHeartbeats.cs
SzymonPobiega: @mikeminutillo these would live in the `Heartbeat engine` so neither host nor component. In this case the `Uptime` component requires this API. If a host/particle has this engine enabled it exposes the API and the component can be initialized.

In the current phase SC contains both the host code and the engines :-(

src/Particular.HealthMonitoring.Uptime/HeartbeatFailureDetector.cs
SzymonPobiega: @mikeminutillo yeah, I am worried about portability and API surface. In my mind the components should only depend on the host/engine APIs. If they need messaging (like a backplane) it needs to be an explicit API in the host.

Since in SC we use NLog I am fine with using NLog directly in components. On the other hand I guess using the Microsoft abstraction would be more future-proof...

src/Particular.HealthMonitoring.Uptime/UptimeMonitoring.cs
SzymonPobiega: The `ComponentInitContext` approach would require that all possible APIs live in the same package and all components depend on all APIs. Since these are only APIs I guess the danger of versioning issues is not that big.

On the other hand I don't see added value. Comparing:

```
context.RegisterPart<EndpointStartedHandler>();
```

versus

```
context.Handlers.Register<EndpointStartedHandler>();
```

Another issue is lifecycle management. I don't think we need a container. We're moving away from container usage in `Features` of the core because of complexity. The model in the core is following:
 * Feature (root, equivalent of `IComponent` here) creates shared objects e.g. `EndpointInstanceMonitoring`)
 * Feature creates various things (parts here) e.g. behaviors as instances and registers them

I think this simple model should be enough for our purposes here as well.

src/ServiceControl/Bootstrapper.cs
SzymonPobiega: I've replied to @mikeminutillo 's similar suggestion here https://github.com/Particular/ServiceControl/pull/1112/files#r175994712. TL;DR I believe we should not start with a container. The new stuff should be based on a simple model we use in the core where a root (feature/component) creates instances of shared components (e.g. `EndpointInstanceMonitoring`) and parts (that do job e.g. behaviors)

src/ServiceControl/HeartbeatMonitoring/RegisterEndpointStartupHandler.cs
WilliamBZA: We get asynchronousity on the incoming message level, but not at the heartbeatProcessors level. If each heartbeat processor handles the messages in a way that it benefits from async processing, then yes.

 Future heartbeat processors might rely heavily on cloud-based operations, where doing them in parallel would be beneficial.

PR #1115, opened 2018-03-27
Use an explicit dedicated thread to stop ReturnToSenderDequeuer

SzymonPobiega:
Based on @mikeminutillo 's idea that using `Task.Run` to stop the `ReturnToSenderDequeuer` might be vulnerable to thread pool starvation problem we've seen in the acceptance tests.


SzymonPobiega:
@Particular/servicecontrol-maintainers do you have any ideas how we could verify this hypothesis *in vitro*? Otherwise the only option is to convince one of the customers to try a private build of SC.

SzymonPobiega:
This theory happened to be incorrect. The real problem is described here https://github.com/Particular/PlatformDevelopment/issues/1876

PR #1121, opened 2018-04-05
Halting Test

mikeminutillo:
This test is halting on the build server.

After the Retry POST was being issued the GET was being called in a tight loop. This caused a hang (still unsure why that is).


mikeminutillo:
Passed once. Running again.

mikeminutillo:
image

src/ServiceControl.AcceptanceTesting/Support/EndpointRunner.cs
```diff
@@ -105,6 +105,10 @@ public class EndpointRunner
                                 {
                                     executedWhens.Add(when.Id);
                                 }
+                                else
```
danielmarbach: Why are we mixing async code path with Thread.Sleep? Would a Task.Delay also work? If yes why would we then need the explicit yield at the end?

danielmarbach: Then how about only doing a Task.Delay ?

danielmarbach:
@mikeminutillo @SzymonPobiega I did a rebase on master and switch to delay for now to verify

mikeminutillo:
Ideally the whole of `Done()` switches to a proper async call. Then all the web stuff can switch to async too. Interestingly the web called are all connected directly to the backend with no actual IO so it'll be async all the way down to Raven.

danielmarbach:
First run with the changes successful. Trigger it a second time

src/ServiceControl/Infrastructure/TimeKeeper.cs
```diff
@@ -33,11 +36,13 @@ public Timer NewTimer(Func<Task<bool>> callback, TimeSpan dueTime, TimeSpan peri
                 {
                     log.Error("Reoccurring timer task failed.", ex);
                 }
-                if (reschedule && timers.ContainsKey(timer))
+
+                var timerSelf = timer[0];
+                if (reschedule && timers.ContainsKey(timerSelf))
```
danielmarbach: Found on TC 

```
[12:46:25][Step 2/2] 2018-04-12 12:46:03,430 [11120] WARN  JetBrains.Container.Log4netDefaultLoggerInitializer - Unhandled exception in current domain: System.ArgumentNullException: Value cannot be null.
[12:46:25][Step 2/2] Parameter name: key
[12:46:25][Step 2/2]    at System.Collections.Concurrent.ConcurrentDictionary`2.ContainsKey(TKey key)
[12:46:25][Step 2/2]    at ServiceControl.Infrastructure.TimeKeeper.<>c__DisplayClass6_0.<NewTimer>b__0(Object _) in C:\BuildAgent\work\7189a56f9f44affc\src\ServiceControl\Infrastructure\TimeKeeper.cs:line 36
[12:46:25][Step 2/2]    at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
[12:46:25][Step 2/2]    at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
[12:46:25][Step 2/2]    at System.Threading.TimerQueueTimer.CallCallback()
[12:46:25][Step 2/2]    at System.Threading.TimerQueueTimer.Fire()
[12:46:25][Step 2/2]    at System.Threading.TimerQueue.FireNextTimers(),    at System.Collections.Concurrent.ConcurrentDictionary`2.ContainsKey(TKey key)
[12:46:25][Step 2/2]    at ServiceControl.Infrastructure.TimeKeeper.<>c__DisplayClass6_0.<NewTimer>b__0(Object _) in C:\BuildAgent\work\7189a56f9f44affc\src\ServiceControl\Infrastructure\TimeKeeper.cs:line 36
[12:46:25][Step 2/2]    at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
[12:46:25][Step 2/2]    at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
[12:46:25][Step 2/2]    at System.Threading.TimerQueueTimer.CallCallback()
[12:46:25][Step 2/2]    at System.Threading.TimerQueueTimer.Fire()
[12:46:25][Step 2/2]    at System.Threading.TimerQueue.FireNextTimers(), (null)
```

due to closure capturing it is possible that the timer callback fires and the closure sees the local as being null and then `ContainsKey` throws

src/ServiceControl.AcceptanceTests/MessageFailures/When_a_pending_retry_is_retried_by_queue_and_timeframe.cs
```diff
@@ -43,7 +44,11 @@ public void Should_succeed()
 
                     if (failedMessage.Status == FailedMessageStatus.RetryIssued)
                     {
-                        return true;
+                        FailedMessage[] failedMessages;
+                        if (TryGet("/api/errors", out failedMessages))
```
tmasternak: We need this to make sure that `FailedMessageIndex` is up to date. Otherwise it's possible that the `FailedMessage` document is returned via `/api/errors/{msgid}` and retried via. `/api/errors/{msgid}/retry`  but when we call retry from `/api/pendingretries/queues/retry` the `FailedMessageIndex` did not catch-up. No `FailedMessageRetry` is deleted and who retry becomes no-op.

PR #1122, opened 2018-04-05
Synchronize Start and Stop of ReturnToSenderDequeuer

SzymonPobiega:
Relates to https://github.com/Particular/PlatformDevelopment/issues/1876

### DIFF
--- a/src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSenderDequeuer.cs
+++ b/src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSenderDequeuer.cs
@@ -318,24 +318,26 @@ namespace ServiceControl.Recoverability
 
         internal class DequeueMessagesWrapper : IDequeueMessages
         {
-            private readonly IDequeueMessages _realDequeuer;
-            private int maximumConcurrencyLevel;
-            private int disposeSignaled;
+            IDequeueMessages realDequeuer;
+            int maximumConcurrencyLevel;
+            object startStopLock = new object();
 
             public DequeueMessagesWrapper(IDequeueMessages realDequeuer)
             {
-                _realDequeuer = realDequeuer;
+                this.realDequeuer = realDequeuer;
             }
 
             public void Init(Address address, TransactionSettings transactionSettings, Func<TransportMessage, bool> tryProcessMessage, Action<TransportMessage, Exception> endProcessMessage)
             {
-                _realDequeuer.Init(address, transactionSettings, tryProcessMessage, endProcessMessage);
+                realDequeuer.Init(address, transactionSettings, tryProcessMessage, endProcessMessage);
             }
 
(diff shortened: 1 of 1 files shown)


src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSenderDequeuer.cs
```diff
@@ -318,24 +318,26 @@ public void Init(Address address)
...
 
             public void Init(Address address, TransactionSettings transactionSettings, Func<TransportMessage, bool> tryProcessMessage, Action<TransportMessage, Exception> endProcessMessage)
             {
-                _realDequeuer.Init(address, transactionSettings, tryProcessMessage, endProcessMessage);
+                realDequeuer.Init(address, transactionSettings, tryProcessMessage, endProcessMessage);
             }
 
             public void StartInternal()
             {
-                Interlocked.Exchange(ref disposeSignaled, 0);
```
ramonsmits: @SzymonPobiega Do you know what this was even used for?

src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSenderDequeuer.cs
```diff
@@ -349,21 +351,9 @@ public void Stop()
...
-                    _realDequeuer.Stop();
-                }
-                catch (Exception)
+                lock (startStopLock)
                 {
-                    // Making build go green.
-                    var r = 1 + 1;
-                    Interlocked.Increment(ref r);
-                    // We are shutting down, race condition can result in an exception in the real dequeuer.
+                    realDequeuer.Stop();
```
ramonsmits: Previous we always did a catch, can the Stop() still throw something? If so, should this be gracefully handled?

SzymonPobiega: Well, if `Stop` throws then what is the condition of the underlying transport? Would we be able to `Start` it again? I guess that if `Stop` throws then we can't make any assumptions on the state of the receiver. It may as well be in an invalid state.

Based on that I'd rather kill SC if `Stop` throws by triggering a critical action.

src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSenderDequeuer.cs
SzymonPobiega: It was used to make sure that `Stop` is not called concurrently. So in the old code `Start` and `Stop` could overlap but two `Stops` could not as one of the `Stops` would detect the flag signalled.

SzymonPobiega:
@WilliamBZA correct. The code that calls stop is invoked in the handler. This ensures that `Start` is called before (because if `Start` was not called then the handler wound not be invoked).

PR #1124, opened 2018-04-05
Enable endpoint detection from audits in multi-instance deployments

danielmarbach:
## Overview

In the multi-instance deployments when an endpoint is detected by a slave node from the audit messages that information is not propagated to the master.

As a results users of `ServicePulse` do not see all endpoints detected by `ServiceControl` instances.

## Fix
`NewEndpointDetected` event is raised on the slaves when new endpoint is found in the audits. Those are handled by the master.


mikeminutillo:
This currently causes some unnecessary events. Here it is on my machine:

```
-> Send from MultiInstance.Sender@MIKE_LAPTOP [0be58ccc-b644-4d79-8a9a-a8ba00d395c1] MyMessage
<- MultiInstance.Sender@MIKE_LAPTOP got [0be58ccc-b644-4d79-8a9a-a8ba00d395c1] MyMessage
-> Publish from slave@MIKE_LAPTOP [e03dfe99-9431-432b-8d81-a8ba00d3973e] NewEndpointDetected
<- slave@MIKE_LAPTOP got [e03dfe99-9431-432b-8d81-a8ba00d3973e] NewEndpointDetected
<- master@MIKE_LAPTOP got [e03dfe99-9431-432b-8d81-a8ba00d3973e] NewEndpointDetected
-> Publish from master@MIKE_LAPTOP [5c512de2-c800-4fc7-b5d3-a8ba00d397c0] NewEndpointDetected
<- master@MIKE_LAPTOP got [5c512de2-c800-4fc7-b5d3-a8ba00d397c0] NewEndpointDetected
```

The slave publishes the initial event but SC subscribes to it's own events so it's also a subscriber. It updates it's internal state but as it's already in place that's OK, it won't actually update anything and it won't publish the event again. 

The master gets the event, updates it's state and that triggers it's own event to get published again. SC subscribes to it's own events so the master gets it again. On a broker, the slave will get the event though. And if those slaves hadn't seen the message, they will trigger the event to get published again. 

I think to fix this we need to:
- Exclude this message type from the SC self-subscribe feature. This prevents slave instances from trying to update state that we know they already have. It will also prevent the slaves from getting events from other slaves in a broker.
- Prevent the master from publishing the message. The master knows it is special but the slaves don't. This prevents the double-processing on the master.

Sound reasonable?

danielmarbach:
We decided to manually test this. The following scenario was tested with RabbitMQ and MSMQ

overview

we also reattached previously seen instances on audit2 to audit1 to see if it properly handles duplicates and it works fine. All the thirty endpoints show up in ServicePulse

image

Issue #1120, opened 2018-04-04
Conflicting logic of the existence of SendingEndpoint information in the message headers


SimonCropp:
the assumption in the logic of `EndpointDetailsParser.SendingEndpoint` implies that `SendingEndpoint` can be null.

https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/Operations/EndpointDetailsParser.cs#L33

This then propogates into the message metadata. https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/EndpointControl/DetectNewEndpointsFromImportsEnricher.cs#L33

This "implied can be null" conlicts with the "implied must exist logic" of `GetErrorById.Map`
https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/MessageFailures/Api/GetErrorById.cs#L60

note the same bug exists for ReceivingEndpoint

tmasternak:
Solved by #1125

PR #1125, opened 2018-04-06
Missing metadata

### LINKED ISSUE #1119: Sending a raw message, that fails, results in SC not being able to return the failed message to SP
So here https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/MessageFailures/Api/GetErrorById.cs#L55

It is assumed that many headers exist and no defensive code is done

When sending a raw message using this approach https://docs.particular.net/transports/sql/operations-scripting#native-send-the-native-send-helper-methods-in-c

The only required header for NSB endpoints to successfully process a message is `NServiceBus.EnclosedMessageTypes`. But, if the message processing fails, this results in `processingAttempt.MessageMetadata` in SC missing the `SendingEndpoint` and `TimeSent` headers.

When SP queries for this failed message SC throws a `KeyNotFoundException`

```
System.Collections.Generic.KeyNotFoundException: The given key was not present in the dictionary.
   at System.ThrowHelper.ThrowKeyNotFoundException()
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   at ServiceControl.MessageFailures.Api.GetErrorById.Map(FailedMessage message, IAsyncDocumentSession session) in C:\BuildAgent\work\7189a56f9f44affc\src\ServiceControl\MessageFailures\Api\GetErrorById.cs:line 55
   at ServiceControl.MessageFailures.Api.GetErrorById.<<-ctor>b__0_1>d.MoveNext() in C:\BuildAgent\work\7189a56f9f44affc\src\ServiceControl\MessageFailures\Api\GetErrorById.cs:line 44
```

mikeminutillo:
Fixes #1119 and #1120

### DIFF
--- a/src/ServiceControl/Infrastructure/MetadataExtensions.cs
+++ b/src/ServiceControl/Infrastructure/MetadataExtensions.cs
@@ -1,11 +1,10 @@
 ﻿namespace ServiceControl
 {
-    using System;
     using System.Collections.Generic;
 
     static class MetadataExtensions
     {
-        public static T MaybeGet<T>(this IDictionary<string, object> metadata, string key)
+        public static T GetOrDefault<T>(this IDictionary<string, object> metadata, string key)
         {
             object foundValue;
             if (metadata.TryGetValue(key, out foundValue))
@@ -15,24 +14,12 @@
             return default(T);
         }
 
-        public static string MaybeGetAsString(this IDictionary<string, object> metadata, string key)
+        public static string GetAsStringOrNull(this IDictionary<string, object> metadata, string key)
         {
             object foundValue;
             if (metadata.TryGetValue(key, out foundValue))
             {
                 return foundValue?.ToString();
-            }
-
-            return null;
-        }
-
-        public static DateTime? AsMaybeDateTime(this string s)
-        {
-            if (s == null) return null;
-            DateTime dt;
-            if (DateTime.TryParse(s, out dt))
-            {
-                return dt;
             }
 
             return null;
--- a/src/ServiceControl/Infrastructure/MetadataExtensions.cs
+++ b/src/ServiceControl/Infrastructure/MetadataExtensions.cs
@@ -1,0 +1,41 @@
+﻿namespace ServiceControl
+{
+    using System;
+    using System.Collections.Generic;
+
+    static class MetadataExtensions
+    {
+        public static T MaybeGet<T>(this IDictionary<string, object> metadata, string key)
+        {
+            object foundValue;
+            if (metadata.TryGetValue(key, out foundValue))
+            {
+                return (T)foundValue;
+            }
+            return default(T);
+        }
+
+        public static string MaybeGetAsString(this IDictionary<string, object> metadata, string key)
+        {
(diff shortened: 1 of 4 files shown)


src/ServiceControl/Infrastructure/MetadataExtensions.cs
```diff
@@ -0,0 +1,41 @@
+﻿namespace ServiceControl
+{
+    using System;
+    using System.Collections.Generic;
+
+    static class MetadataExtensions
+    {
+        public static T MaybeGet<T>(this IDictionary<string, object> metadata, string key)
```
SzymonPobiega: How about `GetOrDefault` ?

danielmarbach: I would prefer GetOrDefault as well

src/ServiceControl/Infrastructure/MetadataExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+            {
+                return foundValue?.ToString();
+            }
+
+            return null;
+        }
+
+        public static DateTime? AsMaybeDateTime(this string s)
+        {
+            if (s == null) return null;
```
danielmarbach: Should this have braces?

src/ServiceControl/Infrastructure/MetadataExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+            DateTime dt;
+            if (DateTime.TryParse(s, out dt))
```
danielmarbach: Wouldn't we want to specify the datetime format we want to see here as well as the datetime style and the invariant culture?

src/ServiceControl/Infrastructure/MetadataExtensions.cs
mikeminutillo: What about the `MaybeGetAsString()`? `GetOrNullAsString()`?

src/ServiceControl/Infrastructure/MetadataExtensions.cs
mikeminutillo: Possibly.

The previous code was just `DateTime.Parse` but it was immediately after a `ToString()`. We might be able to just turn it into a `GetOrDefault<DateTime>()`. I'm not sure why it was being round-tripped through `System.String`.

src/ServiceControl/Infrastructure/MetadataExtensions.cs
WilliamBZA: `GetAsStringOrNull`

src/ServiceControl/Infrastructure/MetadataExtensions.cs
danielmarbach: Maybe then doing TryParse should work. I have not enough knowledge here to say it should work or it won't

danielmarbach:
@mikeminutillo I believe this should also fix #1120 added it to the description

SimonCropp:
@mikeminutillo @danielmarbach thanks guys

PR #1126, opened 2018-04-06
Stop checking for !disable queue name duplicates

WilliamBZA:
Prevents installer errors when more than one instance wants to disable queue creation

### DIFF
--- a/src/ServiceControl.Config/UI/SharedInstanceEditor/SharedServiceControlEditorViewModelValidator.cs
+++ b/src/ServiceControl.Config/UI/SharedInstanceEditor/SharedServiceControlEditorViewModelValidator.cs
@@ -42,7 +42,7 @@ namespace ServiceControl.Config.Validation
                     p.ErrorQueue,
                     p.AuditQueue,
                     p.AuditLogQueue
-                }).Where(queuename => string.Compare(queuename, "!disable", true) != 0)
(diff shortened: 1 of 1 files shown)


src/ServiceControl.Config/UI/SharedInstanceEditor/SharedServiceControlEditorViewModelValidator.cs
```diff
@@ -42,7 +42,7 @@ protected List<string> UsedQueueNames(TransportInfo transportInfo = null, string
                     p.ErrorQueue,
                     p.AuditQueue,
                     p.AuditLogQueue
-                })
+                }).Where(queuename => string.Compare(queuename, "!disable", true) != 0)
```
danielmarbach: From MSDN 

image

https://msdn.microsoft.com/en-us/library/zkcaxw5y(v=vs.110).aspx

danielmarbach:
I think we should bundle this with 1.47.4

src/ServiceControl.Config/UI/SharedInstanceEditor/SharedServiceControlEditorViewModelValidator.cs
WilliamBZA: Nice! I did not know that

PR #1131, opened 2018-04-11
Upgrade Raven 3.5

SzymonPobiega:
* We're going to remove the max throughput limit of 350 -> need to adjust the doco


src/ServiceControl/Recoverability/Archiving/ArchiveDocumentManager.cs
```diff
@@ -38,15 +38,10 @@ public ArchiveOperation CreateArchiveOperation(IDocumentSession session, string
             int documentCount = 0;
             var indexQuery = session.Query<FailureGroupMessageView>(new FailedMessages_ByGroup().IndexName);
 
-            if (cutOff.HasValue)
```
danielmarbach: What are the consequences of this removal?

SzymonPobiega:
I think we are good to merge to `develop` at this stage.

danielmarbach:
@SzymonPobiega the only thing outstanding for me is this question https://github.com/Particular/ServiceControl/pull/1131#discussion_r181426863

danielmarbach:

src/ServiceControl/Infrastructure/RavenDB/Expiration/ExpiredDocumentsCleanerBundle.cs
```diff
@@ -50,10 +50,14 @@ public void Dispose()
         {
             if (timer != null)
             {
-                using (var manualResetEvent = new ManualResetEvent(false))
+                lock (timer)
```
WilliamBZA: Wat dis?

src/ServiceControl/Recoverability/Archiving/ArchiveDocumentManager.cs
WilliamBZA: It means we no longer wait for non stale results as of that cutoff etag. It seems like they've removed this API completely in raven 4, but it [should still be there in 3.5](https://ravendb.net/docs/article-page/3.5/Csharp/client-api/session/querying/how-to-customize-query#waitfornonstaleresultsasof). The cutoff was the current time when the archive all messages message was created.

Perhaps we should re-include it or move closer to the v4 version by switching to [waitfornonstaleresults](https://ravendb.net/docs/article-page/3.5/Csharp/client-api/session/querying/how-to-customize-query#waitfornonstaleresults)

src/ServiceControl/Infrastructure/RavenDB/Expiration/ExpiredDocumentsCleanerBundle.cs
SzymonPobiega: This change is necessary because for some reasons Raven 3.5 calls dispose twice (not sure if sequentially or if it can happen in parallel). If you call

```
timer.Dispose(manualResetEvent);
```

twice it will lock forever. This lock protects the `timer = null` statement to ensure that the underlying timer dispose is attempted only once.

WilliamBZA: :+1:

src/ServiceControl/Recoverability/Archiving/ArchiveDocumentManager.cs
SzymonPobiega: Unfortunately we can't do it. The reason why I removed it was this runtime exceptions:

```
System.NotSupportedException: 'Since Stream() does not wait for indexing (by design), streaming query with WaitForNonStaleResults is not supported.'
```

SzymonPobiega: Damn meaning let's merge it?

PR #1133, opened 2018-04-13
Split test

mikeminutillo:
Splits When_messages_have_been_successfully_processed_by_multiple_instances.cs into 7 individual tests


mikeminutillo:
Not sure what to name each of these fixtures so I've just given them a number for now

PR #1134, opened 2018-04-13
[SPIKE] Simplify Acceptance Test Framework

mikeminutillo:
This is all the stuff we don't use in any of our acceptance tests. The resulting framework is much simpler to follow. 

Note that none of the tests have been touched.

From here we could make the `Done()` call async and properly pass through all of the async calls to the web stack.

We could also return a Task from the the `Run()` call and make all of the acceptance tests async.


mikeminutillo:
I should have done this in a series of commits but I kind of took to it with an axe

mikeminutillo:
First run is a pass

PR #1137, opened 2018-04-16
Reproduce time sent cast exception

danielmarbach:
As part of #1119 a regression was introduced that makes ServiceControl unable to parse the `TimeSent` metadata information

## Who's affected

All customers

## Stacktrace

```
2018-04-13 08:29:47.3031|76|Error|ServiceBus.Management.Infrastructure.Nancy.NServiceBusContainerBootstrapper|Http call failed
System.InvalidCastException: Specified cast is not valid.
at ServiceControl.MetadataExtensions.GetOrDefault[T](IDictionary`2 metadata, String key) in C:\sources\ServiceControl\src\ServiceControl\Infrastructure\MetadataExtensions.cs:line 12
at ServiceControl.MessageFailures.Api.GetErrorById.Map(FailedMessage message, IAsyncDocumentSession session) in C:\sources\ServiceControl\src\ServiceControl\MessageFailures\Api\GetErrorById.cs:line 59
at ServiceControl.MessageFailures.Api.GetErrorById.<<-ctor>b__0_1>d.MoveNext() in C:\sources\ServiceControl\src\ServiceControl\MessageFailures\Api\GetErrorById.cs:line 44
```

### DIFF
--- a/src/ServiceControl/Infrastructure/MetadataExtensions.cs
+++ b/src/ServiceControl/Infrastructure/MetadataExtensions.cs
@@ -29,11 +29,12 @@
         public static DateTime? GetAsNullableDatetime(this IDictionary<string, object> metadata, string key)
         {
             var datetimeAsString = metadata.GetAsStringOrNull(key);
-            DateTime dt;
-            if (datetimeAsString != null && DateTime.TryParse(datetimeAsString, out dt))
+
+            if (datetimeAsString != null)
             {
-                return dt;
+                return DateTime.Parse(datetimeAsString);
             }
+            
             return null;
         }
     }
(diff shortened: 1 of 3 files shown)


src/ServiceControl/Infrastructure/MetadataExtensions.cs
```diff
@@ -24,5 +25,16 @@ public static string GetAsStringOrNull(this IDictionary<string, object> metadata
 
             return null;
         }
+
+        public static DateTime? GetAsNullableDatetime(this IDictionary<string, object> metadata, string key)
+        {
+            var datetimeAsString = metadata.GetAsStringOrNull(key);
+            DateTime dt;
+            if (datetimeAsString != null && DateTime.TryParse(datetimeAsString, out dt))
```
ramonsmits: Why TryParse? If the format is wrong should we then really import it? If we do import, we should log a WARN that this message was imported but that timesent was skipped.

WilliamBZA:
Changed `TryParse` to be a `Parse`

PR #1139, opened 2018-04-17
Expose maintance port

danielmarbach:
image

image

### Edit
image


src/ServiceControl/Infrastructure/RavenDB/RavenBootstrapper.cs
```diff
@@ -74,11 +74,12 @@ public void StartRaven(EmbeddableDocumentStore documentStore, Settings settings,
 
             documentStore.Configuration.DisableClusterDiscovery = true;
             documentStore.Configuration.ResetIndexOnUncleanShutdown = true;
-            documentStore.Configuration.Port = settings.Port;
+            documentStore.Configuration.Port = settings.MaintenancePort;
             documentStore.Configuration.HostName = settings.Hostname == "*" || settings.Hostname == "+"
                 ? "localhost"
                 : settings.Hostname;
-            documentStore.Configuration.VirtualDirectory = $"{settings.VirtualDirectory}/storage";
+            // TODO: Still required?
```
danielmarbach: I'm on the fence about this. The assumption previously was that it is possible to have a virtual directory but it seems to not be fully implemented plus now with the split would we need to introduce a dedicated virtual directory for the maintenance mode since we are no longer sharing the same base URL?

src/ServiceControlInstaller.Engine/Instances/ServiceControlInstance.cs
```diff
@@ -111,11 +113,7 @@ public string StorageUrl
                         host = HostName;
                         break;
                 }
-                if (string.IsNullOrWhiteSpace(VirtualDirectory))
-                {
-                    return $"http://{host}:{Port}/storage/";
-                }
-                return $"http://{host}:{Port}/{VirtualDirectory}{(VirtualDirectory.EndsWith("/") ? String.Empty : "/")}storage/";
+                return $"http://{host}:{MaintenancePort}/";
```
danielmarbach: Should we take the virtual directory into account?

src/ServiceControlInstaller.Engine/Instances/ServiceControlInstance.cs
```diff
@@ -159,6 +157,15 @@ public string AclUrl
             }
         }
 
+        public string AclMaintenanceUrl
+        {
+            get
+            {
+                var baseUrl = $"http://{HostName}:{MaintenancePort}/";
```
danielmarbach: Should we take the virtual directory into account?

src/ServiceControlInstaller.Engine/Instances/ServiceControlInstance.cs
mikeminutillo: > This setting is provided for backward compatibility and should be considered obsolete.

https://docs.particular.net/servicecontrol/creating-config-file#host-settings-servicecontrolvirtualdirectory

I'd say no

src/ServiceControl/Infrastructure/RavenDB/RavenBootstrapper.cs
SzymonPobiega: I'd drop the virtual directory

src/ServiceControlInstaller.Engine/Instances/ServiceControlInstance.cs
SzymonPobiega: Nope

PR #1150, opened 2018-05-24
Remove indexes before performing data migration

WilliamBZA:
Deletes the indexes before migration starts to ensure that they are rebuilt from scratch

### DIFF
--- a/src/ServiceControlInstaller.Engine/Instances/ServiceControlInstance.cs
+++ b/src/ServiceControlInstaller.Engine/Instances/ServiceControlInstance.cs
@@ -435,6 +435,23 @@ namespace ServiceControlInstaller.Engine.Instances
             }
         }
 
+        public void RemoveDatabaseIndexes()
+        {
+            var folders = AppConfig.RavenDataPaths().ToList();
+
+            foreach (var folder in folders.OrderByDescending(p => p.Length))
+            {
+                try
+                {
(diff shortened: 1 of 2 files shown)


src/ServiceControlInstaller.Engine/Instances/ServiceControlInstance.cs
```diff
@@ -428,6 +428,23 @@ public void UpdateDatabase(Action<string> updateProgress)
...
 
+        public void RemoveDatabaseIndexes()
+        {
+            var folders = AppConfig.RavenDataPaths().ToList();
+
+            foreach (var folder in folders.OrderByDescending(p => p.Length))
+            {
+                try
+                {
+                    FileUtils.DeleteDirectory(Path.Combine(folder, "Indexes"), true, false);
```
danielmarbach: I think you also need to delete IndexDefinitions folder. I can very tomorrow

danielmarbach:
Rebased for testing

src/ServiceControlInstaller.Engine/Instances/ServiceControlInstance.cs
danielmarbach: Upgrading the small database confirmed that we don't have to delete the IndexDefinitions folder. Going to merge

PR #1155, opened 2018-05-25
Stale indexes after upgrade

danielmarbach:


danielmarbach:
Very rough draft. Still needs a lot of work and love

src/ServiceControl/Upgrade/StaleIndexAfterUpgrade.cs
```diff
@@ -0,0 +1,84 @@
...
+    using System.Threading;
+    using System.Threading.Tasks;
+    using NServiceBus;
+    using NServiceBus.Features;
+
+    public class StaleIndexAfterUpgrade : Feature
+    {
+        public StaleIndexAfterUpgrade()
+        {
+            EnableByDefault();
```
WilliamBZA: How does the feature pipeline work, could we check for the existence of the file at this point already?

danielmarbach: We could do it in the conditions but I'm not sure whether it is really better in terms of having a feature disabled vs a feature enabled but the startup task noop. Happy to change though

danielmarbach:
Another idea:instead of doing the file it might be better to use the appsettings. Then I don't need to map datetimes, can assume it has always write rights and will be cleaned up when removed. Should I further spike that approach so thag we can compare?

src/ServiceControl/Upgrade/StaleIndexAfterUpgrade.cs
WilliamBZA: The SC log lists all the features at startup. Excluding it here would be different to a noop in that the feature would no longer appear in the log file at all.

I'm trying to decide whether or not that's a good thing...

WilliamBZA:
> can assume it has always write rights

Good point, we can't make that assumption. Any flags will have to be stored in RavenDB. We can store it as a single document with a well-known document ID thereby not needing to wait for indexes before we read or write it.

danielmarbach:
Right now all the flags are stored in the config settings on disk and the
installer has the right to modify it as well as SC itself. Are you
proposing now we should deviat from that? The consequences are the
installer becomes more complex and requires a DB connection as well as
raven needs to run in maintenance mode. Feels akward and unnecessarily
complex for me

WilliamBZA <notifications@github.com> schrieb am Sa., 26. Mai 2018, 08:10:

danielmarbach:
We could use the logpath. That has always write rights. One can say the
upgrade indicator is almost like a log😁

WilliamBZA <notifications@github.com> schrieb am Sa., 26. Mai 2018, 08:23:

WilliamBZA:
Ping @SzymonPobiega @danielmarbach @mikeminutillo and maybe @boblangley. Can I get a review or 2 on this?

src/ServiceControl/Upgrade/StaleIndexAfterUpgrade.cs
```diff
@@ -0,0 +1,84 @@
...
+            protected override void OnStop()
+            {
+                tokenSource?.Cancel();
+                checkTask?.GetAwaiter().GetResult();
+            }
+
+            protected override void OnStart()
+            {
+                var strings = Directory.GetFiles(directory, "*.upgrade");
+                var latestUpgrade = strings.Select(f => Convert.ToInt64(Path.GetFileNameWithoutExtension(f))).OrderByDescending(x => x).ToArray();
```
mikeminutillo: Should this be a `TryParse` just in case there is some other file that matches?

src/ServiceControl/Upgrade/StaleIndexAfterUpgrade.cs
```diff
@@ -0,0 +1,84 @@
...
+                var latest = DateTime.FromFileTimeUtc(fileTime);
+
+                // File exists, so assume in progress
+                staleIndexInfoStore.Store(new StaleIndexInfo { InProgress = true, StartedAt = latest });
+
+                checkTask = Task.Run(async () =>
+                {
+                    while (!tokenSource.IsCancellationRequested)
+                    {
+                        if (await staleIndexChecker.IsReindexingInComplete(latest, tokenSource.Token).ConfigureAwait(false))
```
mikeminutillo: Does this block for a while? If it doesn't, would it make sense to include a Task.Delay in here somewhere?

src/ServiceControl/Upgrade/StaleIndexAfterUpgrade.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            protected override void OnStop()
+            {
+                tokenSource?.Cancel();
+                checkTask?.GetAwaiter().GetResult();
+            }
+
+            protected override void OnStart()
+            {
+                var strings = Directory.GetFiles(directory, "*.upgrade");
```
mikeminutillo: Is it possible for two instances to share a log folder? Do we think that's a common scenario?

src/ServiceControl/Upgrade/StaleIndexAfterUpgrade.cs
```diff
@@ -0,0 +1,84 @@
...
+                // File exists, so assume in progress
+                staleIndexInfoStore.Store(new StaleIndexInfo { InProgress = true, StartedAt = latest });
+
+                checkTask = Task.Run(async () =>
+                {
+                    while (!tokenSource.IsCancellationRequested)
+                    {
+                        if (await staleIndexChecker.IsReindexingInComplete(latest, tokenSource.Token).ConfigureAwait(false))
+                        {
+                            File.Delete(Path.Combine(directory, $"{fileTime}.upgrade"));
```
mikeminutillo: What happens if this fails?

src/ServiceControlInstaller.Engine/Database/DatabaseMigrations.cs
```diff
@@ -67,6 +67,8 @@ public static void RunDataMigration(Action<string> updateProgress, string instal
                     }
                 }
             } while (attempts < 2);
+
+            File.Create(Path.Combine(logPath, $"{DateTime.UtcNow.ToFileTimeUtc().ToString()}.upgrade")).Close();
```
mikeminutillo: Would it make sense to create an Upgrade Marker File service or something to hide/encapsulate the implementation details of the marker file?

src/ServiceControl/Upgrade/StaleIndexAfterUpgrade.cs
mikeminutillo: I think it would still appear but it would be listed as disabled. I think it's fine as is.

src/ServiceControlInstaller.Engine/Database/DatabaseMigrations.cs
WilliamBZA: How would that be different to this file?

src/ServiceControl/Upgrade/StaleIndexAfterUpgrade.cs
WilliamBZA: crash

src/ServiceControl/Upgrade/StaleIndexAfterUpgrade.cs
WilliamBZA: `IsReindexingInComplete` blocks for 1 minute before cancelling, which means this loop iterates once per minute

src/ServiceControl/Upgrade/StaleIndexAfterUpgrade.cs
WilliamBZA: The normal log files would be pretty unusable at that stage, I don't think that's a scenario we can support.

src/ServiceControlInstaller.Engine/Database/DatabaseMigrations.cs
danielmarbach: @mikeminutillo I gave it a try over here https://github.com/Particular/ServiceControl/commit/c68b2d44ff00ee03d2214bcaf41168610dca370d but then I realized that the engine does not reference anything from SC. If we want to go on this path we have to link the source file in the engine. Is it worth it?

danielmarbach:
@mikeminutillo, please review again and then merge if good

src/ServiceControl/Upgrade/StaleIndexAfterUpgrade.cs
```diff
@@ -0,0 +1,104 @@
...
+                this.staleIndexInfoStore = staleIndexInfoStore;
+                this.staleIndexChecker = staleIndexChecker;
+
+                directory = baseDirectory ?? loggingSettings.LogPath;
+            }
+
+            protected override void OnStop()
+            {
+                tokenSource?.Cancel();
+                checkTask?.GetAwaiter().GetResult();
```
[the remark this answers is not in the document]
mikeminutillo: Wow. It doesn't. 

`checkTask?.GetAwaiter().GetResult()` returns null if checkTask is null

`(checkTask?.GetAwaiter()).GetResult()` throws if checkTask is null

PR #1156, opened 2018-05-29
Load ravenstudio directly to system DB

WilliamBZA:
In Raven35 with embedded databases, the form shows no databases as it treats the embedded DB as the system DB. This changes the url that opens to be one that navigates directly to the system DB. The user is presented with a warning that tampering with the system DB is dangerous.


danielmarbach:
Nice change

PR #1157, opened 2018-05-31
Track migration through restarts

mikeminutillo:
Tracks migrations though restarts. If the SCMU stops and restarts during the upgrade process, this will display a label next to the instance showing that it is going through data migration.

### DIFF
--- a/src/ServiceControl/Infrastructure/MarkerFile.cs
+++ b/src/ServiceControl/Infrastructure/MarkerFile.cs
@@ -1,0 +1,45 @@
+﻿namespace ServiceControl.Infrastructure
+{
+    using System;
+    using System.IO;
+
+    public class MarkerFileService
+    {
+        private string rootPath;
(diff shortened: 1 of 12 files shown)


mikeminutillo:
Doesn't currently auto-refresh when the file gets removed. I don't think it's worth adding a File System Watcher or anything here. This is only to catch an edge case.

src/ServiceControl/Infrastructure/MarkerFile.cs
```diff
@@ -0,0 +1,45 @@
...
+        public IDisposable CreateMarker(string name)
+        {
+            var path = Path.Combine(rootPath, name);
+            using (File.Open(path, FileMode.OpenOrCreate))
+            {
+                return new MarkerFile(path);
+            }
+        }
+
+        class MarkerFile : IDisposable
```
WilliamBZA: Ooooh I like this!

Issue #1165, opened 2018-06-07
SCMU instance details didn't update hostname after edit 


gbiellem:
After installed SC version 2.0.   I  added a single instance.
I then edited the instance to details   The only change was to modify the hostname from  localhost to *

The config file was successfully updated and the service restarted but the SCMU still showed the old value in the instance details in the main screen.

After restarting the SCMU the correct value was shown.

WilliamBZA:
I'm on it @gbiellem!  Thanks for finding this

danielmarbach:
Fixed by #1168. Will be released as part of 2.0.1

PR #1168, opened 2018-06-07
Update base instance from vm

WilliamBZA:
Fixes https://github.com/Particular/ServiceControl/issues/1165

Updates the service instance from the edited viewmodel details and only after updating those properties does it publish the `RefreshInstances` command, this ensures that the UI is updated only after the new properties have been applied

### DIFF
--- a/src/ServiceControl.Config/Commands/EditServiceControlInstanceCommand.cs
+++ b/src/ServiceControl.Config/Commands/EditServiceControlInstanceCommand.cs
@@ -1,21 +1,25 @@
 namespace ServiceControl.Config.Commands
 {
     using System;
+    using Caliburn.Micro;
+    using ServiceControl.Config.Events;
     using ServiceControl.Config.Framework;
     using ServiceControl.Config.Framework.Commands;
     using ServiceControl.Config.UI.InstanceDetails;
     using ServiceControl.Config.UI.InstanceEdit;
     using ServiceControlInstaller.Engine.Instances;
-    
+
     class EditServiceControlInstanceCommand : AbstractCommand<InstanceDetailsViewModel>
     {
         private readonly Func<ServiceControlInstance, ServiceControlEditViewModel> editViewModel;
+        private readonly IEventAggregator eventAggregator;
         private readonly IWindowManagerEx windowManager;
 
-        public EditServiceControlInstanceCommand(IWindowManagerEx windowManager, Func<ServiceControlInstance, ServiceControlEditViewModel> editViewModel) : base(null)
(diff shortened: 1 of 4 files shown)


src/ServiceControl.Config/UI/InstanceDetails/InstanceDetailsViewModel.cs
```diff
@@ -210,6 +208,7 @@ public void Handle(RefreshInstances message)
             NotifyOfPropertyChange("Transport");
             NotifyOfPropertyChange("BrowsableUrl");
             NotifyOfPropertyChange("InMaintenanceMode");
+            NotifyOfPropertyChange("InMaintenanceMode");
```
danielmarbach: This is now raised twice

src/ServiceControl.Config/Commands/EditServiceControlInstanceCommand.cs
```diff
@@ -1,28 +1,36 @@
...
 
         public override void Execute(InstanceDetailsViewModel viewModel)
         {
             var editVM = editViewModel((ServiceControlInstance)viewModel.ServiceInstance);
 
             windowManager.ShowInnerDialog(editVM);
+
+            editVM.UpdateInstanceFromViewModel((ServiceControlInstance)viewModel.ServiceInstance);
+
+            eventAggregator.PublishOnUIThread(new RefreshInstances());
```
danielmarbach: Is there a better way to structure this? We create the edit model, show it with the instance and then update it again with the instane and the publish an event that is subscribed by another VM that just triggers property changed. Would it simplify things to send the instance over the dispatched event? Then the edit model could subscribe to that and rebind as well as the instance view model?

WilliamBZA: Not really without reworking the UI command pattern.

When the command is invoked, a `ServiceControlEditViewModel` is created from a `ServiceControlInstance` object. This `ServiceControlInstance` object is the "truth" that needs to be updated. Once we have the `ServiceControlEditViewModel` instance, this is passed to the modal dialog. So the dialog has no reference back to the truth instance, but the command that launched the dialog does. This is why we have to call the `UpdateInstanveFromViewModel` here in order to pull the viewmodel values back into the truth object.

From here, we can bypass the eventAggregator and simply call `instance.Handle(null);` which will trigger the refresh without using the event aggregator, but that's about the only tweak I can come up with at the moment.

Any suggestions?

danielmarbach: Ok then I withdraw my objections. I will test this locally but from a code perspective this looks good

danielmarbach:
Smoke test and it works without restart now.

danielmarbach:
image

Issue #1166, opened 2018-06-07
Maintenance mode option in SCMU throws exception


gbiellem:
## Who's affected

Customers using ServiceControl 2.0

## Symptoms

- With a fresh install of V2.0
- Added  a new instance 
- Click the spanner icon in the instance details
- Click the start maintenance mode option.  

Following exception is shown:

```
System.Configuration.ConfigurationErrorsException: The configuration file has been changed by another program. (C:\Program Files (x86)\Particular Software\Particular.ServiceControl\ServiceControl.exe.config)
   at System.Configuration.MgmtConfigurationRecord.SaveAs(String filename, ConfigurationSaveMode saveMode, Boolean forceUpdateAll)
   at ServiceControlInstaller.Engine.Configuration.ServiceControl.AppConfig.EnableMaintenanceMode() in C:\BuildAgent\work\7189a56f9f44affc\src\ServiceControlInstaller.Engine\Configuration\ServiceControl\AppConfig.cs:line 32
   at ServiceControl.Config.UI.AdvancedOptions.ServiceControlAdvancedViewModel.<>c__DisplayClass34_0.<StartServiceInMaintenanceMode>b__0() in C:\BuildAgent\work\7189a56f9f44affc\src\ServiceControl.Config\UI\AdvancedOptions\ServiceControlAdvancedViewModel.cs:line 75
   at System.Threading.Tasks.Task.Execute()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at ServiceControl.Config.UI.AdvancedOptions.ServiceControlAdvancedViewModel.<StartServiceInMaintenanceMode>d__34.MoveNext() in C:\BuildAgent\work\7189a56f9f44affc\src\ServiceControl.Config\UI\AdvancedOptions\ServiceControlAdvancedViewModel.cs:line 79
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #1169, opened 2018-06-07
Reload config before maintainence mode changes

WilliamBZA:
Fixes https://github.com/Particular/ServiceControl/issues/1166

Reloads the app.config setting before trying to change into and out of maintenance mode


danielmarbach:
Smoked tested.

PR #1176, opened 2018-06-11
Domain events async

danielmarbach:


src/ServiceControl/Infrastructure/SignalR/GlobalEventHandler.cs
```diff
@@ -16,8 +17,8 @@ public void Broadcast(object @event)
             };
             var context = GlobalHost.ConnectionManager.GetConnectionContext<MessageStreamerConnection>();
 
-            context.Connection.Broadcast(new Envelope { Types = types, Message = @event }, emptyArray)
-                 .Wait();
+            await context.Connection.Broadcast(new Envelope { Types = types, Message = @event }, emptyArray)
```
SzymonPobiega: just return?

src/ServiceControl/CustomChecks/CustomCheckNotifications.cs
```diff
@@ -19,26 +19,27 @@ public void OnNext(IndexChangeNotification value)
         {
             try
             {
-                UpdateCount();
+                UpdateCount().GetAwaiter().GetResult();
```
WilliamBZA: Is this signature async in later versions?

src/ServiceControl/CustomChecks/ReportCustomCheckResultHandler.cs
```diff
@@ -37,13 +38,19 @@ public void Handle(ReportCustomCheckResult message)
                 throw new Exception("Received an custom check message without proper initialization of the HostId in the schema");
             }
 
+            HandleAsync(message).GetAwaiter().GetResult();
+        }
+
+        private async Task HandleAsync(ReportCustomCheckResult message)
+        {
```
WilliamBZA: Should the validations also be moved into the async method?

src/ServiceControl/ExternalIntegrations/EventDispatcher.cs
```diff
@@ -143,8 +145,15 @@ private bool TryDispatchEventBatch()
                 {
                     Logger.Debug($"Dispatching {allContexts.Length} events.");
                 }
-                var eventsToBePublished = eventPublishers.SelectMany(p => p.PublishEventsForOwnContexts(allContexts, session));
 
+                var eventsToBePublished = new List<object>();
+                foreach (var publisher in eventPublishers)
```
WilliamBZA: Maybe await the entire set of tasks instead of each one individually? 

```CSharp
var eventsToBePublished = Task.WaitAll(eventPublishers.Select(publisher  => publisher.PublishEventsForOwnContexts(allContexts, session).ConfigureAwait(false)).SelectMany();
```

Or something, I'd have to check the syntax properly

src/ServiceControl/MessageFailures/Handlers/MessageFailureResolvedHandler.cs
```diff
@@ -36,7 +43,8 @@ public void Handle(MessageFailureResolvedByRetry message)
 
             foreach (var alternative in message.AlternativeFailedMessageIds.Where(x => x != message.FailedMessageId))
             {
-                if (MarkMessageAsResolved(alternative))
+                if (await MarkMessageAsResolved(alternative)
```
WilliamBZA: Same as above, WaitAll so the await isn't synchronous per message

src/ServiceControl/Recoverability/Retrying/Handlers/PendingRetriesHandler.cs
```diff
@@ -24,23 +25,34 @@ public PendingRetriesHandler(IBus bus, IDocumentStore store, RetryDocumentManage
...
+        {
+            HandleAsync(message).GetAwaiter().GetResult();
+        }
+
+        private async Task HandleAsync(RetryPendingMessagesById message)
         {
             foreach (var messageUniqueId in message.MessageUniqueIds)
             {
-                manager.RemoveFailedMessageRetryDocument(messageUniqueId);
+                await manager.RemoveFailedMessageRetryDocument(messageUniqueId)
```
WilliamBZA: Await the entire list instead of each message?

src/ServiceControl/MessageFailures/Handlers/MessageFailureResolvedHandler.cs
danielmarbach: I deliberately left out all those kinds of optimizations. Let's first get everything working as is and then start tweaking things like that, ok?

src/ServiceControl/Recoverability/Retrying/Handlers/PendingRetriesHandler.cs
danielmarbach: See https://github.com/Particular/ServiceControl/pull/1176#discussion_r194648515

src/ServiceControl/ExternalIntegrations/EventDispatcher.cs
danielmarbach: See https://github.com/Particular/ServiceControl/pull/1176#discussion_r194648515

src/ServiceControl/CustomChecks/CustomCheckNotifications.cs
danielmarbach: Not sure though I figured it is better to align

src/ServiceControl/CustomChecks/ReportCustomCheckResultHandler.cs
danielmarbach: I can do it while merging if that is ok

PR #1182, opened 2018-06-12
Make subscriptions decoupled from address

danielmarbach:


danielmarbach:
@Particular/ravendb-persistence-maintainers would appreciate a review as well. The goal is to switch the dedicated persistence away from the Address class that was available in v5. I borrow code that I've been writing once for the RavenDB persister which got extended by @DavidBoike and others. Maybe someone can chime and and see if the conversation here is actually good.

DavidBoike:
Looks right to me. You should have it easier here, I think, because we had to worry about what if an endpoint is scaled out and only one instance is updated. What if after the updated instance receives a subscription request and updates the document, then later the non-updated instance receives the same request and updates the document again? At least this should be a one-way, one-time upgrade.

PR #1183, opened 2018-06-13
Make all the handlers async

SzymonPobiega:


src/ServiceControl/HeartbeatMonitoring/LegacyHandler.cs
```diff
@@ -10,12 +11,12 @@ class LegacyHandler : IHandleMessages<RegisterPotentiallyMissingHeartbeats>,
         IHandleMessages<EnableEndpointMonitoring>,
         IHandleMessages<DisableEndpointMonitoring>
     {
-        public void Handle(RegisterPotentiallyMissingHeartbeats message) { }
+        public Task Handle(RegisterPotentiallyMissingHeartbeats message, IMessageHandlerContext context) => Task.FromResult(0);
```
danielmarbach: In the cleanup phase, we should get rid of the `Task.FromResult` here

PR #1187, opened 2018-06-13
Update ingestion satellites

mikeminutillo:
Updates the error/audit ingestion satellites to V6


src/ServiceControl/Operations/AuditImporter.cs
```diff
@@ -18,7 +21,40 @@ public AuditImporterFeature()
...
+            var settings = context.Settings.Get<Settings>();
+
+            if(settings.IngestAuditMessages)
+            {
+                context.Container.ConfigureComponent<AuditImporter>(DependencyLifecycle.SingleInstance);
+                context.Container.ConfigureComponent<AuditIngestor>(DependencyLifecycle.SingleInstance);
+
+                context.AddSatelliteReceiver(
+                    "Audit Import",
+                    settings.AuditQueue,
```
SzymonPobiega: Is the `AuditQueue` expected to be logical address or transport address? If logical than we should translate to transport address here.

mikeminutillo: It is meant to be just a logical queue name. What's the best way to do that translation? Like this?

```csharp
var errorQueueLogicalAddress = LogicalAddress.CreateLocalAddress(settings.ErrorQueue, null);
var transportInfrastructure = context.Settings.Get<TransportInfrastructure>();
var errorQueueTransportAddress = transportInfrastructure.ToTransportAddress(errorQueueLogicalAddress);
```

mikeminutillo: Or should I be doing something more like this?

```csharp
var auditInstance = transportInfrastructure.BindToLocalEndpoint(new EndpointInstance(settings.AuditQueue));

var auditLogicalAddress = LogicalAddress.CreateLocalAddress(settings.AuditQueue, auditInstance.Properties);
var auditTransportAddress = transportInfrastructure.ToTransportAddress(auditLogicalAddress);
```

I'm not sure entirely what the transports use the instance properties for

src/ServiceControl/Operations/NServiceBusSettingsExtensions.cs
```diff
@@ -0,0 +1,16 @@
...
+    using NServiceBus.Settings;
+    using NServiceBus.Transport;
+
+    public static class NServiceBusSettingsExtensions
+    {
+        public static string ToTransportAddress(this ReadOnlySettings settings, string queueName)
+        {
+            var transportInfrastructure = settings.Get<TransportInfrastructure>();
+            var logicalAddress = LogicalAddress.CreateLocalAddress(queueName, null);
+            return transportInfrastructure.ToTransportAddress(logicalAddress);
```
SzymonPobiega: There is already a `GetTransportAddress` address on the settings that takes a logical address.

PR #1189, opened 2018-06-13
Fix FailedMessagesFeature

danielmarbach:


src/ServiceControl/MessageFailures/FailedMessagesFeature.cs
```diff
@@ -18,13 +18,14 @@ public class FailedMessagesFeature : Feature
...
             EnableByDefault();
-            RegisterStartupTask<WireUpFailedMessageNotifications>();
         }
 
         protected override void Setup(FeatureConfigurationContext context)
         {
             context.Container.ConfigureComponent<FailedMessageViewIndexNotifications>(DependencyLifecycle.SingleInstance);
             context.Container.ConfigureComponent<DetectSuccessfullRetriesEnricher>(DependencyLifecycle.SingleInstance);
+
+            context.RegisterStartupTask(b => b.Build<WireUpFailedMessageNotifications>());
```
mikeminutillo: I think that startup tasks are not added to the container unless we explicitly add them

src/ServiceControl/MessageFailures/FailedMessagesFeature.cs
```diff
@@ -79,14 +80,26 @@ private IEnumerable<string> GetAlternativeUniqueMessageId(IReadOnlyDictionary<st
...
+            static string ExtractQueueNameForLegacyReasons(string address)
+            {
+                var atIndex = address?.IndexOf("@", StringComparison.InvariantCulture);
+
+                if (atIndex.HasValue && atIndex.Value > -1)
+                {
+                    return address.Substring(0, atIndex.Value);
                 }
+
+                return address;
```
mikeminutillo: I suspect we're going to want this elsewhere as it will come up a lot

mikeminutillo: We can do that later I guess

src/ServiceControl/MessageFailures/FailedMessagesFeature.cs
mikeminutillo: I guess we'll figure that out easily enough once the tests start

PR #1190, opened 2018-06-13
ReturnToSenderDequeuer

SzymonPobiega:


src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSenderDequeuer.cs
```diff
@@ -1,81 +1,73 @@
...
         {
             if (shouldProcess(message))
             {
-                HandleMessage(message, bodyStorage, sender);
-
-                if (IsCounting)
-                {
-                    CountMessageAndStopIfReachedTarget();
-                }
+                await HandleMessage(message, bodyStorage, sender);
```
danielmarbach: `ConfigureAwait(false)` but we can fix that later

src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSenderDequeuer.cs
```diff
@@ -92,72 +84,59 @@ static byte[] ReadFully(Stream input)
             }
         }
 
-        public static void HandleMessage(TransportMessage message, IBodyStorage bodyStorage, ISendMessages sender) //Public for testing
+        public static async Task HandleMessage(MessageContext message, IBodyStorage bodyStorage, IDispatchMessages sender) //Public for testing
         {
-            message.Headers.Remove("ServiceControl.Retry.StagingId");
+            var body = new byte[0];
```
danielmarbach: This could be static but we can fix that later

src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSenderDequeuer.cs
```diff
@@ -190,11 +166,13 @@ public virtual void Run(Predicate<TransportMessage> filter, CancellationToken ca
                 }
 
                 shouldProcess = filter;
-                resetEvent.Reset();
                 targetMessageCount = expectedMessageCount;
                 actualMessageCount = 0;
                 Log.DebugFormat("Starting receiver");
-                receiver.StartInternal();
+
+                var config = createEndpointConfiguration();
```
danielmarbach: Java style method name? 🚎

src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSenderDequeuer.cs
```diff
@@ -206,6 +184,10 @@ public virtual void Run(Predicate<TransportMessage> filter, CancellationToken ca
             {
                 Log.DebugFormat("Waiting for finish");
                 resetEvent.Wait(cancellationToken);
```
danielmarbach: Ideally, this should be non-blocking but you can argue that right now it is not a big deal because it is done in a continuation of endpoint start which is naturally scheduled and hopefully already away from the thread that entered the call context here

src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSenderDequeuer.cs
SzymonPobiega: That's a func 😂

PR #1193, opened 2018-06-14
More upgrade tweaks

danielmarbach:


src/ServiceControl/Operations/FailedMessagePersister.cs
```diff
@@ -72,7 +74,7 @@ public FailedMessagePersister(IDocumentStore store, BodyStorageFeature.BodyStora
                 // TODO: Do we need to persist all of these things separately still?
                 message.Headers.MessageIntent(),
                 true,
-                message.Headers.CorrelationId(),
+                message.Headers[Headers.CorrelationId],
```
mikeminutillo: We can't assume that any particular headers exist. Otherwise a missing header throws an exception and the message cannot be ingested.

src/ServiceControl/Operations/TransportMessageCleaner.cs
```diff
@@ -1,14 +0,0 @@
...
-
-namespace ServiceControl.Operations
-{
-    using NServiceBus;
-
-    static class TransportMessageCleaner
-    {
-        public static void CleanForForwarding(TransportMessage message)
-        {
-            message.TimeToBeReceived = TimeSpan.MaxValue;
```
mikeminutillo: How is TTBR handled in V7? Is it only applied if there's a delivery constraint or is it implemented as a header? This used to cause a problem with sending messages to the Audit Log Queue that got removed due to TTBR.

src/ServiceControl/Operations/FailedMessagePersister.cs
danielmarbach: I was trying to find the code in question but couldn't. I will go back to the previous branch Nd see how it was implemented. I planned to point this out on the PR but forgot. Thanks for catching it

src/ServiceControl/Operations/TransportMessageCleaner.cs
danielmarbach: This code was never called. Maybe it was not due to the refactoring I'll double check

mikeminutillo:
@danielmarbach these were both caused by my previous PR.

I switched CorrelationId from property access on TransportMessage to an extension method on the headers but never actually wrote that extension method.

I also dropped the call to TransportMessageCleaner. I don't think it's needed in V7 but I wanted to confirm.

mikeminutillo:
I checked and TTBR is applied by the transport according to the presence of a DeliveryConstraint which won't be there when we forward the message. As such, the TransportCleaner is never needed and is safe to drop.

PR #1199, opened 2018-06-19
Don't try to create maintenance port ACL for versions before 2.0

### LINKED ISSUE #1198: Cannot edit V1.* instances of ServiceControl using V2.* of the SCMU
The UrlACL added for the DB port is malformed as the maintenance port is null and is not provided.

The SCMU should not try to recreate the maintenance port URL Acl if it is for a version 1.* instance of ServiceControl

danielmarbach:
Closes #1198


WilliamBZA:
Added a check for the parsing of the maintenance port

PR #1204, opened 2018-06-21
Only change DB port if provided as part of upgrade

WilliamBZA:
Fixes https://github.com/Particular/ServiceControl/issues/1203


danielmarbach:
@WilliamBZA pushed a small change

PR #1206, opened 2018-06-22
Hacks that would get ASQ to pass

danielmarbach:
A demonstration of the evil things I had to do to get ASQ tests to pass and also make SC work again on ASQ new transport. I think this is OK to get going until we have a specific transport extension seam that can be plugged into SC similar to how Monitoring does it.


WilliamBZA:
I feel like this belongs in CustomTransportConfiguration library

src/ServiceControl/Infrastructure/AsqBackwardsCompatibleQueueNameSanitizer.cs
```diff
@@ -0,0 +1,51 @@
+﻿namespace ServiceControl.Infrastructure
+{
+    using System;
+    using System.Security.Cryptography;
+    using System.Text;
+    using System.Text.RegularExpressions;
+
+    public static class BackwardsCompatibleQueueNameSanitizer
```
SzymonPobiega: Align class name with file name

SzymonPobiega:
How about using an approach similar to Monitoring: https://github.com/Particular/ServiceControl.Monitoring/tree/develop/src/ServiceControl.Transports.AzureStorageQueues ?

danielmarbach:
Exactly what I proposed but I think we can look at this extension points
wholistically after the tests run and maybe even expose to customers. Or
are you saying we should go for the proper solution first?

Szymon Pobiega <notifications@github.com> schrieb am Fr., 22. Juni 2018,
17:41:

SzymonPobiega:
@danielmarbach I have no strong opinion about that

PR #1213, opened 2018-06-28
Transport types

WilliamBZA:
Uses the customization as a transport type
All TC build configs need additional env variable called `ServiceControl.AcceptanceTests.TransportCustomization`


src/ServiceControl/Infrastructure/Settings/Settings.cs
```diff
@@ -336,7 +328,7 @@ private static bool GetForwardAuditMessages()
 
         static string GetTransportType()
         {
-            var typeName = SettingsReader<string>.Read("TransportType", typeof(MsmqTransport).AssemblyQualifiedName);
+            var typeName = SettingsReader<string>.Read("TransportType", "ServiceControl.Transports.Msmq.MsmqTransportCustomization, ServiceControl.Transports.Msmq");
```
danielmarbach: I wonder if we should get rid of the default value here. If we leave it then SC actually has to reference that customization because Type.GetType below would fail otherwise. If we remove it then you can't start SC.exe anymore without first copying a transport customization into the bin folder as well as explicitly setting a transport type though

src/ServiceControl.Transports.ASQ/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,35 @@
...
+//      Major Version
+//      Minor Version 
+//      Build Number
+//      Revision
+//
+// You can specify all the values or you can default the Build and Revision Numbers 
+// by using the '*' as shown below:
+// [assembly: AssemblyVersion("1.0.*")]
+[assembly: AssemblyVersion("1.0.0.0")]
+[assembly: AssemblyFileVersion("1.0.0.0")]
```
mikeminutillo: Should these assemblies all have their own version numbers or should they just use GitVersion to do it?

PR #1214, opened 2018-06-29
Monitoring 1.3.0

tmasternak:


danielmarbach:
@tmasternak once this is merged we also need in in develop

PR #1225, opened 2018-07-10
Avoid subscription versioning

timbussmann:
Newer versions of subscription storages and broker based transport no longer use subscription versioning. Using subscription versioning, this would not publish events to subscriber running on an older SC instance (major version) than the publisher as the custom persister uses basically RavenDB persistence's old subscription versioning logic.

This PR adds a hack to always set the `MessageType` version to `1.0.0`. This allows older SC instances still receiving events declared in the SC dll across major versions.

This should be tested somehow but I'm not sure how to setup such a scenario. Maybe @WilliamBZA can help with that?

### DIFF
--- a/src/ServiceControl.UnitTests/Infrastructure/RavenDB/SubscriptionPersisterTests.cs
+++ b/src/ServiceControl.UnitTests/Infrastructure/RavenDB/SubscriptionPersisterTests.cs
@@ -1,0 +1,71 @@
+﻿namespace ServiceControl.UnitTests.Infrastructure.RavenDB
+{
+    using System;
+    using System.Linq;
+    using System.Threading.Tasks;
+    using NServiceBus.Extensibility;
+    using NServiceBus.Settings;
+    using NServiceBus.Unicast.Subscriptions;
+    using NServiceBus.Unicast.Subscriptions.MessageDrivenSubscriptions;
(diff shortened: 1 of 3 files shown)


danielmarbach:
Just discussed this again with @timbussmann and we came to the conclusion that this approach doesn't work in scenarios where a vNext endpoint subscribes to a vPrevious endpoint. So we have to hack the message type version on the headers that we are sending out...

timbussmann:
After some more thinking about it, I came to the conclusion that we don't need any special handling about subscription message versioning (as we removed the hack to subscribe with version=1.0.0 message types on the subscription changes). As `MessageType` equality comparison does not use the version property and the persister implementation does heavily rely on the equality comparison. I've added two unit tests to verify this behavior, I think they are valuable enough to be merged into master.
ping @danielmarbach @WilliamBZA

timbussmann:
the SC 2.0 to 1.0 subscription logic should behave the same, as v5 also didn't take the version into account when comparing `MessageType` (https://github.com/Particular/NServiceBus/blob/support-5.2/src/NServiceBus.Core/Unicast/Subscriptions/MessageDrivenSubscriptions/MessageType.cs) (assuming that the subscription persister hasn't been changed significantly)

PR #1231, opened 2018-07-11
Adding additional data to startup diagnostics

### LINKED ISSUE #1227: Use V7 diagnostics feature to log startup items
Currently all enabled features and settings are only logged to the log file. It would be helpful if we can use the diagnostics feature of V7 to log the key features and settings to a separate diagnostics file

WojcikMike:
fixes #1227

This is WIP I have one thing that I would like to run by you:
I am using the same location for diagnostics as for logs, as it seems like a natural thing to do. what do you think?

At the moment the diagnostics input is a oneliner (meaning no nice break lines etc.) So I will work on this.

### DIFF
--- a/src/ServiceControl/Bootstrapper.cs
+++ b/src/ServiceControl/Bootstrapper.cs
@@ -177,7 +177,35 @@ Selected Transport Customization:   {settings.TransportCustomizationType}
             var logger = LogManager.GetLogger(typeof(Bootstrapper));
             logger.Info(startupMessage);
             endpointConfiguration.SetDiagnosticsPath(loggingSettings.LogPath);
-            endpointConfiguration.GetSettings().AddStartupDiagnosticsSection("Startup", startupMessage);
+            endpointConfiguration.GetSettings().AddStartupDiagnosticsSection("Startup", new
+            {
+                Settings = new
+                {
+                    settings.ApiUrl,
+                    settings.AuditLogQueue,
+                    settings.AuditQueue,
+                    settings.DatabaseMaintenancePort,
+                    settings.ErrorLogQueue,
+                    settings.DisableRavenDBPerformanceCounters,
+                    settings.DbPath,
+                    settings.ErrorQueue,
+                    settings.ForwardAuditMessages,
+                    settings.ForwardErrorMessages,
+                    settings.HttpDefaultConnectionLimit,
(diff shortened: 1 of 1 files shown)


danielmarbach:
The log path definitely makes sense since this is pretty much the only folder we can assume we have write access to. Centralizing the startup diagnostics with the logs is also great because then a customer can just zip the log folder and we have everything we need to know. 

For a starter I would just add the `Settings` and the `LogSettings` as a section. We probably want to exclude the TransportConnectionString from the settings object since it contains potentially sensitive information like credentials.

src/ServiceControl/Bootstrapper.cs
```diff
@@ -175,6 +176,8 @@ private void RecordStartup(LoggingSettings loggingSettings)
 
             var logger = LogManager.GetLogger(typeof(Bootstrapper));
             logger.Info(startupMessage);
+            endpointConfiguration.SetDiagnosticsPath(loggingSettings.LogPath);
+            endpointConfiguration.GetSettings().AddStartupDiagnosticsSection("Startup", startupMessage);
```
WilliamBZA: I would prefer if we added a diagnostic for each of the items in the startup object. This way they are not just a single output object

danielmarbach: @WilliamBZA with the approach of having a section for the settings and having a section for the logsettings this would be covered. @WojcikMike and I agreed that the startup message remains as is and will not be used for a section input

WojcikMike:
When trying to dump settings object into diagnostics I get this error:
2018-07-12 11:59:45.9394|36|Error|NServiceBus.WriteStartupDiagnostics|Failed to serialize startup diagnostics
System.InvalidOperationException: Method may only be called on a Type for which Type.IsGenericParameter is true.

The issue lies in this field:
https://github.com/Particular/ServiceControl/blob/develop/src/ServiceControl/Infrastructure/Settings/Settings.cs#L65

When I comment the whole field out it is fine, but using: [IgnoreDataMember] doesn't help. With that in mind, we probably need to identify important information from the settings class and use them. Unless you have a better idea

danielmarbach:
I think for now we have to create an anonymous object and map every property that we want ourselves to the anonymous object. Something like

```
new {
   settings.Prop1,
   settings.Prop2,
  ...
}
```

PR #1242, opened 2018-07-16
WIP Testing out fixed ASB transport

danielmarbach:


danielmarbach:
Merging this for now. The transport is only on myget! So we can't release that way but at least it gives us some breath in terms of failing tests for now until 8.0.1 is released

PR #1244, opened 2018-07-17
Clean shutdown

danielmarbach:


src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSenderDequeuer.cs
```diff
@@ -41,11 +43,13 @@ public ReturnToSenderDequeuer(TransportDefinition transportDefinition, ReturnToS
             };
 
             faultManager = new CaptureIfMessageSendingFails(store, domainEvents, IncrementCounterOrProlongTimer);
-            timer = new Timer(state => StopInternal());
+            timer = new Timer(state => Stop().GetAwaiter().GetResult());
         }
 
         async Task Handle(MessageContext message, IDispatchMessages sender)
         {
+            Log.Debug($"ReturnToSenderDequeuer.Handle({message.Headers[Headers.MessageId]}, {message.MessageId})");
```
danielmarbach: This should probably be removed

src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSenderDequeuer.cs
```diff
@@ -137,16 +143,17 @@ public Task CreateQueue()
...
-            Task.Run(() => syncEvent.TrySetResult(true));
+            return StopInternal();
         }
 
-        void StopInternal()
+        private async Task StopInternal()
         {
-            Task.Run(() => syncEvent.TrySetResult(true));
+            await Task.Run(() => syncEvent?.TrySetResult(true)).ConfigureAwait(false);
+            await (stopCompletionSource?.Task ?? (Task)Task.FromResult(0)).ConfigureAwait(false);
```
timbussmann: I think there is a potential race condition in case `Stop` is called before `Run` is executed (or when executed concurrently), but that's already the case right now so probably not the most critical thing to address?

PR #1247, opened 2018-07-17
Attempt to fix SatelliteImportFailuresHandler

danielmarbach:


src/ServiceControl/Operations/ErrorImporter.cs
```diff
@@ -77,5 +62,23 @@ private RecoverabilityAction OnError(RecoverabilityConfig config, ErrorContext e
 
             return recoverabilityAction;
         }
+
+        class StartupTask : FeatureStartupTask
+        {
+            public StartupTask(SatelliteImportFailuresHandler importFailuresHandler, ErrorImporter importer)
+            {
+                importer.importFailuresHandler = importFailuresHandler;
```
danielmarbach: i feel dirty now 🚿

WilliamBZA: 😆

WilliamBZA: This is a pity, but if it's the only way to get it to work...

PR #1248, opened 2018-07-17
update transport names for IgnoreTransports attribute

timbussmann:
not sure whether all transport topologies are affected by this, maybe @Particular/rabbitmq-transport-maintainers and @Particular/azure-maintainers can confirm this?

### DIFF
--- a/src/ServiceControl.AcceptanceTests/Recoverability/When_a_message_is_retried_with_a_replyTo_header.cs
+++ b/src/ServiceControl.AcceptanceTests/Recoverability/When_a_message_is_retried_with_a_replyTo_header.cs
@@ -16,7 +16,12 @@
     public class When_a_message_is_retried_with_a_replyTo_header : AcceptanceTest
     {
         // TODO: Add these transports back if/when then are updated to match this behavior
-        [Test, IgnoreTransports("AzureServiceBus", "AzureStorageQueues", "RabbitMq")]
+        [Test, IgnoreTransports(
+             "AzureServiceBus - Endpoint Topology", 
+             "AzureServiceBus - Forwarding Topology", 
+             "AzureStorageQueues", 
+             "RabbitMq - Conventional Routing Topology",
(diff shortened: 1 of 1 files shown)


bording:
Sorry, not sure what you're asking here.

src/ServiceControl.AcceptanceTests/Recoverability/When_a_message_is_retried_with_a_replyTo_header.cs
```diff
@@ -16,7 +16,12 @@
     public class When_a_message_is_retried_with_a_replyTo_header : AcceptanceTest
     {
         // TODO: Add these transports back if/when then are updated to match this behavior
-        [Test, IgnoreTransports("AzureServiceBus", "AzureStorageQueues", "RabbitMq")]
+        [Test, IgnoreTransports(
+             "AzureServiceBus - Endpoint Topology", 
+             "AzureServiceBus - Forwarding Topology",
```
danielmarbach: I would assume these ignores have been written with the topology at that time in mind. Under those circumstances, I would suggest we remove  "AzureServiceBus - Forwarding Topology" and "RabbitMq - Direct Routing Topology" because those have been introduced recently. Or should we say if something did not work under Endpoint Topology it will not work under Forwarding either?

timbussmann:
@bording it seems this tests should be ignored for rabbitmq as it assumes, that rabbit does not properly set the replyto value to the expected value. My question is, whether that is still the case and whether that's the case for both topologies.

danielmarbach:
Well all are green now without exclusions so maybe drop the current ignores?

mikeminutillo:
If they currently pass, I would drop the ignores. These were issues we had with specific V5 versions of these transports.

timbussmann:
see https://github.com/Particular/ServiceControl/pull/1305 to remove the attribute completely

timbussmann:
closing this as #1305 has been merged.

PR #1249, opened 2018-07-18
[WIP] Non concurrent subscribes

mikeminutillo:
Working with Daniel I have a theory that the concurrent subscribe calls are causing build failures in the Azure Service Bus tests


danielmarbach:
No longer needed since we fixed AzureServiceBus

PR #1255, opened 2018-07-19
cleanup SC/CompositeViews namespace

timbussmann:


src/ServiceControl/CompositeViews/Messages/ScatterGatherApi.cs
```diff
@@ -34,12 +34,15 @@ protected override void Load(ContainerBuilder builder)
         }
     }
 
-    public abstract class ScatterGatherApi<TIn, TOut> : IApi
-        where TOut : class
+    // used to host the static jsonSerializer field across the generic instances
+    public abstract class ScatterGatherApiBase
```
timbussmann: please have a closer look at this change when reviewing

src/ServiceControl/CompositeViews/Messages/ScatterGatherApi.cs
```diff
@@ -34,12 +34,15 @@ protected override void Load(ContainerBuilder builder)
         }
     }
 
-    public abstract class ScatterGatherApi<TIn, TOut> : IApi
-        where TOut : class
+    // used to hoist the static jsonSerializer field across the generic instances
+    public abstract class ScatterGatherApiBase
```
WilliamBZA: If I'm reading this right, it means that previously there were 4 instances of `JsonSerilizer` (since there are 4 classes that extend `ScatterGatherApi<TIn, TOut>`), but each class overwrote the static reference that the previous one created meaning only the last one remained?

This would fix that and mean there's only the one instance ever?

danielmarbach: Nothing has been overwritten @WilliamBZA. It was that every concrete closed representation had its own static field. Now we have one. That is good. But I wonder if it wouldn't be better to just outline that static field to another static class and then use it directly instead of having another base type

timbussmann: what @danielmarbach said. Daniel's approach works as well. Do you see any significant difference between those approaches? Happy either way (I went with the recommended approach by R#).

WilliamBZA: Thanks for learning me a thing 🏫  :+1:

danielmarbach: @timbussmann I'm good either way.

PR #1274, opened 2018-07-20
Cleanup ServiceControl.Recoverability namespace

weralabaj:
One thing I'm not sure about is `ArchiveDocumentManager.WaitForIndexUpdateOfArchiveOperation` method, should I remove the unused parameter or is it a mistake and that parameter should be used?


weralabaj:
One thing I'm not sure about is `ArchiveDocumentManager.WaitForIndexUpdateOfArchiveOperation` method, should I remove the unused parameter or is it a mistake and that parameter should be used?

WilliamBZA:
It's not used, so you can remove it :+1:

PR #1289, opened 2018-07-21
Add ability to control upgrades based on current and target versions

### LINKED ISSUE #1228: Prevent upgrading to version 2.*.* of ServiceControl from less than 1.41.3
Add check in the installer to not allow users to upgrade from versions of ServiceControl <= 1.41.3

Users should first upgrade to at least version 1.41.3 so that  internal migrations don't run at the same time as Raven migrations

### LINKED ISSUE #1288: Only delete indexes when upgrading from 1.x to 2.x
## Who's affected

All customers upgrading from 2.x to 2.y

## Symptoms

When a user upgrades from 2.x to 2.y we attempt to migrate the database. As part of that upgrade, indexes will be deleted

boblangley:
Fixes #1228
Fixes #1288 

What it looks like when you are stopped from upgrading. All the text is selectable / copyable.
image

```
You must upgrade to version 1.48.0 before upgrading to version 2.2.0:
1.	Uninstall version 2.2.0.
2.	Download and install version 1.48.0 from https://github.com/Particular/ServiceControl/releases/tag/1.48.0
3.	Upgrade this instance to version 1.48.0.
4.	Download and install the latest version from https://particular.net/start-servicecontrol-download
5.	Upgrade this instance to the latest version of ServiceControl.
```

### DIFF
--- a/src/ServiceControl.Config/Commands/UpgradeServiceControlInstanceCommand.cs
+++ b/src/ServiceControl.Config/Commands/UpgradeServiceControlInstanceCommand.cs
@@ -204,7 +204,7 @@
                     return;
                 }
 
-                reportCard = await Task.Run(() => installer.Upgrade(model.Name, upgradeOptions, progress));
+                reportCard = await Task.Run(() => installer.Upgrade(instance, upgradeOptions, progress));
 
                 if (reportCard.HasErrors || reportCard.HasWarnings)
                 {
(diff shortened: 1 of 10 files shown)


danielmarbach:
Cherry picked the commits against branch that targets master. See https://github.com/Particular/ServiceControl/pull/1290. Closing

PR #1291, opened 2018-07-21
Hotfix 2.1.3

danielmarbach:


src/ServiceControlInstaller.Engine/Unattended/UnattendServiceControlInstaller.cs
```diff
@@ -94,6 +94,12 @@ public bool Add(ServiceControlNewInstance details, Func<PathInfo, bool> promptTo
 
         public bool Upgrade(ServiceControlInstance instance, ServiceControlUpgradeOptions options)
         {
+            if (instance.Version < options.UpgradeInfo.CurrentMinimumVersion)
+            {
+                logger.Error($"Upgrade aborted. An interim upgrade to version {options.UpgradeInfo.RecommendedUpgradeVersion} is required before upgrading to version {ZipInfo.Version}. Download available at https://github.com/Particular/ServiceControl/releases/tag/{options.UpgradeInfo.RecommendedUpgradeVersion}");
```
mikeminutillo: This is problematic. We will direct the customer to the installer and they will download it but it will not install because it is older than the "current" installed version. They need the installation zip that is inside that version to be in the folder where SC can find it. Can we test that?

mikeminutillo: Ah. This will only happen for unattended installations.

Issue #1250, opened 2018-07-18
Align develop with new coding standards


danielmarbach:
- Take latest https://github.com/Particular/RepoStandards
- Update dotSettings and potentially ignore files where necessary
- Fix code inspections
- Reformat all the code to use the newest code style as one dedicated commit
- Where possible and needed use latest csharp feature (optional)

## POA

* [x] Project ServiceControl
  * [x] ServiceControl\CompositeViews @timbussmann
  * [x] ServiceControl\Contracts @timbussmann
  * [x] ServiceControl\CustomChecks @timbussmann
  * [x] ServiceControl\DbMigrations @danielmarbach
  * [x] ServiceControl\ExternalIntegrations @danielmarbach
  * [x] ServiceControl\EventLog @danielmarbach
  * [x] ServiceControl\Hosting @danielmarbach
  * [x] ServiceControl\HeartbeatMonitoring @timbussmann
  * [x] ServiceControl\Infrastructure @timbussmann
  * [x] ServiceControl\Licensing @timbussmann
  * [x] ServiceControl\MessageAuditing @timbussmann
  * [x] ServiceControl\MessageFailures @timbussmann
  * [x] ServiceControl\Monitoring @timbussmann
  * [x] ServiceControl\Operations @weralabaj
  * [x] ServiceControl\Recoverability @weralabaj
  * [x] ServiceControl\SagaAudit @weralabaj
  * [x] ServiceControl\Upgrade @weralabaj

* [x] Project ServiceControl.AcceptanceTests
  * [x] ServiceControl.AcceptanceTests\Audit @timbussmann
  * [x] ServiceControl.AcceptanceTests\CustomChecks @timbussmann
  * [x] ServiceControl.AcceptanceTests\EndpointTemplates @timbussmann
  * [x] ServiceControl.AcceptanceTests\ExternalIntegrations @timbussmann
  * [x] ServiceControl.AcceptanceTests\HeartbeatMonitoring @timbussmann
  * [x] ServiceControl.AcceptanceTests\MessageFailures @timbussmann
  * [x] ServiceControl.AcceptanceTests\MessageRedirects @timbussmann
  * [x] ServiceControl.AcceptanceTests\MultiInstance @timbussmann
  * [x] ServiceControl.AcceptanceTests\Recoverability @weralabaj
  * [x] ServiceControl.AcceptanceTests\SagaAudit @weralabaj
  * [x] ServiceControl.AcceptanceTests @weralabaj

[...]

PR #1296, opened 2018-07-23
Clean up ServiceControl.AcceptanceTests namespace

weralabaj:
https://github.com/Particular/ServiceControl/issues/1250


weralabaj:
@Particular/servicecontrol-maintainers Seems there's a subtle issue related to ServiceControlComponentRunner, line 75, where parameter hides a field. We can either rename the parameter or remove the field, but not sure what was the author's intention ;)

danielmarbach:
@weralabaj I fixed it. That is a leftover from code refactoring. Good catch

PR #1302, opened 2018-07-23
GetTransportType should properly deal with assembly qualified names

danielmarbach:


src/ServiceControl/Infrastructure/Settings/Settings.cs
```diff
@@ -332,7 +332,7 @@ static string GetTransportType()
         {
             var typeName = SettingsReader<string>.Read("TransportType", "ServiceControl.Transports.Msmq.MsmqTransportCustomization, ServiceControl.Transports.Msmq");
             var typeNameAndAssembly = typeName.Split(',');
-            if (typeNameAndAssembly.Length != 2)
+            if (typeNameAndAssembly.Length < 2)
```
danielmarbach: In reality this would almost always be 5 parts but we really only care about the first two.

PR #1307, opened 2018-07-24
.Net framework upgrade 4.6.1

### LINKED ISSUE #1295: Upgrade SC dependency to .NET Framework 4.6.1
- [x] Upgrade TargetFramework
- [x] Verify if installer is affected and change if necessary
- [x] Evaluate if AdvancedInstaller has to be changed
- [x] Smoke test in Win2012 R2 (ships without 4.6.1 installed) and Win2016 (has 4.6.1 installed)

WojcikMike:
fix #1295

Don't be misleaded by the branch name, it is actually upgrade to 4.6.1 ;)


WilliamBZA:
The advanced installer seems to need a tweak to point to 461 instead of 452

WilliamBZA:
Likely all the 452 references here (https://github.com/Particular/ServiceControl/blob/net_framework_upgrade_462/src/Setup/ServiceControl.aip) should be changed to 461

danielmarbach:
- ✔️ Installed Win2012R2 from Azure Marketplace and run the installer

image

- ✔️ Installed Win2016 from Azure Marketplace and run the installer

image

Also checked the reg keys described in 
https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed

Issue #1226, opened 2018-07-11
Review Async Code


WilliamBZA:
A major change was made to switch ServiceControl over to async code. A thorough review of the async code is required to make sure it is valid, and to suggest further refactorings that can be made to improve the performance (e.g. fanout in places)

timbussmann:
the change to async doesn't seem to have intended the usage of `ConfigureAwait(false)`, was that done on purpose or do we have to retrofit that? @WilliamBZA

WilliamBZA:
Seems like all the `ConfigureAwait`s have been added?

timbussmann:
I've had a conversation with @danielmarbach and we decided to add the missing ConfigureAwaits to the SC project. I forgot to update this info on this issue.

timbussmann:
* the particular code rules package has been added to ensure we're not missing any awaits or similar.
* ConfigureAwaits have been added to align with the general best-practice for our packages

I think there's not much more to actively pursue as part of this issue. I'm going to close this. If you feel there is a specific part of the code which needs additional review, please comment or create a new issue.

PR #1319, opened 2018-07-24
Async review

timbussmann:
relates to https://github.com/Particular/ServiceControl/issues/1226

removing some unnecessary `async`s and adding the code rules package to ensure all tasks are captured.

note: I've disabled the missing `ConfigureAwait` warning as there are a lot of them missing. If we want those, I'll create a dedicated PR.

### DIFF
--- a/src/ServiceControl/Infrastructure/Extensions/WaitHandleExtensions.cs
+++ b/src/ServiceControl/Infrastructure/Extensions/WaitHandleExtensions.cs
@@ -6,7 +6,7 @@ namespace ServiceBus.Management.Infrastructure.Extensions
 
     internal static class WaitHandleExtensions
     {
-        public static async Task<bool> WaitOneAsync(this WaitHandle handle, int millisecondsTimeout, CancellationToken cancellationToken)
+        public static Task<bool> WaitOneAsync(this WaitHandle handle, int millisecondsTimeout, CancellationToken cancellationToken)
         {
             RegisteredWaitHandle registeredHandle = null;
             var tokenRegistration = default(CancellationTokenRegistration);
@@ -22,7 +22,7 @@ namespace ServiceBus.Management.Infrastructure.Extensions
                 tokenRegistration = cancellationToken.Register(
                     state => ((TaskCompletionSource<bool>)state).TrySetCanceled(),
                     tcs);
-                return await tcs.Task;
+                return tcs.Task;
             }
             finally
             {
--- a/src/ServiceControl/Infrastructure/Extensions/WaitHandleExtensions.cs
+++ b/src/ServiceControl/Infrastructure/Extensions/WaitHandleExtensions.cs
@@ -6,7 +6,7 @@ namespace ServiceBus.Management.Infrastructure.Extensions
 
     internal static class WaitHandleExtensions
     {
-        public static Task<bool> WaitOneAsync(this WaitHandle handle, int millisecondsTimeout, CancellationToken cancellationToken)
+        public static async Task<bool> WaitOneAsync(this WaitHandle handle, int millisecondsTimeout, CancellationToken cancellationToken)
         {
             RegisteredWaitHandle registeredHandle = null;
             var tokenRegistration = default(CancellationTokenRegistration);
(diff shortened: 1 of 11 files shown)


src/ServiceControl/Infrastructure/Extensions/WaitHandleExtensions.cs
```diff
@@ -22,7 +22,7 @@ internal static class WaitHandleExtensions
                 tokenRegistration = cancellationToken.Register(
                     state => ((TaskCompletionSource<bool>)state).TrySetCanceled(),
                     tcs);
-                return await tcs.Task;
```
danielmarbach: This change has to be reverted because it changes the semantics of the code. Now the finally block is executed directly and registrations are disposed. We only want to execute the finally block after the task is done

src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSenderDequeuer.cs
```diff
@@ -45,37 +46,37 @@ public ReturnToSenderDequeuer(TransportDefinition transportDefinition, ReturnToS
...
-        void CountMessageAndStopIfReachedTarget()
+        Task CountMessageAndStopIfReachedTarget()
         {
             var currentMessageCount = Interlocked.Increment(ref actualMessageCount);
             Log.DebugFormat("Handling message {0} of {1}", currentMessageCount, targetMessageCount);
             if (currentMessageCount >= targetMessageCount.GetValueOrDefault())
             {
                 Log.DebugFormat("Target count reached. Shutting down forwarder");
-                // NOTE: This needs to run on a different thread or a deadlock will happen trying to shut down the receiver
-                Task.Run(StopInternal);
```
danielmarbach: The intent of this line was to preserve the previous behavior. The intent of the previous code was from a caller perspective make this a fire&forget operation and thus offloading from the current thread that is entering and triggering the stop. These changes now remove that and StopInternal is executed on the thread entering StopInternal which can cause issues (if the original comments still holds true). I would vote to not make those changes except for the Ignore introduction

src/ServiceControl/Infrastructure/Extensions/WaitHandleExtensions.cs
timbussmann: that would only be the case if you don't await `WaitOneAsync` though?

src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSenderDequeuer.cs
timbussmann: if there is something causing StopInternal to cause issues, then StopInternal should take care of the `Task.Run` (Which it already does)?

src/ServiceControl/Infrastructure/Extensions/WaitHandleExtensions.cs
danielmarbach: Nope. Similar code like

```
using(var disposable) {
  return await Foo();
}
```

vs

```
using(var disposable) {
  return Foo();
}
```

First example disposable is disposed after Foo is done. Second example disposable is disposed immediately after Foo reached internal continuation but might not be done yet.

timbussmann: ah, I didn't notice the finally block :( Reverting that change then.

src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSenderDequeuer.cs
danielmarbach: Right now it does not. It uses Task.Run to dispatch the TrySetResult of the TCS but then awaits that. The intent here is to offload the whole StopInternal call away from the caller entering the count check. Happy to explain synch if needed.

src/ServiceControl/Infrastructure/Extensions/WaitHandleExtensions.cs
danielmarbach: This is in combination with the other change below might cause that many test timeouts on the ATTs (see build server output)

Issue #1235, opened 2018-07-13
Investigate  When_a_retry_for_a_failed_message_fails.It_should_be_marked_as_unresolved for SQL Transport


danielmarbach:
` When_a_retry_for_a_failed_message_fails.It_should_be_marked_as_unresolved` seems to be flaky on SQL Transport.

Requires SQL Transport and SC skills

PR #1320, opened 2018-07-24
Do not publish `MessageFailed` on the bus

SzymonPobiega:
Fixes https://github.com/Particular/ServiceControl/issues/1235

Use DomainHandler to react on the `MessageFailed` instead of a message handler. Removes the race condition in some tests and removes one round unnecessary roundtrip.


weralabaj:
LGTM let's just wait for the tests, I fixed name in one place :)

danielmarbach:
@mikeminutillo @WilliamBZA It looks like this is safe to merge. External integration is not affected due to the MessageFailed removal since external integration hooks into the domain event publisher. Multi instance should also not be affected since we are only interested in the MessageResolvedByRetry. Can you confirm?

WilliamBZA:
@danielmarbach I'm really not sure 😞 

It would be great if we had a test to confirm this scenario.

PR #1324, opened 2018-07-25
Changed the audit generator to be based on a SC plugin for reporting …

SzymonPobiega:
Improved audit generator uses SC-hosted *enricher* that counts processed audit messages. The SC-hosted reporter task periodically (every 2 seconds) sends, to a configured queue, a message that contains last seen generator host ID and number of processed audits since the host ID has changed.

The generator uses a different host ID for each run. It receives the reports and based on the difference between the number of sent and processed messages estimates the audit queue length. It then ensures that this estimated length is between configured boundary values.

The generator uses SC transport customizations.

This change makes the load generator transport independent of the specific transport technology.


src/ServiceControl/Infrastructure/MarkerFile.cs
```diff
@@ -12,6 +12,7 @@ public MarkerFileService(string rootPath)
 
         public IDisposable CreateMarker(string name)
         {
+            Directory.CreateDirectory(rootPath);
```
danielmarbach: We discussed this before that we cannot do this during runtime. The assumption has to be that the directory got created by the installer.

src/ServiceControl.LoadTests.Reporter/ServiceControl.LoadTests.Reporter.csproj
```diff
@@ -1,17 +1,16 @@
...
   <PropertyGroup>
-    <TargetFramework>net462</TargetFramework>
-    <OutputType>Exe</OutputType>
-    <AssemblyName>QueueLengthMonitor</AssemblyName>
+    <TargetFramework>net461</TargetFramework>
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\ServiceControl.LoadTests.Messages\ServiceControl.LoadTests.Messages.csproj" />
+    <PackageReference Include="NServiceBus" Version="7.0.1" />
```
danielmarbach: ServiceControl references NSB. Transitive dependencies will resolve this so you can drop the package reference

src/ServiceControl.LoadTests.Reporter/Statistics.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace ServiceControl.LoadTests.Reporter
+{
+    using System.Threading;
+
+    class Statistics
+    {
+        long audits;
```
danielmarbach: So the assumption here is when we have messages in the queue and we restart the processing the hostid will be different. we might still be consuming old audit messages but once we see the new hostid we assume we caught up and we will reset the stats. Correct?

src/ServiceControl/Infrastructure/MarkerFile.cs
SzymonPobiega: Hmmm... but it breaks the `F5` experience :-(

src/ServiceControl.LoadTests.Reporter/Statistics.cs
SzymonPobiega: Yeah.

src/ServiceControl/Infrastructure/MarkerFile.cs
WilliamBZA: We can assume SC has write access to the log folder, what about creating the directory in there?

SzymonPobiega: hmmm that sounds good

danielmarbach: The directory IS the log folder and that's why we debated this at length in the past ;) and decided not to add that CreateDirectory. Or if we do for "debugging" purposes we have to properly guard it with try catch and potentially Directory.Exist

SzymonPobiega: 👍 for try-catch

PR #1326, opened 2018-07-25
WIP Att spike

danielmarbach:
- Configures a singleton instance for SC and then throws it away if a specific custom configuration is required such as multiple-instances or custom config.
- Once a customized instance was acquired the next run will create a new and fresh instance

### DIFF
--- a/src/ServiceControl.AcceptanceTests/AcceptanceTest.cs
+++ b/src/ServiceControl.AcceptanceTests/AcceptanceTest.cs
@@ -8,7 +8,6 @@ namespace ServiceBus.Management.AcceptanceTests
     using System.Linq;
     using System.Net;
     using System.Net.Http;
-    using System.Reflection;
     using System.Threading;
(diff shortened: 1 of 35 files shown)


danielmarbach:
@Particular/servicecontrol-maintainers Is this worth pursuing to bring down the test times? When it works it can bring down the test execution time significantly but it has a few risks attached to it and things that would need to change. For example the following challenges need to be solved

- We can only access the context over StaticLoggerFactory.Context because that is a fresh instance for every test. Side effects might occurs
- Registrations on that singleton instance become static and we can no longer inject a derived type of ScenarioContext (because that changes for every test run). So all occurences of MySpecializedContext would need to be replaced by ScenarioContext base class and then casted at runtime. (Currently many tests are failing because this is not yet solved)

danielmarbach:
We decided to close this spike. The additional complexity it introduces and state sharing is not worth taking the risk

PR #1327, opened 2018-07-26
Remove EnableDTC setting

WilliamBZA:


SzymonPobiega:
How about throwing an exception here telling users to see (not yet created) documentation on how to migrate to the transport adapter? We already have a sample for TA in multi-instance scenario. We just need to reference it in SC 2->3 upgrade guide.

The rationale for throwing is that people might forget that they need to change something in order for their multi-instance topology to continue working.

WilliamBZA:
Added an error to the log (I don't like the exception idea very much, but can still be convinced) and changed all the transports that support SendsWithAtomicReceive to use it

Issue #1328, opened 2018-07-26
System.FormatException: Input string was not in a correct format when cancelling the configuration


danielmarbach:
Steps to repro

Happens only on develop

1. Open the config, select a v3 instance, click on the configuration icon
2. cancel the configuration changes

```
System.FormatException: Input string was not in a correct format.
   at System.Number.StringToNumber(String str, NumberStyles options, NumberBuffer& number, NumberFormatInfo info, Boolean parseDecimal)
   at System.Number.ParseInt32(String s, NumberStyles style, NumberFormatInfo info)
   at System.Convert.ToInt32(String value)
   at ServiceControl.Config.UI.InstanceEdit.ServiceControlEditViewModel.UpdateInstanceFromViewModel(ServiceControlInstance instance)
   at ServiceControl.Config.Commands.EditServiceControlInstanceCommand.Execute(InstanceDetailsViewModel viewModel)
   at ServiceControl.Config.Framework.Commands.AbstractCommand`1.ServiceControl.Config.Framework.Commands.ICommand<T>.Execute(T obj)
   at ServiceControl.Config.Framework.Commands.AbstractCommand`1.System.Windows.Input.ICommand.Execute(Object parameter)
   at MS.Internal.Commands.CommandHelpers.CriticalExecuteCommandSource(ICommandSource commandSource, Boolean userInitiated)
   at System.Windows.Controls.Primitives.ButtonBase.OnClick()
   at System.Windows.Controls.Button.OnClick()
   at System.Windows.Controls.Primitives.ButtonBase.OnMouseLeftButtonUp(MouseButtonEventArgs e)
   at System.Windows.RoutedEventArgs.InvokeHandler(Delegate handler, Object target)
   at System.Windows.RoutedEventHandlerInfo.InvokeHandler(Object target, RoutedEventArgs routedEventArgs)
   at System.Windows.EventRoute.InvokeHandlersImpl(Object source, RoutedEventArgs args, Boolean reRaised)
   at System.Windows.UIElement.ReRaiseEventAs(DependencyObject sender, RoutedEventArgs args, RoutedEvent newEvent)
   at System.Windows.UIElement.OnMouseUpThunk(Object sender, MouseButtonEventArgs e)
   at System.Windows.RoutedEventArgs.InvokeHandler(Delegate handler, Object target)
   at System.Windows.RoutedEventHandlerInfo.InvokeHandler(Object target, RoutedEventArgs routedEventArgs)
[...]

PR #1332, opened 2018-07-27
Only update VM on Save

WilliamBZA:
Fixes https://github.com/Particular/ServiceControl/issues/1328


src/ServiceControl.Config/Commands/EditServiceControlInstanceCommand.cs
```diff
@@ -22,11 +22,11 @@ public override void Execute(InstanceDetailsViewModel viewModel)
         {
             var editVM = editViewModel((ServiceControlInstance)viewModel.ServiceInstance);
 
-            windowManager.ShowInnerDialog(editVM);
-
-            editVM.UpdateInstanceFromViewModel((ServiceControlInstance)viewModel.ServiceInstance);
```
SzymonPobiega: Is inverting the order of updating VM and publishing intentional?

PR #1333, opened 2018-07-27
Load testing improvements

SzymonPobiega:
* Get rid of SC reference in generator
 * Add Metrics-based reporting in reporter


src/ServiceControl.Transports.SqlServer/SqlServerTransportCustomization.cs
```diff
@@ -11,7 +11,7 @@ public override void CustomizeEndpoint(EndpointConfiguration endpointConfig, Tra
             var transport = endpointConfig.UseTransport<SqlServerTransport>();
             transport.ConnectionString(transportSettings.ConnectionString);
 
-            if (transportSettings.Get<bool>("TransportSettings.EnableDtc"))
+            if (transportSettings.GetOrDefault<bool>("TransportSettings.EnableDtc"))
```
WilliamBZA: Good catch!

Issue #1317, opened 2018-07-24
Installer: Deprecate usages of AzureStorageQueueConfig


danielmarbach:
The v5 version of AzureStorageQueue has the possibility to define config sections in the app.config of type AzureStorageQueueConfig. Customers are using that to tweak the behavior of the transport. When #1309 is introduced we need to deprecate those settings during the upgrade process of an existing instance.

WilliamBZA:
I have gone back and forth on whether or not an installer change or a warn in the log file is more appropriate. I'm currently leaning towards log file because someone can add the settings after an upgrade because of some internal documentation they might have.

PR #1335, opened 2018-07-30
Add log warning if SC detects if any ConfigSection's are being used

WilliamBZA:
Fixes https://github.com/Particular/ServiceControl/issues/1317

### DIFF
--- a/src/ServiceControl/Infrastructure/Settings/DeprecatedConfigurationCheckFeature.cs
+++ b/src/ServiceControl/Infrastructure/Settings/DeprecatedConfigurationCheckFeature.cs
@@ -7,16 +7,16 @@
     using NServiceBus.Features;
     using NServiceBus.Logging;
 
-    class InvalidConfigurationCheckFeature : Feature
+    class DeprecatedConfigurationCheckFeature : Feature
     {
-        public InvalidConfigurationCheckFeature()
+        public DeprecatedConfigurationCheckFeature()
         {
             EnableByDefault();
         }
 
-        protected override void Setup(FeatureConfigurationContext context) => context.RegisterStartupTask(b => b.Build<InvalidConfigurationCheck>());
+        protected override void Setup(FeatureConfigurationContext context) => context.RegisterStartupTask(b => b.Build<DeprecatedConfigurationCheck>());
 
-        class InvalidConfigurationCheck : FeatureStartupTask
+        class DeprecatedConfigurationCheck : FeatureStartupTask
         {
             protected override Task OnStart(IMessageSession session)
             {
@@ -55,7 +55,7 @@
                 return Task.CompletedTask;
             }
 
-            ILog logging = LogManager.GetLogger(typeof(InvalidConfigurationCheck));
+            ILog logging = LogManager.GetLogger(typeof(DeprecatedConfigurationCheck));
         }
     }
 }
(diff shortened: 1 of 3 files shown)


WilliamBZA:
Does an XML check to avoid having to have the assembly in the dir to be able to load the configSection

src/ServiceControl/ServiceControl.csproj
```diff
@@ -10,6 +10,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\ServiceControl.Transports.Msmq\ServiceControl.Transports.Msmq.csproj" />
```
SzymonPobiega: is this intentional?

WilliamBZA: I could do it as a separate PR, but it fixes the F5 experience

SzymonPobiega: Ok. I have no problem with this change as long as does not affect the packaging etc.

timbussmann: I don't know anything about SC packaging, but this change causes the msmq dlls to appear in the bin folder.

WilliamBZA: Hmm, I thought the other transports were removed during packaging, but that's only in the acceptance tests. I'll revert this commit for now and come up with a batter way to solve this

WilliamBZA:
Reverted and rebased

src/ServiceControl/Infrastructure/Settings/InvalidConfigurationCheckFeature.cs
```diff
@@ -0,0 +1,45 @@
...
+                    logging.Warn("The use of AzureQueueConfig within ServiceControl has been deprecated. A Transport Adapater (https://docs.particular.net/servicecontrol/transport-adapter/) can be used if there are transport defaults that are not suitable for this environment.");
+                }
+            }
+
+            protected override Task OnStop(IMessageSession session)
+            {
+                return Task.CompletedTask;
+            }
+
+            ILog logging = LogManager.GetLogger(typeof(InvalidConfigurationCheck));
```
timbussmann: not a big deal. but I wonder about the naming. Is this invalid or rather a deprecated configuration?

WilliamBZA: I can get on board with a better name. I'll rename

src/ServiceControl/Infrastructure/Settings/DeprecatedConfigurationCheckFeature.cs
```diff
@@ -0,0 +1,61 @@
...
+            EnableByDefault();
+        }
+
+        protected override void Setup(FeatureConfigurationContext context) => context.RegisterStartupTask(b => b.Build<DeprecatedConfigurationCheck>());
+
+        class DeprecatedConfigurationCheck : FeatureStartupTask
+        {
+            protected override Task OnStart(IMessageSession session)
+            {
+                var config = XDocument.Load(AppDomain.CurrentDomain.SetupInformation.ConfigurationFile);
```
timbussmann: is there any chance for that code to throw an exception we need to catch?
e.g. can the configuration file path be null or empty?

WilliamBZA: If there is no app.config file, ServiceControl won't start as there are config settings that are required. If all of the requried settings reside in the machine.config, then this code will return Task.CompletedTask and the user won't get warned about the config sections. [Docs for this property](https://msdn.microsoft.com/en-us/library/system.appdomainsetup.configurationfile.aspx)

If this code can't load the location specified by this property, it means the account that ServiceControl is running under doesn't have access to that disk location, in which case it shouldn't start anyway - at the very least, the codepath wouldn't actually get this far anyway.

So I don't know if there's any specific checks we need to do here, but I suspect not.

Issue #1316, opened 2018-07-24
Installer: Deprecate usages of AzureServiceBusConfig


danielmarbach:
The v5 version of AzureServiceBus has the possibility to define config sections in the app.config of type AzureServiceBusConfig. Customers are using that to tweak the behavior of the transport. When #1308 is introduced we need to deprecate those settings during the upgrade process of an existing instance. 

The `AzureServiceBusConfig` ConfigSection will no longer be supported, and has to be replaced by #1308. As part of this replacement, ServiceControl should log a WARN at startup if it detects that the app.config file contains a ConfigSection for Azure Service Bus to let users know that they should not be using that anymore. This could probably take the shape of a StartupTask that does a simple text search within the app.config file.

WilliamBZA:
Closing as this has been done

PR #1337, opened 2018-07-30
Warn when using ASB config sections

timbussmann:
relates to https://github.com/Particular/ServiceControl/issues/1316


timbussmann:
@WilliamBZA I've used your PR as a base. Maybe it makes sense to bundle all these related issues into a single PR?

PR #1339, opened 2018-07-31
Add defaults to ASQ transport

WilliamBZA:
Fixes https://github.com/Particular/ServiceControl/issues/1309

Adds the defaults identified by @yvesgoeleven in #1309

### DIFF
--- a/src/ServiceControl.Transports.ASQ/ASQTransportCustomization.cs
+++ b/src/ServiceControl.Transports.ASQ/ASQTransportCustomization.cs
@@ -2,6 +2,7 @@
 {
     using NServiceBus;
     using NServiceBus.Raw;
+    using System;
 
     public class ASQTransportCustomization : TransportCustomization
     {
@@ -23,6 +24,10 @@
             transport.SanitizeQueueNamesWith(BackwardsCompatibleQueueNameSanitizer.Sanitize);
             transport.Transactions(TransportTransactionMode.ReceiveOnly);
             transport.ConnectionString(transportSettings.ConnectionString);
+
+            transport.MessageInvisibleTime(TimeSpan.FromMinutes(5));
+            transport.BatchSize(1);
+            transport.DegreeOfReceiveParallelism((int)Math.Sqrt(Math.Min(Environment.ProcessorCount, transportSettings.MaxConcurrency)));
         }
     }
 }
--- a/src/ServiceControl.Transports.ASQ/ASQTransportCustomization.cs
+++ b/src/ServiceControl.Transports.ASQ/ASQTransportCustomization.cs
@@ -25,9 +25,8 @@
             transport.Transactions(TransportTransactionMode.ReceiveOnly);
             transport.ConnectionString(transportSettings.ConnectionString);
 
-            transport.MessageInvisibleTime(TimeSpan.FromMinutes(5));
-            transport.BatchSize(1);
-            transport.DegreeOfReceiveParallelism((int)Math.Sqrt(Math.Min(Environment.ProcessorCount, transportSettings.MaxConcurrency)));
+            transport.MessageInvisibleTime(TimeSpan.FromMinutes(1));
+            transport.DegreeOfReceiveParallelism((int)Math.Sqrt(transportSettings.MaxConcurrency));
         }
     }
 }
--- a/src/ServiceControl.Transports.ASQ/ASQTransportCustomization.cs
+++ b/src/ServiceControl.Transports.ASQ/ASQTransportCustomization.cs
@@ -26,7 +26,6 @@
             transport.ConnectionString(transportSettings.ConnectionString);
 
             transport.MessageInvisibleTime(TimeSpan.FromMinutes(1));
-            transport.DegreeOfReceiveParallelism((int)Math.Sqrt(transportSettings.MaxConcurrency));
         }
     }
 }
--- a/src/ServiceControl.Transports/TransportSettings.cs
+++ b/src/ServiceControl.Transports/TransportSettings.cs
@@ -7,5 +7,7 @@
         public string ConnectionString { get; set; }
 
         public string EndpointName { get; set; }
+
+        public int MaxConcurrency { get; set; }
     }
 }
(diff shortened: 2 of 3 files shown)


yvesgoeleven:
Looks good to me, but would like to have @Scooletz or @SeanFeldman to review it as well

src/ServiceControl.Transports.ASQ/ASQTransportCustomization.cs
```diff
@@ -23,6 +24,10 @@ static void ConfigureTransport(TransportExtensions<AzureStorageQueueTransport> t
             transport.SanitizeQueueNamesWith(BackwardsCompatibleQueueNameSanitizer.Sanitize);
             transport.Transactions(TransportTransactionMode.ReceiveOnly);
             transport.ConnectionString(transportSettings.ConnectionString);
+
+            transport.MessageInvisibleTime(TimeSpan.FromMinutes(5));
+            transport.BatchSize(1);
+            transport.DegreeOfReceiveParallelism((int)Math.Sqrt(Math.Min(Environment.ProcessorCount, transportSettings.MaxConcurrency)));
```
mikeminutillo: Is that right? If we have 16 processors then we can only have Receive Parallelism of 4?

SeanFeldman: [Doco](https://docs.particular.net/transports/azure-storage-queues/configuration#configuration-parameters-degreeofreceiveparallelism) says `sqrt(maxConcurrency)`. And concurrency is `max(Number of logical processors, 2)`. Therefore, shouldn't it be `sqrt(settings.Concurrency)` only?

src/ServiceControl.Transports.ASQ/ASQTransportCustomization.cs
```diff
@@ -23,6 +24,10 @@ static void ConfigureTransport(TransportExtensions<AzureStorageQueueTransport> t
             transport.SanitizeQueueNamesWith(BackwardsCompatibleQueueNameSanitizer.Sanitize);
             transport.Transactions(TransportTransactionMode.ReceiveOnly);
             transport.ConnectionString(transportSettings.ConnectionString);
+
+            transport.MessageInvisibleTime(TimeSpan.FromMinutes(5));
+            transport.BatchSize(1);
```
SeanFeldman: Batch size of one?

WilliamBZA: I'm not sure if it's applicable with ASQ, but we have had support cases (most recently with H&M) where they had to set the batch size for ASB to 1 in order to prevent Lock Lost exceptions. Daniel suggested it has something to do with prefetching and a bunch of other words I didn't understand.

I don't know what the correct values are here 😞

SzymonPobiega: I guess I can run a test with batch size set to 1 to verify the impact

Scooletz: I don't understand setting this to `1`. In ASQ a message could reappear (as there are no locks) but what would take more than the configured 5min of  the  timeout invisibility?

yvesgoeleven: As the issue description says, it is expected for a handler to take several seconds, in that case it's a good practice to dial down the batch size.

@WilliamBZA correct values do not exist, for optimal performance it needs testing and adjustments to all the variables

SzymonPobiega: So there is no danger of messages going to the DLQ?

SeanFeldman: ASQ has no DLQ. Messages will exhaust all the retries and will be sent to the Error queue.

boblangley: After talking with @SeanFeldman, larger batch sizes lower the number of requests to Azure Storage, which is a key billing metric.

The concern with a large batch size is timeouts. If the last message in the batch isn't processed before the timeout, another thread can pick it up.

Based on @SzymonPobiega's early testing of V3 I think default batch size is probably fine, and maybe bumping up the default timeout to 60 seconds would give us plenty of breathing room for those systems that are at the slower end of the spectrum, while allowing "abandoned" messages to reappear in the queue before or shortly after ServiceControl has restarted.

SzymonPobiega: I can always retest. Our goals here are, in this order:
 - Minimize the chances of moving messages to the error queue (because SC has no good support for moving them back)
 - Have acceptable throughput at least as high as the max processing rate we've seen in Azure (160 msg/s)
 - Minimize the cost

src/ServiceControl.Transports.ASQ/ASQTransportCustomization.cs
SzymonPobiega: If we don't set anything the default is `sqrt(settings.Concurrency)` so I vote for removing this line. Let's stick with whatever transport default it.

SzymonPobiega:
To summarize my comments, I'd leave the `DegreeOfReceiveParallelism` and `BatchSize` to transport defaults and only bump the `MessageInvisibleTime` as the default value of 30 seconds might be to small for slow IO beast such as SC

src/ServiceControl.Transports.ASQ/ASQTransportCustomization.cs
```diff
@@ -25,9 +25,8 @@ static void ConfigureTransport(TransportExtensions<AzureStorageQueueTransport> t
             transport.Transactions(TransportTransactionMode.ReceiveOnly);
             transport.ConnectionString(transportSettings.ConnectionString);
 
-            transport.MessageInvisibleTime(TimeSpan.FromMinutes(5));
-            transport.BatchSize(1);
-            transport.DegreeOfReceiveParallelism((int)Math.Sqrt(Math.Min(Environment.ProcessorCount, transportSettings.MaxConcurrency)));
+            transport.MessageInvisibleTime(TimeSpan.FromMinutes(1));
+            transport.DegreeOfReceiveParallelism((int)Math.Sqrt(transportSettings.MaxConcurrency));
```
SzymonPobiega: This is the transport default so I think we should not duplicate it here.

PR #1340, opened 2018-08-03
Stop using metadata in audit cleaner

SzymonPobiega:
Add a computed field to the index instead


mikeminutillo:
Does this significantly improve anything? Making a change to an index will cause a re-index at startup.

Issue #1341, opened 2018-08-06
Ensure we didn't break mixed version Master-Slave configuration in the V3 upgrade


mikeminutillo:
## Context

As a part of making audit processing more scalable we introduced a [scale-out mode for ServiceControl](https://docs.particular.net/servicecontrol/servicecontrol-instances/distributed-instances). This mode allows users to run multiple ServiceControl instances which present a single audit data store to tools like ServiceInsight. 

As a part of the upgrade process, customers will need to upgrade one instance at a time to V3 and potentially run in mixed mode where some instances are v3 and some are not.

We need to ensure that ServiceControl works properly when deployed in a mixed version configuration like this.

## Out of Scope

- Multi-instance deployments where all instances are running on v3 (this is covered by the Acceptance Tests)

## Plan of attack

- [x] Document the features of a Multi-Instance deployment (Audit message aggregation, Saga Audit aggregation, Endpoint detection aggregation, successful retry detection, ...) https://github.com/Particular/docs.particular.net/pull/3953
- [x] Decide on the best strategy to test these (automated/manual/etc.) -> manual
- [x] Brainstorm the configurations that need to be tested (Master V3/Slave V2, Master V2/Slave V3, etc.)
- [x] Document the scenarios to be tested (error ingested into master - retried - audited into slave - detected successful retry in master)
- [x] Run test scenarios
- [ ] Document results

## Required Knowledge

- Knowledge of ServiceControl Multi-instance mode
- Nothing else

This is a good opportunity for someone without much SC knowledge to pair with someone who does in order to learn

SzymonPobiega:
I've started the part that really required deep SC knowledge and now it makes sense to recruit somebody who would like to learn more about how multi-instance SC works.

mauroservienti:
I’m willing to join this one

seanfarmar:
I’d like to join this one

SzymonPobiega:
@seanfarmar I am afraid we've already staffed the TF with @Scooletz and @mauroservienti :-(

mauroservienti:
took the liberty of removing the `Needs taskforce` label, as the task force is formed.

SzymonPobiega:
After reviewing the code we decided to try a hybrid manual/automatic approach where we use the existing automated test cases but we start with manual setup procedure that involves installing on the test machine all required versions of ServiceControl.

The PR that refactors the SC acceptance tests so that they can run against an external instance of SC is in progress.

[...]

PR #1348, opened 2018-08-16
Create multiversion-multiinstance-smoke-test.md

boblangley:
Connects to #1341

### DIFF
--- a/docs/multiversion-multiinstance-smoke-test.md
+++ b/docs/multiversion-multiinstance-smoke-test.md
@@ -1,0 +1,53 @@
+I am not sure a PR is necessary:
+
+## Setup
+
+1. Install ServiceControl version 2.
+1. Add 1 instance of ServiceControl v2 with MSMQ. It would help to use a name relating it to v3, e.g. `v2_SC`
+   - Configure a unique error queue name. It would help to use a name relating it to v2, e.g. `v2_error`
+   - Configure a unique audit queue name. It would help to use a name relating it to v2, e.g. `v2_audit`
+1. Install ServiceControl version 3
+1. Add 1 instance of ServiceControl v3 with MSMQ. It would help to use a name relating it to v3, e.g. `v3_SC`
+   - Configure a unique error queue name. It would help to use a name relating it to v3, e.g. `v3_error`
+   - Configure a unique audit queue name. It would help to use a name relating it to v3, e.g. `v3_audit`
+1. Download or checkout the [FaultTolerance](https://docs.particular.net/samples/faulttolerance/) sample.
+1. Edit the configuration of the sample project
+   - Reconfigure the endpoint to use the MSMQ transport
+   
+## Scenario 1   
+1. Configure ServiceControl v3 as Remote.
+1. Restart v3 SC
+1. Configure ServiceControl v2 as Master. 
+1. Restart v2 SC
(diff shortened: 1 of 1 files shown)


docs/multiversion-multiinstance-smoke-test.md
```diff
@@ -0,0 +1,53 @@
...
+1. Restart v2 SC
+1. Edit the configuration of the sample project
+   - Configure the error queue to the unique error queue assigned to v2 instance of SC
+   - Configure auditing to the unique audit queue assigned to v3 instance of SC.
+1. Run the sample, sending error messages to the v2 instance of SC
+1. Set the sample to successfully process messages
+1. Connect ServiceInsight to the v2 instance of ServiceControl
+1. Retry one or more failed messages in ServiceInsight
+1. Confirm the message(s) successfully processed in the sample
+1. Confirm the message status is Resolved in ServiceInsight
```
boblangley: I am actually not sure this is correct, or if it just looks like an audit message now. I do not it will show initially as a Failed message, and then Retry Pending after a retry is issue. I just didn't have time to test it today.

mikeminutillo: Correct. In SI it will appear as a regular audited message. I wonder if we should raise an issue to change this. It could be useful to know that a message succeeded after being retried. Then again, the same could be said of Immediate and Delayed Retries.

The real issue is whether the SC instance thinks the failed message is resolved. You might need to resort to querying the API directly to determine this.

danielmarbach: Interesting I was assuming we'd be using SP foe this but SI certainly works as well

docs/multiversion-multiinstance-smoke-test.md
```diff
@@ -0,0 +1,53 @@
...
+   - Configure a unique audit queue name. It would help to use a name relating it to v2, e.g. `v2_audit`
+1. Install ServiceControl version 3
+1. Add 1 instance of ServiceControl v3 with MSMQ. It would help to use a name relating it to v3, e.g. `v3_SC`
+   - Configure a unique error queue name. It would help to use a name relating it to v3, e.g. `v3_error`
+   - Configure a unique audit queue name. It would help to use a name relating it to v3, e.g. `v3_audit`
+1. Download or checkout the [FaultTolerance](https://docs.particular.net/samples/faulttolerance/) sample.
+1. Edit the configuration of the sample project
+   - Reconfigure the endpoint to use the MSMQ transport
+   
+## Scenario 1
```
SzymonPobiega: Should we label these scenarios somehow? i.e. "V2 remote notifies V3 master about successful retry"

docs/multiversion-multiinstance-smoke-test.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Reset 
+1. Stop the v2 instance of SC
+1. Remove the database directory of the v2 instance of SC
+1. Start the v2 instance of SC
+1. Stop the v3 instance of SC
+1. Remove the database directory of the v3 instance of SC
+1. Start the v3 instance of SC
+
+## Scenario 2
```
SzymonPobiega: Should we label these scenarios somehow? e.g. "V3 remote notifies V2 master about successful retry"

PR #1351, opened 2018-08-17
MessageForwarder resets TTBR if present

danielmarbach:
The previous version did a reset of the TTBR, see

https://github.com/Particular/ServiceControl/blob/33b4e74e7357e3daef5f16ebe80dbc56f3a7b461/src/ServiceControl/Operations/ErrorQueueImport.cs#L160
and
https://github.com/Particular/ServiceControl/blob/7f46be452a66877eca8c960e545c46dee40c16f3/src/ServiceControl/Operations/AuditQueueImport.cs#L105

See documented core behavior https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Headers.cs#L221-L222

This does it now consistently in the forwarder and removes the TODO

### DIFF
--- a/src/ServiceControl.UnitTests/Operations/MessageForwarderTests.cs
+++ b/src/ServiceControl.UnitTests/Operations/MessageForwarderTests.cs
@@ -1,0 +1,51 @@
+﻿namespace ServiceControl.UnitTests.Operations
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using NServiceBus;
+    using NServiceBus.Extensibility;
+    using NServiceBus.Transport;
+    using NUnit.Framework;
+    using ServiceControl.Operations;
+
+    [TestFixture]
+    public class MessageForwarderTests
+    {
+        FakeDispatcher dispatcher;
+        MessageForwarder forwarder;
+
+        [SetUp]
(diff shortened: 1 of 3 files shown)


danielmarbach:
The previous code always set the TimeToBeReceived property which did not set a header but was used by the transport. So effectively we can say the TimeToBeReceived was always set to `TimeSpan.Max`. The question is should we always set the header as well and therefore even remove the `if(headers.ContainsKey())` part?

danielmarbach:
Given that forwarded messages should last as long as possible and the previous version did always set the TimeToBeReceived property I'm now always setting the TTBR header since this is the replacement in the modern core for TimeToBeReceived property.

mikeminutillo:
I think the previous implementation set the property to MAX because it was not nullable. We don't get a TTBR property on incoming messages anymore. Shouldn't we just remove the header, if present?

danielmarbach:
That is one way to look at it. Another perspective is to look at the
semantics that was there. Setting it to maxvalue means "lives forever".
With the property beikg replaced by TTBR header the semantic replacement
woulf be to set the header to maxvalue. If no TTBR header is available it
means "lives forever" we can definitely remove the header if present

danielmarbach:
Update the PR to remove the header if present. @Particular/servicecontrol-maintainers please review and merge

Issue #1359, opened 2018-08-20
[TODO] FailedMessagePersister - fan out enrichers?


SzymonPobiega:

SzymonPobiega:
Similar to https://github.com/Particular/ServiceControl/issues/1356 I guess we should not do it because of thread safety.

danielmarbach:
I agree especially because there is an implicit assumption of execution order over some shared state. I'll send in PR which replaces the TODO with a comment

danielmarbach:
I double checked [what @WilliamBZA said](https://github.com/Particular/ServiceControl/issues/1356#issuecomment-414283780) and can confirm they use all their unique keys. Sent in https://github.com/Particular/ServiceControl/pull/1361 to fix this TODO

danielmarbach:
Merged the PR. Closing

Issue #1356, opened 2018-08-20
[TODO] AuditImporter - fan out enrichers


SzymonPobiega:

SzymonPobiega:
I am not sure this is a good idea since enrichers operate on the metadata dictionary which is not thread-safe. @Particular/servicecontrol-maintainers thoughts?

WilliamBZA:
The written object is the `metadata` `IDictionary`, and all of the enrichers seem to write their own values instead of duplicating others (they use the `Add` method which would throw if there were already a key with the same value).

I think it would be ok to fan this out. It would be better to make the concurrent nature of this explicit though by using a concurrent dictionary.

WilliamBZA:
Also, not necessarily required for a 3.0.0 release

danielmarbach:
I double checked what @WilliamBZA said and can confirm they use all their unique keys. Sent in https://github.com/Particular/ServiceControl/pull/1361 to fix this TODO

danielmarbach:
Merge the PR. Closing

PR #1361, opened 2018-08-20
Fanout enrichers

danielmarbach:
Would address #1356 and #1359

### DIFF
--- a/src/ServiceControl/Operations/AuditImporter.cs
+++ b/src/ServiceControl/Operations/AuditImporter.cs
@@ -1,6 +1,7 @@
 ﻿namespace ServiceControl.Operations
 {
     using System;
+    using System.Collections.Concurrent;
     using System.Collections.Generic;
     using System.IO;
     using System.Linq;
@@ -104,24 +105,27 @@
                 messageId = DeterministicGuid.MakeId(message.MessageId).ToString();
             }
 
-            var metadata = new Dictionary<string, object>
+            var metadata = new ConcurrentDictionary<string, object>
             {
                 ["MessageId"] = messageId,
                 ["MessageIntent"] = message.Headers.MessageIntent(),
                 ["HeadersForSearching"] = string.Join(" ", message.Headers.Values)
             };
 
-            // TODO: Fan out?
+            var enricherTasks = new List<Task>(enrichers.Length);
+            // ReSharper disable once LoopCanBeConvertedToQuery
             foreach (var enricher in enrichers)
             {
-                await enricher.Enrich(message.Headers, metadata)
-                    .ConfigureAwait(false);
+                enricherTasks.Add(enricher.Enrich(message.Headers, metadata));
             }
+
+            await Task.WhenAll(enricherTasks)
+                .ConfigureAwait(false);
 
             await bodyStorageEnricher.StoreAuditMessageBody(message.Body, message.Headers, metadata)
                 .ConfigureAwait(false);
 
-            var auditMessage = new ProcessedMessage(message.Headers, metadata)
+            var auditMessage = new ProcessedMessage(message.Headers, new Dictionary<string, object>(metadata))
             {
                 // We do this so Raven does not spend time assigning a hilo key
                 Id = $"ProcessedMessages/{Guid.NewGuid()}"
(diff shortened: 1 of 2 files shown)


src/ServiceControl/Operations/AuditImporter.cs
```diff
@@ -104,24 +105,27 @@ public async Task<ProcessedMessage> ConvertToSaveMessage(MessageContext message)
...
             }
 
+            await Task.WhenAll(enricherTasks)
+                .ConfigureAwait(false);
+
             await bodyStorageEnricher.StoreAuditMessageBody(message.Body, message.Headers, metadata)
                 .ConfigureAwait(false);
 
-            var auditMessage = new ProcessedMessage(message.Headers, metadata)
+            var auditMessage = new ProcessedMessage(message.Headers, new Dictionary<string, object>(metadata))
```
danielmarbach: This would be the consequence but I think this is an OK price to pay given that we can fan out the enrichers which do domain event raise and other IO bound stuff?

src/ServiceControl/Operations/FailedMessagePersister.cs
```diff
@@ -46,28 +47,31 @@ public async Task<FailureDetails> Persist(MessageContext message)
...
+
             await bodyStorageEnricher.StoreErrorMessageBody(message.Body, message.Headers, metadata)
                 .ConfigureAwait(false);
 
             var failureDetails = failedMessageFactory.ParseFailureDetails(message.Headers);
 
             var processingAttempt = failedMessageFactory.CreateProcessingAttempt(
                 message.Headers,
-                metadata,
+                new Dictionary<string, object>(metadata),
```
danielmarbach: This would be the consequence but I think this is an OK price to pay given that we can fan out the enrichers which do domain event raise and other IO bound stuff?

mikeminutillo:
Given that we are already processing multiple audit messages in parallel, is there much to be gained here or will we run into task starvation?

I guess that reducing the non-Esent latency for each message can lead to greater throughput but do we know how much?

SzymonPobiega:
I don't like passing the concurrent dictionary to the enrichers because it would be hard to detect if somebody decides to make two enrichers communicate via metadata.

I'd rather pass an `Action<string, object>` that adds an item to the metadata to the enrichers to enforce that the metadata at this stage is write-only.

mikeminutillo:
> I don't like passing the concurrent dictionary to the enrichers because it would be hard to detect if somebody decides to make two enrichers communicate via metadata.
>
>I'd rather pass an Action<string, object> that adds an item to the metadata to the enrichers to enforce that the metadata at this stage is write-only.

That's not bad. The original (original) version had each enricher returning a dictionary and then the calling class aggregated them into a single dictionary. We moved to passing the dictionary in as an optimization when we reviewed that original implementation.

src/ServiceControl/Operations/AuditImporter.cs
```diff
@@ -104,24 +105,27 @@ public async Task<ProcessedMessage> ConvertToSaveMessage(MessageContext message)
...
-            var metadata = new Dictionary<string, object>
+            var metadata = new ConcurrentDictionary<string, object>
             {
                 ["MessageId"] = messageId,
                 ["MessageIntent"] = message.Headers.MessageIntent(),
                 ["HeadersForSearching"] = string.Join(" ", message.Headers.Values)
             };
 
-            // TODO: Fan out?
+            var enricherTasks = new List<Task>(enrichers.Length);
```
WilliamBZA: Is there a benefit to doing it this way instead of 

```CSharp
var enricherTasks = enrichers.Select(enricher => enricher.Enrich(message.Headers, metadata));

await Task.WhenAll(enricherTasks).ConfigureAwait(false);
```

WilliamBZA: Also, don't we need to add `ConfigureAwait(false)` after calling `enricher.Enrich(...)`?

danielmarbach: Yes. One doesn't use LINQ which is good in terms of perf. Plus the LINQ version materializes on WhenAll while the list version materializes and thus schedules the tasks for execution when they are added

A task not awaited doesn't need to be configured with `ConfigureAwait`

Issue #1360, opened 2018-08-20
[TODO] Processing attempt - do we need to persist intent, correlation ID and reply-to at the document level


SzymonPobiega:

SzymonPobiega:
Reply-to seems to be unused so I am 👍 for removing it. The other properties are used when staging a message in the `RetryProcessor`. I'd rather keep them then try to get these values from the headers.

SzymonPobiega:
@Particular/servicecontrol-maintainers thoughts?

WilliamBZA:
`ReplyToAddress` seems to be only used by the v1.27.0 migration. Which is not applicable anymore. Even if we have to support a v1 - v3 scenario. 👍 for removing `ReplyToAddress`.

This would change the storage format used though. What does removing stored properties do to backwards compatibility I wonder.

PR #1362, opened 2018-08-20
Remove ReplyToAddress and Migration tests since we force an upgrade to 2.1.x latest

danielmarbach:
Would address #1360


mikeminutillo:
I wonder if we can remove some of the Migrations now too.
If you are on V2 then you've either run those migrations or you don't need to because you were never on an affected version.

mikeminutillo:
Do we still need the property `FailedMessage.ProcessingAttempt.RetryAddress`?
If we drop it, does the storage just ignore it when it deserializes?

WilliamBZA:
> I wonder if we can remove some of the Migrations now too.

I am 💯 behind doing so

danielmarbach:
@mikeminutillo RavenDB is tolerant when reading records. So effectively old documents will be read without that property but not modified

WilliamBZA:
The only bug-free code is deleted code

Issue #1358, opened 2018-08-20
[TODO] ErrorImporter - fail startup if can't write to error forwarding queue


SzymonPobiega:

WilliamBZA:
Wouldn't this fail anyway once we tried to write? I guess failing at startup would be a clearer message to the users though.

How would we check if we can write or not anyway?

danielmarbach:
Similar comment to https://github.com/Particular/ServiceControl/issues/1355#issuecomment-414424151

Issue #1355, opened 2018-08-20
[TODO] AuditImporterFeature - fail startup if can't write to the forwarding queue


SzymonPobiega:

danielmarbach:
This one is tricky. How do you want to do this in a transport agnostic way? Can we even do this? Probably would add a startup task and then try to send into the forwarding queue but that feels very ugly. I'd rather see us adding a circuit breaker and shut down if the user request us to forward and we can't write.

mikeminutillo:
> Probably would add a startup task and then try to send into the forwarding queue but that feels very ugly.

That's exactly how we used to do it 

https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/Operations/AuditQueueImport.cs#L110-L129

I raised https://github.com/Particular/ServiceControl/pull/1363 with a similar implementation.

PR #1363, opened 2018-08-21
Verify forwarding address at startup

mikeminutillo:
Verify can write to audit and error log queue at startup

Fixes https://github.com/Particular/ServiceControl/issues/1355
Fixes https://github.com/Particular/ServiceControl/issues/1358

### DIFF
--- a/src/ServiceControl/Operations/ErrorImporter.cs
+++ b/src/ServiceControl/Operations/ErrorImporter.cs
@@ -40,9 +40,13 @@
                 );
                 
                 context.RegisterStartupTask(b => new StartupTask(b.Build<SatelliteImportFailuresHandler>(), this));
-            }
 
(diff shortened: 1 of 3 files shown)


src/ServiceControl/Operations/ErrorImporter.cs
```diff
@@ -81,5 +85,27 @@ protected override Task OnStop(IMessageSession session)
                 return Task.FromResult(0);
             }
         }
+
+        class EnsureCanWriteToForwardingAddress : FeatureStartupTask
```
SzymonPobiega: How about making this one non-nested and sharing it between audit and error forwarding?

mikeminutillo: Can do. The only reason I thought it might make sense to keep them separate is that the startup diagnostics will record them by type name and nesting them means it will report Audit+Ensure... and Error+Ensure...

PR #1365, opened 2018-08-21
Remove todos from ATs

SzymonPobiega:


src/ServiceControl.AcceptanceTests/MessageFailures/When_a_retry_for_a_empty_body_message_is_successful.cs
```diff
@@ -110,26 +110,22 @@ protected override TransportOperations CreateMessage(MyContext context)
                 }
             }
 
-            public class LookForControlMessage : IMutateIncomingTransportMessages
+            public class LookForControlMessage : Behavior<IIncomingPhysicalMessageContext>
```
danielmarbach: Please inherit this from IBehavior. It is more performant and less allocaty

SzymonPobiega: @danielmarbach that's an acceptance tests. Shouldn't we use public API in these?

danielmarbach: You are right. Missed that. Approved

PR #1366, opened 2018-08-21
Allow disable ingestion

danielmarbach:
Found as part of investigating #1341

### DIFF
--- a/src/ServiceControl/Bootstrapper.cs
+++ b/src/ServiceControl/Bootstrapper.cs
@@ -158,9 +158,11 @@ namespace Particular.ServiceControl
 ServiceControl Version:             {version}
 Audit Retention Period:             {settings.AuditRetentionPeriod}
 Error Retention Period:             {settings.ErrorRetentionPeriod}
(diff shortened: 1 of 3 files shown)


SzymonPobiega:
@danielmarbach do I understand correctly that we were not setting the flags properly and, as a result, feature has been activated with `null` queue?

If so, could we move the check if the queue is `null` to the feature's prerequisite to make it more explicit?

danielmarbach:
@SzymonPobiega the previous installer approach handled null properly and disabled the queue installation and everything (I think that was part of the core behavior). I thought about the settings approach in the prerequisite but I decide against it because of the two following reasons:

- We need access to settings and while possible to get the SC settings out of the Settings bag in the prerequisite phase we are not using it that way
- In general, SC seems to follow the approach of having features enabled by default and features deciding to not do anything. For example, the existing feature checks the IngestXYZMessages flag and doesn't register the infrastructure instead of disabling the feature itself

Because of that I felt it is more aligned with what we already have and assume to be true

Issue #1350, opened 2018-08-17
Add support for V7 transports to ServiceControl


danielmarbach:
As part of the 3.1 release of ServiceControl, we'll be introducing support for transports and topologies that are not yet supported by SC. This issue represents the high-level PoA to get this effort going

TF: @kbaley @tmasternak @danielmarbach 
Reviewer Azure: @SeanFeldman 
Reviewer RabbitMQ: @bording 
Reviewer SC: @SzymonPobiega 

## Plan of Attack

The transport verification and integration steps might be possible to parallelize onces the SCMU / UI changes have been implemented

- [x] Implement SCMU support for transport selection. Define meaningful names for all the to be supported transports (don't forget to change PowerShell as well) (Dependency for all other items)
  - [x] Document SCMU names including links to the transport sections on the docs site @kbaley ([PR](https://github.com/Particular/docs.particular.net/pull/3999))
  - [x] Update https://docs.particular.net/servicecontrol/transport-adapter/incompatible-features. Make sure it is clear that in many scenarios transport adapters are no longer necessary @kbaley ([PR](https://github.com/Particular/docs.particular.net/pull/3999))
  - [x] Add guidance to https://docs.particular.net/samples/servicecontrol/events-subscription/ for ASBS and ASB to show how `ValidateAndHashIfNeedded` or Hashing rules for subscriptions and rule greater than 50 chars is required ([PR](https://github.com/Particular/docs.particular.net/pull/4000))

- [x] Support Azure Service Bus Forwarding Topology
  - [x] Verify or create build infrastructure to run ATTs under that topology
  - [x] Verify or come up with sensible defaults if not already present
  - [x] Smoke test
- [x] Support RabbitMQ Direct Topology
  - [x] Verify or create build infrastructure to run ATTs under that topology
  - [x] Verify or come up with sensible defaults if not already present
  - [x] Smoke test
- [x] Support SQS
  - [x] Verify or create build infrastructure to run ATTs under that transport
  - [x] Verify or come up with sensible defaults if not already present
  - [x] Smoke test
- [x] Support ASBS if deemed ready enough
  - [x] Verify or create build infrastructure to run ATTs under that transport
  - [x] Create transport integration assemblies
  - [x] Verify or come up with sensible defaults if not already present
[...]

PR #1378, opened 2018-08-23
ASB forwarding topology UI

kbaley:
Connects to #1350 

- Adds ASB forwarding topology to the list of transports
- Renames existing ASB to clarify that it's endpoint-oriented topology

I smoke tested with https://docs.particular.net/samples/servicecontrol/events-subscription/


SeanFeldman:
This will reflect in UI, but how will this change be communicated to the transport?

boblangley:
@SeanFeldman this was added previously, this PR just wires it up the SCMU UI:
Endpoint Oriented: https://github.com/Particular/ServiceControl/blob/develop/src/ServiceControl.Transports.ASB/ASBEndpointTopologyTransportCustomization.cs
Forwarding: https://github.com/Particular/ServiceControl/blob/develop/src/ServiceControl.Transports.ASB/ASBForwardingTopologyTransportCustomization.cs

PR #1386, opened 2018-08-24
ASBS support

kbaley:
#1350


src/ServiceControlInstaller.Engine/Instances/ServiceControlCoreTransports.cs
```diff
@@ -35,6 +35,14 @@ public class ServiceControlCoreTransports
                 Matches = name => name.Equals("ServiceControl.Transports.ASB.ASBForwardingTopologyTransportCustomization, ServiceControl.Transports.ASB", StringComparison.OrdinalIgnoreCase)
             },
             new TransportInfo
+            {
+                Name = "AzureServiceBus - .Net Standard",
```
bording: Why is there a hyphen in the name?

bording: Also, it should be ".NET", not ".Net"

src/ServiceControl.AcceptanceTests/ServiceControl.AcceptanceTests.csproj
```diff
@@ -11,12 +11,23 @@
     <ProjectReference Include="..\ServiceControl.Transports.SQS\ServiceControl.Transports.SQS.csproj" />
     <ProjectReference Include="..\ServiceControl\ServiceControl.csproj" />
     <ProjectReference Include="..\ServiceControl.Transports.ASB\ServiceControl.Transports.ASB.csproj" />
+    <ProjectReference Include="..\ServiceControl.Transports.ASBS\ServiceControl.Transports.ASBS.csproj" />
+
```
danielmarbach: 🔥

src/ServiceControl.Transports.ASBS/Helper.cs
```diff
@@ -0,0 +1,40 @@
...
+    using System.Text;
+    using NServiceBus;
+
+    public static class Helper
+    {
+        public static void ConfigureTransport(this TransportExtensions<AzureServiceBusTransport> transport, TransportSettings transportSettings)
+        {
+            transport.ConfigureSanitization();
+
+            transport.Transactions(TransportTransactionMode.SendsAtomicWithReceive);
```
bording: This the default transaction mode, but I guess you're just specifying it regardless?

src/ServiceControl.Transports.ASBS/Helper.cs
```diff
@@ -0,0 +1,40 @@
...
+            transport.ConnectionString(transportSettings.ConnectionString);
+        }
+
+        public static void ConfigureSanitization(this TransportExtensions<AzureServiceBusTransport> transport)
```
bording: We've intentionally moved away from the term "Sanitization" in the new transport. This might be better named as `ConfigureNameShorteners`.

src/ServiceControl.Transports.ASBS/Helper.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            transport.SubscriptionNameShortener(n => n.Length > MaxEntityName ? MD5DeterministicNameBuilder.Build(n) : n);
```
bording: I assume these are the same rules that are default for the old transport?

src/ServiceControl.AcceptanceTests/ServiceControl.AcceptanceTests.csproj
bording: 🔥

src/ServiceControl.AcceptanceTests/ConfigureEndpointAzureServiceBusStandardTopologyTransport.g.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        return Task.FromResult(0);
+    }
+
+    public Task Cleanup()
+    {
+        return Task.FromResult(0);
+    }
+
+    public string Name => "AzureServiceBus - .NET Standard";
```
bording: Same comments about name: Why the hyphen?

src/ServiceControl.AcceptanceTests/ConfigureEndpointAzureServiceBusStandardTopologyTransport.g.cs
```diff
@@ -0,0 +1,36 @@
+﻿extern alias TransportASBS;
+using System.Threading.Tasks;
+using NServiceBus;
+using NServiceBus.AcceptanceTesting.Support;
+using ServiceBus.Management.AcceptanceTests;
+using ServiceControl.Transports.ASBS;
+
+public class ConfigureEndpointAzureServiceBusStandardTransport : ITransportIntegration
```
bording: Should this be named something like `ConfigureEndpointAzureServiceBusNetStandardTransport` instead, or is there something requiring this specific name?

src/ServiceControl.AcceptanceTests/ConfigureEndpointAzureServiceBusStandardTopologyTransport.g.cs
```diff
@@ -0,0 +1,36 @@
+﻿extern alias TransportASBS;
```
bording: This comment is about the name of the file "ConfigureEndpointAzureServiceBusStandardTopologyTransport.g.cs"

This doesn't seem to be matching the name of the class.

Also, there is only one topology in ASBS, so I'm not sure it needs to be in the name.

bording: Also, why does this file have a ".g.cs" extension?

bording:
When adding an SC instance, how are the required queues, etc created? The reason I'm asking is because of settings like `EntityMaximumSize` and `EnablePartitioning`, which could need to be changed based on the namespace settings.

src/ServiceControlInstaller.Engine/Instances/ServiceControlCoreTransports.cs
danielmarbach: That name has no real relevance other than it is displayed ik the test output. This reflects whatever that has been typed in there first and can be considered a braindump

src/ServiceControl.Transports.ASBS/Helper.cs
danielmarbach: The intention is to make it backward compat with forwarding in ASB yes.

src/ServiceControl.Transports.ASBS/Helper.cs
danielmarbach: I chose this name because of the desired backward compat so I felt having the naming of ASB makes more sense. Happy to change tough

src/ServiceControl.Transports.ASBS/Helper.cs
danielmarbach: We specify it in all transports, ergo here as well

src/ServiceControl.AcceptanceTests/ConfigureEndpointAzureServiceBusStandardTopologyTransport.g.cs
danielmarbach: Can be any name. What ever makes us happy and singing in unison 🎶😊

danielmarbach:
@bording SC is executed during SCMU bootstrapping with EnableInstallers. Any nonstandard configuration that does not work with sensible defaults that can be pre-configured is out of scope unfortunately. Can you elaborate a bit?

src/ServiceControl.AcceptanceTests/ConfigureEndpointAzureServiceBusStandardTopologyTransport.g.cs
danielmarbach: g.cs or Generated.cs removes files from inspection thus preventing ReSharper inspection in failing the build

bording: I guess that works, but it seems like a weird way to prevent inspection errors.

bording:
> @bording SC is executed during SCMU bootstrapping with EnableInstallers. Any nonstandard configuration that does not work with sensible defaults that can be pre-configured is out of scope unfortunately. Can you elaborate a bit?

@danielmarbach okay, that makes sense. I wasn't sure if installers were used or not.

Looking closer at the settings I mentioned, if those aren't able to be set, that won't really be a blocker for  usage.

One that could be a blocker would be `UseWebSockets`. If someone is in a scenario where the AMQP ports are blocked, then they won't be able to use SC.

src/ServiceControl.AcceptanceTests/ConfigureEndpointAzureServiceBusStandardTopologyTransport.g.cs
kbaley: I picked it to coincide with the ASBS acronym we've chosen

src/ServiceControlInstaller.Engine/Instances/ServiceControlCoreTransports.cs
danielmarbach: I made a mistake here. I commented on the wrong file. Here it definitely is relevant what we write! Will change

danielmarbach:
> One that could be a blocker would be UseWebSockets. If someone is in a scenario where the AMQP ports are blocked, then they won't be able to use SC.

Correct or use a transport adapter if that works

Issue #1367, opened 2018-08-21
Spike RavenDB 4.1 embedded mode


danielmarbach:
In this spike, we'd like to investigate the throughput and performance improvements we'd get when switching to RavenDB 4.1 embedded.

RavenDB 4.1 will bring back the [embedded mode](https://ayende.com/blog/183555-A/ravendb-4-1-features-running-ravendb-embedded). As another benefit, the instance will be running as a sidecar along with the main process and no longer need to share resources like the thread pool. Recent efforts already made significant resource utilization improvements in ServiceControl. Performance tests showed that ESENT's single writer thread can be the bottleneck in some scenarios. 

Timebox: 3 days

## Plan of attack

- Switch to RavenDB 4.1 embedded (hack)
- Setup loadtest resources on Azure on a single transport
- Run load test scenarios 1-2 days
- Document outcome under /docs

## Out of scope

- Migration from Esent to Voron 
- Tweak on the codebase in an attempt to improve performance

## Skills required

- Familiarity with RavenDB or willingness to learn
- Tolerance with pre-release software
- At least one member of @Particular/servicecontrol-maintainers who is familiar with the load test generator

## You'll learn

- RavenDB 4.1 and its new embedded mode
- ServiceControl and its marriage with RavenDB
- LoadTestGenerator and how to click through the Azure portal :)

boblangley:
I am interested in joining this task force.

SzymonPobiega:
After upgrading the package reference compilation of the `ServiceControl` project resulted in 200+ errors. Some of them are simple namespace changes but some are much more severe:
 * The `IndexQuery` API changed a lot (missing stuff)
 * The `AbstractTransformerCreationTask` does not exist any more
 * The `Etag` is no longer surfaced via the API. Instead we can access the change vector which contains etag and cluster node information.
   * A lot of advanced features was based on assumption that you can query the index Etag to see if it includes the most recent changes. This is probably not possible any more
   * No easy way to compare change vectors in our code as they are exposed as strings. We would probably have to re-implement the comparison algo described in RavenDB docs

I haven't even got through 20% of the errors. I am pretty sure we won't be able to get it to compile, let alone run and pass the tests in the time slot assigned. 

@boblangley I propose we try configuring our well known 3.5 RavenDB to use Voron and check if this setup is runnable and passes the tests. I don't expect much benefits in terms of performance though because of the architecture of RavenDB 3.5.

danielmarbach:
[...]

PR #1387, opened 2018-08-25
[SPIKE] Raven35 Voron Spike

boblangley:
Connects to #1367 
Switch to Voron without upgraded RavenDb so we can test if Voron has performance improvements


danielmarbach:
@SzymonPobiega I guess this can be closed?

PR #1391, opened 2018-08-29
Support new transports in PowerShell

kbaley:
Connects to #1350 

Marked as WIP because I haven't tested this locally yet.


danielmarbach:
I'm bringing this in. We can add more fixes if required as additional PRs

PR #1393, opened 2018-08-30
Change asbs zip folder so that asb folder is not it's prefix

tmasternak:
The code which is responsible for extracting the transport specific files from zip package assumes that there is a single subfolder that has transport name as a prefix.

Before changes in this PR that was not true. `AzureServiceBusNetStandard` subfolder had a prefix `AzureServiceBus` that was also a valid transport name. When installing any of the two ASB topologies unzipping would incorrectly try to extract ASBS subfolder in [FileUtils.cs](https://github.com/Particular/ServiceControl/blob/develop/src/ServiceControlInstaller.Engine/FileSystem/FileUtils.cs#L891)


SzymonPobiega:
Aaaa sorry for jumping so quick on this. I should have waited for Daniel or Kyle.

Issue #1354, opened 2018-08-20
Come up with a way to ensure Web/Message contracts don't change between ServiceControl releases


WilliamBZA:
## Overview

* Facilitator: @mauroservienti
* TF Slack channel: https://particularsoftware.slack.com/messages/CCHB3D2D7

ServiceControl currently exposes functionality via 4 channels:

* A RESTful api through NancyFx
* A message-based integration through the [ServiceControl.Contracts Package](https://www.nuget.org/packages/ServiceControl.Contracts/)
* Via messaging internally with itself (This is technically not exposed, but anyone with access can put a message on the ServiceControl queue to invoke that functionality)
* Via SignalR to ServicePulse

Currently there are no mechanisms in place to ensure that the three API surfaces are not changed in a breaking manner between releases.

Focus: Internal benefit and quality assurance

## Proposed PoA

* [x] Confirm that the ServiceControl.Contracts has Particular Approvals running as part of it's build and test
  _Raised https://github.com/Particular/ServiceControl/pull/1421 to start running approval tests_
* [x] Decide if all the internal message classes need to be moved into an assembly so that it can be run against Particular Approvals or if one of the API Builder overloads can be used to determine if the type is an internal message or simply an internal class
  _API Generator allows to select types to generate public API for, no need for a separate assembly_
* [x] Add a test to ensure the internal messages class API doesn't change between releases
  _https://github.com/Particular/ServiceControl/pull/1421 to start running approval tests_
* [x] Decide on a way to generate a text representation of the RESTful API (Swagger? OpenAPI?)
  _we're using the Nancy test infrastructure in https://github.com/Particular/ServiceControl/pull/1421_
* [x] Add a test that compares the test representation of the RESTful API against an approved one
  _https://github.com/Particular/ServiceControl/pull/1421_
* ~Decide on a way to generate a text representation of the SignalR API~
   _We couldn't find any tool to achieve this. We realized it's not needed as we're using 1 single SignalR method (`Broadcast` on the `IConnection` interface) to route all messages from Control to Pulse_
* ~Add a test that compares the test representation of the SignalR API against an approved one~
  _withdrawn_
* [x] raise separate issue related to test Nancy headers https://github.com/Particular/ServiceControl/issues/1431
* [ ] retrospective 

## What you will learn

By joining this taskforce, you will learn:

* Knowledge of API Approvals
* How NancyFx deals with headers and parameter bindings
* SignalR experience

## Required specialists

[...]

PR #1401, opened 2018-09-05
[WIP] Message contracts approvals

mauroservienti:
Connects to https://github.com/Particular/ServiceControl/issues/1354

### DIFF
--- a/src/ServiceControl.UnitTests/API/APIApprovals.cs
+++ b/src/ServiceControl.UnitTests/API/APIApprovals.cs
@@ -32,5 +32,21 @@
             var publicApi = ApiGenerator.GeneratePublicApi(serviceControlAssembly, domainEventTypes, shouldIncludeAssemblyAttributes: false);
             Approver.Verify(publicApi);
         }
+
+        [Test]
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        public void ApproveMessageTypes()
+        {
+            var serviceControlAssembly = typeof(Particular.ServiceControl.Bootstrapper).Assembly;
+            var notDomainEventNorUserInterfaceEventTypes = serviceControlAssembly.DefinedTypes.Where(t =>
+                t.Namespace != null
+                && t.Namespace.StartsWith("ServiceControl.Contracts")
+                && !typeof(IDomainEvent).IsAssignableFrom(t)
+                && !typeof(IUserInterfaceEvent).IsAssignableFrom(t)
+                ).ToArray();
+
+            var publicApi = ApiGenerator.GeneratePublicApi(serviceControlAssembly, notDomainEventNorUserInterfaceEventTypes, shouldIncludeAssemblyAttributes: false);
+            Approver.Verify(publicApi);
+        }
     }
 }
--- a/src/ServiceControl.UnitTests/API/APIApprovals.cs
+++ b/src/ServiceControl.UnitTests/API/APIApprovals.cs
@@ -3,6 +3,7 @@
     using NUnit.Framework;
     using Particular.Approvals;
     using PublicApiGenerator;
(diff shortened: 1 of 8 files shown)


mauroservienti:
I'm not sure about the last commit https://github.com/Particular/ServiceControl/pull/1401/commits/4b2f980c2f4b3c40a3e54ed085295e6bb4270011

src/ServiceControl.UnitTests/ApprovalFiles/APIApprovals.ApproveDomainEventTypes.approved.txt
```diff
@@ -0,0 +1,288 @@
...
+{
+    public class CustomCheckFailed : ServiceControl.Infrastructure.DomainEvents.IDomainEvent, ServiceControl.Infrastructure.SignalR.IUserInterfaceEvent
+    {
+        public CustomCheckFailed() { }
+        public string Category { get; set; }
+        public string CustomCheckId { get; set; }
+        public System.DateTime FailedAt { get; set; }
+        public string FailureReason { get; set; }
+        public System.Guid Id { get; set; }
+        public ServiceControl.Contracts.Operations.EndpointDetails OriginatingEndpoint { get; set; }
```
mikeminutillo: Hmmm. We need to find a way to find all of the child types too as they form a part of the contract. I wonder if the ApiApprovals library is the best way to do this or if some basic reflection code would be better suited to our needs here.

mikeminutillo: For instance, we probably don't care about interfaces or constructors for this approval.

mikeminutillo:
> I'm not sure about the last commit 4b2f980

I wonder if we should be diffing this against the SC Contracts assembly. In fact, why aren't we just using the SC Contract assembly?

mauroservienti:
Where is the ServiceControl.Contracts assembly?
The thing is that we want to test the public API of all the messages that we send via SignalR as well.

mikeminutillo:
@mauroservienti SC Contracts has it's own Repo with it's own approval tests https://github.com/Particular/ServiceControl.Contracts

It's not the same set of classes though.

mauroservienti:
@mikeminutillo I can find any approval tests for [ServiceControl.Contracts](https://github.com/Particular/ServiceControl.Contracts). Are they supposed to be in the same repo or anywhere else?

src/ServiceControl.UnitTests/ApprovalFiles/APIApprovals.ApproveDomainEventTypes.approved.txt
bording: To clarify, the content of this file is created by PublicApiGenerator, not Particular.Approvals. If this file has unwanted stuff in it, the call to `ApiGenerator.GeneratePublicApi` is what needs to be tweaked/replaced, not `Approver.Verify`.

mauroservienti:
@boblangley can this one be closed in favor of #1406 ?

src/ServiceControl.UnitTests/API/APIApprovals.cs
```diff
@@ -0,0 +1,52 @@
...
+    using ServiceControl.Infrastructure.DomainEvents;
+    using ServiceControl.Infrastructure.SignalR;
+    using System.Linq;
+    using System.Runtime.CompilerServices;
+
+    [TestFixture]
+    class APIApprovals
+    {
+        [Test]
+        [MethodImpl(MethodImplOptions.NoInlining)]
```
bording: FYI These attributes aren't required when using Particular.Approvals

boblangley:
Yes, but we can remove the attributes @bording mentioned in that PR.

PR #1402, opened 2018-09-07
Adds queue name to main window

dvdstelt:


dvdstelt:
@boblangley Can you review for now? I'll discuss the icon with Hadi next week.

PR #1404, opened 2018-09-07
Reduce allocations when expiring documents

danielmarbach:
While investigating a customer case I stumbled over large portions of allocations in the expired documents cleaner.

This PR removes some of them. The closure allocations are removed as well as the MessageMetadata allocations are reduced to only what is needed

### DIFF
--- a/src/ServiceControl.UnitTests/Expiration/ChunkerTests.cs
+++ b/src/ServiceControl.UnitTests/Expiration/ChunkerTests.cs
@@ -13,11 +13,12 @@
             var starts = new List<int>();
             var ends = new List<int>();
 
-            Chunker.ExecuteInChunks(1500, (s, e) =>
+            var count = Chunker.ExecuteInChunks(1500, (startList, endList, s, e) =>
             {
-                starts.Add(s);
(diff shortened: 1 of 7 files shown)


danielmarbach:
I did not see the tests in ProcessedMessageExpirationTests. I was looking under the Infrastructure folder and did not find the other one. Happy to remove the additional tests from this PR.

SzymonPobiega:
Looks good. Did you check if it helps? My previous attempt to limit allocations had no visible effect. But since this one does not change the index I think we should merge it even if we can't see huge benefits.

danielmarbach:
@SzymonPobiega I did run dotMemory benchmarks but I ran out of time on Friday. The closure allocation changes definitely helped. I was able to free up dozens of megabytes of Gen0 allocs. I was not able to pinpoint in the short time if the projection changes really made a big difference. At the end I figured since this change is small and doesn't affect an index I'll send it in as is and do not spend more time on it

danielmarbach:
@SzymonPobiega @mikeminutillo I removed the redundant tests and rebased it on top of develop

PR #1406, opened 2018-09-08
[WIP] Change sc access modifiers

boblangley:
Connects to #1354

Move types to internal in order to use APIApprovals for stabilizing public (both customer and Particular use only) contracts.


boblangley:
53 acceptance tests are still failing.

danielmarbach:
@boblangley why is it necessary to change the access modifiers of the Acceptance Tests?

boblangley:
@danielmarbach So far because they expose previously public types. The first pass was very crude, I may have to reverse some of the access modifier changes.

boblangley:
Closed in favor of #1421 which reorganizes these commits into something more reviewable.

PR #1410, opened 2018-09-11
Remove HeadersForSearching from metadata and compute query array on the index

danielmarbach:
For ingested audit messages and error messages we created a metadata field called `HeadersForSearching`. That metadata field is stored in the `ProcessedMessage` collection as well as the FailedMessages collection. In the `FailedMessages` collection, we might store this field redundantly for each processing attempt. The value of that field is never directly used. It is used in the `MessagesViewIndex` to precompute the `Query` array which is then stored in the index itself. This PR removes this redundant storing from the document and modifies the `MessagesViewIndex` to compute the concatenated headers field and directly store it in the index.

This means we only store the headers two times (in the document and the index) per document or processing attempt instead of three times (in the document, the metadata and the index). With that change, we are saving the kbytes of data per ingested audit and error message while still being able to fully query

This will not modify existing documents but just change newly ingested once. Old documents will expire with the expiry period. Reindexing will happen on the `MessagesViewIndex` though.

The larger the headers the bigger the savings plus we no longer allocate the concatenated strings in ServiceControl to shuffle it to RavenDB.

### Before

image

image

## After

image

image

### Before

image

image

### After

image

image

## Searching

Demonstrates how searching for custom header still works

image

image

### DIFF
--- a/src/ServiceControl.UnitTests/CompositeViews/MessagesViewTests.cs
+++ b/src/ServiceControl.UnitTests/CompositeViews/MessagesViewTests.cs
@@ -36,6 +36,8 @@
 
                 session.SaveChanges();
             }
+            
+            documentStore.WaitForIndexing();
 
             using (var session = documentStore.OpenSession())
             {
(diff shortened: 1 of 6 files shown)


danielmarbach:
Btw. the above optimization trick could be done also for other fields that we don't query against the metadata and thus even more storage savings could be gained

PR #1411, opened 2018-09-12
Hotfix 3.1.2

mikeminutillo:


SzymonPobiega:
Re-running the acceptance tests now to see if the failures were caused by flaky tests

SzymonPobiega:
Almost all tests are green (except two in ASB EOT and 1 in ASBS). I think we can :shipit:

PR #1412, opened 2018-09-12
WIP Make CheckDeadLetterQueue work on non-english versions of windows

danielmarbach:


boblangley:
Looks good so far, but I will review properly when it is not WIP

danielmarbach:
Something is completely off with performance counter localization both on Windows 10 and Windows 2016. Based on my research the assumption is the following:

`HKEY_LOCAL_MACHINE\Software\Microsoft\Windows NT\Current Version\Perflib\` contains location ID specific subfolder and a `CurrentLanguage` folder. The `CurrentLanguage` folder has to point to the currently selected primary display language. There should be always a `009` subfolder which contains the English neutral names for the performance counters. Once you switch to a different locale let's say German (ID 007) it should create a new subfolder 007 containing the localized german performance counters and categories and switch the `CurrentLanguage` folder to point to the German names. In reality, this is NOT what is happening. See the screenshots

## German

image

image

## Norwegian

image

image


## Polish

image

image


The tests have been conducted on @janovesk workstation, my workstation and laptop, Azure Cloud VM and @Scooletz laptop.

I tested also French in the Azure VM and it behaves exactly like German and Norwegian in the wrong way. Polish is the only one that works correctly as specified.

Issue #1433, opened 2018-09-25
AmazonSQS Transport : Retries don't work


AbdoHassanUMG:
Message Replay doesn't work when using AmazonSQS transport because of inconsistency in the casing of Error Queue. It is called "{InstanceName}.Errors" when writing the error message and "{InstanceName}.errors" when forwarding it. 
This should cause no problem if tested using MSMQ but causes trouble in SQS because queue names are case-sensitive. 
I have created a fix for this and I want to send it as a pull request

AbdoHassanUMG:
I think the problem is here
`public class RawEndpointFactory
    {
        public RawEndpointFactory(Settings settings, TransportSettings transportSettings, TransportCustomization transportCustomization)
        {
            this.transportSettings = transportSettings;
            this.settings = settings;
            this.transportCustomization = transportCustomization;
        }

        public RawEndpointConfiguration CreateRawEndpointConfiguration(string name, Func<MessageContext, IDispatchMessages, Task> onMessage, TransportDefinition transportDefinition)
        {
// the problem is on the line below !
            var config = RawEndpointConfiguration.Create(name, onMessage, $"{transportSettings.EndpointName}.errors");
            config.LimitMessageProcessingConcurrencyTo(settings.MaximumConcurrencyLevel);

            transportCustomization.CustomizeRawEndpoint(config, transportSettings);
            return config;
        }

        Settings settings;
        TransportCustomization transportCustomization;
        private TransportSettings transportSettings;
    }`

danielmarbach:
Correct AmazonSQS is case sensitive. Really weird that our acceptance tests didn't catch that one. We'll have to do a closer investigation, especially because we have tests that do retry and they are green in our CI environment. Otherwise, we wouldn't have released this functionality

Thanks for reporting

danielmarbach:
@AbdoHassanUMG Can you explain why you think retries don't work? The queue we are talking about is `Particular-ServiceControl-Errors` which ServiceControl sends internally failed messages to. That queue is not used for message retrying. While we still want to fix the casing there and your PR is valid the retry functionality will use the staging queue `Particular-ServiceControl-staging`.

AbdoHassanUMG:
Without handling the case-sensitivity I get the following error at SC console when attempting a "retry"
Looking at trafic to SQS using fiddler I discovered the cases sensitivity issue as one of SQS API calls returns a 400 Bad request.

[...]

PR #1434, opened 2018-09-25
Fix for AmazonSQS Transport : Retries don't work #1433

AbdoHassanUMG:


danielmarbach:
I approved the PR because it is a valid change that we want to apply. I pointed out in comment https://github.com/Particular/ServiceControl/issues/1433#issuecomment-424611029 that retries cannot be broken because of that casing issue. The retries capability doesn't use that queue. It is the internal error queue of ServiceControl

danielmarbach:
@AbdoHassanUMG can you please sign https://particular.net/contributors-license-agreement-consent? Thank you

danielmarbach:
CLA signed. Pulling in

PR #1436, opened 2018-09-26
Improve License Expiry Experience

mikeminutillo:


mikeminutillo:
@boblangley the only outstanding thing is that some of the text messages may change

src/ServiceControl.Config/UI/License/LicenseComponentFactory.cs
```diff
@@ -0,0 +1,148 @@
...
+                    Label = "Platform license type:",
+                    Value = $"{details.LicenseType}, {details.Edition}"
+                };
+
+                if (details.ExpirationDate.HasValue)
+                {
+                    yield return PlatformExpiryComponent(details.ExpirationDate.Value);
+                }
+
+                if (details.UpgradeProtectionExpiration.HasValue)
```
boblangley: Shouldn't this be before expiration date? All of the perpetual licenses I have seen have both.

mikeminutillo: They have both in the file but we don't load expiration date for perpetual licenses

mikeminutillo: And because these are doing a `yield return`, it won't short circuit anyway. If your license reports having both, we will display both.

mikeminutillo:
> Could the license component portion of this be moved to the Engine so we can unit test and eventually make the non wpf experience better as well? I am thinking the factory could use so basic tests and we might want some messaging in the powershell scripts.

I discussed this with @HEskandari and we both feel that the "logic" in these changes is really a UI concern. In SI, the change calculates days until expiry and then uses that to show/hide stuff. In SCMU, I convert the License into a dictionary of components and then bind the UI to that.

The only thing that maybe belongs in the licensing package is the calculation of days remaining.

mikeminutillo:
@boblangley as SCMU is the only code to currently rely on the license component factory, can we skip extraction until someone else needs the code?

I can add some unit (approval) tests around them.

SzymonPobiega:
@mikeminutillo is there anything here I can help with to merge this one?

mikeminutillo:
@SzymonPobiega I have tested with a bunch of different licenses after the last round of changes. I am happy to merge and release.

SzymonPobiega:
@mikeminutillo do we need to release it or would merging to develop be just fine?

mikeminutillo:
@SzymonPobiega it can go in the next maintenance release.

SzymonPobiega:
Cool. Then I am merging it 🎉

PR #1437, opened 2018-09-26
Allow to specify localized category name, counter name and instance name for CheckDeadLetterQueue custom check

danielmarbach:
## Who's affected

Customers using MSMQ transport together with ServiceControl on a localized (non-english) version of Windows

## Symptoms

System performance counters get localized on non-english versions of Windows. When the `CheckDeadLetterQueue` custom check is created or when it is executed an `InvalidOperationException` might be thrown in cases the counter under category `MSMQ Queues` with name `Messages in Queue` and instance name `Computer Queues` could not be found. 

## Fix

When the counter cannot be created or the value cannot be fetched due to localization issues the custom check will log a meaningful message that explains how to resolve the issue containing the following information

```
Unable to read the Dead Letter Queue length. The performance counter with category 'MSMQ Queues' and name 'Messages in Queue' and instance name 'Computer Queues' is not available. It is possible that the counter category, name and instance name have been localized into different languages. Consider overriding the counter category, name and instance name in the application configuration file by adding:
   <appSettings>
     <add key="ServiceControl/Msmq/PerformanceCounterCategoryName" value="LocalizedCategoryName" />
     <add key="ServiceControl/Msmq/PerformanceCounterName" value="LocalizedCounterName" />
     <add key="ServiceControl/Msmq/PerformanceCounterInstanceName" value="LocalizedCounterInstanceName" />
   </appSettings>
```

The failed custom check will also contain this helpful information and present it in ServicePulse without line breaks

image

### DIFF
--- a/src/ServiceControl.Transports.Msmq/CheckDeadLetterQueue.cs
+++ b/src/ServiceControl.Transports.Msmq/CheckDeadLetterQueue.cs
@@ -1,6 +1,7 @@
 ﻿namespace ServiceControl.MSMQ.DLQMonitor
 {
     using System;
+    using System.Configuration;
     using System.Diagnostics;
     using System.Threading.Tasks;
     using NServiceBus.CustomChecks;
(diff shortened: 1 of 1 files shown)


src/ServiceControl.Transports.Msmq/CheckDeadLetterQueue.cs
```diff
@@ -13,31 +14,96 @@ public class CheckDeadLetterQueue : CustomCheck
...
+            }
 
             if (currentValue <= 0)
             {
                 Logger.Debug("No messages in Dead Letter Queue");
                 return CheckResult.Pass;
             }
 
-            var result = $"{currentValue} messages in the Dead Letter Queue on {Environment.MachineName}. This could indicate a problem with ServiceControl's retries. Please submit a support ticket to Particular using support@particular.net if you would like help from our engineers to ensure no message loss while resolving these dead letter messages.";
+            result = $"{currentValue} messages in the Dead Letter Queue on {Environment.MachineName}. This could indicate a problem with ServiceControl's retries. Please submit a support ticket to Particular using support@particular.net if you would like help from our engineers to ensure no message loss while resolving these dead letter messages.";
```
mikeminutillo: Is it worth pulling this out into a method like the `ContainerMightBeLocalized()`? Only for readability.

PR #1454, opened 2018-10-11
Upgrade to NServiceBus 7.1.4 and acceptance testing

danielmarbach:


danielmarbach:
Pushed an update @boblangley

andreasohlund:
This one looks good to go, should this one make into the upcoming 3.3.0 release?

danielmarbach:
@andreasohlund ideally yes

andreasohlund:
@Particular/servicecontrol-maintainers please merge so that we can include this in the upcoming 3.3.0

WilliamBZA:
SQS tests are still failing, should that prevent this PR from being merged?

danielmarbach:
@WilliamBZA I rebased the PR on top of develop and remove the explicit timeouts that are short. There is a client issue that we hit sometimes in the ATTs that causes the message to not be visible until the visibility timeout is reached (more than 30 seconds). When the test hits that and has a runtime of less than 20 seconds the test is aborted before the message can arrive.

PR #1460, opened 2018-10-12
WIP Fix registry key lower bound for prerequisite

danielmarbach:
See https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/how-to-determine-which-versions-are-installed


WilliamBZA:
I'll set up a blank VM to test this on

danielmarbach:
I think the culprit is somewhere else

PR #1465, opened 2018-10-19
Filter out null failedmessages

WilliamBZA:
Fixes https://github.com/Particular/ServiceControl/pull/1465

2 Commits. One to bump the zip package version (I can remove this?) and a second to filter out null failed messages.

I can't figure out how to force the expired documents to run in an acceptance test so am not sure how to add a test around this. Any suggestions?

### DIFF
--- a/src/ServiceControl/ExternalIntegrations/MessageFailedPublisher.cs
+++ b/src/ServiceControl/ExternalIntegrations/MessageFailedPublisher.cs
@@ -27,10 +27,9 @@ namespace ServiceControl.ExternalIntegrations
             var failedMessages = new List<object>(failedMessageData.Length);
             foreach (var entity in failedMessageData)
             {
-                session.Advanced.Evict(entity);
-
                 if (entity != null)
                 {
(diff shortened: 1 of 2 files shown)


src/ServiceControl/ExternalIntegrations/MessageFailedPublisher.cs
```diff
@@ -27,10 +27,9 @@ protected override async Task<IEnumerable<object>> PublishEvents(IEnumerable<Dis
             var failedMessages = new List<object>(failedMessageData.Length);
             foreach (var entity in failedMessageData)
             {
-                session.Advanced.Evict(entity);
-
                 if (entity != null)
```
danielmarbach: Another option would be to add a `where` clause to the list and then iterate over that but I prefer the `if` approach

WilliamBZA: I had that originally, but then saw the if that was in the loop already and thought it looked neater than including the where in the loop above.

PR #1473, opened 2018-11-05
Add asbs dlq check

andreasohlund:


andreasohlund:
There is no good place to close the the client (need to check if we dispose the custom checks) similar to https://github.com/Particular/ServiceControl.Monitoring/blob/develop/src/ServiceControl.Transports.AzureServiceBus/QueueLengthProvider.cs#L105 so I wonder if we should open a new client for each call to perform the check?

src/ServiceControl.Transports.ASBS/CheckDeadLetterQueue.cs
```diff
@@ -0,0 +1,46 @@
...
+            managementClient = new ManagementClient(transportConnectionString);
+            stagingQueue = $"{settings.EndpointName}.staging";
+        }
+
+        public async override Task<CheckResult> PerformCheck()
+        {
+            Logger.Debug("Checking Dead Letter Queue length");
+
+            var queueRuntimeInfo = await managementClient.GetQueueRuntimeInfoAsync(stagingQueue).ConfigureAwait(false);
+            var deadLetterMessageCount = queueRuntimeInfo.MessageCountDetails.DeadLetterMessageCount;
```
tmasternak: What happens when this fails?

andreasohlund: The custom check infra will try catch and mark it as a failure

https://github.com/Particular/NServiceBus.CustomChecks/blob/develop/src/NServiceBus.CustomChecks/TimerBasedPeriodicCheck.cs#L38

andreasohlund:
As far as I can tell custom checks are not disposed

https://github.com/Particular/NServiceBus.CustomChecks/blob/develop/src/NServiceBus.CustomChecks/TimerBasedPeriodicCheck.cs#L26

so if closing the management client is a must we need to open and close in the actual check. 

It runs every hour so I don't see any perf issues there? cc @SeanFeldman

andreasohlund:
Looking at the management client code not closing the client means that the underlying HttpClient won't be disposed on SC shutdown

https://github.com/Azure/azure-service-bus-dotnet/blob/dev/src/Microsoft.Azure.ServiceBus/Management/ManagementClient.cs#L876

Not sure how critical that is?

SeanFeldman:
> It runs every hour so I don't see any perf issues there

I'd vote for creating a client and disposing (closing). Once an hour is not going to impact you that much. But better than keeping that client running and if it's faulting, giving you a false negative.

andreasohlund:
>I'd vote for creating a client and disposing (closing). Once an hour is not going to impact you that much. But better than keeping that client running and if it's faulting, giving you a false negative.

Done

PR #1474, opened 2018-11-05
Fix FormComboBox overflow issue

tmasternak:
## Problem
For long values of the selected item, the textblock overflows the drop-down.

image

## Fix
This fix wraps the `ContentPresenter` with `StackPanel` that has the `MaxWidth` property bound to the `ActualWidth` of the dropdown border.


andreasohlund:
So with this the "Old" part won't show in the dropbox?

tmasternak:
@mauroservienti here is what (I think) is happening. 
 * When the control is rendered the item selected is added as TextBlock 
 * The size of the TextBlock is disconnected from the size of the Boarder which is above it in the logical tree but overlays it when rendered. This is what is the root cause of the problematic behavior,
 * The solution which is implemented is about enforcing MaxWidth on whatever will be rendered in the content presenter (in our case the TextBlock)

tmasternak:
> So with this the "Old" part won't show in the dropbox?

@andreasohlund no unless we increase the dropbox width

andreasohlund:
Will it show in the tool tip? Does it show at any later stage?

tmasternak:
@andreasohlund we don't have tooltip. It shows in the dropdown when expanded and later on the instance list view.

andreasohlund:
Will the user see the "old" part before he's actually installed the instance?

tmasternak:
@andreasohlund yes. When selecting it from the dropdown for the list of available options 

image

andreasohlund:
Got it!

Lgtm then!

mauroservienti:
What I don’t get is why we’re limiting the ComboBox size in any way. Shouldn’t it be allowed to stretch to full width like all other controls so?

danielmarbach:
I agree with @mauroservienti why are we limiting the width @Particular/servicecontrol-maintainers ?

WilliamBZA:
That question doesn't really make sense. The names were never long enough to require more space. Seeing as the stackpanel is inside a grid, there are differences in how width is defaulted. Much like css. So asking why the width was limited is like asking why the default behaviour wasn't changed when it didn't have to be changed.

danielmarbach:
@WilliamBZA I understood @mauroservienti questions in the following way:

image

Given that most of the controls use the available width on the page, why is the width limited in the combo box? And that question makes sense to ask right?

HEskandari:
You really don't need to bind to ActualWidth/ActualHeight properties. The control's size is either set strictly (like in this case), or constrained by the parent container so you can:

- Set the width of the combobox to be larger, to accommodate the largest text.
- Let the parent container control the width (which is what Mauro is suggesting).

I'm +1 on the second option.

mauroservienti:
(It should really be s template binding and not a hardcoded value)

tmasternak:
> (It should really be s template binding and not a hardcoded value)

I agree. That said there are other places when this is set to a constant and I do want to make the minimal change needed.

mauroservienti:
Yeah, mine was more a kind of general comment. Not something that we should enhance now.

PR #1475, opened 2018-11-05
Align required ravendb.database package version

timbussmann:
allowing the build to work correctly on a machine without the patch version being in the nuget cache.

### DIFF
--- a/src/ServiceControl/ServiceControl.csproj
+++ b/src/ServiceControl/ServiceControl.csproj
@@ -42,10 +42,6 @@
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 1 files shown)


bording:
While this is an immediate fix, this will just break again when the RavenDB package is updated. Instead, we can do something where the Item references the version from the package reference, so they will always be in sync.

bording:
I believe something like the following should work:

```
  <ItemGroup>
    <None Include="$(NuGetPackageRoot)ravendb.database\%(PackageReference->Version)\tools\Raven.Studio.Html5.zip" CopyToOutputDirectory="PreserveNewest" Condition="'@(PackageReference)' == 'RavenDB.Database'" />
  </ItemGroup>
```

danielmarbach:
@bording thanks added

Issue #1448, opened 2018-10-09
[Minor / Limited] ServiceInsight result paging is broken when Service Control secured via a reverse proxy


janovesk:
Paging message search results is done by including `next`, `prev`, `first` and `last` URLs in the response send from SC. The problem is these links are absolute (as opposed to e.g. message body link which is relative). When ServiceControl is secured behind a reverse proxy the public URL used to access it is different from the one SC listens on. The paging links contain the private URL which is useless in SI.

Example would be `/api/messages/` that returns a header like "Link=<http://localhost:33333/api/messages/?include_system_messages=False&per_page=50&sort=time_sent&direction=desc&page=1786>; rel=\"last\", <http://localhost:33333/api/messages/?include_system_messages=False&per_page=50&sort=time_sent&direction=desc&page=2>; rel=\"next\"" with "next" and "last" links. 

The code building these links assume that the incoming hostname can be used for the host part of the link. This is not true in scenarios where Service Control is accessed through a reverse proxy. (https://docs.particular.net/servicepulse/install-servicepulse-in-iis#advanced-configuration). In these scenarios, the request to the main hostname is processed by IIS before being forwarded to SC running at localhost. 

So the problem is:

1) A request is made to http://mysuperhost.com:33333/api/messages
2) IIS responsible for mysuperhost.com is a reverse proxy. It looks at the routing rules and finds that anything coming into http://mysuperhost.com:33333/api/ should be forwarded to http://localhost:33333/api/. It does.
3) SC gets the request as "http://localhost:33333/api/messages". 
4) SC assembles response with link header pointing to "http://localhost:33333/api/messages&whatever" because localhost is the incoming host for ServiceControl.
5) IIS returns the link header with "http://localhost:33333/api/messages&whatever" links, instead of links pointing to "http://mysuperhost.com:33333"

This, of course, is incorrect and anybody following these links, like ServiceInsight, will get a failure when they do.

By default, ServiceControl provides no access control. Using a reverse proxy is required to add basic access control to ServicePulse/ServiceControl. It is not uncommon or unreasonable scenario to support given the potentially sensitive nature of messages content available through the ServiceControl API.

We have a customer hit by this: https://c.eu1.visual.force.com/apex/Case?num=00039844

### Plan of action

 - [x] Decide on the solution
    - Modify ServiceInsight to replace the prefix on the absolute links with the connection URL
    - Modify ServiceControl to return relative links
[...]

PR #1476, opened 2018-11-05
Use relative paging links

timbussmann:
relates to https://github.com/Particular/ServiceControl/issues/1448
Added a few unit tests + a first naive implementation

### DIFF
--- a/src/ServiceControl.UnitTests/Infrastructure/Extensions/NegotiatorExtensionsTests.cs
+++ b/src/ServiceControl.UnitTests/Infrastructure/Extensions/NegotiatorExtensionsTests.cs
@@ -13,7 +13,7 @@
         [Test]
         public void WithPagingLinks_ReturnsLinksWithRelativeUri()
         {
-            var pagingHeaders = GetLinks(200, currentPage: 3, path: "/api/test");
+            var pagingHeaders = GetLinks(totalResults: 200, currentPage: 3, path: "/api/test");
 
             Assert.Contains("</api/test?page=4>; rel=\"next\"", pagingHeaders);
             Assert.Contains("</api/test?page=4>; rel=\"last\"", pagingHeaders);
@@ -24,7 +24,7 @@
         [Test]
         public void WithPagingLinks_KeepsExistingQueryParams()
         {
-            var pagingHeaders = GetLinks(100, path: "/api", queryParams: "token=abc&id=42");
+            var pagingHeaders = GetLinks(totalResults: 100, path: "/api", queryParams: "token=abc&id=42");
 
             Assert.Contains("</api?token=abc&id=42&page=2>; rel=\"next\"", pagingHeaders);
             Assert.Contains("</api?token=abc&id=42&page=2>; rel=\"last\"", pagingHeaders);
@@ -33,7 +33,7 @@
         [Test]
         public void WithPagingLinks_WhenHasNextPage_AddNextPageLink()
         {
-            var pagingHeaders = GetLinks(51);
+            var pagingHeaders = GetLinks(totalResults: 51);
 
             Assert.Contains("</?page=2>; rel=\"next\"", pagingHeaders);
         }
@@ -41,7 +41,7 @@
         [Test]
         public void WithPagingLinks_WhenHasNoNextPage_AddNoNextPageLink()
         {
-            var pagingHeaders = GetLinks(50);
+            var pagingHeaders = GetLinks(totalResults: 50);
 
             Assert.IsEmpty(pagingHeaders);
         }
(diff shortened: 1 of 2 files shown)


src/ServiceControl/Infrastructure/Extensions/NegotiatorExtensions.cs
```diff
@@ -110,32 +111,32 @@ public static Negotiator WithPagingLinks(this Negotiator negotiator, int totalRe
...
 
-        static void AddLink(ICollection<string> links, int page, string rel, string queryParams, Url url)
+        static void AddLink(ICollection<string> links, int page, string rel, string queryParams, Uri uri)
         {
-            url.Query = queryParams + "page=" + page;
-
-            links.Add($"<{url}>; rel=\"{rel}\"");
+            var query = queryParams + "page=" + page;
+        
+            links.Add($"<{uri.AbsolutePath + query}>; rel=\"{rel}\"");
```
danielmarbach: I had to look up what `AbsolutePath` does https://docs.microsoft.com/en-us/dotnet/api/system.uri.absolutepath?view=netframework-4.7.2 because I was like "hu? we are trying to avoid using the absolute path"

src/ServiceControl.UnitTests/Infrastructure/Extensions/NegotiatorExtensionsTests.cs
```diff
@@ -0,0 +1,145 @@
...
+    using NUnit.Framework;
+    using ServiceBus.Management.Infrastructure.Extensions;
+
+    [TestFixture]
+    public class NegotiatorExtensionsTests
+    {
+        [TestCase]
+        public void WithPagingLinks_ReturnsLinksWithRelativeUri()
+        {
+            var pagingHeaders = GetLinks(200, currentPage: 3, path: "/api/test");
```
danielmarbach: I would prefer to use named parameters in all the test to make it a bit more readable.

timbussmann: @danielmarbach you mean for the first parameter?

danielmarbach: yes

src/ServiceControl/Infrastructure/Extensions/NegotiatorExtensions.cs
timbussmann: strictly speaking, we want the absolute path as it's not relative to the "current location". We want to omit the scheme + authority on the URI spec:
> URI = scheme:[//authority]path[?query][#fragment]

 which `AbsolutePath` does:

> The path information does not include the scheme, host name, or query portion of the URI.

danielmarbach: Yeah exactly, like I said I had to look up what it does because in my mind I thought we have to use relative path. Your code makes sense, I was just reflecting about my thought process publicly, ignore me 💃

danielmarbach:
@timbussmann while testing I removed the ravendb changes and rebased on develop.

I found that we broke the paging with ServiceInsight. When we connect with ServiceInsight we also connect to `http://domain:port/api` when the links contain `/api` ServiceInsight uses that as part of the relative link and sends the request to `http://domain:port/api/api` which then doesn't work. `/api` is a hardcoded OWIN map, see 

https://github.com/Particular/ServiceControl/blob/develop/src/ServiceControl/Infrastructure/OWIN/Startup.cs#L34

ServicePulse works like a charm

src/ServiceControl/Infrastructure/Extensions/NegotiatorExtensions.cs
```diff
@@ -110,32 +111,32 @@ public static Negotiator WithPagingLinks(this Negotiator negotiator, int totalRe
...
         }
 
-        static void AddLink(ICollection<string> links, int page, string rel, string queryParams, Url url)
+        static void AddLink(ICollection<string> links, int page, string rel, string uriPath, string queryParams)
         {
-            url.Query = queryParams + "page=" + page;
-
-            links.Add($"<{url}>; rel=\"{rel}\"");
+            var query = queryParams + "page=" + page;
+
```
danielmarbach: @timbussmann 🔥 ;)

janovesk:
Just tested the reverse proxy scenario. Works as expected. Good stuff!

danielmarbach:
Merged into release 3.3 via https://github.com/Particular/ServiceControl/pull/1480/commits/065986dad9db7de6a07f54ebc9f21f4609c620d9

PR #1477, opened 2018-11-06
Fix concurrency issues in retrying manager

SzymonPobiega:


src/ServiceControl/Recoverability/Retrying/RetryingManager.cs
```diff
@@ -85,7 +85,7 @@ public async Task Forwarding(string requestId, RetryType retryType)
                 return;
             }
 
-            var summary = Get(requestId, retryType);
```
SzymonPobiega: No idea why this was `Get` instead of `GetOrCreate`. I looked at the invocation path and it seemed like it could (after SC restart) be invoked without prior call to `GetOrCreate` so I changed it to `GetOrCreate` to be safe.

src/ServiceControl/Recoverability/Retrying/RetryingManager.cs
```diff
@@ -97,7 +97,7 @@ public async Task ForwardedBatch(string requestId, RetryType retryType, int numb
...
```
SzymonPobiega: Same as above

src/ServiceControl/Recoverability/Retrying/RetryingManager.cs
```diff
@@ -37,7 +37,7 @@ public bool IsOperationInProgressFor(string requestId, RetryType retryType)
 
         public bool IsRetryInProgressFor(string requestId)
         {
-            return retryOperations.Keys.Where(key => key.EndsWith($"/{requestId}")).Any(key => retryOperations[key].IsInProgress());
```
SzymonPobiega: Accessing the `retryOperations` by key, even if `retryOperations` is a concurrent dict would not be safe here because there is no lock that makes sure the keys returned by `Keys` are still present when actually enumerating the returned collection. I've changed it to enumerating the values.

danielmarbach: I think you can leave it as is with the keys, `Keys` and `Values` both acquire all the locks, see https://referencesource.microsoft.com/#mscorlib/system/Collections/Concurrent/ConcurrentDictionary.cs,2036

`GetEnumerator` would be lock-free but might return dirty reads

PR #1487, opened 2018-11-13
Fix failed error import handling

SzymonPobiega:
The first commit tries to address the root cause issue.
The second commit tries to do some damage control and adds handling of failed errors that have already been moved to failed audits collection when trying to re-import the audits.


src/ServiceControl.AcceptanceTests/Audit/When_an_error_message_fails_to_import.cs
```diff
@@ -0,0 +1,137 @@
...
+                .Done(async c =>
+                {
+                    if (c.UniqueMessageId == null)
+                    {
+                        return false;
+                    }
+
+                    if (!c.WasReimported)
+                    {
+                        var countResult = await this.TryGet<FailedAuditsCountReponse>("/api/failedaudits/count");
```
danielmarbach: I'm wondering if it would be possible to make this test step ID based. As long as we create a new instance of SC with a temporary folder the count based approach will work but I think the ID based one, in this case, would be more robust.

src/ServiceControl.AcceptanceTests/Audit/When_an_error_message_fails_to_import.cs
```diff
@@ -0,0 +1,137 @@
...
+                        failedImportsCountReponse = countResult;
+                        if (countResult)
+                        {
+                            if (failedImportsCountReponse.Count > 0)
+                            {
+                                c.FailedImport = true;
+                                await this.Post<object>("/api/failedaudits/import", null, code =>
```
danielmarbach: I think this can be simplified to

```
if(await this.Post<object>("/api/failedaudits/import")) {
   c.WasReimported = true;
}
```

the Post method already checks for

`response.IsSuccessStatusCode`

src/ServiceControl/Operations/ImportFailedAudits.cs
```diff
@@ -40,14 +42,16 @@ public Task Run(CancellationTokenSource tokenSource)
                         try
                         {
                             var messageContext = new MessageContext(dto.Id, dto.Headers, dto.Body, EmptyTransaction, EmptyTokenSource, EmptyContextBag);
-                            var entity = await auditImporter.ConvertToSaveMessage(messageContext)
-                                .ConfigureAwait(false);
-                            using (var storeSession = store.OpenAsyncSession())
+
+                            //Due to bug https://github.com/Particular/PlatformDevelopment/issues/2464 some failed error could have ended up in failed audits collection
+                            if (messageContext.Headers.ContainsKey(FaultsHeaderKeys.FailedQ))
```
danielmarbach: This works under the assumption that only error messages might have been imported into the audit collection and not audit messages into the error collection. Given that assembly scanning with features and the container is pretty deterministic it sounds like a valid assumption to make. We only remove the `FaultsHeaderKeys.FailedQ` when we retry messages so we should be safe?

SzymonPobiega: The header is removed in `RetryProcessor::StageMessage` method so it safe to use it here.

SzymonPobiega: I added a failing test as the first commit to prove the issue was deterministic. Also, there is `When_a_message_fails_to_import` test in the `Audit` category that proves that the failed audits are stored in the proper collection so we're covered.

src/ServiceControl.AcceptanceTests/Audit/When_an_error_message_fails_to_import.cs
SzymonPobiega: I looked at it but it would be hard because the failed audit/error documents have auto-generated IDs so accessing them by ID would require adding an index :\

SzymonPobiega:
@danielmarbach I've pushed the failing test as the first commit and rearranged the files a bit. I also removed the reference to non-public issue.

src/ServiceControl/Operations/AuditPersister.cs
```diff
@@ -0,0 +1,70 @@
...
+    using System.Linq;
+    using System.Threading.Tasks;
+    using BodyStorage;
+    using Infrastructure;
+    using MessageAuditing;
+    using NServiceBus;
+    using NServiceBus.Transport;
+    using Raven.Client;
+
+
```
danielmarbach: 🔥

src/ServiceControl.AcceptanceTests/MessageFailures/When_a_message_fails_to_import.cs
```diff
@@ -0,0 +1,110 @@
...
+    using NServiceBus.Settings;
+    using NUnit.Framework;
+    using ServiceControl.Operations;
+
+    class When_a_message_fails_to_import : AcceptanceTest
+    {
+        [Test]
+        public async Task It_is_stored_in_the_document_collection()
+        {
+            //Make sure the audit import attempt fails
```
danielmarbach: Shouldn't the comment say

```
// Make sure the error import attempt fails
```

src/ServiceControl.AcceptanceTests/MessageFailures/When_a_message_fails_to_import.cs
```diff
@@ -0,0 +1,110 @@
...
+    using NServiceBus.AcceptanceTesting.Customization;
+    using NServiceBus.AcceptanceTests;
+    using NServiceBus.Settings;
+    using NUnit.Framework;
+    using ServiceControl.Operations;
+
+    class When_a_message_fails_to_import : AcceptanceTest
+    {
+        [Test]
+        public async Task It_is_stored_in_the_document_collection()
```
danielmarbach: Would it be helpful to say in which collection?

PR #1503, opened 2018-11-28
Hotfix 3.3.4

### LINKED ISSUE #1489: Azure Service Bus instances can't be installed on net461-only machines
As [reported by @markgould](https://github.com/Particular/ServiceControl/issues/1488) version `3.3.0` contains a bug that prevents Azure Service Bus instances to be installed due `netstandard.dll` missing from the packaged transport binaries.

### Symptoms

Instance installation fails with "Could not load file or assembly netstandard.dll"

### Who's affected

All v3.3.0 users trying to install Azure Service Bus instances on environments with .net framework prior to 4.7.2

### Workaround

Install net4.7.2 on the server.

andreasohlund:
Bump asb transport to 1.0.0-alpha174 to make sure we get the multi targeted ASB client (3.2.0). Also included Particular.Monitoring 2.0.2 to get the equivalent fix.

Closes #1489

### DIFF
--- a/src/ServiceControlInstaller.Packaging/ServiceControlInstaller.Packaging.csproj
+++ b/src/ServiceControlInstaller.Packaging/ServiceControlInstaller.Packaging.csproj
@@ -16,7 +16,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Packaging.ServiceControl.Monitoring" Version="2.0.1" PrivateAssets="All" />
+    <PackageReference Include="Packaging.ServiceControl.Monitoring" Version="2.0.2" PrivateAssets="All" />
   </ItemGroup>
 
   <UsingTask AssemblyFile="..\..\buildsupport\DeploymentZipTask.Dll" TaskName="DeploymentZipTask.AddToZip" />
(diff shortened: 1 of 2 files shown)


andreasohlund:
I've verified that this fixes the issue. @Particular/servicecontrol-maintainers please review and merge

src/ServiceControlInstaller.Packaging/ServiceControlInstaller.Packaging.csproj
```diff
@@ -16,7 +16,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Packaging.ServiceControl.Monitoring" Version="2.0.1" PrivateAssets="All" />
+    <PackageReference Include="Packaging.ServiceControl.Monitoring" Version="2.0.2" PrivateAssets="All" />
```
mikeminutillo: Why is this change bundled into the PR?

andreasohlund: Monitorid had the same issue and needs to be released as well so that’s why this is here?

andreasohlund: So 3.3.4 will contain both the SC fix and the monitoring fix

mikeminutillo: :+1:

PR #1508, opened 2018-12-03
If message consistently fails in staging phase of retries, ServiceControl is unable to retry any messages

SzymonPobiega:


src/ServiceControl/Recoverability/Retrying/RetryProcessor.cs
```diff
@@ -195,6 +198,57 @@ async Task<int> Stage(RetryBatch stagingBatch, IAsyncDocumentSession session)
...
+            catch (Exception e)
+            {
+                if (failedMessageRetry.StageAttempts < 5)
+                {
+                    await IncrementAttemptCounter(failedMessageRetry).ConfigureAwait(false);
+                    //We need to rethrow here to stop the current staging process
+                    throw new Exception($"Error while trying to stage failed message {message.UniqueMessageId}", e);
+                }
+
+                Log.Error($"Failed message {message.UniqueMessageId} reached its staging retry limit ({failedMessageRetry.StageAttempts}) and is going to be removed from the batch.", e);
```
SzymonPobiega: Not sure if I should use `Id` or `UniqueMessageId` when logging 😕

src/ServiceControl/Recoverability/Retrying/RetryProcessor.cs
```diff
@@ -195,6 +198,57 @@ async Task<int> Stage(RetryBatch stagingBatch, IAsyncDocumentSession session)
...
+        }
+
+        async Task IncrementAttemptCounter(FailedMessageRetry message)
+        {
+            try
+            {
+                await store.AsyncDatabaseCommands.PatchAsync(message.Id,
+                    new[]
+                    {
+                        new PatchRequest
```
SzymonPobiega: Existing document would not have this property set but this way of setting/incrementing it should be fine

src/ServiceControl/Recoverability/Retrying/RetryProcessor.cs
```diff
@@ -195,6 +198,57 @@ async Task<int> Stage(RetryBatch stagingBatch, IAsyncDocumentSession session)
...
+                if (failedMessageRetry.StageAttempts < 5)
+                {
+                    await IncrementAttemptCounter(failedMessageRetry).ConfigureAwait(false);
+                    //We need to rethrow here to stop the current staging process
+                    throw new Exception($"Error while trying to stage failed message {message.UniqueMessageId}", e);
+                }
+
+                Log.Error($"Failed message {message.UniqueMessageId} reached its staging retry limit ({failedMessageRetry.StageAttempts}) and is going to be removed from the batch.", e);
+
+                await store.AsyncDatabaseCommands.DeleteAsync(FailedMessageRetry.MakeDocumentId(message.UniqueMessageId), null)
```
WilliamBZA: Should the delete be the last statement perhaps? This is an atomic command meaning the document is out of the DB, so what would it mean if the domain events fails for some reason? Is it better to re-try the staging or better to not raise a domain event?

src/ServiceControl/Recoverability/Retrying/RetryProcessor.cs
mikeminutillo: UniqueMessageId is the document Id isn't it? It will make it easier to find in Maintenance mode. Also, Message Id is non-unique. You could have an event (one Id) handled by multiple subscribers. Each one gets it's own UniqueMessageId. You can think of it as a Logical Message Processing Id.

src/ServiceControl/Recoverability/Retrying/RetryProcessor.cs
```diff
@@ -195,6 +198,57 @@ async Task<int> Stage(RetryBatch stagingBatch, IAsyncDocumentSession session)
...
+            {
+                await StageMessage(message, stagingId).ConfigureAwait(false);
+            }
+            catch (Exception e)
+            {
+                if (failedMessageRetry.StageAttempts < 5)
+                {
+                    await IncrementAttemptCounter(failedMessageRetry).ConfigureAwait(false);
+                    //We need to rethrow here to stop the current staging process
+                    throw new Exception($"Error while trying to stage failed message {message.UniqueMessageId}", e);
```
mikeminutillo: Can this exception message and the one below be differentiated (so we can tell which one was thrown).
Should we also include the attempt count?
Out of interest, what happens to this exception? Is it logged? Is the inner exception logged correctly?

mikeminutillo: At the moment this exception will cause the system to wait 30 seconds before trying again. As we're going to retry up to 5 times, a truly poison message that cannot be staged will halt the retries system for 3 minutes before being able to proceed.

src/ServiceControl/Recoverability/Retrying/RetryProcessor.cs
SzymonPobiega: Not sure here. If we flip that then we can have a situation where we announced that the document has been removed from the batch while in reality it might succeed in the next staging attempt. Currently the event is only used to created an activity log entry. I don't have any preference here. @mikeminutillo ?

src/ServiceControl/Recoverability/Retrying/RetryProcessor.cs
SzymonPobiega: I tried to redesign this. Now throwing and catching a custom exception to trigger immediate retry of staging.

SzymonPobiega:
@WilliamBZA @mikeminutillo ready for another round

src/ServiceControl/Recoverability/Retrying/RetryProcessor.cs
```diff
@@ -195,6 +207,59 @@ async Task<int> Stage(RetryBatch stagingBatch, IAsyncDocumentSession session)
...
+        {
+            try
+            {
+                await StageMessage(message, stagingId).ConfigureAwait(false);
+            }
+            catch (Exception e)
+            {
+                if (failedMessageRetry.StageAttempts < MaxStagingAttempts)
+                {
+                    Log.Error($"Attempt {failedMessageRetry.StageAttempts + 1} of {MaxStagingAttempts} to stage a retry message {message.UniqueMessageId} failed", e);
```
WilliamBZA: Pedantic now, but perhaps a Warn here and an error when breaking out of the retry attempts?

SzymonPobiega: Fixed

SzymonPobiega:
@mikeminutillo @WilliamBZA how do you feel about bundling this with the maintenance release to save time & effort related to deploying SC?

WilliamBZA:
:+1: from my side

PR #1519, opened 2018-12-06
Include additional license details

WilliamBZA:
Adds the LicenseType as well as the ServiceName


src/ServiceControl/Licensing/LicenseModule.cs
```diff
@@ -23,7 +23,9 @@ public LicenseModule()
                     RegisteredTo = ActiveLicense.Details.RegisteredTo ?? string.Empty,
                     UpgradeProtectionExpiration = ActiveLicense.Details.UpgradeProtectionExpiration?.ToString("O") ?? string.Empty,
                     ExpirationDate = ActiveLicense.Details.ExpirationDate?.ToString("O") ?? string.Empty,
-                    Status = ActiveLicense.IsValid ? "valid" : "invalid"
+                    Status = ActiveLicense.IsValid ? "valid" : "invalid",
+                    LicenseType = ActiveLicense.Details.LicenseType,
```
mikeminutillo: All the other strings returned here replace a null with an empty string. Should that be done here too?

Is this because the serializer skips properties with null values?

WilliamBZA: It's because I didn't think to do it. Updated

PR #1523, opened 2018-12-07
Upgrade license package

WilliamBZA:
Upgrades the licensing package so that SP can simply look at the `GetLicenseStatus()` method without us having to re-implement the logic. This is likely going to cause conflicts with the maintenance release as @SzymonPobiega did the same upgrade on that branch. Should this wait until the maintenance release is merged to develop to make the conflicts easier to deal with?


boblangley:
Today is pencil's down for the maintenance release. We will be releasing it on Monday. With that in mind I am going to change the base to the MR branch and merge (since we already approved this change).

PR #1525, opened 2018-12-07
Support IAM Roles or EC2 metadata when using SQS transport

### LINKED ISSUE #1459: SQS Transport to accept EC2-Profile as  Credentials  Source
Allow SQS Transport customisation to accept EC2 Profile ( Instance Role ) as a credential source.  Currently the transport **requires** API Keys to be passed in connection string. Which is less secure and introduces ops complexity around key rotation.

boblangley:
Fixes #1459

### DIFF
--- a/src/ServiceControl.Transports.SQS/SQSTransportCustomization.cs
+++ b/src/ServiceControl.Transports.SQS/SQSTransportCustomization.cs
@@ -7,6 +7,7 @@
     using Amazon;
     using NServiceBus;
     using NServiceBus.Configuration.AdvancedExtensibility;
+    using NServiceBus.Logging;
     using NServiceBus.Raw;
 
     public class SQSTransportCustomization : TransportCustomization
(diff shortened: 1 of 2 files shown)


src/ServiceControl.Transports.SQS/SQSTransportCustomization.cs
```diff
@@ -29,8 +30,17 @@ static void ConfigureTransport(TransportExtensions<SqsTransport> transport, Tran
         {
             var builder = new DbConnectionStringBuilder { ConnectionString = transportSettings.ConnectionString };
 
-            PromoteEnvironmentVariableFromConnectionString(builder, "AccessKeyId", "AWS_ACCESS_KEY_ID");
-            PromoteEnvironmentVariableFromConnectionString(builder, "SecretAccessKey", "AWS_SECRET_ACCESS_KEY");
+            if (builder.ContainsKey("AccessKeyId") || builder.ContainsKey("SecretAccessKey"))
```
danielmarbach: Given that we required either both or none should we make here both mandatory and thus change the bool condition to `&&`?

danielmarbach: yeah but if we make both mandatory then we'd fall into the else part and thus fall back to the else if one is missing. Then the SDK constructor would fail. Isn't that more aligned with the case if both are not provide IAM role either then also the SDK constructor would fail?

boblangley: Using else allows the exception they receive to be more specific than the generic auth error they will receive from the SDK constructor. The else will also log something about using environment variables or IAM roles, when they were attempting to provide the credentials in the connection string. Wouldn't this be more confusing?

danielmarbach: I'm ok to follow your logic. Let's move ahead with or

AbdoHassanUMG:
Thanks guys !

boblangley:
Release 3.4.0, which now supports using IAM roles, is now available: https://particular.net/start-servicecontrol-download

PR #1545, opened 2018-12-19
Replace Old with Legacy for ASB

danielmarbach:


danielmarbach:
I'm guessing with the constants I don't need to change the matching rule to incorporate the previous name right? Because in the settings we store the customization and nothing else correct?

SeanFeldman:
@Particular/servicecontrol-maintainers this hint is very confusing:

https://github.com/Particular/ServiceControl/blob/92a7288f21b31773c848548a54eaafc84d945a39/src/ServiceControl.UnitTests/API/APIApprovals.cs#L83

Do you have any idea what is it?

Issue #1577, opened 2019-01-31
Duplicate FailedMessageIds can cause retries to fail


WilliamBZA:
If a failed message ID is repeated for some reason within a single retry batch (this could be due to indexes being corrupt, or duplicate retry attempts being received within a very short time period), then the retry processor fails as it is unable to convert the list of messages to a dictionary.

PR #1578, opened 2019-01-31
Deduplicate failed messages

WilliamBZA:
Fixes https://github.com/Particular/ServiceControl/issues/1577

### DIFF
--- a/src/ServiceControl/Recoverability/Retrying/Infrastructure/FailedMessageEqualityComparer.cs
+++ b/src/ServiceControl/Recoverability/Retrying/Infrastructure/FailedMessageEqualityComparer.cs
@@ -4,6 +4,8 @@ namespace ServiceControl.Recoverability
 {
     class FailedMessageEqualityComparer : IEqualityComparer<FailedMessageRetry>
     {
+        private static FailedMessageEqualityComparer instance = new FailedMessageEqualityComparer();
+
         public bool Equals(FailedMessageRetry x, FailedMessageRetry y)
         {
             return x.FailedMessageId == y.FailedMessageId;
@@ -12,6 +14,14 @@ namespace ServiceControl.Recoverability
         public int GetHashCode(FailedMessageRetry obj)
         {
             return obj.FailedMessageId.GetHashCode();
+        }
+
+        public static FailedMessageEqualityComparer Instance
+        {
+            get
+            {
+                return instance;
+            }
         }
     }
 }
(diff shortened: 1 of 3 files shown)


src/ServiceControl/Recoverability/Retrying/Infrastructure/FailedMessageEqualityComparer.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System.Collections.Generic;
+
+namespace ServiceControl.Recoverability
+{
+    class FailedMessageEqualityComparer : IEqualityComparer<FailedMessageRetry>
```
danielmarbach: Introduce singleton pattern on the comparer

src/ServiceControl/Recoverability/Retrying/RetryProcessor.cs
```diff
@@ -179,6 +179,7 @@ async Task<int> Stage(RetryBatch stagingBatch, IAsyncDocumentSession session)
             var failedMessageRetryDocs = await session.LoadAsync<FailedMessageRetry>(stagingBatch.FailureRetries).ConfigureAwait(false);
             var matchingFailures = failedMessageRetryDocs
                 .Where(r => r != null && r.RetryBatchId == stagingBatch.Id)
+                .Distinct(new FailedMessageEqualityComparer())
```
danielmarbach: Use singleton here

src/ServiceControl/Recoverability/Retrying/RetriesGateway.cs
```diff
@@ -41,7 +41,7 @@ public RetriesGateway(IDocumentStore store, RetryDocumentManager documentManager
 
                     if (currentBatch.Count == BatchSize)
                     {
-                        response.Add(currentBatch.ToArray());
+                        response.Add(currentBatch.Distinct().ToArray());
```
danielmarbach: Why not simply use HashSet with string comparer in this method?

WilliamBZA: What is the difference?

SzymonPobiega:
Nothing more to add besides what @danielmarbach pointed add

src/ServiceControl/Recoverability/Retrying/RetriesGateway.cs
danielmarbach: The currentBatch will never contain duplicates because the strings get deduped while adding. Plus we rely on HashSet dedup mechanism instead of LINQ generic distinct

WilliamBZA:
Changes made :+1:

danielmarbach:
This should probably target a proper hotfix branch. Other than that 👍

PR #1583, opened 2019-02-06
(GH-1505) fix bootstrapper dispose leak

### LINKED ISSUE #1505: ESent error on importing of failed messages
We currently have 2 custom checks failing for our service control instance stating that we had audit and error messages that were unable to be imported.  We followed the documentation for importing failed audit messages.  The cli said it processed the 15 failed audit messages we had but then bombed out with an esent error shown below

```
2018-11-28 18:20:11.3995|12|Info|ServiceControl.Operations.ImportFailedAudits|Done re-importing failed audits. Successfu
lly re-imported 0 messaged. Failed re-importing 0 messages.
Disposing esent resources from finalizer! You should call TransactionalStorage.Dispose() instead!
Failed to dispose esent instance from finalizer because: Microsoft.Isam.Esent.Interop.EsentTooManyActiveUsersException:
Too many active database users
   at Microsoft.Isam.Esent.Interop.Api.JetTerm2(JET_INSTANCE instance, TermGrbit grbit)
   at Raven.Storage.Esent.TransactionalStorage.Finalize()
PS C:\Windows\system32>

```

We went to look at servicepulse and noticed the custom check for audit messages that need to be imported went away but the custom check for failed error messages that need to be imported is still there.  We tried running running service control using the —import-failed-audits which says nothing needs to happen and bombs out on esent again.

My questions are:

1. Does passing the —import-failed-audits also import failed error messages as well and is just poorly named?

2. How do we resolve this esent error from happening when needing to manually import failed messages?

3. How do I get my failed error messages imported?

4. When I put servicecontrol into maintenance mode and go to the port ravendb is listening on it shows no databases which seems completely wrong.  Where is the database servicecontrol is using?

I believe our version of servicecontrol is 2.1.2.  I can confirm this in the morning.

Any help is greatly appreciated

### Root cause

The `ImportFailedAuditsCommand` does not stop/dispose the `Bootstrapper` object.

TraGicCode:
I could use some help with someone review these changes to see if they look right in regards to the using IDisposable.  Also, i'm unable to test the issue actually solves #1505 since i'm unsure how to easily stage failed audit imports locally.

Closes #1505 

/cc @WilliamBZA @danielmarbach @SzymonPobiega

### DIFF
--- a/src/ServiceControl/Bootstrapper.cs
+++ b/src/ServiceControl/Bootstrapper.cs
@@ -30,7 +30,7 @@ namespace Particular.ServiceControl
     using ServiceBus.Management.Infrastructure.OWIN;
     using ServiceBus.Management.Infrastructure.Settings;
 
-    class Bootstrapper
+    class Bootstrapper : IDisposable
     {
         // Windows Service
         public Bootstrapper(Action<ICriticalErrorContext> onCriticalError, Settings settings, EndpointConfiguration configuration, LoggingSettings loggingSettings, Action<ContainerBuilder> additionalRegistrationActions = null)
@@ -221,5 +221,13 @@ Selected Transport Customization:   {settings.TransportCustomizationType}
         private TransportSettings transportSettings;
         TransportCustomization transportCustomization;
         private static HttpClient httpClient;
+
+        public void Dispose()
+        {
+            WebApp?.Dispose();
+            documentStore?.Dispose();
+            notifier?.Dispose();
+            container?.Dispose();
+        }
     }
 }
--- a/src/ServiceControl/Bootstrapper.cs
+++ b/src/ServiceControl/Bootstrapper.cs
@@ -129,15 +129,10 @@ namespace Particular.ServiceControl
 
         public async Task Stop()
         {
-            notifier.Dispose();
             if (bus != null)
             {
                 await bus.Stop().ConfigureAwait(false);
             }
-
-            documentStore.Dispose();
-            WebApp?.Dispose();
-            container.Dispose();
         }
 
         private long DataSize()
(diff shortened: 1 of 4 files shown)


src/ServiceControl/Hosting/Host.cs
```diff
@@ -55,6 +55,7 @@ protected override void OnStop()
         protected override void Dispose(bool disposing)
         {
             OnStop();
+            bootstrapper?.Dispose();
```
TraGicCode: is this the right place to call this?  Do we also need to call bootstrapper Stop in OnPause or OnStop?

src/ServiceControl/Bootstrapper.cs
```diff
@@ -221,5 +216,14 @@ private void RecordStartup(LoggingSettings loggingSettings, EndpointConfiguratio
         private TransportSettings transportSettings;
         TransportCustomization transportCustomization;
         private static HttpClient httpClient;
+
+        public async void Dispose()
```
TraGicCode: Should i be calling Stop in the dispose?  Also because i'm calling an async method i had to add async to the dispose method signature.  Is this even valid?

danielmarbach: It is a valid signature but very dangerous because once the first await statement is reached for the caller calling dispose the call is over and the appdomain would shut down

TraGicCode:
Hey @danielmarbach,

I’m trying to solve the issue that @szymonpobiega believes is causing #1505 that is believed to happen because the bootstrapper class is using  idisposable members that never get dispose called because it’s not implementing that interface and placed in a using block throughout the code base

danielmarbach:
I think I don't understand why implementing IDisposable even matters here given that stop did the same as your newly introduced dispose. Would it be possible to get a good stack trace or a process dump when it happens so that we can have a deeper look and track down the root cause?

But I guess typing this made me realize what the problem is and I sent in a PR to cover it 

https://github.com/Particular/ServiceControl/pull/1588

TraGicCode:
Thanks @danielmarbach ,

Closing this.

PR #1601, opened 2019-03-02
Fix bug, api, and revert

boblangley:
Fixes a bug in GetBodyByIdApi
Makes the IBodyStore API more like 3.x
Reverts urlencoding (tests pass)


boblangley:
The `BodyStorageTests.Attachments_with_ids_that_contain_backslash_should_be_readable` test passes reverting the URL encoding: https://github.com/Particular/ServiceControl/blob/de3efee8cc314c254536edc0abe75b3f3700acdd/src/ServiceControl/Operations/BodyStorage/RavenAttachments/RavenAttachmentsBodyStorage.cs#L42

PR #1610, opened 2019-03-28
Update Msmq transport to 1.0.1

indualagarsamy:


andreasohlund:
Should we tweak the timeout to make the tests complete faster?

see https://github.com/Particular/NServiceBus.Transport.Msmq/blob/master/src/NServiceBus.Transport.Msmq.TransportTests/ConfigureMsmqTransportInfrastructure.cs#L17

PR #1614, opened 2019-03-29
Update to 3.5.9-patch-35280 which fixes the cancellation warnings in the logs

danielmarbach:
Should contain

https://github.com/ravendb/ravendb/pull/8512
and 
https://github.com/ravendb/ravendb/pull/8452

### DIFF
--- a/src/ServiceControl/Operations/BodyStorage/RavenAttachments/RavenAttachmentsBodyStorage.cs
+++ b/src/ServiceControl/Operations/BodyStorage/RavenAttachments/RavenAttachmentsBodyStorage.cs
@@ -5,7 +5,6 @@
     using System.Linq;
     using System.Threading;
     using System.Threading.Tasks;
-    using Nancy.Helpers;
     using Raven.Client;
     using Raven.Json.Linq;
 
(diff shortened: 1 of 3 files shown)


src/ServiceControl/Operations/BodyStorage/RavenAttachments/RavenAttachmentsBodyStorage.cs
```diff
@@ -52,7 +51,7 @@ public async Task<StreamResult> TryFetch(string bodyId)
         {
             //We want to continue using attachments for now
 #pragma warning disable 618
-            var attachment = await DocumentStore.AsyncDatabaseCommands.GetAttachmentAsync($"messagebodies/{HttpUtility.UrlEncode(bodyId)}").ConfigureAwait(false);
+            var attachment = await DocumentStore.AsyncDatabaseCommands.GetAttachmentAsync($"messagebodies/{bodyId}").ConfigureAwait(false);
```
danielmarbach: @mikeminutillo @boblangley does any of you know why we used `HttpUtility.UrlEncode` here? I was wondering if we observed a difference between how `PutAttachment` behaves vs `GetAttachment` and that's why we wrote `BodyStorageTests.cs` but I'm not sure because TryFetch is used by a nancy module called `GetBodyById`. But it seems with 3.5.9 if we store with `bodyId` and try to retrieve with the url-encoded body id RavenDB doesn't find that attachment because it is a different id now. So I feel this is a safe change but I wanted to verify with you

mikeminutillo: There was a bug in Raven that caused us to be unable to read back attachments once they had been written if the Url contained a `/`. Put worked, Get did not. We ran some tests and proved that the attachments were actually there and that urlencoding the id allowed us to retrieve them. I would guess that this Raven client has fixed the original bug and is now urlencoding the id you give it, which results in a double urlencode. If so, this fix should be OK.

mikeminutillo: Confirmed, it's here https://github.com/ravendb/ravendb/pull/8834/files

PR #1617, opened 2019-04-08
Opt-out packaging

mikeminutillo:
Converts ServiceControl Packaging to opt-out.

Rather than having to specify every single dll to include for every transport, just grab everything and rely on the ability of zip to deduplicate.

### DIFF
--- a/src/ServiceControl.Transports/ServiceControl.Transports.csproj
+++ b/src/ServiceControl.Transports/ServiceControl.Transports.csproj
@@ -6,6 +6,7 @@
 
   <ItemGroup>
     <PackageReference Include="NServiceBus.Raw" Version="3.0.0" />
+    <PackageReference Include="NServiceBus" Version="7.1.6" />
   </ItemGroup>
 
 </Project>
--- a/src/ServiceControlInstaller.Packaging/ServiceControlInstaller.Packaging.csproj
+++ b/src/ServiceControlInstaller.Packaging/ServiceControlInstaller.Packaging.csproj
@@ -35,41 +35,23 @@
     </ItemGroup>
     <Delete Files="@(OldZips)" />
     <!-- SQL Server -->
-    <AddToZip IncludeMask="NServiceBus.Transport.SQLServer.*" ZipFolder="Transports\SQLServer" SourceFolder="$(OutputPath)" ZipFileName="$(ZipToCreate)" />
-    <AddToZip IncludeMask="NServiceBus.Transports.SQLServer.*" ZipFolder="Transports\SQLServer" SourceFolder="$(OutputPath)" ZipFileName="$(ZipToCreate)" />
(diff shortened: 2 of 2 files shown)


SzymonPobiega:
@mikeminutillo why WIP?

mikeminutillo:
I was testing it. It's behaving weird for me. On this branch I cannot use the SCMU to create a new instance or upgrade an existing instance. I run into RavenDB issues of all things.

mikeminutillo:
Figured it out. The Transport packages all pulled in NServiceBus.Core 7.0.0 so that overwrote the version in ServiceControl which is 7.1.6.

SzymonPobiega:
@mikeminutillo so the idea is that the transport zips include everything that the transport contains while we exclude stuff from the SC zip, right?

It looks correct. Do you want me do do smoke tests?

mikeminutillo:
@SzymonPobiega yep. That way if a transport package as a new DLL we get it as well without having to remember to update the transport packaging project. 

 I have run some smoke tests across all of the transports in the last week. Some independent tests won't hurt.

SzymonPobiega:
@mikeminutillo I looked at the contents of the packages to make sure they match the output of a regular NSB project that uses a given transport. It looks OK so I am 👍 for it.

One outstanding thing is what to do with NSB.Core, NSB.Raw and SC.Transport dlls. I think that for the sake of consistency we should exclude them from the transport packages to make sure that SC always uses the versions it was compiled for. The potential risk might be that the transport uses a different version than it was compiled for but I'd rather track problems with the transport than with SC. What do you think?

I am a bit on the fence regarding the custom checks. I lean towards saying that the package should be included in the transports project as a dependency and that individual transport packages should not include it.

mikeminutillo:
@SzymonPobiega the only way to do that is with changes to the AddToZip deployment task to enable multiple exclusion masks. The way it works today, we couldn't come with a single mask that only targets those three dlls.

The transport projects need to have a reference to the custom checks package because we have to include a dll from them in our code. If we include in the transports projects dependencies, we'd still have to mask them out of the AddToZip task.

mikeminutillo:
@SzymonPobiega After talking to @andreasohlund, we decided that taking ownership of the AddToZip deployment task and putting into the ServiceControl repo was the best way to go. I raised a separate PR to show that https://github.com/Particular/ServiceControl/pull/1617

src/ServiceControl.Transports/ServiceControl.Transports.csproj
```diff
@@ -6,6 +6,7 @@
 
   <ItemGroup>
     <PackageReference Include="NServiceBus.Raw" Version="3.0.0" />
+    <PackageReference Include="NServiceBus" Version="7.1.6" />
```
danielmarbach: latest would be 7.1.9

mikeminutillo:
Closed in favor of https://github.com/Particular/ServiceControl/pull/1646

PR #1621, opened 2019-04-16
Add basic editing functionality

timbussmann:
This PR adds a new HTTP API to allow editing failed messages.

The basic process goes like this:
* An edit request for an existing failure is sent via HTTP API (containing both body and headers to use on the edited message).
* Once verified the given failure exists, a message is sent locally, capturing the edit intent. The client request completes.
* Once the local message is handled, the handler verifies the following first:
  * make sure that no other request edited the same message before. This happens via a dedicated document for the edit intent. It might be possible that this is a retry from a previous failed attempt in which case the handler continues (and potentially dispatches the edited message again).
  * make sure the failed message is still unresolved, otherwise the edit is discarded
* If those conditions are met, the failure is marked as resolved to prevent further edits and retries.
* A new message with the edited body and headers is created and dispatched.

### Not addressed
* ~~The API currently expects the client to always send a body and headers, regardless whether those have been edited or not. It would make sense to allow to omit any of those to let SC to fallback to the original headers/body of the failed message instead.~~ addressed: both required
* ~~Once a failed message has been edited, the failure will be marked as Resolved. There is no further state visible on that original message. Since an edited message receives a new message id, it will appear as a new failure in case it fails again (with the new body/headers). It might be worthwhile to introduce an additional state but for this basic implementation, this didn't seem to be absolutely necessary.~~ addressed: out of scope
* ~~Once the client side is more clear, the exact details of the HTTP API might need to be adjusted to match better. Currently the body is transmitted as a string but it would potentially make sense to encode it differently.~~ addressed
* ~~The edited message contains one additional header which points at the originally failed message id. There might be value in linking to a different or additional values related to the edit.~~ addressed: link to the original failed message document


src/ServiceControl.AcceptanceTests/Recoverability/EditMessageHelper.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace ServiceBus.Management.AcceptanceTests.Recoverability
+{
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Text.RegularExpressions;
+
+    public static class EditMessageHelper
+    {
+        public static List<KeyValuePair<string, string>> TryRestoreOriginalHeaderKeys(Dictionary<string, string> headers)
```
timbussmann: this wouldn't be necessary if we fixed #1629 , e.g. as suggested in this PR: https://github.com/Particular/ServiceControl/pull/1631

src/ServiceControl/Recoverability/Editing/EditHandler.cs
```diff
@@ -0,0 +1,123 @@
...
+                        return;
+                    }
+
+                    // create a retries document to prevent concurrent edits
+                    await session.StoreAsync(new FailedMessageEdit
+                    {
+                        Id = FailedMessageEdit.MakeDocumentId(message.FailedMessageId),
+                        FailedMessageId = message.FailedMessageId,
+                        EditId = context.MessageId
+                    }, Etag.Empty).ConfigureAwait(false);
```
timbussmann: I'm not sure the Etag.Empty approach indeed works. The goal was to make sure that it's not possible to overwrite an existing document (as it would have a different etag). Mabye Raven experts can verify that.

andreasohlund:
With the audit split merged to develop this will either have to be retargeted towards master (and a release-3.X) if is needs to be released immediately. If it can wait for SC 4.0 then this can be rebased against develop

mauroservienti:
@dbelcham ☝️

andreasohlund:
>and a release-3.X

Ie to create a `release-3.9` based on `master` and make this PR target the new release branch. Let me know if you want me to do this

src/ServiceControl/MessageFailures/Api/EditFailedMessages.cs
```diff
@@ -0,0 +1,128 @@
...
+                    "NServiceBus.ExceptionInfo.ExceptionType",
+                    "NServiceBus.ExceptionInfo.HelpLink",
+                    "NServiceBus.ExceptionInfo.Message",
+                    "NServiceBus.ExceptionInfo.Source",
+                    "NServiceBus.ExceptionInfo.StackTrace"
+                },
+                SensitiveHeaders = new[]
+                {
+                    "NServiceBus.ConversationId",
+                    "NServiceBus.MessageIntent"
```
DavidBoike: I think I would add

* NServiceBus.CorrelationId
* NServiceBus.ReplyToAddress
* NServiceBus.SagaId
* NServiceBus.ControlMessage
* NServiceBus.OriginatingSagaId
* NServiceBus.OriginatingSagaType
* NServiceBus.RelatedTo

https://docs.particular.net/nservicebus/messaging/headers

src/ServiceControl/MessageFailures/Api/GetErrorById.cs
```diff
@@ -54,6 +54,7 @@ private static FailedMessageView Map(FailedMessage message, IAsyncDocumentSessio
 
             var metadata = processingAttempt.MessageMetadata;
             var failureDetails = processingAttempt.FailureDetails;
+            var wasEdited = message.ProcessingAttempts.Last().Headers.ContainsKey("ServiceControl.EditOf");
```
DavidBoike: no `const` somewhere?

src/ServiceControl/MessageFailures/Api/LockedHeaderModificationValidator.cs
```diff
@@ -0,0 +1,35 @@
...
+    using System.Collections.Generic;
+    using System.Linq;
+
+    public class LockedHeaderModificationValidator
+    {
+        public bool Check(string[] lockedHeaders, IList<KeyValuePair<string, string>> editedMessageHeaders, Dictionary<string, string> originalMessageHeaders)
+        {
+            foreach (var header in lockedHeaders)
+            {
+                var originalHeader = originalMessageHeaders.FirstOrDefault(x => string.Compare(x.Key, header, StringComparison.InvariantCulture) == 0);
```
DavidBoike: Why can't the dictionary be a one that enforces this equality comparer so TryGetValue can be used? Below too, why does it have to come in as an IList of KeyValuePairs?

src/ServiceControl/Recoverability/Editing/EditHandler.cs
DavidBoike: Yep, specifying Etag.Empty is how you go about getting a concurrency exception if you're trying to create a new document but someone beat you to it.

src/ServiceControl/MessageFailures/Api/GetErrorById.cs
dbelcham: It appears that there are no constants for `ServiceControl.` prefixed headers anywhere in the codebase. Probably something that should be cleaned up. Not sure if it should be done as part of this work.

src/ServiceControl/MessageFailures/Api/EditFailedMessages.cs
```diff
@@ -0,0 +1,151 @@
...
+                    Headers.IsDeferredMessage,
+                    Headers.SagaType,
+                    Headers.OriginatingSagaType,
+                    Headers.DelayedRetries,
+                    Headers.DelayedRetriesTimestamp,
+                    Headers.ImmediateRetries,
+                    Headers.ProcessingStarted,
+                    Headers.ProcessingEnded,
+                    Headers.TimeSent,
+                    "Header"
```
mauroservienti: is `header` expected to be a real key?

dbelcham: Unfortunately it is. https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Headers.cs#L61

mauroservienti: Scary 😱

src/ServiceControl/MessageFailures/Api/EditFailedMessages.cs
```diff
@@ -0,0 +1,151 @@
...
+                    "NServiceBus.ExceptionInfo.StackTrace"
+                },
+                SensitiveHeaders = new[]
+                {
+                    Headers.RouteTo,
+                    Headers.DestinationSites,
+                    Headers.OriginatingSite,
+                    Headers.HttpTo,
+                    Headers.ReplyToAddress,
+                    Headers.NServiceBusVersion,
```
mauroservienti: Makes no sense to modify this header

src/ServiceControl/MessageFailures/Api/EditFailedMessages.cs
```diff
@@ -0,0 +1,151 @@
...
+                    Headers.ReturnMessageErrorCodeHeader,
+                    Headers.IsSagaTimeoutMessage,
```
mauroservienti: With native delayed deliveries this is becoming less and less useful, still makes no sense to modify this header

src/ServiceControl/MessageFailures/Api/EditFailedMessages.cs
```diff
@@ -0,0 +1,151 @@
...
+                    Headers.IsDeferredMessage,
```
mauroservienti: With native delayed deliveries this is becoming less and less useful, still makes no sense to modify this header

src/ServiceControl/MessageFailures/Api/EditFailedMessages.cs
```diff
@@ -0,0 +1,151 @@
...
+                    Headers.SagaType,
+                    Headers.OriginatingSagaType,
+                    Headers.DelayedRetries,
+                    Headers.DelayedRetriesTimestamp,
+                    Headers.ImmediateRetries,
```
mauroservienti: makes no sense to modify headers related to retries: delayed or immediate.

src/ServiceControl/MessageFailures/Api/EditFailedMessages.cs
```diff
@@ -0,0 +1,151 @@
...
+                    Headers.ProcessingStarted,
+                    Headers.ProcessingEnded,
```
mauroservienti: makes no sense to modify processing time related headers

src/ServiceControl/MessageFailures/Api/EditFailedMessages.cs
```diff
@@ -0,0 +1,151 @@
...
+                    Headers.RouteTo,
+                    Headers.DestinationSites,
+                    Headers.OriginatingSite,
+                    Headers.HttpTo,
+                    Headers.ReplyToAddress,
+                    Headers.ReturnMessageErrorCodeHeader,
+                    Headers.SagaType,
+                    Headers.OriginatingSagaType,
+                    Headers.TimeSent,
+                    "Header"
```
adamralph: Is there any explanation written down for why each of these of these headers should be editable?

dbelcham: No there isn't. Where would we put such a thing? It's not the normal kind of doco we would write.

adamralph: It doesn't necessarily have to be doco, but I'd like to see an explanation and justification for each field. Previously we agreed that only custom headers would be editable and all system (NSB) headers would be locked. I'd like to understand why we are allowing editing for each of these fields.

mauroservienti: The following headers, currently categorized as `sensitive` contains user owned information, that's the reasoning behind the decision to allow them to be edited.

Gateway related headers might need to be edited in case a site is renamed

- Headers.DestinationSites
- Headers.OriginatingSite
- Headers.HttpTo

Routing headers might need to be renamed in case an endpoint is renamed

- Headers.RouteTo
- Headers.ReplyToAddress

Saga types related headers might be edited if a saga type changes over time

- Headers.SagaType
- Headers.OriginatingSagaType
- Headers.TimeSent

`Headers.ReturnMessageErrorCodeHeader` is used to store (if I understood correctly) the return value when using callbacks

@dbelcham in theory `"Header"` can be moved to locked headers, it was used as a prefix, thus is not a proper header. It's probably worth consulting a Core expert to better understand what this is.

adamralph: Thanks @mauroservienti, that helps a lot. Perhaps it would be good to group the list in this way, and include those explanations as comments?

I understand the justification for most of them, but I'm struggling a bit with the saga-related headers. Why would we need to change each of those?

mauroservienti: >include those explanations as comments?

@dbelcham what do you think?

>I understand the justification for most of them, but I'm struggling a bit with the saga-related headers. Why would we need to change each of those?

User saga type changed namespace for example due to a refactoring. I'm not sure how they are used by Core. I'm not how they are used, they are populated by Core but then as far as I can tell never used.

dbelcham: > include those explanations as comments?

Is this something we normally do? I've never seen documentation about decisions included in code as a comment before in any of our codebases. My past experience says that this type of documentation will rot in the codebase over time so I would vote no to including it.

adamralph: I wonder if `Headers.SagaType` and `Headers.OriginatingSagaType` are used by SP/SI in some way? BTW do we know what the difference is between them?

I guess `Headers.TimeSent` is not related to sagas? Do we know how that's used and why it might need to be edited?

mauroservienti: If message processing depends on time and there is a large clock drift they may want to adjust that to accommodate clock drift

SagaType is the destination saga type of message I’d addressed to a saga. Originating Saga Type is if message originally come from a saga. I guess it’s used in the reply to originator scenario

PR #1631, opened 2019-04-26
Do not serialize dictionary keys using the property name strategy

### LINKED ISSUE #1629: FailedMessages API headers are serialized incorrectly
The headers on `FailedMessage` are declared as `Dictionary<string, string>`. When retrieving this data via HTTP API e.g. using `/errors/<messageId>` this dictionary is translated to an object with the dictionary keys as properties as specified here: https://www.newtonsoft.com/json/help/html/SerializeDictionary.htm.
Due to the custom serialization settings using `UnderscoreMappingResolver`, property names get changed. e.g.
`NServiceBus.MessageId` becomes `n_service_bus.message_id`

Using this approach, we're losing critical information in case we want to rebuild the headers again based on that information due to the lowecasing.
e.g. when looking at `sc.session_id` we don't know whether the header should be `SC.SessionId`, `SC.SessionID`,  or `Sc.SessionId` ...

when using message search API, the headers on `MessagesView` are defined as `IEnumerable<KeyValuePair<string, object>>` which serializes them similar to this:
```
[
  {
    "key": "SC.SessionID",
    "value": "abc-123-..."
  },
  {
     "key": "NServiceBus.MessageId",
     "value": "ec57cbd9-1f35-4af9-b739-aa31012d1937"
   },
  ...
]
```

which preserves all important information.

With the `UnderscoreMappingResolver` convention I'd argue that serializing dictionary<string, X> values should always be treated as `IEnumerable<KeyValuePair<string, X>>` to avoid loosing essential information on the dictionary (e.g. string dictionaries are case sensitive by default)

timbussmann:
This is a bit of a long shot but it actually seems to work. That way, dictionary keys won't be changed by the `ResolvePropertyName` method (as this is what the default implementation does) and it will retain the original key value.

I didn't find any other place where this would have an impact on the API but I need to admit that I'm not sure about that.

### DIFF
--- a/src/ServiceControl/Infrastructure/SignalR/UnderscoreMappingResolver.cs
+++ b/src/ServiceControl/Infrastructure/SignalR/UnderscoreMappingResolver.cs
@@ -6,6 +6,11 @@
 
     class UnderscoreMappingResolver : DefaultContractResolver
     {
+        protected override string ResolveDictionaryKey(string dictionaryKey)
+        {
+            return dictionaryKey;
+        }
+
         protected override string ResolvePropertyName(string propertyName)
         {
             return Regex.Replace(


timbussmann:
closing this PR after talking with the "edit and retry message" TF and they see no need for it.

PR #1633, opened 2019-05-01
Check remaining disk space

mikeminutillo:
If disk space runs out on the drive where ServiceControl stores it's data directory, then it can no longer ingest new messages. It also causes the retention indexes to get stopped so old data is retained forever. This custom check informs the user when the disk space remaining drops below 20%

This is what it looks like when it fires:

Running out of Disk Space

And if you click on it you get additional details:

Running out of Disk Space - Details

For the purposes of testing and taking the screenshots, I upped the limit to 40%.


mikeminutillo:
@danielmarbach if a custom check throws an exception, the infrastructure catches it and reports it as a custom check failure.

PR #1639, opened 2019-05-07
Hotfix 3.7.2

andreasohlund:
Do not add headers when retrying

### DIFF
--- a/src/ServiceControl.AcceptanceTests/Recoverability/When_a_native_integration_message_is_retried.cs
+++ b/src/ServiceControl.AcceptanceTests/Recoverability/When_a_native_integration_message_is_retried.cs
@@ -1,0 +1,99 @@
+﻿namespace ServiceBus.Management.AcceptanceTests.Recoverability
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+    using EndpointTemplates;
+    using NServiceBus;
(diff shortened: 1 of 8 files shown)


src/ServiceControl.AcceptanceTests/Recoverability/When_a_native_integration_message_is_retried.cs
```diff
@@ -0,0 +1,99 @@
...
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.MessageMutator;
+    using NServiceBus.Routing;
+    using NServiceBus.Transport;
+    using NUnit.Framework;
+    using ServiceControl.MessageFailures.Api;
+    using ServiceControlInstaller.Engine.Instances;
+    using Conventions = NServiceBus.AcceptanceTesting.Customization.Conventions;
+
+    class When_a_native_integration_message_is_retried : AcceptanceTest
```
WilliamBZA: I don't want to derail this PR, but could these not be unit tests instead of acceptance tests given how flaky the acceptance tests seem to be?

andreasohlund: I couldn't find a good way to do it via a unit test:(  Also not sure if that's a valid argument for not testing things properly. I'll bring this up in the retro notes to help making fixing the tests a prio.

PR #1641, opened 2019-05-09
[SPIKE] Print logs to console when --portable is set

SzymonPobiega:


danielmarbach:
This PR seems to be a spike that is now quite far from what develop was. Closing for now. Feel free to reopen

PR #1644, opened 2019-05-21
Exclude common files from opt-out transport packaging

mikeminutillo:
This PR takes ownership of the `AddToZip` deployment task, adds the ability to include multiple inclusion and exclusion masks, and updates the packaging project to exclude common dlls that are used across all transport packages.


danielmarbach:
Why not simply prepare a directory that has the right folders and content and then use

https://docs.microsoft.com/en-us/visualstudio/msbuild/zipdirectory-task

?

mikeminutillo:
@danielmarbach good point. I tried that and it works: https://github.com/Particular/ServiceControl/pull/1646

I definitely think that's a better approach.

mikeminutillo:
Closed in favor of https://github.com/Particular/ServiceControl/pull/1646

PR #1647, opened 2019-05-27
Add setting for free disk space threhsold

mikeminutillo:
Adds a setting for setting the threshold of the free disk space custom check. This setting is an integer percentage (0-100). If the amount of remaining disk space drops below that percentage (of total space) then the custom check triggers an error warning the user to check their space.

Doco https://github.com/Particular/docs.particular.net/pull/4374

### DIFF
--- a/src/ServiceControl/Infrastructure/Settings/Settings.cs
+++ b/src/ServiceControl/Infrastructure/Settings/Settings.cs
@@ -51,6 +51,7 @@
             HttpDefaultConnectionLimit = SettingsReader<int>.Read("HttpDefaultConnectionLimit", 100);
             DisableRavenDBPerformanceCounters = SettingsReader<bool>.Read("DisableRavenDBPerformanceCounters", true);
             RemoteInstances = GetRemoteInstances();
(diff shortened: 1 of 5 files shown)


src/ServiceControl/Infrastructure/Settings/Settings.cs
```diff
@@ -501,6 +504,27 @@ static string Subscope(string address)
...
 
+        int GetDataSpaceRemainingThreshold()
+        {
+            string message;
+            var threshold = SettingsReader<int>.Read("DataSpaceRemainingThreshold", DataSpaceRemainingThresholdDefault);
+            if (threshold < 0)
+            {
+                message = $"{nameof(DataSpaceRemainingThreshold)} is invalid, minimum value is 0.";
+                logger.Fatal(message);
+                throw new Exception(message);
```
WilliamBZA: Pedantic comment, but is `Exception` the correct type?

mikeminutillo: We could add a custom exception type. I was just following the pattern that is already in place. All of the other settings throw Exception if they are invalid. There's 17 other instances.

PR #1648, opened 2019-05-27
Remove unused licensing code

andreasohlund:
@mikeminutillo as part of the work to split audit out we found the code that looks for unlicensed endpoints via the audit and error queues are not used (same on `develop`) do you remember if this was intentional?


mikeminutillo:
It appears that we dropped it as a part of https://github.com/Particular/ServiceControl/pull/963 and didn't notice. That's probably because neither Pulse or Insight actually shows license status and there were no tests asserting it's correctness. Let me confirm that it's not used. If so, it's probably worth dropping  `EndpointsView.LicenseStatus` as well .

mikeminutillo:
ServiceInsight does not use this info. I'm fairly certain that ServicePulse doesn't either.

andreasohlund:
@mikeminutillo I've dropped the license status from the endpoints view as well since SP doesn't use it https://github.com/Particular/ServicePulse/search?q=LicenseStatus&unscoped_q=LicenseStatus

PR #1651, opened 2019-05-27
[WIP]Fix failed audit import custom check

andreasohlund:
Still WIP since I still need to trigger the custom check properly. 

@danielmarbach thoughts on this approach with a bit more generic event? (lets talk it through on Wednesday)


danielmarbach:
What else do you see us piggybacking in the dictionary? I'm a bit sceptic
because it requires more complex code instead of just publishing a concrete
event in the startup class. Plus I'm more of a fan of concrete events if we
can but happy to hear out your arguments

Andreas Öhlund <notifications@github.com> schrieb am Mo., 27. Mai 2019,
22:03:

andreasohlund:
The FreeDiskSpace check would also use this
https://github.com/Particular/ServiceControl/pull/1647 and I think there is
a few more candidates. Let’s discuss when U’re back?

src/ServiceControl.Audit/Infrastructure/PublishFullTypeNameOnlyBehavior.cs
```diff
@@ -0,0 +1,29 @@
+﻿namespace Particular.ServiceControl
+{
+    using System;
+    using System.Threading.Tasks;
+    using NServiceBus;
+    using NServiceBus.Pipeline;
+
+    class PublishFullTypeNameOnlyBehavior : Behavior<IOutgoingPhysicalMessageContext>
```
andreasohlund: FYI This is needed for any integration events to work so perhaps I should make this a separate PR?

danielmarbach:
I guess we can close this @andreasohlund ?

PR #1661, opened 2019-06-04
Fix ASQ Acceptence Test

mikeminutillo:
[This test](https://github.com/Particular/ServiceControl/blob/develop/src/ServiceControl.AcceptanceTests/Recoverability/When_a_native_integration_message_is_retried.cs#L37) is trying to skip an assertion for the Azure Storage Queues transport but the strings didn't match because they were coming from different sources. 

This PR should fix that test.

I moved all of the other transports across too so that this sort of issue doesn't happen again.

I didn't move the Learning transport because it's not in the source list.

### DIFF
--- a/src/ServiceControl.AcceptanceTests/ConfigureEndpointAzureServiceBusNetStandardTransport.g.cs
+++ b/src/ServiceControl.AcceptanceTests/ConfigureEndpointAzureServiceBusNetStandardTransport.g.cs
@@ -4,6 +4,7 @@ using NServiceBus;
 using NServiceBus.AcceptanceTesting.Support;
 using ServiceBus.Management.AcceptanceTests;
 using ServiceControl.Transports.ASBS;
+using ServiceControlInstaller.Engine.Instances;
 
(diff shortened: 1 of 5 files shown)


src/ServiceControl.AcceptanceTests/ConfigureEndpointAzureServiceBusNetStandardTransport.g.cs
```diff
@@ -29,7 +29,7 @@ public Task Cleanup()
         return Task.FromResult(0);
     }
 
-    public string Name => "AzureServiceBus .NET Standard";
+    public string Name => ServiceControlInstaller.Engine.Instances.TransportNames.AzureServiceBus;
```
danielmarbach: How about adding

```
using ServiceControlInstaller.Engine.Instances
```

mikeminutillo: Done. Resharper wouldn't do it automatically (maybe because of the extern statement?) and I forgot to fix it before I pushed.

PR #1674, opened 2019-06-10
Remove domain events from audit instance

andreasohlund:
This also adds an enricher context so we can pass the `IMessageSession` to the enrichers better


src/ServiceControl.Audit/Auditing/AuditEnricherContext.cs
```diff
@@ -0,0 +1,26 @@
...
+            Headers = headers;
+            this.outgoingEvents = outgoingEvents;
+            Metadata = metadata;
+        }
+
+        public IReadOnlyDictionary<string, string> Headers { get; }
+
+        public IDictionary<string, object> Metadata { get; }
+
+        public void Emit(IEvent @event)
```
danielmarbach: Would `Add`, `AddForPublish`, `AddForEmit` or something similar be a better name?

andreasohlund: I like `AddForPublish` since we are still tied to a publish operation in the end. Renamed...

PR #1675, opened 2019-06-11
Add connection switching

mikeminutillo:
Lets all of the acceptance test projects get their connection details from a file called `connection.txt` in the same folder as the solution. If the file is missing, then it falls back to the old behavior (check environment variables and then fall back to MSMQ).

This lets you switch transport/connection quickly in development. You will need to rebuild after a change of transport type as the ATTs delete the non-selected transport assemblies at startup.


danielmarbach:
Love this change

PR #1676, opened 2019-06-11
Send messages to ServiceControl instead of doing pub/sub

danielmarbach:


src/ServiceControl.Audit/Infrastructure/NServiceBusFactory.cs
```diff
@@ -56,7 +62,7 @@ public static Task<IStartableEndpoint> Create(Settings.Settings settings, Transp
             if (!isRunningAcceptanceTests)
             {
                 // Sending the custom check to ourselves and then republish it to the primary
```
andreasohlund: This comment can be removed

PR #1686, opened 2019-06-14
Powershell to Convert or Split an instance (introducing Audit instances)

mikeminutillo:
Powershell cmdlets to
- convert an existing ServiceControl instance into an Audit instance (Invoke-ServiceControlInstanceConvert)
- split an existing ServiceControl instance into a Main and an Audit instance (Invoke-ServiceControlInstanceSplit)

Additionally, there is code preventing a user from performing the wrong type of upgrade (Upgrade, Convert, or Split) that directs them to the correct Powershell

Alos added cmdlets and upgrade code to handle RemoteInstances as we need to be able to manage them during the upgrade to 4.0 and during a split operation

- Get-ServiceControlRemotes gives you a list of all of the remote instances for a given servicecontrol instance
- Add-ServiceControlRemote lets you add one or more remote instances to a ServiceControl instance
- Remote-ServiceControlRemote lets you remove one or more remote instances from a ServiceControl instance


src/ServiceControlInstaller.Engine/Configuration/ServiceControl/SettingsList.cs
```diff
@@ -65,5 +65,11 @@ public static class ServiceControlSettings
             Name = "ServiceControl/MaintenanceMode",
             SupportedFrom = new Version(1, 18, 1)
         };
+
+        public static SettingInfo RemoteInstances = new SettingInfo
+        {
+            Name = "ServiceControl/RemoteInstances",
+            SupportedFrom = new Version(1, 47, 0)
```
danielmarbach: Shouldn't this be 2.0?

mikeminutillo: Not according to https://github.com/Particular/ServiceControl/releases/tag/1.47.0

src/ServiceControlInstaller.Engine/Unattended/UnattendServiceControlSplitter.cs
```diff
@@ -0,0 +1,187 @@
...
+                AuditQueue = source.AuditQueue,
+                AuditLogQueue = source.AuditLogQueue,
+                ForwardAuditMessages = source.ForwardAuditMessages,
+                AuditRetentionPeriod = source.AuditRetentionPeriod,
+                TransportPackage = source.TransportPackage,
+                ConnectionString = source.ConnectionString,
+                HostName = source.HostName,
+                Name = $"{source.Name}.Audit",
+                ServiceAccount = source.Service.Account,
+                // NOTE: This should always be blank
```
HEskandari: hmmm what should be blank? ☝️

src/ServiceControlInstaller.Engine/Unattended/UnattendServiceControlToAuditInstanceConverter.cs
```diff
@@ -0,0 +1,176 @@
...
+                ConnectionString = source.ConnectionString,
+                InstallPath = source.InstallPath,
+                DBPath = source.DBPath,
+                LogPath = source.LogPath,
+                HostName = source.HostName,
+                Port = source.Port,
+                DatabaseMaintenancePort = source.DatabaseMaintenancePort,
+                Name = source.Name,
+                ServiceAccount = source.Service.Account,
+                // NOTE: This should always be blank
```
HEskandari: Sam as above

src/ServiceControlInstaller.Engine/Unattended/UnattendServiceControlSplitter.cs
mikeminutillo: The password. We never read it from the instance so it should always be blank

PR #1691, opened 2019-06-20
Make audit expiry optional

andreasohlund:
Myself and @mikeminutillo thought that it would be best to make audit expiry optional in SC so that the SCMU can omit it when installing new instances

### DIFF
--- a/src/ServiceControl.UnitTests/ApprovalFiles/APIApprovals.PlatformSampleSettings.approved.txt
+++ b/src/ServiceControl.UnitTests/ApprovalFiles/APIApprovals.PlatformSampleSettings.approved.txt
@@ -35,7 +35,7 @@
   "IngestErrorMessages": true,
   "RunRetryProcessor": true,
   "ExpirationProcessTimerInSeconds": 600,
-  "AuditRetentionPeriod": "10.00:00:00",
+  "AuditRetentionPeriod": null,
   "ErrorRetentionPeriod": "10.00:00:00",
   "EventsRetentionPeriod": "14.00:00:00",
   "ExpirationProcessBatchSize": 65512,
--- a/src/ServiceControl.UnitTests/ApprovalFiles/APIApprovals.PlatformSampleSettings.approved.txt
+++ b/src/ServiceControl.UnitTests/ApprovalFiles/APIApprovals.PlatformSampleSettings.approved.txt
@@ -17,18 +17,18 @@
   "DisableRavenDBPerformanceCounters": true,
   "SkipQueueCreation": false,
   "RunCleanupBundle": false,
-  "RootUrl": "http://localhost:8888/",
+  "RootUrl": "http://localhost:33333/",
   "DatabaseMaintenanceUrl": "http://localhost:33334",
-  "ApiUrl": "http://localhost:8888/api",
-  "StorageUrl": "http://localhost:8888/storage",
-  "Port": 8888,
+  "ApiUrl": "http://localhost:33333/api",
+  "StorageUrl": "http://localhost:33333/storage",
+  "Port": 33333,
   "DatabaseMaintenancePort": 33334,
   "ExposeRavenDB": false,
   "Hostname": "localhost",
   "VirtualDirectory": "",
   "HeartbeatGracePeriod": "00:00:40",
-  "TransportCustomizationType": "ServiceControl.Transports.Learning.LearningTransportCustomization, ServiceControl.Transports.Learning",
-  "DbPath": "C:\\DB",
+  "TransportCustomizationType": "ServiceControl.Transports.Msmq.MsmqTransportCustomization, ServiceControl.Transports.Msmq",
+  "DbPath": "C:\\ProgramData\\Particular\\ServiceControl\\localhost-33333",
   "ErrorLogQueue": "error.log",
   "ErrorQueue": "error",
   "ForwardErrorMessages": false,
@@ -46,6 +46,13 @@
   "ProcessRetryBatchesFrequency": "00:00:30",
   "MaximumConcurrencyLevel": 10,
   "RetryHistoryDepth": 10,
-  "RemoteInstances": [],
+  "RemoteInstances": [
+    {
+      "ApiUri": "http://instance1"
+    },
+    {
+      "ApiUri": "http://instance2"
+    }
+  ],
   "DataSpaceRemainingThreshold": 20
 }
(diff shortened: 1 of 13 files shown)


src/ServiceControl/Infrastructure/RavenDB/Expiration/ExpiredDocumentsCleaner.cs
```diff
@@ -12,13 +12,7 @@ class ExpiredDocumentsCleaner
...
         public static Task<TimerJobExecutionResult> RunCleanup(int deletionBatchSize, DocumentDatabase database, Settings settings, CancellationToken token)
         {
-            var threshold = SystemTime.UtcNow.Add(-settings.AuditRetentionPeriod);
-
-            logger.Debug("Trying to find expired ProcessedMessage and SagaHistory documents to delete (with threshold {0})", threshold.ToString(Default.DateTimeFormatsToWrite, CultureInfo.InvariantCulture));
-            AuditMessageCleaner.Clean(deletionBatchSize, database, threshold, token);
-            SagaHistoryCleaner.Clean(deletionBatchSize, database, threshold, token);
-
-            threshold = SystemTime.UtcNow.Add(-settings.ErrorRetentionPeriod);
+            var threshold = SystemTime.UtcNow.Add(-settings.ErrorRetentionPeriod);
```
danielmarbach: I know it was there before but what is the purpose of using RavenDB SystemTime vs DateTime.UtcNow?

src/ServiceControl/Infrastructure/RavenDB/Expiration/ExpiredDocumentsCleaner.cs
```diff
@@ -28,6 +22,15 @@ public static Task<TimerJobExecutionResult> RunCleanup(int deletionBatchSize, Do
             logger.Debug("Trying to find expired EventLogItem documents to delete (with threshold {0})", threshold.ToString(Default.DateTimeFormatsToWrite, CultureInfo.InvariantCulture));
             EventLogItemsCleaner.Clean(deletionBatchSize, database, threshold, token);
 
+            if (settings.AuditRetentionPeriod.HasValue)
+            {
+                threshold = SystemTime.UtcNow.Add(-settings.AuditRetentionPeriod.Value);
```
danielmarbach: I know it was there before but what is the purpose of using RavenDB SystemTime vs DateTime.UtcNow?

src/ServiceControl/Bootstrapper.cs
```diff
@@ -164,7 +164,7 @@ private void RecordStartup(LoggingSettings loggingSettings, EndpointConfiguratio
             var startupMessage = $@"
 -------------------------------------------------------------
 ServiceControl Version:             {version}
-Audit Retention Period:             {settings.AuditRetentionPeriod}
+Audit Retention Period (optional):             {settings.AuditRetentionPeriod}
```
danielmarbach: Fix spacing?

src/ServiceControl.UnitTests/ApprovalFiles/APIApprovals.PlatformSampleSettings.approved.txt
```diff
@@ -17,25 +17,25 @@
   "DisableRavenDBPerformanceCounters": true,
   "SkipQueueCreation": false,
   "RunCleanupBundle": false,
-  "RootUrl": "http://localhost:8888/",
+  "RootUrl": "http://localhost:33333/",
```
danielmarbach: Why his this whole file be changed? I'm pretty sure originally we set explicit values for the test. The only thing that should be changed in this file is the AuditRentionPeriod right?

src/ServiceControl.UnitTests/ApprovalFiles/APIApprovals.RootPathValue.approved.txt
```diff
@@ -1,5 +1,5 @@
 {
-  "description": "testDescription",
+  "description": "The management backend for the Particular Service Platform",
```
danielmarbach: I don't think these changes are necessary. The test sets explicit fake values.

src/ServiceControl.UnitTests/ApprovalFiles/APIApprovals.PlatformSampleSettings.approved.txt
mikeminutillo: There is something weird going on here. I reverted https://github.com/Particular/ServiceControl/pull/1691/commits/76f4de67b659c692e17e521093e0f31085085d40 locally and ran into this approval test failing on the RootUrl. I looked and the appSettings that in the unit test folder are nothing like the ones in the app.config of the unit tests (and I'm not entirely sure how they ever were). We also have some tests that explicitly write to the ServiceControl.exe.config file and then make assertions about their state, so there is a timing issue there.

mikeminutillo: Ah. Because we don't use the ServiceControl.exe.config ones. We use the ServiceControl.UnitTests.dll.config.
We still destroy that file in a later test and it's not recopied (because the one in the bin folder is newer) on a build. Anyway, that last commit is wrong. What's the best way forward? Force-push after a revert?

src/ServiceControl/Infrastructure/RavenDB/Expiration/ExpiredDocumentsCleaner.cs
mikeminutillo: It has been this way since the expiration bundle was first written (Feb 2014). As far as I know, only the saga audit tests use SystemTime.

andreasohlund: ok, lets keep as is for now?

src/ServiceControl/Infrastructure/RavenDB/Expiration/ExpiredDocumentsCleaner.cs
andreasohlund: Same as above?

src/ServiceControl.UnitTests/ApprovalFiles/APIApprovals.RootPathValue.approved.txt
andreasohlund: Turned out that some test muck around with the config file, did a clean rebuild and only ran the approvals so they should be back to the original state now

andreasohlund:
This one should be good to go now

PR #1694, opened 2019-06-26
Upgrading SC instance through the SCMU

HEskandari:


src/ServiceControl.Config/Commands/UpgradeServiceControlInstanceCommand.cs
```diff
@@ -171,15 +181,103 @@ public override async Task ExecuteAsync(InstanceDetailsViewModel model)
                 }
             }
 
+            if (!instance.VersionHasServiceControlAuditFeatures)
+            {
+                auditViewModel = auditUpgradeViewModelFactory(instance.Name);
+                if (windowManager.ShowInnerDialog(auditViewModel) == true)
+                {
+                    upgradeOptions.InstallNewAuditSidecar = true;
```
mikeminutillo: Does this need to be added to the upgrade options? It only ever gets referred to from this method

src/ServiceControl.Config/Commands/UpgradeServiceControlInstanceCommand.cs
```diff
@@ -171,15 +181,103 @@ public override async Task ExecuteAsync(InstanceDetailsViewModel model)
                 }
             }
 
+            if (!instance.VersionHasServiceControlAuditFeatures)
```
mikeminutillo: What about the case where audit was already disabled?

src/ServiceControl.Config/UI/Upgrades/AddNewAuditInstanceViewModelValidator.cs
```diff
@@ -0,0 +1,68 @@
...
+                .WithMessage(Validations.MSG_MUST_BE_UNIQUE, "Paths")
+                .When(x => x.SubmitAttempted);
+
+            RuleFor(x => x.ServiceControlAudit.AuditForwarding)
+                .NotNull().WithMessage(Validations.MSG_SELECTAUDITFORWARDING);
+
+            RuleFor(x => x.ServiceControlAudit.AuditQueueName)
+                .NotEmpty()
+                .NotEqual(x => x.ServiceControlAudit.AuditForwardingQueueName).WithMessage(Validations.MSG_UNIQUEQUEUENAME, "Audit Forwarding")
+                .When(x => x.SubmitAttempted && x.ServiceControlAudit.AuditQueueName != "!disable");
```
mikeminutillo: We wouldn't have an Audit Queue `!disable` in an Audit Instance. You'd just remove the audit instance

src/ServiceControl.Config/Commands/UpgradeServiceControlInstanceCommand.cs
HEskandari: Possibly not. I wasn't sure if it is needed for other purposes. I can remove for now.

src/ServiceControl.Config/Commands/UpgradeServiceControlInstanceCommand.cs
HEskandari: That's in the PoA to be done ;) We can include the check it in this PR though.

src/ServiceControl.Config/UI/Upgrades/AddNewAuditInstanceViewModelValidator.cs
HEskandari: I've copied the validator code verbatim from the existing validation logic of the Add screen. Should it be removed altogether or just here?

PR #1700, opened 2019-06-28
storing remote instance address when installing

HEskandari:


src/ServiceControl.Config/UI/InstanceAdd/ServiceControlAddAttachment.cs
```diff
@@ -85,7 +85,6 @@ async Task Add(object arg)
                 HostName = viewModel.ServiceControlAudit.HostName,
                 Port = Convert.ToInt32(viewModel.ServiceControlAudit.PortNumber),
                 DatabaseMaintenancePort = Convert.ToInt32(viewModel.ServiceControlAudit.DatabaseMaintenancePortNumber),
-                VirtualDirectory = null, // TODO
```
danielmarbach: So the reason to remove this is

> This setting is provided for backward compatibility and should be considered obsolete.

https://docs.particular.net/servicecontrol/creating-config-file#host-settings-servicecontrolvirtualdirectory

so should it be removed completely then (happy for a different PR)

src/ServiceControlInstaller.Engine/Instances/ServiceControlNewInstance.cs
```diff
@@ -37,8 +37,22 @@ public override void WriteConfigurationFile()
 
         public override string DirectoryName => "ServiceControl";
 
+        public List<RemoteInstanceSetting> RemoteInstances { get; set; } = new List<RemoteInstanceSetting>();
```
danielmarbach: I was thinking if it would be better to use a HashSet and more the logic for filtering into the comparer but I guess it doesn't matter really here

src/ServiceControl.Config/UI/InstanceAdd/ServiceControlAddAttachment.cs
HEskandari: I'm happy to remove that (and some other code around the virtual directory thingy) altogether in a separate PR.

src/ServiceControlInstaller.Engine/Instances/ServiceControlNewInstance.cs
HEskandari: This 'pattern' has been used in 2-3 different places. We can refactor it and move everything into a custom collection of sort, but considering there isn't much logic around it, I thought it'd be overkill.

PR #1702, opened 2019-06-28
Support installing audit instances with all mandatory parameters

danielmarbach:


danielmarbach:
Discussed with @mikeminutillo and we decided to not merge this because users have the ability to install a service control instance and an audit instance by orchestrating two Powershell scripts

PR #1706, opened 2019-07-02
Unify upgrade into a single cmdlet

mikeminutillo:


src/ServiceControlInstaller.PowerShell/Cmdlets/ServiceControlInstances/InvokeServiceControlInstanceUpgrade.cs
```diff
@@ -9,24 +9,44 @@ namespace ServiceControlInstaller.PowerShell
...
 
     [Cmdlet(VerbsLifecycle.Invoke, "ServiceControlInstanceUpgrade")]
     public class InvokeServiceControlInstanceUpgrade : PSCmdlet
     {
-        [Parameter(HelpMessage = "Specify the timespan to keep Audit Data")]
-        [ValidateTimeSpanRange(MinimumHours = 1, MaximumHours = 8760)] //1 hour to 365 days
-        public TimeSpan? AuditRetentionPeriod { get; set; }
+        [ValidateNotNullOrEmpty]
+        [Parameter(Mandatory = true, ValueFromPipeline = true, ValueFromPipelineByPropertyName = true, Position = 0, HelpMessage = "Specify the name of the ServiceControl Instance to update")]
+        public string[] Name;
```
danielmarbach: I find this one a bit of an oddball. The script assumed that you can update multiple instances, how would that work with mandatory InstallPath, DBPath etc.?

I can see the benefit of having the name array but then we have to have InstallPath, DBPath and other required settings for the audit instance to be arrays as well or change the behavior so that multiple names can be provided as long as no split is required, if a split is required we ask for an invocation with the mandatory arguments

PR #1709, opened 2019-07-05
Powershell changes

danielmarbach:


src/ServiceControlInstaller.Engine/Instances/ServiceControlInstance.cs
```diff
@@ -147,7 +147,12 @@ protected override void ApplySettingsChanges(KeyValueConfigurationCollection set
             settings.Set(ServiceControlSettings.LogPath, LogPath);
             settings.Set(ServiceControlSettings.ForwardAuditMessages, ForwardAuditMessages.ToString(), Version);
             settings.Set(ServiceControlSettings.ForwardErrorMessages, ForwardErrorMessages.ToString(), Version);
-            settings.Set(ServiceControlSettings.AuditRetentionPeriod, TimeSpan.FromHours(1).ToString(), Version); //TODO: Remove in future
+
+            if (AuditRetentionPeriod.HasValue)
+            {
+                settings.Set(ServiceControlSettings.AuditRetentionPeriod, AuditRetentionPeriod.Value.ToString(), Version);
```
mikeminutillo: `settings.Set` is how we remove config values that aren't there. Maybe add an `else` that calls `RemoveIfRetired`?

src/ServiceControlInstaller.Engine/Instances/ServiceControlInstance.cs
```diff
@@ -101,7 +101,7 @@ public override void Reload()
             LogPath = AppConfig.Read(ServiceControlSettings.LogPath, DefaultLogPath());
             DBPath = AppConfig.Read(ServiceControlSettings.DBPath, DefaultDBPath());
             AuditQueue = AppConfig.Read(ServiceControlSettings.AuditQueue, (string)null);
-            AuditLogQueue = AppConfig.Read(ServiceControlSettings.AuditLogQueue, (string)null);
+            AuditLogQueue = AppConfig.Read(ServiceControlSettings.AuditLogQueue, string.IsNullOrEmpty(AuditQueue) ? null : $"{AuditQueue}.log");
```
mikeminutillo: Why do we need to do this? Is it possible there are still instances that are relying on this default somewhere? If you've ever been modified by SCMU then we should have saved this setting into the file long ago

src/ServiceControlInstaller.Engine/Instances/ServiceControlInstance.cs
danielmarbach: Would

```
settings.Set(ServiceControlSettings.AuditRetentionPeriod, AuditRetentionPeriod.ToString(), Version); 
```

work as well?

mikeminutillo: Yes it would

danielmarbach: Apparently, we have not retired the setting because existing instances that got converted still need it. Maybe we should retire the setting then?

mikeminutillo: Ah. Now that you mention it, we can't actually retire the setting. Existing instances can still have Audit messages in them after the upgrade so we need AuditRetentionPeriod to stay. 

I think `settings.Set` is the best option because it will remove the setting and then only re-add it if it is there

src/ServiceControlInstaller.Engine/Configuration/ServiceControl/SettingsList.cs
```diff
@@ -45,7 +45,8 @@ public static class ServiceControlSettings
         public static SettingInfo AuditRetentionPeriod = new SettingInfo
         {
             Name = "ServiceControl/AuditRetentionPeriod",
-            SupportedFrom = new Version(1, 12, 1)
+            SupportedFrom = new Version(1, 12, 1),
+            RemovedFrom = new Version(4, 0, 0)
```
mikeminutillo: We need to drop this. We can't deprecate the setting until we remove audit data from the main instance

danielmarbach: Good catch forgot to remove

PR #1710, opened 2019-07-05
Refactoring validation

HEskandari:
This PR has a few things:

- Using one set of validation for Ports and Paths between different instances.
- Checking all instances for used folders, ports and queues except for audit. 
- Slightly better validation messages
- Excluding Audit queue from duplication check
- Using [InjectValidation] for the upgrade view model (to make it consistent with other VMs)
- Not writing the AuditQueueLog/ErrorQueueLog if the forwarding options are not turned on.
- Fixing an issue with the auto-refresh of the instances

### DIFF
--- a/src/ServiceControl.Config/Extensions/Validations.cs
+++ b/src/ServiceControl.Config/Extensions/Validations.cs
@@ -2,45 +2,52 @@
 {
     using System;
     using System.Collections.Generic;
-    using System.Collections.ObjectModel;
     using System.Linq;
     using ServiceControlInstaller.Engine.Instances;
 
-    public static class ServiceControlValidationExtension
+    public static class Validations
     {
-        // We need this to ignore the instance that represents the edit screen
-        public static List<string> UsedPaths(this ReadOnlyCollection<ServiceControlInstance> ServiceControlInstances, string instanceName = null)
+        public static List<string> UsedPaths(string instanceName = null)
         {
-            return ServiceControlInstances
+            var monitoringInstances = InstanceFinder.MonitoringInstances();
+            var serviceControlAuditInstances = InstanceFinder.ServiceControlAuditInstances();
+            var serviceControlInstances = InstanceFinder.ServiceControlInstances();
+            var result = new List<string>();
+
+            result.AddRange(monitoringInstances
                 .Where(p => string.IsNullOrWhiteSpace(instanceName) || p.Name != instanceName)
                 .SelectMany(p => new[]
                 {
-                    p.DBPath,
                     p.LogPath,
                     p.InstallPath
-                })
-                .Distinct()
-                .ToList();
-        }
+                }));
 
-        public static List<string> UsedPaths(this ReadOnlyCollection<ServiceControlAuditInstance> ServiceControlAuditInstances, string instanceName = null)
-        {
-            return ServiceControlAuditInstances
+            result.AddRange(serviceControlAuditInstances
                 .Where(p => string.IsNullOrWhiteSpace(instanceName) || p.Name != instanceName)
                 .SelectMany(p => new[]
                 {
-                    p.DBPath,
                     p.LogPath,
+                    p.DBPath,
                     p.InstallPath
-                })
-                .Distinct()
-                .ToList();
+                }));
+
+            result.AddRange(serviceControlInstances
+                .Where(p => string.IsNullOrWhiteSpace(instanceName) || p.Name != instanceName)
+                .SelectMany(p => new[]
+                {
+                    p.LogPath,
+                    p.DBPath,
+                    p.InstallPath
+                }));
+
+            return result.Distinct().ToList();
         }
 
         // We need this to ignore the instance that represents the edit screen
-        public static List<string> UsedQueueNames(this ReadOnlyCollection<ServiceControlInstance> ServiceControlInstances, TransportInfo transportInfo = null, string instanceName = null, string connectionString = null)
+        public static List<string> UsedErrorQueueNames(TransportInfo transportInfo = null, string instanceName = null, string connectionString = null)
         {
-            var instancesByTransport = ServiceControlInstances.Where(p => p.TransportPackage.Equals(transportInfo) &&
+            var serviceControlInstances = InstanceFinder.ServiceControlInstances();
+            var instancesByTransport = serviceControlInstances.Where(p => p.TransportPackage.Equals(transportInfo) &&
                                                                           string.Equals(p.ConnectionString, connectionString, StringComparison.OrdinalIgnoreCase)).ToList();
 
             return instancesByTransport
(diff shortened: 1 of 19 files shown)


src/ServiceControl.Config/UI/Shell/ShellView.xaml.cs
```diff
@@ -11,7 +11,11 @@ public ShellView()
         {
             InitializeComponent();
 
-            Activated += (s, e) => IoC.Get<IEventAggregator>().PublishOnUIThread(new RefreshInstances());
+            Activated += (s, e) =>
+            {
+                //IoC.Get<IEventAggregator>().PublishOnUIThread(new RefreshInstances());
```
danielmarbach: Why no longer over the event broker?

src/ServiceControl.Config/UI/Shell/ShellViewModel.cs
```diff
@@ -91,7 +91,7 @@ protected override void OnInitialize()
             RefreshInstances();
         }
 
-        void RefreshInstances()
+        public void RefreshInstances()
```
danielmarbach: I guess this was needed because we no longer dispatch over the event broker?

src/ServiceControl.Config/UI/Upgrades/AddNewAuditInstanceViewModel.cs
```diff
@@ -3,8 +3,10 @@
     using System.Windows.Input;
     using Framework.Rx;
     using InstanceAdd;
+    using Validar;
     using Validation;
 
+    [InjectValidation]
```
danielmarbach: What sorcery is this?

src/ServiceControlInstaller.Engine/Interfaces.cs
```diff
@@ -73,7 +73,16 @@ public interface IDatabaseMaintenanceSupport : IVersionInfo
         int? DatabaseMaintenancePort { get; }
     }
 
-    public interface IServiceControlAuditInstance : IServiceInstance, IServiceControlPaths, IHttpInstance, IInstallable, IDatabaseMaintenanceSupport, ITransportConfig
+    public interface IServiceControlBaseInstance : IHttpInstance,
```
danielmarbach: Looks like this interface is nowhere really needed but here. So this is really just convenience to make sure we don't have to declare all those types?

src/ServiceControlInstaller.Engine/Validation/QueueNameValidator.cs
```diff
@@ -112,17 +118,23 @@ internal void CheckQueueNamesAreUniqueWithinInstance()
...
             var allQueues = AuditInstances.SelectMany(instance => new QueueNameValidator(instance).queues);
-            CheckQueueNamesAreNotTaken(allQueues);
+            var duplicates = (
+                from queue in queues
+                where allQueues.Any(p =>
+                    p.QueueType != QueueType.Audit &&
+                    string.Equals(p.ConnectionString, queue.ConnectionString, StringComparison.OrdinalIgnoreCase) &&
+                    string.Equals(p.QueueName, queue.QueueName, StringComparison.OrdinalIgnoreCase) &&
+                    string.Compare("!disable", queue.QueueName, StringComparison.OrdinalIgnoreCase) != 0 &&
+                    string.Compare("!disable.log", queue.QueueName, StringComparison.OrdinalIgnoreCase) != 0)
```
danielmarbach: Why was `!disable.log` introduced. So far we never needed this

src/ServiceControl.Config/Extensions/Validations.cs
```diff
@@ -55,9 +62,10 @@ public static List<string> UsedQueueNames(this ReadOnlyCollection<ServiceControl
                 .ToList();
         }
 
-        public static List<string> UsedQueueNames(this ReadOnlyCollection<ServiceControlAuditInstance> ServiceControlInstances, TransportInfo transportInfo = null, string instanceName = null, string connectionString = null)
+        public static List<string> UsedAuditQueueNames(TransportInfo transportInfo = null, string instanceName = null, string connectionString = null)
         {
-            var instancesByTransport = ServiceControlInstances.Where(p => p.TransportPackage.Equals(transportInfo) &&
+            var serviceControlInstances = InstanceFinder.ServiceControlInstances();
```
mikeminutillo: Does this need to take into account ServiceControl Audit instances?

src/ServiceControl.Config/UI/Upgrades/AddNewAuditInstanceViewModel.cs
mikeminutillo: It is the [Validar Fody plugin](https://github.com/Fody/Validar)

src/ServiceControlInstaller.Engine/Validation/QueueNameValidator.cs
```diff
@@ -112,17 +118,23 @@ internal void CheckQueueNamesAreUniqueWithinInstance()
         internal void CheckQueueNamesAreNotTakenByAnotherAuditInstance()
         {
             var allQueues = AuditInstances.SelectMany(instance => new QueueNameValidator(instance).queues);
-            CheckQueueNamesAreNotTaken(allQueues);
+            var duplicates = (
+                from queue in queues
+                where allQueues.Any(p =>
+                    p.QueueType != QueueType.Audit &&
```
mikeminutillo: I'm not sure that this quite captures every case. What if `queue` is not an audit queue but `p` is?

i.e. 

- First instance has:
  - Audit queue `audit`
  - Audit log queue `audit.log`
- Second instance has:
  - Audit queue `audit2`
  - Audit log queue `audit`

src/ServiceControlInstaller.Engine/Interfaces.cs
HEskandari: The interface is needed for removing a bunch of validation logic. Currently, it is duplicated for Monitoring/SC/SCA because of the lack of common interface

src/ServiceControl.Config/UI/Upgrades/AddNewAuditInstanceViewModel.cs
HEskandari: Yes, this injects IDataErrorInfo into the class.

src/ServiceControl.Config/UI/Shell/ShellView.xaml.cs
HEskandari: Because it shortcuts a check that's done in the Shell. If you have no active view (no instance installed) and install a new license this code will blow up with a NRE. 

The code that goes through the shell (below) has this check and is the right way to do it anyway (making a call to its ViewModel rather than an event being raised directly from the View).

src/ServiceControl.Config/UI/Shell/ShellViewModel.cs
HEskandari: Commented above.

src/ServiceControlInstaller.Engine/Validation/QueueNameValidator.cs
HEskandari: Don't know :) legacy code?

src/ServiceControl.Config/Extensions/Validations.cs
HEskandari: Good catch. That was actually incorrect. Updated.

src/ServiceControlInstaller.Engine/Validation/QueueNameValidator.cs
HEskandari: @mikeminutillo 	within the context of this method:

- The 'queues' will always be from SCA, so either an Audit queue or Audit Log queueu.
- The method `CheckQueueNamesAreNotTakenByAnotherAuditInstance` itself is just for checking audit queues. The check for ServiceControl queues is a different one `CheckQueueNamesAreNotTakenByAnotherServiceControlInstance` and that goes over `SCInstances`.

So I think the `queue` within this method is always either audit or audit log queue. The check compares the name (Audit + Audit Log) against all the other queues (everything "Audit" queues) of other instances installed.

PR #1716, opened 2019-07-26
Scatter gather endpoints

mikeminutillo:
In order to support [Multi-Region Deployment](https://docs.particular.net/servicecontrol/servicecontrol-instances/distributed-instances#advanced-scenarios-multi-region-deployments) we need to scatter-gather over known endpoints. 

ServiceControl instances will Scatter-Gather results over it's set of remotes

ServiceControl Audit instances index results from the ingested audits. 

This means in a multi-region deployment, the Global instance will return it's own results and the aggregated results from the primary regional instances. Each regional primary instance will return it's own results and the aggregated results from their remotes (usually audit instances). This means that each query might have 2 hops to get a full result set.

### DIFF
--- a/src/ServiceControl.Audit/Monitoring/KnownEndpoints/GetKnownEndpointsApi.cs
+++ b/src/ServiceControl.Audit/Monitoring/KnownEndpoints/GetKnownEndpointsApi.cs
@@ -1,0 +1,35 @@
+namespace ServiceControl.Audit.Monitoring
+{
+    using System.Linq;
+    using System.Threading.Tasks;
+    using Auditing.MessagesView;
+    using Infrastructure;
+    using Infrastructure.Extensions;
+    using Nancy;
+    using Raven.Client;
+
+    class GetKnownEndpointsApi : ApiBase<NoInput, KnownEndpointsView[]>
+    {
+        public override async Task<QueryResult<KnownEndpointsView[]>> Query(Request request, NoInput input)
+        {
+            using (var session = Store.OpenAsyncSession())
+            {
+                var endpoints = await session.Query<EndpointDetails, EndpointsIndex>()
+                    .Statistics(out var stats)
+                    .ToListAsync()
+                    .ConfigureAwait(false);
+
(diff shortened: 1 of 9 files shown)


src/ServiceControl.Audit/Monitoring/KnownEndpoints/GetKnownEndpointsApi.cs
```diff
@@ -0,0 +1,35 @@
...
+                var knownEndpoints = endpoints
+                    .Select(x => new KnownEndpointsView
+                    {
+                        Id = DeterministicGuid.MakeId(x.Name, x.HostId.ToString()),
+                        EndpointDetails = x,
+                        HostDisplayName = x.Host
+                    })
+                    .ToArray();
+
+                return new QueryResult<KnownEndpointsView[]>(knownEndpoints, stats.ToQueryStatsInfo());
```
danielmarbach: Could we return the IList instead to avoid ToArray?

src/ServiceControl/Monitoring/Web/GetKnownEndpointsApi.cs
```diff
@@ -0,0 +1,33 @@
...
+
+        public GetKnownEndpointsApi(EndpointInstanceMonitoring monitoring)
+        {
+            this.monitoring = monitoring;
+        }
+
+        public override Task<QueryResult<KnownEndpointsView[]>> LocalQuery(Request request, NoInput input)
+        {
+            return Task.FromResult(
+                new QueryResult<KnownEndpointsView[]>(
```
danielmarbach: Same comment as above

src/ServiceControl.Audit/Monitoring/KnownEndpoints/GetKnownEndpointsApi.cs
danielmarbach: I just realized that my suggestion was useless given we have to transform with select and select many. Sorry for that.

PR #1718, opened 2019-07-29
Remove message from stacktrace

SzymonPobiega:


danielmarbach:
I kicked the build once more. Weird that the `When_heartbeat_loss_is_detected.Should_publish_notification` starts suddenly failing

PR #1726, opened 2019-08-08
Hotfix 4.0.2

mikeminutillo:


src/ServiceControl.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.28803.452
+# Visual Studio 15
```
danielmarbach: Should we revert this?

src/ServiceControlInstaller.Packaging.UnitTests/DeploymentPackageTests.cs
```diff
@@ -0,0 +1,34 @@
...
+                var mainEntries = zip.Entries.Where(x => x.FullName.StartsWith(deploymentPackage.ServiceName)).ToList();
+
+                CollectionAssert.IsNotEmpty(mainEntries, $"Expected a {deploymentPackage.ServiceName} folder in {deploymentPackage.FullName}");
+
+                CollectionAssert.IsEmpty(
+                    from mainEntry in mainEntries
+                    join entry in zip.Entries.Except(mainEntries) on mainEntry.Name equals entry.Name
+                    where entry.Length != mainEntry.Length
+                    select entry.FullName, 
+                    $"File sizes should match the ones in the {deploymentPackage.ServiceName} folder. Check versions of dependencies.");
```
danielmarbach: Is it sizes or file names?

mikeminutillo: The file names match. We're doing the join on file name. The sizes do not match.

So if you have `Newtonsoft.Json.dll` in the `ServiceControl` folder and in the `Transports/SqlServer` folder but the file sizes don't match, then we have a problem the files are obviously different. One of them will overwrite the other at install time.

src/ServiceControl.sln
mikeminutillo: Done

Issue #1729, opened 2019-08-12
Default service account to the same account used for SC for the new Audit process


gbiellem:
Hi There :)

Just a small feature request, 

When upgrading ServiceControl to 4.x  it would be nice if the audit process input screen defaulted to the same service account that was used under the existing servicecontrol service.  I realize that the user would still need to supply the password for a standard account. (although if it's an AD managed account then you don't)

HEskandari:
I'm not sure if pre-filling the account field is much helpful? You have to provide a password as well, or as you stated it wouldn't work for AD cases. Especially since the upgrade is a one-time operation only.

gbiellem:
HI @HEskandari 

I can see your point about a one time operation but it does feel a little sloppy to have not done it.

>> You stated it wouldn't work for AD cases.
Actually I didn't.  In AD domains  Group Managed Service Accounts are best practice (Since 2012) and they do not have passwords. Standard accounts require passwords, so yeah on a standalone box or if you are ignoring best practice and not using AD GMSA then yes you need to know the password.

Also you can't edit or view the current service account used for an existing instance within the SCMU until after the upgrade. I presume this is because the GUI isn't compatible with the old setup   So you have to leave the tool and go look it up in the Windows Services Snapin to complete the upgrade.  Pre-populating the account would  have made this a more user friendly user experience.

HEskandari:
@gbiellem ah I see your point. I'll have a look into it, feel that it can be done easily.

HEskandari:
@gbiellem I have submitted a PR to fix the issue.

gbiellem:
Great! Thanks Hadi

danielmarbach:
Merged and released as part of 4.1

PR #1732, opened 2019-08-14
Provide service account information when doing an upgrade

HEskandari:
Connects to https://github.com/Particular/ServiceControl/issues/1729


danielmarbach:
Smoke tested

image

Works

PR #1765, opened 2019-09-23
Fix ServiceControl shutdown detection logic

### LINKED ISSUE #907: Upgrade fails because files still in use
In Raygun we have several reports of exceptions when upgrading ServiceControl instances. In each case it look like the existing service has not completely shutdown and the old files are locked.  This exhibits itself as the following exception:

'Access to the path 'C:\Program Files (x86)\Particular Software\ServiceControl\Autofac.dll' is denied.'

fourpastmidnight:
Fixes #907.

The return value from `Task.Wait(Int32)` was never used. Even if it had been used, it would always have returned `false`. The return value from `HasUnderlyingProcessExited` was never used to complete the task. (I may not be understanding the original code correctly, and in the case that I'm not, then the opposite would be true: the task always completes successfully and `Task.Wait(Int32)` would always return `true`--either way, this code does not behave as its intended.)

Second, given the above, and the fact that the method blocks to wait for the task to complete, why use a task at all?

Lastly, I happened to remember while reviewing all of this code that the only thing that happens when `instance.TryStopService()` is called in all of the `Upgrade` methods of the various installers is that a message is logged, but the `Upgrade` method continues on. So this PR changes the behavior so the message is logged and `false` is returned.

### DIFF
--- a/src/ServiceControlInstaller.Engine/Instances/BaseService.cs
+++ b/src/ServiceControlInstaller.Engine/Instances/BaseService.cs
@@ -68,14 +68,13 @@
                         Thread.Sleep(100);
                     }
                 });
-                t.Wait(5000);
+
+                return t.Wait(5000);
             }
             catch (TimeoutException)
             {
                 return false;
             }
-
-            return true;
         }
 
         public bool TryStartService()
--- a/src/ServiceControlInstaller.Engine/Unattended/UnattendAuditInstaller.cs
+++ b/src/ServiceControlInstaller.Engine/Unattended/UnattendAuditInstaller.cs
@@ -112,6 +112,7 @@ namespace ServiceControlInstaller.Engine.Unattended
             if (!instance.TryStopService())
             {
                 logger.Error("Service failed to stop or service stop timed out");
+                return false;
             }
 
             try
--- a/src/ServiceControlInstaller.Engine/Unattended/UnattendMonitoringInstaller.cs
+++ b/src/ServiceControlInstaller.Engine/Unattended/UnattendMonitoringInstaller.cs
@@ -111,6 +111,7 @@ namespace ServiceControlInstaller.Engine.Unattended
             if (!instance.TryStopService())
             {
                 logger.Error("Service failed to stop or service stop timed out");
+                return false;
             }
 
             try
--- a/src/ServiceControlInstaller.Engine/Unattended/UnattendServiceControlInstaller.cs
+++ b/src/ServiceControlInstaller.Engine/Unattended/UnattendServiceControlInstaller.cs
@@ -120,6 +120,7 @@ namespace ServiceControlInstaller.Engine.Unattended
             if (!instance.TryStopService())
             {
                 logger.Error("Service failed to stop or service stop timed out");
+                return false;
             }
 
             try


fourpastmidnight:
BTW, would it be possible, if this is approved, to get this back-ported to the 1.x version of ServiceControl? We're trying to get our instances upgraded as part of the first step in our migration process to a more current version of ServiceControl.

fourpastmidnight:
I'm unable to see the build log for the acceptance tests to understand why these changes would've caused a failure. Any help would be appreciated.

dbelcham:
@fourpastmidnight looks like all of the tests and other things in our build pipeline have passed. Thanks for the PR. Like I said on the issue, we've added it to the list for a future release.

fourpastmidnight:
Perfect thanks! What are the chances of having this applied retroactively to all major versions of ServiceControl, since this has been an issue since the 1.x era? As (I may have) mentioned in the issue, we are currently on the 1.x series and upgrading is part of our overall plan for migrating to the latest/greatest version of SC.

indualagarsamy:
While we backport fixes for NServiceBus, we don't do that for ServiceControl, ServiceInsight or ServicePulse. We recommend that you upgrade to the latest. 
Our support policy in general is described here: 
https://docs.particular.net/nservicebus/upgrades/support-policy

fourpastmidnight:
OK--I understand that policy, in general, but this is clearly a pain point _for even getting upgraded to current_!

dbelcham:
@fourpastmidnight We have to walk a fine line when we prioritize issues for fixing. We have an internal list that we judge every issue against to determine a prioritization outcome. One of the things that puts an issue at a lower level of priority is the availability and awkwardness of a work around. As you pointed out in the [associated issue](https://github.com/Particular/ServiceControl/issues/907#issuecomment-534678976), there is a way to complete the task at hand (shutting down ServiceControl, and/or upgrading ServiceControl) using a work around (attempting the upgrade twice in a row).

When there is a work around, we consider an issue to have less impact (not zero impact though) and, thus, a lower priority compared to other issues (i.e. ones with data loss, or having no work around, or one of many other situations). We definitely don't ignore these lower priority issues. They are worked on when more severe bugs/issues have all been resolved, or if we determine that there are enough lower priority issues to warrant work on a bundle of them being initiated.

We apply a similar set of logic to back-porting bug fixes. If this bug were blocking the upgrade path (i.e. there was absolutely no way to upgrade to a later version) we would consider it for back-porting. Again, that doesn't appear to be the case so back-porting isn't something we would be doing with this fix.

One thing that we don't have in our prioritization is a modifier to increase the priority when a customer provides a PR, such as you have done. I am going to initiate a discussion about that.

src/ServiceControlInstaller.Engine/Instances/BaseService.cs
```diff
@@ -61,21 +61,27 @@ public bool TryStopService()
...
-                });
-                t.Wait(5000);
             }
             catch (TimeoutException)
             {
                 return false;
             }
 
-            return true;
+            var processHasExited = false;
```
andreasohlund: Did you observe the previous code being wrong?

src/ServiceControlInstaller.Engine/Instances/BaseService.cs
```diff
@@ -61,21 +61,27 @@ public bool TryStopService()
...
             {
                 Service.WaitForStatus(ServiceControllerStatus.Stopped, TimeSpan.FromSeconds(60));
-                var t = new Task(() =>
-                {
-                    while (!HasUnderlyingProcessExited())
-                    {
-                        Thread.Sleep(100);
-                    }
-                });
-                t.Wait(5000);
```
andreasohlund: @fourpastmidnight The other changes look ok but would it be easier to just change this line to `return t.Wait(5000);`?

src/ServiceControlInstaller.Engine/Instances/BaseService.cs
andreasohlund: Strike that, I see that it was wrong now. See my comment above

fourpastmidnight: I just saw this now. Are we OK here, or are you awaiting an answer to your question?

andreasohlund: Kind of, I was thinking that the original code was more clean? (assuming we fix the return)

andreasohlund: Ie https://github.com/Particular/ServiceControl/pull/1765#discussion_r335466963

fourpastmidnight: See my reply above, too. 😄

fourpastmidnight:
@andreasohlund Updated this PR as discussed. Thanks!

andreasohlund:
Thank you @fourpastmidnight much appreciated

Would you be able to send your mail address + phone to andreas dot ohlund at particular dot net so that we can send you a little token of our appreciation :) ?

PR #1776, opened 2019-10-09
ServiceControl process stops when transport/broker becomes unavailable

mikeminutillo:
## Old behavior

When ServiceControl encounters an error connecting to the transport/broker it raises an [NServiceBus critical error](https://docs.particular.net/nservicebus/hosting/critical-errors) and shuts the process down. Each instance is configured to automatically restart after one minute.

This has a couple of issues:

1. This type of error can be transient. A very short term network error or restarting the transport/broker might cause ServiceControl to stop for up to a minute. In an environment with an unstable network this can cause a lot of delays and downtime.
2. This type of error can limited in scope. For instance, if error forwarding is disabled and an operator removes the configured error log queue, then the ServiceControl process will be halted. While error message ingestion cannot continue in this state, endpoint monitoring (heartbeats and custom checks) and recoverability (retrying and archiving previously ingested failed messages) should be able to continue.
2. This type of failure should be more visible. Once the ServiceControl process exits, the rest of the platform is not able to get any visibility into what is happening. It's not possible to see the failure in ServicePulse (the production monitoring tool) for example. Although each ServiceControl instance will automatically restart, there is nothing to show that this happened unless an operator investigates the ServiceControl logs.

## New behavior

When ServiceControl encounters an error connecting to the transport it will trigger an NServiceBus custom check and disconnect the error ingestion pipeline for up to 1 minute. The custom check failure is reported to operators via ServicePulse. Depending on the nature of the error, it may be possible for ServiceControl to publish an integration event for the failure as well.

After the connection to the transport has been successfully re-established the custom check status status is updated (and reported via ServicePulse and integration events).


src/ServiceControl/Infrastructure/NServiceBusFactory.cs
```diff
@@ -54,16 +55,12 @@ public static Task<IStartableEndpoint> Create(Settings.Settings settings, Transp
...
             {
-                configuration.ReportCustomChecksTo(endpointName);
+                configuration.EnableFeature<InternalCustomChecks>();
             }
 
             configuration.UseContainer<AutofacBuilder>(c => c.ExistingLifetimeScope(container));
 
-            configuration.DefineCriticalErrorAction(criticalErrorContext =>
-            {
-                onCriticalError(criticalErrorContext);
```
mikeminutillo: We are still passing this in but it doesn't get called anywhere. For instance, it was once used to shut down an acceptance test on critical failure. I'm guessing we no longer need to do that and can rely on the test timeouts to shut the ATT down.

src/ServiceControl/Operations/ImportFailedErrors.cs
```diff
@@ -5,69 +5,84 @@
...
         }
 
-        public async Task Run(CancellationTokenSource tokenSource)
+        public async Task Run(CancellationToken token)
         {
-            var token = tokenSource.Token;
-            var succeeded = 0;
-            var failed = 0;
-            using (var session = store.OpenAsyncSession())
+            var config = rawEndpointFactory.CreateSendOnlyRawEndpointConfiguration("ImportFailedErrors");
```
mikeminutillo: I'm assuming this name is arbitrary as we shouldn't be changing any headers or anything.

src/ServiceControl/Infrastructure/CriticalErrorCustomCheck.cs
```diff
@@ -0,0 +1,35 @@
...
+    using System;
+    using System.Threading.Tasks;
+    using NServiceBus;
+    using NServiceBus.CustomChecks;
+
+    class CriticalErrorCustomCheck : CustomCheck
+    {
+        static volatile string recentFailure;
+
+        public CriticalErrorCustomCheck()
```
mikeminutillo: Do we need both this custom check and [this one](https://github.com/Particular/ServiceControl/pull/1776/files#diff-29ed3273db53178460f61089f4fb8a84R11)

They seem to be trying to accomplish the same thing.

src/ServiceControl/Operations/ErrorImporter.cs
```diff
@@ -25,95 +58,34 @@ protected override void Setup(FeatureConfigurationContext context)
         {
             var settings = context.Settings.Get<Settings>("ServiceControl.Settings");
 
+            var queueBindings = context.Settings.Get<QueueBindings>();
+            queueBindings.BindReceiving(settings.ErrorQueue);
+
             context.Container.ConfigureComponent<ErrorIngestor>(DependencyLifecycle.SingleInstance);
             context.Container.ConfigureComponent<ErrorPersister>(DependencyLifecycle.SingleInstance);
             context.Container.ConfigureComponent<FailedMessageAnnouncer>(DependencyLifecycle.SingleInstance);
```
mikeminutillo: We should probably remove these from the container as they are internal to the component now.

src/ServiceControl/Operations/ImportFailedErrors.cs
SzymonPobiega: Yes. For a send-only endpoint this name is arbitrary.

src/ServiceControl/Operations/ErrorImporter.cs
SzymonPobiega: We can't because we depend on the domain events and the `IDocumentSession`. If we had these two available outside of the container, we could do this but I think it would require a bigger componentization effort.

src/ServiceControl/Infrastructure/NServiceBusFactory.cs
SzymonPobiega: Yeah. I'll remove it.

src/ServiceControl/Infrastructure/CriticalErrorCustomCheck.cs
SzymonPobiega: This one would be signalled if there is something wrong with the main endpoint so a slightly different purpose. The error ingestion might be signalled e.g. when we have a bug that causes each error ingestion to fail. In that case this check would be OK.

src/ServiceControl/Operations/ErrorImporter.cs
mikeminutillo: But we create these manually inside the component. Who is resolving them from the container?

mikeminutillo: Also, we still allow [resolution of any concrete type with no interfaces](https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/Bootstrapper.cs#L79) so I think this is redundant.

PR #1777, opened 2019-10-09
Queue length delay exposed as a cs setting

tmasternak:


src/ServiceControl.Monitoring.Transports.AzureServiceBus/QueueLengthProvider.cs
```diff
@@ -112,6 +127,7 @@ public async Task Stop()
             await managementClient.CloseAsync().ConfigureAwait(false);
         }
 
+        //TODO: agree on the default
```
andreasohlund: Once per seconds at least I'd say? Should we also switch to a task.whenall to make the queries concurrently? (to make sure that lots of queues doesn't cause delays)

andreasohlund:
Do we have any test coverage on the queue length providers?

tmasternak:
> Do we have any test coverage on the queue length providers?

Yes, we do. But those were never running on the TC build agent :(.

andreasohlund:
>Yes, we do. But those were never running on the TC build agent :(.

The upcoming TF that will merge the Seams have this on their TODO list

src/ServiceControl.Monitoring.Transports.AzureServiceBus/QueueLengthProvider.cs
```diff
@@ -80,28 +97,52 @@ public Task Start()
...
                 }
-                else
+
+                pageNo++;
+            }
+
+            return queues.ToDictionary(q => q.Path, q => q);
+        }
+
+        Task UpdateAllQueueLengths(ConcurrentDictionary<string, QueueDescription> queues, CancellationToken token) => Task.WhenAll(endpointQueueMappings.Select(eq => UpdateQueueLength(eq, queues, token)));
```
janovesk: Do we know that this is safe? Is it okay to hit Azure with 1000 concurrent requests if you have 1000 queues?

andreasohlund: Good point, I have a pet project that hits NuGet concurrently and at about 20 concurrent calls I get throttled and things start to fail. Not the same service here but I think we should test this? (A little app to gen 1000 queues and then make the calls?)

tmasternak: Did some testings. With the current implementation:
 1. We will grab all the queues from the namespace (paged, 100 queues at a time)
 2. We will ask of queue length only if we have nsb endpoint reporting to the sc. So the number of call per queue is equal to number of endpoints.

Test:
 1. 800 queues in the namespace and 200 nsb endpoints
 2. No errors/exceptions with the current approach. The execution of one query loop (queue list + length per queue) takes ~ 10 sec.

I propose to leave this as is. We can enhance this if customers need it.

PR #1781, opened 2019-10-11
Allow changing RabbitMQ UseExternalAuthMechanism and DisableRemoteCertificateValidation via connection string

### LINKED ISSUE #1743: RabbitMQ transport allow activating ExternalAuthMechanism
Hi

We are currently using RMQ for our NSB transport layer with SSL/TLS activated and external auth via ssl client certificates.
For our NSB services this means we have to call 'UseExternalAuthMechanism' on our transport to enable this.
```
var transportExtensions = endpointConfiguration.UseTransport<RabbitMQTransport>();
transportExtensions.ConnectionString(serviceConfig.ConnectionStrings.Messaging);
transportExtensions.UseExternalAuthMechanism();
```
Would it be possible to provide a setting for ServiceControl.Monitoring so this can be enabled?
Looks like the ConnectionFactory always defaults to 'false' in this project:
https://github.com/Particular/ServiceControl.Monitoring/blob/43f9a6a4532ea9a7251d4f12ed2cbb0c7f079f91/src/ServiceControl.Transports.RabbitMQ/QueueLengthProvider.cs#L142
This would allow us to use client certificates by providing them in the connection string.

Thanks

andreasohlund:
Closes #1743

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ nugets
 deploy
 build32
 binaries
+zip
 *.vshost.*
 .nu
 _UpgradeReport.*
--- a/src/ServiceControl.Monitoring.Transports.RabbitMQ/ConnectionFactory.cs
+++ b/src/ServiceControl.Monitoring.Transports.RabbitMQ/ConnectionFactory.cs
@@ -3,14 +3,15 @@
     using System;
     using System.Net.Security;
     using System.Security.Authentication;
-   using global::RabbitMQ.Client;
+    using System.Security.Cryptography.X509Certificates;
+    using global::RabbitMQ.Client;
 
     class ConnectionFactory
     {
         readonly global::RabbitMQ.Client.ConnectionFactory connectionFactory;
         readonly object lockObject = new object();
 
-        public ConnectionFactory(ConnectionConfiguration connectionConfiguration, bool disableRemoteCertificateValidation, bool useExternalAuthMechanism)
+        public ConnectionFactory(ConnectionConfiguration connectionConfiguration, X509CertificateCollection clientCertificates, bool disableRemoteCertificateValidation, bool useExternalAuthMechanism)
         {
             if (connectionConfiguration == null)
             {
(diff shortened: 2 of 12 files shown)


andreasohlund:
Tested against my local broker and this works as expected. Should be good for final review

.gitignore
```diff
@@ -2,6 +2,7 @@ nugets
 deploy
 build32
 binaries
+zip
```
bording: Why was this changed?

src/ServiceControl.Monitoring.Transports.RabbitMQ/ConnectionFactory.cs
```diff
@@ -3,15 +3,14 @@
...
-    using global::RabbitMQ.Client;
+   using global::RabbitMQ.Client;
 
     class ConnectionFactory
     {
         readonly global::RabbitMQ.Client.ConnectionFactory connectionFactory;
         readonly object lockObject = new object();
 
-        public ConnectionFactory(ConnectionConfiguration connectionConfiguration, X509CertificateCollection clientCertificates, bool disableRemoteCertificateValidation, bool useExternalAuthMechanism)
+        public ConnectionFactory(ConnectionConfiguration connectionConfiguration, bool disableRemoteCertificateValidation, bool useExternalAuthMechanism)
```
bording: I know the original intention of this class was to be an exact copy of the one from the transport, so that's why the `clientCertificates` parameter was there.  Perhaps it would be better to leave it as-is?

.gitignore
andreasohlund: To avoid including the /zip folder that gets created if you build in release mode

src/ServiceControl.Monitoring.Transports.RabbitMQ/ConnectionFactory.cs
andreasohlund: Ah, I didn't know that. What was the rationale for that?  Easier copy/paste should the transport change?

bording: Yes, that was the primary reason I believe.

andreasohlund: Ok, I've reverted the code now

bording:
Looks like this needs to be rebased to resolve a conflict. 

Also:

image

What? How did that happen?

andreasohlund:
Hmm, I wonder if that is GitHub for Windows doing something funky. I’ll sort out the conflict

PR #1782, opened 2019-10-11
Add support for deleting endpoints from the heartbeat list

janovesk:
Fixes https://github.com/Particular/DeveloperExperience/issues/536.

Or at least the SC bits. SP bits fixed [here](https://github.com/Particular/ServicePulse/pull/791).

### DIFF
--- a/src/ServiceControl.AcceptanceTests/Monitoring/When_an_endpoint_is_removed.cs
+++ b/src/ServiceControl.AcceptanceTests/Monitoring/When_an_endpoint_is_removed.cs
@@ -1,0 +1,71 @@
+namespace ServiceControl.AcceptanceTests.Monitoring
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Net.Http;
+    using System.Threading.Tasks;
+    using CompositeViews.Endpoints;
+    using NServiceBus;
+    using NServiceBus.AcceptanceTesting;
+    using NUnit.Framework;
+    using ServiceBus.Management.AcceptanceTests;
+    using ServiceBus.Management.AcceptanceTests.EndpointTemplates;
+    using ServiceBus.Management.Infrastructure.Settings;
+
+    class When_an_endpoint_is_removed : AcceptanceTest
+    {
+        [Test]
+        public async Task Should_signal_support_for_delete()
+        {
+            HttpResponseMessage response = null;
+
+            await Define<Context>()
+                .Done(async c =>
+                {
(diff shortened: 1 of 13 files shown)


src/ServiceControl.AcceptanceTests/Monitoring/When_an_endpoint_is_removed.cs
```diff
@@ -0,0 +1,70 @@
...
+
+        class Context : ScenarioContext
+        {
+        }
+
+        public class StartingEndpoint : EndpointConfigurationBuilder
+        {
+            public StartingEndpoint()
+            {
+                EndpointSetup<DefaultServer>(c => c.SendHeartbeatTo(Settings.DEFAULT_SERVICE_NAME));
```
tmasternak: What is the default interval for heartbeats? Are we risking a flaky test i.e. the endpoint gets readded to the collection just after we delete it and before we query for the collection?

src/ServiceControl/Monitoring/Web/EndpointsMonitoringController.cs
```diff
@@ -29,10 +34,51 @@ internal EndpointsMonitoringController(EndpointInstanceMonitoring monitoring, Ge
...
+        [HttpDelete]
+        public async Task<StatusCodeResult> DeleteEndpoint(Guid endpointId)
+        {
+            var removedFromCache = endpointInstanceMonitoring.RemoveEndpoint(endpointId);
+            if (removedFromCache == null)
+            {
+                return StatusCode(HttpStatusCode.NotFound);
+            }
+
+            await DeletePersistedEndpoint(endpointId).ConfigureAwait(false);
```
tmasternak: Should we swap `DeletePersistedEndpoint` and `RemoveEndpoint` from the cache? I think the Raven based is more likely to fail (due to network issues) and that would leave the cache and the storage in an inconsistent state until restart.

src/ServiceControl.AcceptanceTests/Monitoring/When_an_endpoint_is_removed.cs
janovesk: Good point. I'll adjut the heatbeat frequency up to avoid it.

src/ServiceControl/Monitoring/Web/EndpointsMonitoringController.cs
janovesk: Sure, why not.

PR #1783, opened 2019-10-11
Build on 3.0 SDK also

bording:
While the proper fix for this project is to get VS 2019 on the build servers and start using the `Microsoft.NET.Sdk.WindowsDesktop` SDK, here's a way to keep the existing project building regardless of what version of the .NET Core SDK is being used.


bording:
Test failure should be unrelated to this change.

PR #1785, opened 2019-10-14
ServiceName must be set before Run is called

andreasohlund:
The allows monitoring to start properly when running as a Windows Service. The problem was likely introduced in the WebAPI conversion so no users are affected. I have verified that 4.3.3 is ok

### DIFF
--- a/src/ServiceControl.Monitoring/Hosting/Commands/RunCommand.cs
+++ b/src/ServiceControl.Monitoring/Hosting/Commands/RunCommand.cs
@@ -17,7 +17,11 @@ namespace ServiceControl.Monitoring
 
         Task RunNonBlocking(Settings settings)
         {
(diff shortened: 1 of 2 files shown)


danielmarbach:
Cool so SC and SC.Audit is not affected right?

https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/Hosting/Commands/RunCommand.cs#L29

andreasohlund:
Yea, only monitoring

tmasternak:
I've looked into the version from before Web Api migration. The name was set in the `Run` method https://github.com/Particular/ServiceControl/blob/49c24f0bd60dd5f5d1915296f3ea2f7c12d76ed0/src/ServiceControl.Monitoring/Hosting/HostService.cs#L14. Which confirms this is the right approach.

PR #1786, opened 2019-10-16
Revert "Remove network drive restriction"

WilliamBZA:
Reverts Particular/ServiceControl#1773


WilliamBZA:
I've decided to revert this until a proper investigation can be done.

PR #1787, opened 2019-10-16
Force MenuDropAlignment on startup

### LINKED ISSUE #1374: SCMU menu dropdown menus are misaligned
After upgrading to 3.0.0, I'm seeing the following in SCMU:

image1
image2
image3

@danielmarbach has said he is seeing the same behavior, but @boblangley is not seeing it on his computer.

tmasternak:
fixes #1374

### DIFF
--- a/src/ServiceControl.Config/App.xaml.cs
+++ b/src/ServiceControl.Config/App.xaml.cs
@@ -4,6 +4,7 @@
     using System.Diagnostics;
     using System.IO;
     using System.Reflection;
+    using System.Windows;
     using UI.MessageBox;
 
     public partial class App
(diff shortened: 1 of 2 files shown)


tmasternak:
@HEskandari could you have a look an see if this makes sense? We need your WPF skills 🦇

HEskandari:
@tmasternak I think it is also good to ask @mikeminutillo to see if it really fixes the issue, as he can consistently reproduce that issue on his machine, which can't be said for me. 

I'm more than happy to review the code though.

janovesk:
@tmasternak could both repo and see that the fix worked on his tablet/latop @HEskandari

src/ServiceControl.Config/App.xaml.cs
```diff
@@ -16,6 +17,23 @@ public App()
             {
                 ExceptionMessageBox.Attach();
             }
+
+            EnsureStandardPopupAlignment();
+        }
+
+        static void EnsureStandardPopupAlignment()
```
HEskandari: A bit nitpicking here, but could you put the logic in a class and use it in App? This is a workaround and clutters the App.cs code. e.g.

```StandardPopup.ApplyAlignment();```

The  hint/comment can also be moved into the class implementing the workaround.

tmasternak: @HEskandari moved.

PR #1794, opened 2019-10-17
Add back support for /api/upgrade

janovesk:
Fixes https://github.com/Particular/ServicePulse/issues/793

It alsways returns false since we don't have any db migrations now. Returning something will make SP stop complaining with a 404 on a regular basis.

### DIFF
--- a/src/ServiceControl.AcceptanceTests/Upgrade/When_upgrade_status_requested.cs
+++ b/src/ServiceControl.AcceptanceTests/Upgrade/When_upgrade_status_requested.cs
@@ -1,0 +1,33 @@
+﻿namespace ServiceControl.AcceptanceTests.Upgrade
+{
+    using System.Threading.Tasks;
+    using NServiceBus.AcceptanceTesting;
+    using NUnit.Framework;
+    using ServiceBus.Management.AcceptanceTests;
+    using ServiceControl.Upgrade;
+
+    class When_upgrade_status_requested : AcceptanceTest
+    {
+        [Test]
+        public async Task Should_return_false()
+        {
+            StaleIndexInfo indexInfo = null;
+
+            await Define<MyContext>()
+                .Done(async c =>
+                {
+                    var result = await this.TryGet<StaleIndexInfo>("/api/upgrade");
+                    indexInfo = result;
+                    return indexInfo != null;
+                })
+                .Run();
+
+            Assert.AreEqual(false, indexInfo.InProgress);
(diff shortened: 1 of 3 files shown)


janovesk:
Had to add a constructor with at least one argument to make WebApi find the controller.. complains when if you have a empy constructor. Not sure why.. 

Thoughts on this @danielmarbach ?

WilliamBZA:
The upgrade endpoint was used after the database migration from v1.x to v2.x. It was a way to let users know that the messages they are seeing could be wildly out of sync with what is actually there. A warning, if you will. Since users can only upgrade from one major version to another, it only makes sense for that endpoint to exist in v2.x of ServiceControl as there are no data migrations running from v2 to v3, or v3 to v4. To me, this means that I agree with the original idea to remove the HTTP endpoint completely. It makes no sense to still exist in v4.

However, it gets complicated because we may have users that still have old instances of SC mixed with newer instances of SC. A believable (albeit a stretch and this is likely grasping at straws) scenario is: user has a large database and can't risk the downtime to migrate a v1.x version of SC to a v2.x version of SC. Our guidance in this case is to deploy a new V2.x instance and leave the v1.x running until audits are cleared up and the DB is small enough to migrate. If the bulk of that DB was errors, they could have been running with an up to date audit/heartbeats/custom check instance of SC and an old v1.x instance of SC for errors. SP could be updated here too - and being able to switch to different SC environments from within SP enables this. In this fictional scenario, it could be feasible that they want to upgrade the old instance now. Meaning there might still be value in including the feature within SP, but not within the current version of SC.

I am not terribly attached to this though. Just wanted to explain why I don't think it makes sense to include the check in SC, but it still makes sense (to me) to include it in SP. As you suggested, we could also just make the decision to remove the feature from SP too and we'd only impact users on really old versions of SC - and the impact is usability, not damaging.

danielmarbach:
I agree we should focus on fixing something in SP. As far as I remember SC can have a minimum required SC version. 

https://github.com/Particular/ServicePulse/blob/b43e8c67f06a824eee1b2cba9cdce2469cf00a61/src/ServicePulse.Host/app/js/services/factory.shareddata.js#L44

The problem we'll always have is that SC and SP are decoupled and users can decide to update that as they wish. So why not remove the check from SP like suggested and bump the SC version to 4.x?

I do agree with William we should not reintroduce that endpoint in SC

janovesk:
Don't have a strong opinion either way, so I'm fine with following the recommendations of @danielmarbach and @WilliamBZA who have more context on this. Closing.

PR #1796, opened 2019-10-21
Properly call the task that checks for process exit

andreasohlund:
It turned out that we didn't actually start the task that makes sure that the service process have exited correctly and since we fixed the bug where the return value is ignored the SCMU always reported a failing stop. This is now fixed

Side note: I previous versions we unnecessary waited for 5s on every service shutdown

### DIFF
--- a/src/ServiceControlInstaller.Engine/Instances/BaseService.cs
+++ b/src/ServiceControlInstaller.Engine/Instances/BaseService.cs
@@ -6,9 +6,9 @@
     using System.Linq;
     using System.ServiceProcess;
     using System.Threading;
-    using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


andreasohlund:
@danielmarbach the installer is for some reason targeting `net40` so `Task.Run` isn't available. I guess `new Task(..)` + `t.Start()` is the equivalent?

danielmarbach:
Create a static method like

```
public static class TaskHelpers
{
   public static Task Run(Action action)
   {
       return Task.Factory.StartNew(null, action, default(CancellationToken), TaskCreationOptions.DenyChildAttach, TaskScheduler.Default);
    }
}
```

PR #1799, opened 2019-10-23
Do not shutdown service control when failed message ingestion fails

SzymonPobiega:


src/ServiceControl/Operations/ErrorIngestion.cs
```diff
@@ -0,0 +1,86 @@
...
+        public ErrorIngestion(ErrorIngestor errorIngestor, string errorQueue, RawEndpointFactory rawEndpointFactory, IDocumentStore documentStore, LoggingSettings loggingSettings, Func<string, Exception, Task> onCriticalError)
+        {
+            this.errorIngestor = errorIngestor;
+            this.errorQueue = errorQueue;
+            this.rawEndpointFactory = rawEndpointFactory;
+            this.onCriticalError = onCriticalError;
+            importFailuresHandler = new SatelliteImportFailuresHandler(documentStore, loggingSettings, onCriticalError);
+        }
+
+        public async Task EnsureStarted()
```
danielmarbach: Why not flow the cancellation token from watchdog into this method so that the waitasync could be cancelled if needed and we avoid indefinite hanging?

src/ServiceControl/Operations/ErrorIngestion.cs
```diff
@@ -0,0 +1,86 @@
...
+            }
+            finally
+            {
+                startStopSemaphore.Release();
+            }
+        }
+
+        Task OnCriticalErrorAction(ICriticalErrorContext ctx) => onCriticalError(ctx.Error, ctx.Exception);
+
+        public async Task EnsureStopped()
```
danielmarbach: Why not flow the cancellation token from watchdog into this method so that the waitasync could be cancelled if needed and we avoid indefinite hanging?

src/ServiceControl/Operations/ErrorIngestionCustomCheck.cs
```diff
@@ -0,0 +1,40 @@
...
+        {
+            this.criticalErrorHolder = criticalErrorHolder;
+        }
+
+        public override Task<CheckResult> PerformCheck()
+        {
+            var failure = criticalErrorHolder.GetLastFailure();
+            if (failure == null)
+            {
+                return Task.FromResult(CheckResult.Pass);
```
danielmarbach: Can be cached

src/ServiceControl/Operations/ErrorIngestionCustomCheck.cs
```diff
@@ -0,0 +1,40 @@
...
+            }
+
+            return Task.FromResult(CheckResult.Failed(failure));
```
danielmarbach: If failure is cached this task can be as well

src/ServiceControl/Operations/ErrorIngestionCustomCheck.cs
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        public class State
+        {
+            volatile string lastFailure;
```
danielmarbach: If you make this a Task<string> you can optimize this for the customer. This would be actually an excellent case for ValueTask :(

src/ServiceControl/Operations/Watchdog.cs
```diff
@@ -0,0 +1,86 @@
...
+            watchdog = Task.Run(async () =>
+            {
+                while (!shutdownTokenSource.IsCancellationRequested)
+                {
+                    try
+                    {
+                        await ensureStarted().ConfigureAwait(false);
+                        clearFailure();
+                    }
+                    catch (OperationCanceledException)
```
danielmarbach: In which case would this raise such an exception given that this API doesn't officially support cancellation?

src/ServiceControl/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
```
danielmarbach: Why is this needed?

src/ServiceControl/Hosting/Commands/ImportFailedErrorsCommand.cs
```diff
@@ -26,15 +26,15 @@ public override async Task Execute(HostArguments args)
             var tokenSource = new CancellationTokenSource();
 
             var loggingSettings = new LoggingSettings(settings.ServiceName, LogLevel.Info, LogLevel.Info);
-            var bootstrapper = new Bootstrapper(ctx => { tokenSource.Cancel(); }, settings, busConfiguration, loggingSettings);
```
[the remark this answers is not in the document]
SzymonPobiega: That means that if we get a critical error raised here, the process won't just stop automatically. This should not be a problem as if a message is failed to be re-imported, it stays in the database and there is no negative consequences to attempting to re-import them again.

src/ServiceControl/Operations/ErrorIngestion.cs
```diff
@@ -0,0 +1,86 @@
...
+                if (ingestionEndpoint != null)
+                {
+                    return; //Already started
+                }
+
+                var rawConfiguration = rawEndpointFactory.CreateRawEndpointConfiguration(
+                    errorQueue,
+                    (messageContext, dispatcher) => errorIngestor.Ingest(messageContext));
+
+                rawConfiguration.Settings.Set("onCriticalErrorAction", (Func<ICriticalErrorContext, Task>)OnCriticalErrorAction);
```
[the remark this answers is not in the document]
SzymonPobiega: Nope. This is unfortunately the only was to provide on critical error callback to NSB.Raw as it does not have a public API for this :( I need to talk to the maintainers of that piece of c...ode

src/ServiceControl/Operations/ErrorIngestionCustomCheck.cs
SzymonPobiega: I think as we expect this state to be not that common I think we should be fine with only caching the success result (for readability of the code).

src/ServiceControl/Operations/ErrorIngestionCustomCheck.cs
SzymonPobiega: You mean to have a `ValueTask` instead of `string` here?

src/ServiceControl/Operations/Watchdog.cs
SzymonPobiega: You're right. But now that I added the cancellation token to `ensureStarted` it makes sense to have a catch here, right?

src/ServiceControl/Properties/launchSettings.json
SzymonPobiega: Committed by mistake :(

src/ServiceControl/Operations/Watchdog.cs
danielmarbach: Then yes

src/ServiceControl/Operations/ErrorIngestionCustomCheck.cs
danielmarbach: yeah but we don't have it available so I would basically directly cache `Task<string>` and use `Task.FromResult` here

PR #1801, opened 2019-10-24
Review of 4.1.1

SzymonPobiega:
I created a PR to the `master` branch that attempts to address the issues raised by @danielmarbach . When we are done with this PR, I'll cherry pick the result to the hotfix branch and merge it to the release.

### DIFF
--- a/src/ServiceControl/Infrastructure/Settings/Settings.cs
+++ b/src/ServiceControl/Infrastructure/Settings/Settings.cs
@@ -39,7 +39,7 @@ namespace ServiceBus.Management.Infrastructure.Settings
             Port = SettingsReader<int>.Read("Port", 33333);
             DatabaseMaintenancePort = SettingsReader<int>.Read("DatabaseMaintenancePort", 33334);
             ProcessRetryBatchesFrequency = TimeSpan.FromSeconds(30);
-            TimeToRestartAfterCriticalFailure = TimeSpan.FromSeconds(60);
+            TimeToRestartErrorIngestionAfterFailure = TimeSpan.FromSeconds(60);
             MaximumConcurrencyLevel = SettingsReader<int>.Read("MaximumConcurrencyLevel", 10);
             RetryHistoryDepth = SettingsReader<int>.Read("RetryHistoryDepth", 10);
             HttpDefaultConnectionLimit = SettingsReader<int>.Read("HttpDefaultConnectionLimit", 100);
@@ -48,6 +48,7 @@ namespace ServiceBus.Management.Infrastructure.Settings
             RemoteInstances = GetRemoteInstances();
             DataSpaceRemainingThreshold = GetDataSpaceRemainingThreshold();
             DbPath = GetDbPath();
+            TimeToRestartErrorIngestionAfterFailure = GetTimeToRestartErrorIngestionAfterFailure();
         }
 
         public bool AllowMessageEditing { get; set; }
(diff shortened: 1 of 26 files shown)


src/ServiceControl/Infrastructure/Settings/Settings.cs
```diff
@@ -39,7 +39,7 @@ public Settings(string serviceName = null)
             Port = SettingsReader<int>.Read("Port", 33333);
             DatabaseMaintenancePort = SettingsReader<int>.Read("DatabaseMaintenancePort", 33334);
             ProcessRetryBatchesFrequency = TimeSpan.FromSeconds(30);
-            TimeToRestartAfterCriticalFailure = TimeSpan.FromSeconds(60);
+            TimeToRestartErrorIngestionAfterFailure = TimeSpan.FromSeconds(60);
```
danielmarbach: Why are we setting this twice? Shouldn't the default be applied with `GetTimeToRestartErrorIngestionAfterFailure`

src/ServiceControl/ServiceControl.csproj
```diff
@@ -51,6 +51,10 @@
     <EmbeddedResource Include="Infrastructure\RavenDB\RavenLicense.xml" />
   </ItemGroup>
 
+  <ItemGroup>
```
danielmarbach: 🔥

src/ServiceControl/Operations/ErrorIngestion.cs
```diff
@@ -20,9 +20,9 @@ public ErrorIngestion(ErrorIngestor errorIngestor, string errorQueue, RawEndpoin
             importFailuresHandler = new SatelliteImportFailuresHandler(documentStore, loggingSettings, onCriticalError);
         }
 
-        public async Task EnsureStarted()
+        public async Task EnsureStarted(CancellationToken cancellationToken)
         {
-            await startStopSemaphore.WaitAsync().ConfigureAwait(false);
+            await startStopSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
```
danielmarbach: Would it make sense to include this into the try / catch finally to make sure we always release?

src/ServiceControl/Operations/ErrorIngestion.cs
```diff
@@ -54,9 +54,9 @@ public async Task EnsureStarted()
 
         Task OnCriticalErrorAction(ICriticalErrorContext ctx) => onCriticalError(ctx.Error, ctx.Exception);
 
-        public async Task EnsureStopped()
+        public async Task EnsureStopped(CancellationToken cancellationToken)
         {
-            await startStopSemaphore.WaitAsync().ConfigureAwait(false);
+            await startStopSemaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
```
danielmarbach: Would it make sense to include this into the try / catch finally to make sure we always release?

src/ServiceControl/Infrastructure/Settings/Settings.cs
mikeminutillo: Correct. In fact we already do. Fixed

src/ServiceControl/ServiceControl.csproj
mikeminutillo: 🔥 d

src/ServiceControl/Operations/ErrorIngestion.cs
mikeminutillo: I wondered about this. What's the semantics of releasing a semaphore we don't hold? If this method throws, do we have the lock?

danielmarbach: In all the other code paths where we have a semaphore we always have the WaitAsync and Release pattern in a try finally because that is how it is intended to be used.

There is only one case when the counter is not decrement

https://referencesource.microsoft.com/#mscorlib/system/threading/SemaphoreSlim.cs,612

This happens when you enter with the token already canceled.

Release has explicit lock semantics

https://referencesource.microsoft.com/#mscorlib/system/threading/SemaphoreSlim.cs,771

so it is guaranteed those operations are in sync. Therefore going with WaitAsync finally Release seems to be the most straight forward way

PR #1819, opened 2019-11-11
Convert old asb to new seam

andreasohlund:


aleksandr-samila:
Hm... I don't see where/how this ***QueueLengthProvider*** should be applied. 
And it not used in the old implementation. In this case do we need to have it at all?

Or did I miss something?

aleksandr-samila:
Never mind. Forget my previous comment.

PR #1829, opened 2019-11-18
Update to SQL Transport version 5

mikeminutillo:


mikeminutillo:
Ran into a bug. Should not be returning `null` here.

https://github.com/Particular/NServiceBus.SqlServer/blob/master/src/NServiceBus.SqlServer/SubscriptionTableInstaller.cs#L23

SzymonPobiega:
Looks good :)

PR #1841, opened 2019-12-04
[POC] Host ServicePulse UI directly from ServiceControl

mikeminutillo:
To test this, extract ServicePulse and then zip it's contents into C:\Code\Particular\ServicePulse-1.22.0.zip

At the moment the filename is hardcoded just to get it going.

All this does is serve up the js/html/css/img assets in ServicePulse directly from the ServiceControl endpoint at the /web location. No additional setup of ServicePulse is needed. 

It also sets up index.html as the default document, and redirects requests for the root (/) url to /web.

NOTE: This won't handle SC instances that aren't hosted at localhost without some additional mucking about but it's really just a Proof of Concept.


andreasohlund:
Can we close this for now?

mikeminutillo:
Done

PR #1852, opened 2019-12-20
Remove unneeded GitVersionTask reference

bording:
This older version isn't needed since it is included in all projects via Directory.Build.props.


bording:
Test failures are unrelated to this change. Merging.

Issue #1808, opened 2019-10-30
Paging links use route template without replacing route variables


boblangley:
When messages for a specific endpoint are queried via `/api/endpoints/{endpoint}/messages/` the endpoint variable is not replaced.

### Symptoms

ServiceInsight is unable to correctly navigate message view pages when viewing an individual endpoint.

### Who's affected

* All users using ServiceControl 4.4.0 or greater

### Root Cause

We are using the template, not taking into consideration variables, resulting in broken next/prev links for the endpoint:

https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/Infrastructure/WebApi/Negotiator.cs#L108

mikeminutillo:
Corresponding issue in ServiceInsight repo https://github.com/Particular/ServiceInsight/issues/844

PR #1855, opened 2020-01-03
Replace template values in paging links

mikeminutillo:
Fixes https://github.com/Particular/ServiceControl/issues/1808

### DIFF
--- a/src/ServiceControl/Infrastructure/WebApi/Negotiator.cs
+++ b/src/ServiceControl/Infrastructure/WebApi/Negotiator.cs
@@ -105,7 +105,7 @@ namespace ServiceControl.Infrastructure.WebApi
                 return response;
             }
 
-            var path = request.GetRouteData().Route.RouteTemplate; //should ideally be request.RequestUri.AbsolutePath, but using route instead to keep it backwards compatible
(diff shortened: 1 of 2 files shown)


src/ServiceControl/Infrastructure/WebApi/Negotiator.cs
```diff
@@ -105,7 +105,7 @@ public static HttpResponseMessage WithPagingLinks(this HttpResponseMessage respo
                 return response;
             }
 
-            var path = request.GetRouteData().Route.RouteTemplate; //should ideally be request.RequestUri.AbsolutePath, but using route instead to keep it backwards compatible
+            var path = request.RequestUri.AbsolutePath.Substring(5); // NOTE: Strips off the /api/ for backwards compat
```
timbussmann: I wonder whether we should guard this substring call with a check if the `RequestUri` actually starts with `/api/`. If we ever change the prefix, it will be hard to find this code again.

timbussmann:
I removed the additional tests as these aren't necessary anymore with the suggested changes to use the request URI. By removing the route template information on the tests, the tests will fail if somebody tries to access the route information so that should prevent those being used by accident.

PR #1857, opened 2020-01-07
Guard trim with check for /api path

timbussmann:
Helps to prevent potential bugs in case the /api/ prefix will ever be changed or dropped.

This might be unnecessary if we assume /api/ will always be there I guess but on the other hand I didn't see any tests regarding this assumption nor do I have a good idea how we could tests this instead.

Thoughts @mikeminutillo ?


timbussmann:
After some more thinking I'm closing this one as I'm not convinced it's worth being so defensive here. Most of the potential changes would still break the behavior (e.g. chagning the api prefix) and the chance of this being changed seems to be very low anyway. Let's keep it simple for now.

PR #1934, opened 2020-03-10
Cleanup

danielmarbach:
While looking through the code @timbussmann and I found two things that can be cleaned up

### DIFF
--- a/src/ServiceControl.Transports.ASQ/Helper.cs
+++ b/src/ServiceControl.Transports.ASQ/Helper.cs
@@ -1,13 +1,10 @@
 ﻿namespace ServiceControl.Transports.ASQ
 {
     using System;
-    using System.Reflection;
     using NServiceBus;
-    using NServiceBus.Azure.Transports.WindowsAzureStorageQueues;
     using NServiceBus.Configuration.AdvancedExtensibility;
     using NServiceBus.Serialization;
     using NServiceBus.Settings;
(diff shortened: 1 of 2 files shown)


src/ServiceControl.Transports.ASQ/Helper.cs
```diff
@@ -16,11 +13,6 @@ public static void ApplyHacksForNsbRaw(this TransportExtensions<AzureStorageQueu
             var settings = extensions.GetSettings();
             var serializer = Tuple.Create(new NewtonsoftSerializer() as SerializationDefinition, new SettingsHolder());
             settings.Set("MainSerializer", serializer);
-
-            var ctor = typeof(MessageMetadataRegistry).GetConstructor(BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.Instance, null, new[] {typeof(Func<Type, bool>)}, null);
```
danielmarbach: Pretty sure this is no longer needed because of this change

https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/304

danielmarbach:
@mikeminutillo @SzymonPobiega would appreciate a look over this

src/ServiceControl.Transports.ASQ/Helper.cs
SzymonPobiega: If the tests pass then I am fine with this change ;-)

PR #1936, opened 2020-03-11
Update EULA

WilliamBZA:
Updates the EULA to the current non-production version


ramonsmits:
Close this? Probably needs to be managed via the RepoStandards.

PR #1959, opened 2020-04-07
GitHubSync update

particularbot:
This is an automated synchronization PR.

The following source template repositories were used:
* particular/RepoStandards


andreasohlund:
Snk needs to be removed

PR #1980, opened 2020-04-22
WIP App.config for debug and production plus transformation

danielmarbach:
This makes sure the debug app.config contains relevant stuff for local development while the release config only contains the previously added gc section. By including it into the packaging we can make sure the auto generated assembly redirects are properly added to the config.

It seems this only works for new installations of SC. Old installations will not get the assembly redirects because we restore the old config again in RestoreAppConfig.


src/ServiceControl.Audit/ServiceControl.Audit.csproj
```diff
@@ -35,6 +39,12 @@
     <PackageReference Include="Particular.CodeRules" Version="0.2.1" PrivateAssets="All" />
   </ItemGroup>
 
+  <ItemGroup>
```
bording, line 42: Not sure I see the benefit of having this nested like this, since it just hides one of the two files. Removing the `ItemGroup` would also simplify the changes in the PR.

src/ServiceControl.Audit/ServiceControl.Audit.csproj
```diff
@@ -7,6 +7,10 @@
     <ApplicationIcon>Operations.ico</ApplicationIcon>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <AppConfig>App.$(Configuration).config</AppConfig>
```
bording, line 11: I'd just move this up into the `PropertyGroup` above instead of defining a separate one just for this property.

bording:
The unfortunate thing is with the current project structure, there's no way to incorporate any binding redirects that could be required from the transport projects.

danielmarbach:
Unfortunately upgrade scenarios also don't work. So new assembly redirects don't get added the way the whole installation procedure is written.

bording:
> Unfortunately upgrade scenarios also don't work. So new assembly redirects don't get added the way the whole installation procedure is written.

Yeah that's a problem. I wonder how that could be solved.

PR #1987, opened 2020-04-29
Allow for the request of archived groups

WilliamBZA:
Adds the ability to request the groups of archived messages.
ServicePulse uses this to display the information. If ServicePulse connects to a version of ServiceControl that does not support this then the ServicePulse behaviour does not change.

### DIFF
--- a/src/ServiceControl.AcceptanceTests/Recoverability/Groups/When_a_group_is_archived.cs
+++ b/src/ServiceControl.AcceptanceTests/Recoverability/Groups/When_a_group_is_archived.cs
@@ -63,6 +63,59 @@
         }
 
         [Test]
+        public async Task All_archived_messages_should_be_grouped()
+        {
+            await Define<MyContext>()
+                .WithEndpoint<Receiver>(b => b.When(async bus =>
+                {
+                    await bus.SendLocal<MyMessage>(m => m.MessageNumber = 1)
+                        .ConfigureAwait(false);
+                    await bus.SendLocal<MyMessage>(m => m.MessageNumber = 2)
+                        .ConfigureAwait(false);
+                }).DoNotFailOnErrorMessages())
+                .Do("WaitUntilGrouped", async ctx =>
+                {
+                    if (ctx.FirstMessageId == null || ctx.SecondMessageId == null)
+                    {
+                        return false;
+                    }
+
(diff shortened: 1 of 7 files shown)


WilliamBZA:
Ping @ramonsmits and @WojcikMike

danielmarbach:
I personally would like to see an acceptance test that covers those new APIs. We have plenty of acceptance tests in Recoverability and group that also test the archival that should be extended or new tests added to verify future changes to the product don't break the functionality

src/ServiceControl/Infrastructure/WebApi/RootController.cs
```diff
@@ -39,7 +39,8 @@ public OkNegotiatedContentResult<RootUrls> Urls()
                 LicenseStatus = license.IsValid ? "valid" : "invalid",
                 LicenseDetails = baseUrl + "license",
                 Configuration = baseUrl + "configuration",
-                EventLogItems = baseUrl + "eventlogitems"
+                EventLogItems = baseUrl + "eventlogitems",
+                ArchivedGroupsUrl = baseUrl + "errors/groups/{classifier?}"
```
danielmarbach, line 43: Are we deliberately not adding `archive/groups/id/{groupId}` here or did I miss that?

src/ServiceControl.AcceptanceTests/Recoverability/Groups/When_a_group_is_archived.cs
```diff
@@ -62,6 +62,59 @@ await bus.SendLocal<MyMessage>(m => m.MessageNumber = 2)
                 .Run();
         }
 
+        [Test]
+        public async Task All_archived_messages_should_be_grouped()
```
danielmarbach, line 66: Looks good. Should we have another test that also exercises the `archive/groups/id/{groupId}` route?

WilliamBZA: Added

src/ServiceControl/Infrastructure/WebApi/RootController.cs
WilliamBZA: Nope, added

PR #1989, opened 2020-04-29
Extend the monitoring API to include the connected and disconnected instance counts

ramonsmits:


danielmarbach:
Why has this been merged without any tests added or existing tests augmented given that this is new functionaly?

ramonsmits:
You are right @danielmarbach , I'll check if there are existing tests and extend them.

PR #1995, opened 2020-04-30
If no values are present instance must return as being stale

ramonsmits:
While creating a unit test to test the `GetAllEndpointsMetrics` API I stumbled upon this API. Normally this state probably is not possible chronologically but either it should return `true`,`null`, or an exception but not `false`.

Also, this potentially has a memory leak. If a large amount of instances is tracked that are shortlived this collection never releases this data. IMHO this collection should periodically cleanup.


ramonsmits:
This was a POC for discussion with @tmasternak which gave his blessing on the change :-)

Closing as its addressed in https://github.com/Particular/ServiceControl/pull/1996

PR #1998, opened 2020-05-01
Expand environment variables in the learning transport connection string

ramonsmits:
This make it possible to do for example `%TEMP%\.learningTransport`

### DIFF
--- a/src/ServiceControl.Transports.Learning/LearningTransportCustomization.cs
+++ b/src/ServiceControl.Transports.Learning/LearningTransportCustomization.cs
@@ -1,5 +1,6 @@
 ﻿namespace ServiceControl.Transports.Learning
 {
+    using System;
     using LearningTransport;
     using NServiceBus;
     using NServiceBus.Raw;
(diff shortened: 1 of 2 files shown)


danielmarbach:
This will only properly work if you do it also for `LearningTransportCustomization` connection strings where we set the storage directory. Pushed an update

danielmarbach:
I actually think this would be a great feature enhancement to Core @ramonsmits Maybe send in a to the core learning transport?

ramonsmits:
@danielmarbach I considered that first but as learning transport is initialized via code in core and not via a connection string (I think..) adding ENVVAR expanding is up to the developer.

src/ServiceControl.Transports.Learning/LearningTransportCustomization.cs
```diff
@@ -49,14 +50,14 @@ public override IProvideQueueLength CreateQueueLengthProvider()
         static void CustomizeEndpoint(EndpointConfiguration endpointConfig, TransportSettings transportSettings, TransportTransactionMode transportTransactionMode)
         {
             var transport = endpointConfig.UseTransport<LearningTransport>();
-            transport.StorageDirectory(transportSettings.ConnectionString);
+            transport.StorageDirectory(Environment.ExpandEnvironmentVariables(transportSettings.ConnectionString));
```
danielmarbach, line 53: With this we can even support it via the config

Issue #1622, opened 2019-04-18
Allow ServiceControl to log to STDOUT


fourpastmidnight:
When running ServiceControl as a console application (versus as a Windows Service), allow ServiceControl to log output to STDOUT. This would make ServiceControl more "container-friendly".

fourpastmidnight:
OK, this is working already. Yay!

fourpastmidnight:
OK, this is _NOT_ working. (Well, not anymore, as seen by running 3.7.0; I'm running 1.48 locally on my machine, which does log to STDOUT).

fourpastmidnight:
OK, looking at the ServiceControl code in question, it appears it is set up to log to the console, but at some point, the configuration was changed to only log errors to the console (with an exception for RavenDb). But, still, I maintain that this is a step backward for containerization.

Perhaps the logging level could be specified via a command line switch and/or an app.config setting?

fourpastmidnight:
Could this be due to the fact that the instance in the container isn't licensed? I haven't mounted a volume containing a license for ServiceControl to use just yet, as I'm just putting the final finishing touches on my container image, but the lack of console output was bothering me.

SzymonPobiega:
@fourpastmidnight nope, the logging level is not dependent on license checking. I'll take a look at this and let you know of my findings.

fourpastmidnight:
So, I meant to post this last night, but never hit the Comment button. Oops.

> Ah, I mis-read the code, only RavenDb errors are loged to the console. Everything else goes into the log files. 

Anyway, thanks @SzymonPobiega . That's the code I was looking through.

It's a shame that the logging level cannot be changed for the console, as I was saying above. For docker, it would be nice to log at the Warn or Info level to the console so you can run a command like `docker container logs svccontrol`, for example. Because now, if you run this command, you get the very helpful output:

```
Press Ctrl+C to exit
```

That's it. Nothing, nada, zilch. (Unless, of course, you have errors--but then you have other issues....)

SzymonPobiega:
@fourpastmidnight OK, I don't follow right now. The default console logging level is `Info` which means all `Info`, `Warn` and `Error` message should be logged to both console and the file. At least this is how it is supposed to work.

fourpastmidnight:
@SzymonPobiega OK, wow, that's what I thought should be happening. But then I thought you said only Error and above would log to the console (I missed _to the console **and**_)--so I misread what you said.

[...]

PR #2000, opened 2020-05-02
Fix logging in non-user-interactive environments (e.g. docker) where logs were not sent to STDOUT

WilliamBZA:
Fixes https://github.com/Particular/ServiceControl/issues/1622

Adds a flag so that STDOUT can be used when running with docker containers.


SzymonPobiega:
PR 2000? Should we have some celebration?

WilliamBZA:
Let's just be very careful about what PR 2020 will bring 😨

PR #2002, opened 2020-05-05
Interactive console logging was broken by 05662e5 which is now restored

ramonsmits:
This will log to the console for interactive sessions `printToConsole: Environment.UserInteractive`. This ensures logging to console is restored when launching from VS or Rider.

Should not conflict with "environment detection" for docker vs window service. These permutations would be:

- interactive console
- non-interactive console
- windows service

### DIFF
--- a/src/ServiceControl.Audit/Program.cs
+++ b/src/ServiceControl.Audit/Program.cs
@@ -23,7 +23,7 @@
                 return;
             }
 
-            var loggingSettings = new LoggingSettings(arguments.ServiceName, false);
+            var loggingSettings = new LoggingSettings(arguments.ServiceName, printToConsole: Environment.UserInteractive);
             LoggingConfigurator.ConfigureLogging(loggingSettings);
 
             await new CommandRunner(arguments.Commands).Execute(arguments)
--- a/src/ServiceControl.Monitoring/Program.cs
+++ b/src/ServiceControl.Monitoring/Program.cs
@@ -18,7 +18,7 @@ namespace ServiceControl.Monitoring
 
                 var settings = LoadSettings(ConfigurationManager.AppSettings, arguments);
 
-                MonitorLogs.Configure(settings, false);
+                MonitorLogs.Configure(settings, printToConsole: Environment.UserInteractive);
 
                 var runner = new CommandRunner(arguments.Commands);
 
--- a/src/ServiceControl/Program.cs
+++ b/src/ServiceControl/Program.cs
@@ -22,7 +22,7 @@
                 return;
             }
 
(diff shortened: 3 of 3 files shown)


danielmarbach:
@WilliamBZA do you know what the portable host argument is supposed to do?

https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/Hosting/HostArguments.cs#L76-L83

Wouldn't it be better to set the portable argument instead of introducing another flag? I confess I don't yet fully understand all those nobs though

WilliamBZA:
The cause here is this line: https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl.Audit/Infrastructure/Hosting/Commands/RunCommand.cs#L10

There might be something wrong with the logic here.

* The application runs as STDOUT if the application is portable or is user interactive.
* The application runs as console out otherwise

I think what @ramonsmits is seeing here is because he is running as user interactive. I can see an argument to invert that logic so that user interactive runs as console out instead of STDOUT, but would have to validate that it does not break the docker experience again.

ramonsmits:
@WilliamBZA I'm just launching it from VS or Rider and I want to see logging in the console. That is interactive, but that argument is always false so how can it ever be true. 

Just launch all 3 and you'll see no logging.

WilliamBZA:
I'm not disputing what you're saying you're seeing. I'm saying we need make sure we don't break it for docker again.

WilliamBZA:
@ramonsmits like I said, I think the fix is probably https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl.Audit/Infrastructure/Hosting/Commands/RunCommand.cs#L10 and not in the program.cs

TraGicCode:
Hey @WilliamBZA ,

At this point i don't think it works with docker as it looks like things are worse.  

I got the release with the fix ( 4.9.0) and i see nothing being reported in stdout.  I also no longer see anything printing to my powershell console when i run servicecontrol fior audit/error import. 

Before

image


After

image


I tried adding the --portable flag and still saw nothing

WilliamBZA:
@ramonsmits @TraGicCode I've moved this to a [full issue](https://github.com/Particular/ServiceControl/issues/2017) to track better.

We'll get this sorted.

ramonsmits:
Closing in favor of https://github.com/Particular/ServiceControl/pull/2039

SzymonPobiega:
We shipped a fix to logging logic in the latest 4.13.0, adding the `portable` command line option that forces SC to run (and log) as in the console mode.

TraGicCode:
Hey @SzymonPobiega ,

I will give this a shot when i got time and if i find any issues i will let you know.  This is alot cleaner approach then providing some sort of custom nlog.config and is less error prone.

PR #2013, opened 2020-05-14
Add a more helpful exception message when the FailedQ header is missing

ramonsmits:


src/ServiceControl/Operations/FailedMessageFactory.cs
```diff
@@ -35,7 +36,10 @@ public FailureDetails ParseFailureDetails(IReadOnlyDictionary<string, string> he
 
             result.Exception = GetException(headers);
 
-            result.AddressOfFailingEndpoint = headers[FaultsHeaderKeys.FailedQ];
+            if (!headers.ContainsKey(FaultsHeaderKeys.FailedQ))
+            {
+                throw new Exception($"Missing '{FaultsHeaderKeys.FailedQ}' header. Message is poison message or incorrectly send to (error) queue.");
```
WilliamBZA, line 41:
```suggestion
                throw new Exception($"Missing '{FaultsHeaderKeys.FailedQ}' header. Message is poison message or incorrectly sent to (error) queue.");
```

PR #2041, opened 2020-06-03
Learning transport for Debug builds

ramonsmits:
This makes it easy to just clone the repository and run all projects using the learning transport.


danielmarbach:
I was several times about to do the same and everytime I reminded myself there are valid reasons to have a debug build with a different transport for debugging purposes. Every-time I came back to not do it. I would much rather prefer to have something similar like the `connection.txt` support that we have in the ATTs. I guess we could even build a compile step that looks at the existing `connection.txt` and copy the transport definition including the connection string from the `connection.txt` into the output folder on debug builds, but even then it would be dangerous because on the Debug build you get automagically bundled a transport where in release you don't and never should get one. Maybe then the better option is to have a dedicated debug configuration like `DebugWithLearningTransport`? So basically your PR but instead have a different compile configuration and switch all the conditions to that?

danielmarbach:
Closing due to no activity since June

PR #2043, opened 2020-06-03
Fix audit instances to start under Local System account

### LINKED ISSUE #2047: Audit instance in version 4.7 and higher does not start using Local Service account
## Who's affected

Customers trying to install or upgrade to ServiceControl 4.7 and higher with the Local Service account

## Symptoms

When trying to install ServiceControl 4.7 and higher with the Local Service Account the installation or upgrade will not succeed due to ServiceControl Audit not starting. 

The event log might show

```
Application: ServiceControl.Audit.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: Microsoft.Isam.Esent.Interop.EsentTempPathInUseException
   at Microsoft.Isam.Esent.Interop.Api.JetInit(Microsoft.Isam.Esent.Interop.JET_INSTANCE ByRef)
   at Raven.Storage.Esent.TransactionalStorage.Initialize(Raven.Database.Impl.IUuidGenerator, Raven.Abstractions.MEF.OrderedPartCollection`1<Raven.Database.Plugins.AbstractDocumentCodec>, System.Action`1<System.String>, System.Action`2<System.Object,System.Exception>)
```

the detailed exception contains

```
System.InvalidOperationException
  HResult=0x80131509
  Message=Could not open transactional storage: C:\ProgramData\Particular\ServiceControl\ServiceControl.Audit\DB\Data
  Source=<Cannot evaluate the exception source>
  StackTrace:
<Cannot evaluate the exception stack trace>
Inner Exception 1:
EsentTempPathInUseException: Temp path already used by another database instance
```

danielmarbach:
Fixes #2047

Reintroduces running the send-only endpoint of ServiceControl.Audit with installers enabled in the `SetupBootstrapper` to make sure all required infrastructure is created including `RavenBoostrapper` which relies on `INeedInitialization`. Ideally ServiceControl.Audit Host should be in charge of creating and orchestrating the startup process and not NServiceBus. For historic reasons NServiceBus was in charge and still is so we went for still relying on `INeedInitialization`. Embarking a journey of refactoring the startup of ServiceControl is not a risk we were willing to take.

This PR also aligns the acceptance test components to rely on the setup bootstrappers to create the necessary infrastructure. In order to make sure assembly scanning doesn't discover additional assemblies that can cause undesired side effects the setup bootstrapper can be configured with additional optional exclusions

### DIFF
--- a/src/ServiceControl.Audit/Infrastructure/SetupBootstrapper.cs
+++ b/src/ServiceControl.Audit/Infrastructure/SetupBootstrapper.cs
@@ -2,8 +2,13 @@ namespace ServiceControl.Audit.Infrastructure
 {
     using System.Collections.Generic;
     using System.Threading.Tasks;
+    using Autofac;
+    using NServiceBus;
     using NServiceBus.Logging;
     using NServiceBus.Raw;
+    using Raven.Client;
+    using Raven.Client.Embedded;
+    using Settings;
     using Transports;
 
     class SetupBootstrapper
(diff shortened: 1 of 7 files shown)


src/ServiceControl.Audit/Infrastructure/SetupBootstrapper.cs
```diff
@@ -40,6 +45,35 @@ public async Task Run(string username)
 
             //No need to start the raw endpoint to create queues
             await RawEndpoint.Create(config).ConfigureAwait(false);
+
+            var configuration = new EndpointConfiguration(settings.ServiceName);
```
ramonsmits, line 49: This feels wrong, basically doing the RawEndpoint create and the normal Create is also run....

ramonsmits: Also, instead of doing this can't we then just not only create the EndpointConfiguration object and invoke `INeedInitialization` without doing `NServiceBusFactory.Create`?

PR #2067, opened 2020-07-01
Return end count instead of zero to represent the number of attachments cleaned

danielmarbach:
Fixes https://github.com/Particular/ServiceControl/issues/2072


src/ServiceControl/Infrastructure/RavenDB/Expiration/AuditMessageCleaner.cs
```diff
@@ -116,7 +116,7 @@ public static void Clean(int deletionBatchSize, DocumentDatabase database, DateT
                     logger.Debug($"Batching deletion of {s}-{e} attachment audit documents completed.");
                 }
 
-                return 0;
+                return e;
```
mikeminutillo, line 119: I don't think this is right because `e` is the end count. So if you have a batch of 25 things and you chunk it into batches of 10, you will get 3 batches and e will be 10, 20, 25. If you return those, it will sum them up and return 55 which is not what you'd expect.

PR #2068, opened 2020-07-01
Disable InMemoryIndexing on EndpointsIndex

danielmarbach:
RunInMemory means the whole index is stored in memory and is redone from scratch when the server restarts. That makes no sense for any of our indexes. Not even for the endpoint index because it would mean the endpoint index has to run through the entire processed message collection again. In addition to that it unnecessarily increases the memory pressure.


WilliamBZA:
Not needed anymore. Closing.

PR #2078, opened 2020-07-10
Updated to NServiceBus.RabbitMQ 6.0.0

ramonsmits:
Closes https://github.com/Particular/ServiceControl/pull/2077

I've chosen to not break the connection string parsing as that would result in a breaking change.

### DIFF
--- a/src/ServiceControl.Transports.RabbitMQ/ConnectionConfiguration.cs
+++ b/src/ServiceControl.Transports.RabbitMQ/ConnectionConfiguration.cs
@@ -33,7 +33,7 @@
 
         public string Password { get; }
 
-        public ushort RequestedHeartbeat { get; }
+        public TimeSpan RequestedHeartbeat { get; }
 
         public TimeSpan RetryDelay { get; }
 
@@ -51,7 +51,7 @@
             string virtualHost,
             string userName,
             string password,
-            ushort requestedHeartbeat,
+            TimeSpan requestedHeartbeat,
             TimeSpan retryDelay,
             bool useTls,
             string certPath,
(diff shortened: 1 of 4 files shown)


bording:
> I've chosen to not break the connection string parsing as that would result in a breaking change.

Not sure what you mean by this. The way we implemented the connection string parsing in the transport let's both formats be used. We should definitely copy the update code over so that both formats can be used in ServiceControl as well.

bording:
Of course, looking more at the code in here, there are a lot of assumptions around the connection string format baked in here, so there's going to be more work than I initially thought to enable both of them. 😢 

I've at least gone ahead and updated the classes copied from the transport with the latest versions to ensure we are keeping things like default values in sync.

For now, it doesn't seem like we could actually do the additional work to enable the amqp connection string format because of ServiceControl's increasingly outdated assumption that all configuration is passed in via a connection string.

src/ServiceControl.Transports.RabbitMQ/ConnectionConfiguration.cs
```diff
@@ -18,7 +18,7 @@ class ConnectionConfiguration
         const string defaultVirtualHost = "/";
         const string defaultUserName = "guest";
         const string defaultPassword = "guest";
-        const ushort defaultRequestedHeartbeat = 5;
+        const ushort defaultRequestedHeartbeat = 60;
```
tmasternak, line 21: Why did we change this?

bording: This was changed in https://github.com/Particular/NServiceBus.RabbitMQ/pull/608 to align our default value with the client's actual default. Five seconds was way too short, and we should never have been setting it that low.

Issue #2019, opened 2020-05-20
FailedMessageRetries are not cleaned up and can grow to be unbounded


WilliamBZA:
RetryBatches are removed as part of the retry process cleanup, but the FailedMessageRetries documents are not. This could result in a collection of extremely large documents that have no use anymore and add strain to RavenDB for no reason.

These should be cleaned up once batches are removed too.

PR #2095, opened 2020-08-05
Remove FailedMessageRetry when message is successfully retried

WilliamBZA:
A partial fix for #2019 

The FailedMessageRetry document collection is used for the Pending Retries screen, but it seems that the entries here are only cleared when a messages is a repeated failure. When a message is successfully re-processed and ServiceControl receives and audit message confirming it, it should remove the corresponding entry so that the collection does not end up being unbounded in size.

### DIFF
--- a/src/ServiceControl.AcceptanceTests/Recoverability/MessageFailures/FailedMessageRetriesController.cs
+++ b/src/ServiceControl.AcceptanceTests/Recoverability/MessageFailures/FailedMessageRetriesController.cs
@@ -1,13 +1,10 @@
 ﻿namespace ServiceControl.AcceptanceTests.Recoverability.MessageFailures
 {
-    using System;
     using System.Net;
     using System.Net.Http;
-    using System.Threading;
     using System.Threading.Tasks;
     using System.Web.Http;
     using Infrastructure.WebApi;
-    using Operations;
     using Raven.Client;
     using ServiceControl.Recoverability;
 
@@ -25,20 +22,18 @@
 
         [Route("failedmessageretries/count")]
         [HttpGet]
-        public async Task<HttpResponseMessage> GetFailedMessageRetriesCount()
+        public Task<HttpResponseMessage> GetFailedMessageRetriesCount()
         {
             using (var session = store.OpenAsyncSession())
             {
                 var query =
                     session.Query<FailedMessageRetry>().Statistics(out var stats);
 
-                var count = await query.CountAsync();
-
-                return Request.CreateResponse(HttpStatusCode.OK, new FailedMessageRetriesCountReponse
+                return Task.FromResult(Request.CreateResponse(HttpStatusCode.OK, new FailedMessageRetriesCountReponse
                 {
-                    Count = count
+                    Count = stats.TotalResults
                 })
-                .WithEtag(stats);
+                .WithEtag(stats));
             }
         }
 
--- a/src/ServiceControl.AcceptanceTests/Recoverability/MessageFailures/FailedMessageRetriesController.cs
+++ b/src/ServiceControl.AcceptanceTests/Recoverability/MessageFailures/FailedMessageRetriesController.cs
@@ -1,0 +1,47 @@
+﻿namespace ServiceControl.AcceptanceTests.Recoverability.MessageFailures
+{
+    using System;
+    using System.Net;
+    using System.Net.Http;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using System.Web.Http;
+    using Infrastructure.WebApi;
+    using Operations;
+    using Raven.Client;
+    using ServiceControl.Recoverability;
+
+    public class FailedMessageRetriesCountReponse
+    {
+        public int Count { get; set; }
+    }
+
+    public class FailedMessageRetriesController : ApiController
+    {
+        internal FailedMessageRetriesController(IDocumentStore store)
+        {
+            this.store = store;
+        }
+
+        [Route("failedmessageretries/count")]
+        [HttpGet]
+        public async Task<HttpResponseMessage> GetFailedMessageRetriesCount()
+        {
+            using (var session = store.OpenAsyncSession())
+            {
+                var query =
+                    session.Query<FailedMessageRetry>().Statistics(out var stats);
+
+                var count = await query.CountAsync();
+
+                return Request.CreateResponse(HttpStatusCode.OK, new FailedMessageRetriesCountReponse
+                {
+                    Count = count
+                })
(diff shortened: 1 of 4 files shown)


src/ServiceControl.AcceptanceTests/Recoverability/MessageFailures/FailedMessageRetriesController.cs
```diff
@@ -0,0 +1,47 @@
...
+        }
+
+        [Route("failedmessageretries/count")]
+        [HttpGet]
+        public async Task<HttpResponseMessage> GetFailedMessageRetriesCount()
+        {
+            using (var session = store.OpenAsyncSession())
+            {
+                var query =
+                    session.Query<FailedMessageRetry>().Statistics(out var stats);
```
DavidBoike, line 33: Will this reuse the `FailedMessageRetries/ByBatch` index, or create a new one? Or is it assumed that the volume will be low enough that it doesn't matter?

Though I don't totally get why a controller is in the acceptance tests in the first place. Do we not care because it's just for a test?

Also, the `stats` should have `TotalResults` so it shouldn't be necessary to enumerate the results with `CountAsync()`. Though I see this is also in use in FailedErrorsController.

WilliamBZA: > Will this reuse the FailedMessageRetries/ByBatch index, or create a new one? Or is it assumed that the volume will be low enough that it doesn't matter?

Probably create a new one? It's only for the tests and only the actual test instances that use this controller, so it should be fine.

> Though I don't totally get why a controller is in the acceptance tests in the first place. Do we not care because it's just for a test?

I'm following the convention set by the `FailedErrorsController`, I assume the thinking behind it is so that the tests don't have to have raven code directly in them because reasons?

> Also, the stats should have TotalResults so it shouldn't be necessary to enumerate the results with CountAsync(). Though I see this is also in use in FailedErrorsController.

Updated.

danielmarbach: Part of it is a bit legacy reasons. Since the beginning it WA possible to bring in Nancy modules and when we switched to Web api controllers that run inside the instance and therefore have access to everything available on the container. The tests have always used a wired up http client to communicate via the http api with SC and I assume that's why the obvious way expose internal details was also using http by injecting controllers. That being said though I'm pretty sure the new acceptance test base magic I wrote would allow to expose the doc store client directly but probably overkill and not worth diverging from the current way plus I actually like that all the ATTs use one consistent way to interact with the instance

src/ServiceControl.AcceptanceTests/Recoverability/MessageFailures/When_a_failed_message_is_retried.cs
```diff
@@ -0,0 +1,187 @@
...
+                        failedMessageRetries = await this.TryGet<FailedMessageRetriesCountReponse>("/api/failedmessageretries/count");
+
+                        return failedMessageRetries.Count == 0;
+                    }
+
+                    return false;
+                })
+                .Run();
+
+            Assert.AreEqual(failedMessageRetries.Count, 0, "FaileMessageRetries not removed");
```
mikeminutillo, line 67:
```suggestion
            Assert.AreEqual(failedMessageRetries.Count, 0, "FailedMessageRetries not removed");
```

src/ServiceControl.AcceptanceTests/Recoverability/MessageFailures/When_a_failed_message_is_retried.cs
```diff
@@ -0,0 +1,187 @@
...
+        {
+            public FailingEndpoint()
+            {
+                EndpointSetup<DefaultServer>(c =>
+                {
+                    c.EnableFeature<Outbox>();
+                    c.ReportSuccessfulRetriesToServiceControl();
+
+                    var recoverability = c.Recoverability();
+                    recoverability.Immediate(s => s.NumberOfRetries(1));
```
mikeminutillo, line 121: Why do we need a retry?

src/ServiceControl.AcceptanceTests/TestSupport/ServiceControlComponentRunner.cs
```diff
@@ -156,6 +156,7 @@ async Task InitializeServiceControl(ScenarioContext context)
                 bootstrapper = new Bootstrapper(settings, configuration, loggingSettings, builder =>
                 {
                     builder.RegisterType<FailedErrorsController>().FindConstructorsWith(t => t.GetTypeInfo().DeclaredConstructors.ToArray());
+                    builder.RegisterType<FailedMessageRetriesController>().FindConstructorsWith(t => t.GetTypeInfo().DeclaredConstructors.ToArray());
```
mikeminutillo, line 159: If we're going to keep following this pattern, we can replace it with a generic catch-all registration that will pick up all testing controllers:

```csharp
builder.RegisterAssemblyTypes(typeof(FailedErrorsController).Assembly)
  .AssignableTo<ApiController>()
  .FindConstructorsWith(t => t.GetTypeInfo().DeclaredConstructors.ToArray())
  .AsSelf();
```

That way the next test that needs to do it can just add the controller and it will just work.

danielmarbach: This is a nice addition. How about a dedicated PR?

src/ServiceControl.AcceptanceTests/Recoverability/MessageFailures/When_a_failed_message_is_retried.cs
danielmarbach: Could this be a left over that got copy-pasted due to the fact that previously tests run across all the transports and we wanted to make sure things don't get deadlettered?

WilliamBZA: Copy-paste indeed

src/ServiceControl.AcceptanceTests/TestSupport/ServiceControlComponentRunner.cs
WilliamBZA: Added

danielmarbach: I double-checked and the other component runners don't need this code so we should be fine

PR #2102, opened 2020-08-07
Clean FailedMessageRetry when FailedMessage is cleaned

WilliamBZA:


src/ServiceControl/MessageFailures/FailedMessage.cs
```diff
@@ -26,6 +26,11 @@ public static string MakeDocumentId(string messageUniqueId)
             return $"{CollectionName}/{messageUniqueId}";
         }
 
+        public static string GetDocumentIdFromCollectionId(string collectionId)
```
danielmarbach, line 29: Is collection id the correct term? Wouldn't the better term be something like document id?

WilliamBZA: Yep, you're correct

WilliamBZA: Renamed

danielmarbach:
Rebased

src/ServiceControl/Infrastructure/RavenDB/Expiration/ErrorMessageCleaner.cs
```diff
@@ -88,6 +104,22 @@ public static void Clean(int deletionBatchSize, DocumentDatabase database, DateT
                 return results.Count(x => x.Deleted == true);
             }, items, database, token);
 
+            var deletedFailedMessageRetry = Chunker.ExecuteInChunks(failedRetryItems.Count, (itemsForBatch, db, s, e) =>
```
mikeminutillo, line 107: Do we want to include this count in the log message below?

Should we consider putting this (and the attachment deletion) before the message deletion? That would prevent orphaned docs if we clean up the message and then crash.

src/ServiceControl/MessageFailures/FailedMessage.cs
```diff
@@ -26,6 +26,11 @@ public static string MakeDocumentId(string messageUniqueId)
             return $"{CollectionName}/{messageUniqueId}";
         }
 
+        public static string GetMessageIdFromDocumentId(string failedMessageDocumentId)
+        {
+            return failedMessageDocumentId.Substring(CollectionName.Length + 1);
```
mikeminutillo, line 31: If, for whatever reason, the strings are not what we expect, we will throw an exception and stop the whole cleanup process here. Is it worth putting some defensive checks around this?

src/ServiceControl.UnitTests/ApprovalFiles/APIApprovals.PublicClr.approved.txt
```diff
@@ -846,6 +846,7 @@ namespace ServiceControl.MessageFailures
         public System.Collections.Generic.List<ServiceControl.MessageFailures.FailedMessage.ProcessingAttempt> ProcessingAttempts { get; set; }
         public ServiceControl.MessageFailures.FailedMessageStatus Status { get; set; }
         public string UniqueMessageId { get; set; }
+        public static string GetMessageIdFromDocumentId(string failedMessageDocumentId) { }
```
mikeminutillo, line 849: Could this be internal? I'm not sure that we expose this outside of the SC anywhere anyway so it's probably not that important.

src/ServiceControl/MessageFailures/FailedMessage.cs
danielmarbach: I think for perf reasons it would be fair to assume it is only used when the collection name is a guaranteed match. Any defensive check has associated runtime costs

src/ServiceControl.AcceptanceTests/Recoverability/MessageFailures/FailedMessageRetriesController.cs
```diff
@@ -22,17 +22,17 @@ internal FailedMessageRetriesController(IDocumentStore store)
 
         [Route("failedmessageretries/count")]
         [HttpGet]
-        public Task<HttpResponseMessage> GetFailedMessageRetriesCount()
+        public async Task<HttpResponseMessage> GetFailedMessageRetriesCount()
         {
             using (var session = store.OpenAsyncSession())
             {
-                session.Query<FailedMessageRetry>().Statistics(out var stats);
+                await session.Query<FailedMessageRetry>().Statistics(out var stats).ToListAsync().ConfigureAwait(false);
```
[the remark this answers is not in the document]
WilliamBZA: When I tried the code without the `ToListAsync` the stats.TotalCount was always 0. I assume without actually executing something, the stats object is never populated.

src/ServiceControl/Infrastructure/RavenDB/Expiration/ErrorMessageCleaner.cs
WilliamBZA: Good catch!

src/ServiceControl.UnitTests/ApprovalFiles/APIApprovals.PublicClr.approved.txt
WilliamBZA: :+1: Changed

Issue #1945, opened 2020-03-20
ServiceControl Audit crashes after version store reaches maximum size


mikeminutillo:
Under under high load, ServiceControl Audit will run for a while and then crash with the following error:

> The version store for this instance (0) has reached its maximum size of 1023Mb. It is likely that a long-running transaction is preventing cleanup of the version store and causing it to build up in size. Updates will be rejected until the long-running transaction has been completely committed or rolled back.

## Workaround

If you run into this problem you can reduce the concurrency on the ServiceControl Audit instance to make it more stable. Add the following into the ServiceControl.Audit.exe.config appSettings:

```xml
<add key="ServiceControl.Audit/MaximumConcurrencyLevel" value="5" />
```

If this key is not present, the system will use the default value of 10. Reducing concurrency to 5 will reduce the pressure on the embedded database and stabilize the system.

This may cause the audit queue to grow continuously as the audit instance is no longer able to keep up with load. If that happens, consider [adding a 2nd audit instance](https://docs.particular.net/servicecontrol/servicecontrol-instances/remotes) to your installation.

PR #2106, opened 2020-08-10
Gather known endpoints during batch ingestion

danielmarbach:
Closes https://github.com/Particular/ServiceControl/issues/1945


src/ServiceControl.Audit/Auditing/AuditIngestionComponent.cs
```diff
@@ -43,6 +57,14 @@ AuditIngestionCustomCheck.State ingestionState
             watchdog = new Watchdog(ingestion.EnsureStarted, ingestion.EnsureStopped, ingestionState.ReportError,
                 ingestionState.Clear, settings.TimeToRestartAuditIngestionAfterFailure, log, "failed message ingestion");
 
+            channel = Channel.CreateBounded<MessageContext>(new BoundedChannelOptions(settings.MaximumConcurrencyLevel * 2)
```
DavidBoike, line 60: Is MaxConcurrencyLevel equivalent of how many in-flight messages are allowed by the semaphore? Wouldn't it then be impossible for the channel to fill with any more than the MaxConcurrencyLevel of messages?

Either way some comments on this line about those assumptions would be really good.

src/ServiceControl.Audit/Auditing/AuditIngestionComponent.cs
```diff
@@ -64,9 +86,33 @@ public Task Start(IMessageSession session)
...
+
+        async Task Loop()
+        {
+            var contexts = new List<MessageContext>(settings.MaximumConcurrencyLevel);
+
+            while (await channel.Reader.WaitToReadAsync().ConfigureAwait(false))
+            {
+                try
+                {
+                    while (channel.Reader.TryRead(out var context))
```
DavidBoike, line 104: Comments here about what conditions lead to the TryRead returning false would be good. Is it bound by the in-flight messages semaphore? Does it return false if there's a lag in being able to read something (which would require another WaitToReadAsync to return true) and then what is that caused by? Are the batches here firmly constrained to a specific number?

src/ServiceControl.Audit/Auditing/AuditPersister.cs
```diff
@@ -24,55 +29,157 @@ public void Initialize(IMessageSession messageSession)
...
+            var stopwatch = Stopwatch.StartNew();
+
+            if (Logger.IsDebugEnabled)
             {
-                messageId = DeterministicGuid.MakeId(message.MessageId).ToString();
+                Logger.Debug($"Batch size {contexts.Count}");
             }
 
-            var metadata = new ConcurrentDictionary<string, object>
+            using (var bulkInsert = store.BulkInsert(options: new BulkInsertOptions
```
DavidBoike, line 41: Since bulkInsert.Dispose() effectively completes the bulk insert, should we be concerned that a random exception halfway through would fall out of the using scope and complete that bulk insert?

danielmarbach: All the operations are idempotent so I see no big deal here. Still, it is a good comment. I checked the `DisposeAsync` and the `Dispose` implementation of the bulk insert operations. If you explicitly call `DisposeAsync` then `Dispose` does nothing. I saw that ConcurrencyViolation and other exceptions are raised either on `Dispose` or `DisposeAsync`. For now, I removed the `using` block and expanded the try-catch just in case eventhough the `ProcessedMessage` method does a general try-catch and early release the transport transaction

src/ServiceControl.Audit/Auditing/AuditIngestionComponent.cs
danielmarbach: I verified again and setting it to `MaxConcurrencyLevel` is good enough. I pushed a commit that changes it and does also specify the full mode enum even though it uses the default value. I've also added a few comments to the loop. 

In case you are interested to see how it behaves here is a sample

https://github.com/danielmarbach/RawAndChannel

src/ServiceControl.Audit/Auditing/AuditIngestionComponent.cs
danielmarbach: I've added a few comments. See if that is enough

PR #2107, opened 2020-08-11
Migrate endpoints from index to collection

WilliamBZA:


src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
```diff
@@ -0,0 +1,88 @@
...
+            if (knownEndpointsIndex == null)
+            {
+                return;
+            }
+
+            await WaitForNonStaleIndex(knownEndpointsIndex.Name).ConfigureAwait(false);
+            
+            using (var session = store.OpenAsyncSession())
+            {
+                var endpointsFromIndex = await session.Query<dynamic>(knownEndpointsIndex.Name, true).ToListAsync().ConfigureAwait(false);
```
WilliamBZA, line 47: Assumption: The list returned will be small enough to work in memory

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
```diff
@@ -0,0 +1,88 @@
...
+                var knownEndpoints = endpointsFromIndex.Select(endpoint => new KnownEndpoint
+                {
+                    Id = KnownEndpoint.MakeDocumentId(endpoint.Name, Guid.Parse(endpoint.HostId)),
+                    Host = endpoint.Host,
+                    HostId = Guid.Parse(endpoint.HostId),
+                    Name = endpoint.Name,
+                    LastSeen = DateTime.Now // We don't store it anywhere so....?
```
WilliamBZA, line 54: This isn't recorded anywhere on the index

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
```diff
@@ -0,0 +1,88 @@
...
+using ServiceControl.Audit.Monitoring;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace ServiceControl.Audit.Infrastructure.RavenDB
+{
+    class MigrateKnownEndpoints : Feature
```
danielmarbach, line 13: So this works as long as we have the sendonly endpoint still present in the audit instance.

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
```diff
@@ -0,0 +1,88 @@
+﻿using NServiceBus;
```
danielmarbach, line 1: Usings should go inside

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
danielmarbach: We do in `ProcessedMessage`

```
            if (Headers.TryGetValue(NServiceBus.Headers.ProcessingEnded, out var processedAt))
            {
                ProcessedAt = DateTimeExtensions.ToUtcDateTime(processedAt);
            }
            else
            {
                ProcessedAt = DateTime.UtcNow; // best guess
            }
```

so how about using `DateTime.UtcNow` as well?

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        protected override async Task OnStart(IMessageSession messageSession)
+        {
+            var knownEndpointsIndex = await store.AsyncDatabaseCommands.GetIndexAsync("EndpointsIndex").ConfigureAwait(false);
+            if (knownEndpointsIndex == null)
+            {
+                return;
+            }
+
+            await WaitForNonStaleIndex(knownEndpointsIndex.Name).ConfigureAwait(false);
```
danielmarbach, line 43: Isn't it fair to assume we can even operate on stale indexes? Given the index is a fanout index I'm sceptic it is good to wait for it to be non-stale. I would say the migration can be a best effort and we assume it to catch up over time

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
danielmarbach: I think that is a fair assumption considering the complexity of running multiple endpoints as well as our new licensing model. Should we set an explicit limit so that it doesn't crash should there be more though?

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
```diff
@@ -0,0 +1,88 @@
...
+
+                foreach (var endpoint in knownEndpoints)
+                {
+                    await session.StoreAsync(endpoint).ConfigureAwait(false);
+                }
+
+                await session.SaveChangesAsync().ConfigureAwait(false);
+            }
+
+            await store.AsyncDatabaseCommands.DeleteIndexAsync(knownEndpointsIndex.Name).ConfigureAwait(false);
```
danielmarbach, line 65: Is this idempotent?

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
```diff
@@ -0,0 +1,88 @@
...
+                var knownEndpoints = endpointsFromIndex.Select(endpoint => new KnownEndpoint
+                {
+                    Id = KnownEndpoint.MakeDocumentId(endpoint.Name, Guid.Parse(endpoint.HostId)),
+                    Host = endpoint.Host,
+                    HostId = Guid.Parse(endpoint.HostId),
+                    Name = endpoint.Name,
+                    LastSeen = DateTime.Now // We don't store it anywhere so....?
+                });
+
+                foreach (var endpoint in knownEndpoints)
```
danielmarbach, line 57: Maybe use a bulk insert session?

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
WilliamBZA: Correct. There are other Feature's in the audit instance so I feel it's safe enough as removing the endpoint will be a much larger process and won't likely result in forgotten features running.

I don't like that this is essentially a one-off while all the other features in the audit instance are long-running.

Do you have a suggestion for a better place to put this?

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
WilliamBZA: The problem is the index will be deleted at the end of this code, so if it's stale it will never be caught up. We could wait for an hour before proceeding to give it a decent chance to update but not completely block.

danielmarbach: On a system with continuous load it will never be caught up. I think it is to dangerous to assume we ever can catch up

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
WilliamBZA: Looks like it. If it deletes an index that doesn't exist (or has been previously deleted), it appears to be a NOOP

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
WilliamBZA: It now waits an hour before giving up and just yolo'ing through the list

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
WilliamBZA: Paging has been added

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
danielmarbach: That would hold up SC Audit from starting fro over an hour. That doesn't seem to be a good solution. Do it entirely in the background?

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
```diff
@@ -0,0 +1,111 @@
...
+                            OverwriteExisting = true
+                        }))
+                        {
+                            foreach (var endpoint in knownEndpoints)
+                            {
+                                bulkInsert.Store(endpoint);
+                            }
+                        }
+
+                        await session.SaveChangesAsync().ConfigureAwait(false);
```
DavidBoike, line 82: What changes are being saved here? The bulk insert is separate, right? (So if my suspicion is correct, this line is not needed, as the only thing the session is doing is querying.)

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
mikeminutillo: I think it's fine to run off of non-stale data. The known endpoints collection likely doesn't change very often anyway. The intent of doing the migration is to try and capture the state before the switch. The stale index _is_ the state before the switch. The only case we aren't covering is the one where a endpoint has been added but has not yet appeared in the index. Chances are, that endpoint will still be sending/processing messages and will get picked up in the first few minutes of execution. In the rare case that an endpoint is not yet in the index and is no long sending/processing messages then I think it's OK for us to miss it.

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
```diff
@@ -0,0 +1,111 @@
...
+
+            await WaitForNonStaleIndex(knownEndpointsIndex.Name).ConfigureAwait(false);
+
+            int previouslyDone = 0;
+            do {
+                using (var session = store.OpenAsyncSession())
+                {
+                    var endpointsFromIndex = await session.Query<dynamic>(knownEndpointsIndex.Name, true)
+                        .Take(1024)
+                        .Skip(previouslyDone)
```
mikeminutillo, line 50: Shouldn't this be `Skip` then `Take`?

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
```diff
@@ -0,0 +1,111 @@
...
+                        .ToListAsync()
+                        .ConfigureAwait(false);
+
+                    if (endpointsFromIndex.Count == 0)
+                    {
+                        previouslyDone = -1;
```
mikeminutillo, line 56: Could also `break` out of the loop here and make this a `while(true)`.

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
mikeminutillo: Given that we create the endpoint during the upgrade process to run the installers, and that will execute this feature, we could consider moving this to an installer.

WilliamBZA: I'm moving this out of a feature anyway

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
WilliamBZA: It's now an installer

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
WilliamBZA: Done

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
WilliamBZA: The order doesn't matter until the expression is evaluated, but I've change it around for readability

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
```diff
@@ -0,0 +1,89 @@
...
+                    }
+                }
+            } while (true);
+
+            await Store.AsyncDatabaseCommands.DeleteIndexAsync(knownEndpointsIndex.Name).ConfigureAwait(false);
+        }
+
+        async Task WaitForNonStaleIndex(string indexName)
+        {
+            var maxWaitTime = DateTime.UtcNow.AddMinutes(5);
```
danielmarbach, line 75: Will the customers have the patience to wait for 5 minutes while the only thing they see on the UI is "Creating queues..." ?

I still think we should just remove the `WaitForNonStaleIndex`. Happy to be convinced otherwise though

WilliamBZA: Removed

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
```diff
@@ -0,0 +1,71 @@
...
+
+                    using (var bulkInsert = Store.BulkInsert(options: new BulkInsertOptions
+                    {
+                        OverwriteExisting = true
+                    }))
+                    {
+                        foreach (var endpoint in knownEndpoints)
+                        {
+                            bulkInsert.Store(endpoint);
+                        }
```
danielmarbach, line 63: I would explicitly call `await bulkInsert.DisposeAsync().ConfigureAwait(false)` to be intention revealing

WilliamBZA: Done

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
```diff
@@ -0,0 +1,71 @@
...
+    using Raven.Abstractions.Data;
+    using Raven.Client;
+    using ServiceControl.Audit.Monitoring;
+    using System;
+    using System.Linq;
+    using System.Threading.Tasks;
+
+    class MigrateKnownEndpoints : INeedToInstallSomething
+    {
+        public IDocumentStore Store { get; set; }
```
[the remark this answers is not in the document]
WilliamBZA: Changed to ctor injection

src/ServiceControl.Audit.UnitTests/KnownEndpointsMigration/EndpointMigrationTests.cs
```diff
@@ -0,0 +1,223 @@
...
+    using System.Linq;
+    using System.Threading.Tasks;
+    using Audit.Auditing;
+    using NUnit.Framework;
+    using Raven.Client.Indexes;
+    using ServiceControl.Audit.Infrastructure.RavenDB;
+    using ServiceControl.Audit.Monitoring;
+
+    [TestFixture]
+    public class EndpointMigrationTests
```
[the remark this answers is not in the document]
WilliamBZA: I've added a test for that

src/ServiceControl.Audit.UnitTests/KnownEndpointsMigration/EndpointMigrationTests.cs
```diff
@@ -0,0 +1,223 @@
...
+                    var loadedReceiverEndpoint = await session.LoadAsync<KnownEndpoint>(KnownEndpoint.MakeDocumentId("ReceivingHostName", receiverHostId)).ConfigureAwait(false);
+
+                    Assert.NotNull(loadedReceiverEndpoint);
+                    Assert.NotNull(loadedSenderEndpoint);
+                }
+            }
+        }
+
+        [Test]
+        public async Task Should_migrate_idempotently()
```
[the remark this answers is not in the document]
WilliamBZA: There is a dedicated test for both disabled and for deleted.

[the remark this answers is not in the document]
WilliamBZA: Correction: I changed the test to run the migration 5 times to make sure

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
```diff
@@ -0,0 +1,90 @@
...
+                            bulkInsert.Store(endpoint);
+                        }
+
+                        await bulkInsert.DisposeAsync().ConfigureAwait(false);
+                    }
+                }
+            } while (true);
+
+            // Disable the index so it can be safely deleted in the next migration run
+            await store.AsyncDatabaseCommands.SetIndexPriorityAsync(knownEndpointsIndex.Name, IndexingPriority.Disabled).ConfigureAwait(false);
```
danielmarbach, line 87: image

Say you can query a disabled index. If that is also true for 3.5 we might even consider switching the order? What do you think?

src/ServiceControl.Audit/Infrastructure/SetupBootstrapper.cs
```diff
@@ -72,6 +73,7 @@ public async Task Run(string username)
             var documentStore = new EmbeddableDocumentStore();
             containerBuilder.RegisterInstance(documentStore).As<IDocumentStore>().ExternallyOwned();
             containerBuilder.RegisterInstance(settings).SingleInstance();
+            containerBuilder.RegisterType<MigrateKnownEndpoints>();
```
danielmarbach, line 76: Was this necessary for it to work? I thought installer are automatically picked up by scanning and registered by NServiceBus. Just curious not suggestion you have to change it

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
DavidBoike: Then you could disable the index, fail, run again, and the index would get deleted with no conversion.

danielmarbach: Good point, so obvious. Wonder why I haven't thought about that.

src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
mikeminutillo: That's Raven weirdness then as order should matter. `Enumerable.Range(1, 6).Take(2).Skip(2)` returns an empty collection, `Enumerable.Range(1,6).Skip(2).Take(2)` returns `[3, 4]`

src/ServiceControl.Audit/Infrastructure/SetupBootstrapper.cs
WilliamBZA: Error when creating the instance without doing this.

PR #2109, opened 2020-08-11
Fix retrieval of bodies larger than the large object heap but below the maximum body size to store

danielmarbach:
Fixes regression introduced in https://github.com/Particular/ServiceControl/pull/1843

```
            private static readonly byte[] body = Encoding.UTF8.GetBytes($@"<?xml version=""1.0""?>
<MyMessage xmlns:xsi=""http://www.w3.org/2001/XMLSchema-instance"" xmlns:xsd=""http://www.w3.org/2001/XMLSchema""
            xmlns=""http://tempuri.net/"">
   <MyProperty>{new string('a', 87 * 1024)}</MyProperty>
</MyMessage>");
```

## Before

image

## After

image


WilliamBZA:
`When_processed_message_is_imported.Should_be_accessible_via_the_rest_api` is failing now

danielmarbach:
@WilliamBZA just pushed a fix

PR #2111, opened 2020-08-13
Add logging

WilliamBZA:


src/ServiceControl.Audit/Infrastructure/RavenDB/MigrateKnownEndpoints.cs
```diff
@@ -27,6 +28,10 @@ internal async Task MigrateEndpoints(int pageSize = 1024)
             var knownEndpointsIndex = await store.AsyncDatabaseCommands.GetIndexAsync("EndpointsIndex").ConfigureAwait(false);
             if (knownEndpointsIndex == null)
             {
+                if (Logger.IsDebugEnabled)
```
danielmarbach, line 31: Given this is not on the ingestion path technically you could also just go straight forward to use

```
 Logger.Debug
```

WilliamBZA: Removed

PR #2114, opened 2020-08-13
Interleave bulk inserts if possible

danielmarbach:


danielmarbach:
Not worth the complexity yet

image

vs

image

PR #2139, opened 2020-09-03
Make archive group operation idempotent

### LINKED ISSUE #2138: Race condition could result in exception during ArchiveAllInGroup
Requires a certain set of timings:

* Multiple requests to `api/recoverability/groups/{groupId}/errors/archive` for the same group
* Both requests are processed at near the same time interleaving:
  * One request does the check on [line 28](https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/Recoverability/API/FailureGroupsArchiveController.cs#L28) to see if an operation is already in progress
  * Second request then does the same check, but before [line 30](https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/Recoverability/API/FailureGroupsArchiveController.cs#L30) sets an operation
  * At this point, 2 `ArchiveAllInGroup` messages will be sent for processing.
* If one message is processed only after another has completed, then when the second tries to [load the group details](https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/Recoverability/Archiving/ArchiveAllInGroupHandler.cs#L41), the group will not exist and an exception will be raised.
* Recoverability will kick in, but this message will never be processed and will end up in ServiceControl's error queue.

WilliamBZA:
Fixes #2138

### DIFF
--- a/src/ServiceControl.UnitTests/Archiving/ArchiveGroupTests.cs
+++ b/src/ServiceControl.UnitTests/Archiving/ArchiveGroupTests.cs
@@ -21,7 +21,7 @@
 
                 using (var session = documentStore.OpenAsyncSession())
                 {
-                    var previousAchiveBatch = new ArchiveBatch {Id = previousArchiveBatchId};
+                    var previousAchiveBatch = new ArchiveBatch { Id = previousArchiveBatchId };
                     await session.StoreAsync(previousAchiveBatch)
                         .ConfigureAwait(false);
 
@@ -51,7 +51,7 @@
                 var handler = new ArchiveAllInGroupHandler(documentStore, domainEvents, documentManager, archivingManager, retryingManager);
 
                 var context = new TestableMessageHandlerContext();
-                var message = new ArchiveAllInGroup {GroupId = groupId};
+                var message = new ArchiveAllInGroup { GroupId = groupId };
 
                 // Act
                 await handler.Handle(message, context)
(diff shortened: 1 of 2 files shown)


src/ServiceControl.UnitTests/Archiving/ArchiveGroupTests.cs
```diff
@@ -65,5 +65,60 @@ await handler.Handle(message, context)
...
+                var domainEvents = new FakeDomainEvents();
+                var archivingManager = new ArchivingManager(domainEvents);
+                var retryingManager = new RetryingManager(domainEvents);
+                var handler = new ArchiveAllInGroupHandler(documentStore, domainEvents, documentManager, archivingManager, retryingManager);
+
+                var context = new TestableMessageHandlerContext();
+                var message = new ArchiveAllInGroup { GroupId = groupId + "Invalid" };
+
+                // Act
+                await handler.Handle(message, context)
```
danielmarbach, line 116: You could use `Assert.DoesNotThrow`

src/ServiceControl/Recoverability/Archiving/ArchiveDocumentManager.cs
```diff
@@ -87,8 +87,8 @@ public async Task<GroupDetails> GetGroupDetails(IAsyncDocumentSession session, s
 
             return new GroupDetails
             {
-                NumberOfMessagesInGroup = group.Count,
-                GroupName = group.Title ?? "Undefined"
+                NumberOfMessagesInGroup = group?.Count ?? 0,
+                GroupName = group?.Title ?? "Undefined"
```
danielmarbach, line 91: What is the consequence of adding Undefined here from the UI perspective?

src/ServiceControl.UnitTests/Archiving/ArchiveGroupTests.cs
```diff
@@ -65,5 +65,60 @@ await handler.Handle(message, context)
...
+            {
+                var failureGroupsViewIndex = new FailureGroupsViewIndex();
+                await failureGroupsViewIndex.ExecuteAsync(documentStore);
+
+                var groupId = "TestGroup";
+                var previousArchiveBatchId = ArchiveBatch.MakeId(groupId, ArchiveType.FailureGroup, 1);
+
+                using (var session = documentStore.OpenAsyncSession())
+                {
+                    var previousAchiveBatch = new ArchiveBatch { Id = previousArchiveBatchId };
```
danielmarbach, line 83: Spelling in `previousAchiveBatch `

src/ServiceControl.UnitTests/Archiving/ArchiveGroupTests.cs
```diff
@@ -65,5 +65,60 @@ await handler.Handle(message, context)
...
+
+                    await session.SaveChangesAsync()
+                        .ConfigureAwait(false);
+                }
+
+                var documentManager = new ArchiveDocumentManager();
+                var domainEvents = new FakeDomainEvents();
+                var archivingManager = new ArchivingManager(domainEvents);
+                var retryingManager = new RetryingManager(domainEvents);
+                var handler = new ArchiveAllInGroupHandler(documentStore, domainEvents, documentManager, archivingManager, retryingManager);
```
danielmarbach, line 110: Given that we only care about the handler I would tend to inline the things than can be inlined

src/ServiceControl/Recoverability/Archiving/ArchiveDocumentManager.cs
WilliamBZA: No change in UI behaviour

src/ServiceControl.UnitTests/Archiving/ArchiveGroupTests.cs
WilliamBZA: Done

src/ServiceControl.UnitTests/Archiving/ArchiveGroupTests.cs
WilliamBZA: Thanks, also fixed spelling in other test

src/ServiceControl.UnitTests/Archiving/ArchiveGroupTests.cs
WilliamBZA: Inlined. Also did it for the other test

danielmarbach (approved):

PR #2142, opened 2020-09-04
Fix stopping error and audit ingestion when poison message is received

SzymonPobiega:
- Ensure `onCriticalError` callback is invoked on a separate thread similarly to what NServiceBus does in `CriticalError` class to prevent deadlocks when the critical error is triggered from within message processing code
 - Remove the code that tried to store the message that failed to be ingested in an error queue as this was never intended. The correct behavior of ServiceControl is to store these messages in the RavenDB database
 - Add debug-level logging in the `Watchdog` class
 - Fix command-line audit ingestion by changing the root type resolved from the container

### DIFF
--- a/src/ServiceControl.Audit.AcceptanceTests/Auditing/FailedAuditsController.cs
+++ b/src/ServiceControl.Audit.AcceptanceTests/Auditing/FailedAuditsController.cs
@@ -12,10 +12,10 @@
 
     public class FailedAuditsController : ApiController
     {
-        internal FailedAuditsController(IDocumentStore store, Lazy<ImportFailedAudits> importFailedAudits)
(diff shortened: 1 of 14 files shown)


SzymonPobiega:
While investigating the test failure I found [another bug](https://github.com/Particular/ServiceControl/issues/2143) that should be fixed in 5d443c1. This bug was there since we split the audits but was never reported by customers because failure to ingest audit messages is rare and the problem only manifests itself if audit forwarding is turned on.

mikeminutillo (approved):

SzymonPobiega:
@tmasternak @seanfarmar want to review this one or can I merge it as-is? :) Waiting for your answer till the afternoon today :P

PR #2144, opened 2020-09-07
`FailedMessage` patch with failure de-duplication by AttemptedAt

tmasternak:
This does de-duplication of the `ProcessingAttempts` based on `TimeOfFailure` value. This is stored to the microsecond so a chance that we will have two failures for the same message happening at the same time (to the microsecond) is slim.

That said there are are two (a least) additional angles to this change:
  * de-duplication based on `TimeOfFailure` is a hack. I think we should introduce a proper `ProcessingAttemptId`
  * I'm not an expert in Raven but my hunch is that Javascript patch might be significantly slower than non-Javascript patch. I propose to do some testing around that

Thoughts?

### DIFF
--- a/src/ServiceControl/Operations/ErrorPersister.cs
+++ b/src/ServiceControl/Operations/ErrorPersister.cs
@@ -84,67 +84,47 @@
         {
             var documentId = FailedMessage.MakeDocumentId(uniqueMessageId);
 
-            await store.AsyncDatabaseCommands.PatchAsync(documentId,
-                new[]
-                {
-                    new PatchRequest
-                    {
-                        Name = nameof(FailedMessage.Status),
-                        Type = PatchCommandType.Set,
-                        Value = (int)FailedMessageStatus.Unresolved
-                    },
-                    new PatchRequest
-                    {
-                        Name = nameof(FailedMessage.FailureGroups),
-                        Type = PatchCommandType.Set,
-                        Value = RavenJToken.FromObject(groups)
-                    }
-                },
-                new[]
-                {
-                    new PatchRequest
-                    {
-                        Name = nameof(FailedMessage.UniqueMessageId),
-                        Type = PatchCommandType.Set,
-                        Value = uniqueMessageId
-                    },
-                    new PatchRequest
-                    {
-                        Name = nameof(FailedMessage.Status),
-                        Type = PatchCommandType.Set,
-                        Value = (int)FailedMessageStatus.Unresolved
-                    },
-                    new PatchRequest
-                    {
-                        Name = nameof(FailedMessage.FailureGroups),
-                        Type = PatchCommandType.Set,
-                        Value = RavenJToken.FromObject(groups)
-                    }
-                }, JObjectMetadata
-            ).ConfigureAwait(false);
-
             var attemptedAtField = nameof(FailedMessage.ProcessingAttempt.AttemptedAt);
             var processingAttemptsField = nameof(FailedMessage.ProcessingAttempts);
+            var statusField = nameof(FailedMessage.Status);
+            var failureGroupsField = nameof(FailedMessage.FailureGroups);
+            var uniqueMessageIdField = nameof(FailedMessage.UniqueMessageId);
 
-            var script = $@"var duplicateIndex = _.findIndex(this.{processingAttemptsField}, function(a){{ 
-                                return a.{attemptedAtField} === attempt.{attemptedAtField};
-                            }});
-
-                            output(attempt.{attemptedAtField});
-
-                            if(duplicateIndex === -1){{
-                                this.{processingAttemptsField} = _.union(this.{processingAttemptsField}, [attempt]);
-                            }}";
-
-            var response =    await store.AsyncDatabaseCommands.PatchAsync(documentId,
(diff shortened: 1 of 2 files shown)


src/ServiceControl/Operations/ErrorPersister.cs
```diff
@@ -80,60 +80,52 @@ await SaveToDb(message.Headers.UniqueId(), processingAttempt, groups)
...
-                        Value = RavenJToken.FromObject(processingAttempt, Serializer) // Need to specify serilaizer here because otherwise the $type for EndpointDetails is missing and this causes EventDispatcher to blow up!
-                    },
-                    new PatchRequest
-                    {
-                        Name = nameof(FailedMessage.FailureGroups),
-                        Type = PatchCommandType.Set,
-                        Value = RavenJToken.FromObject(groups)
+                        { "status", (int)FailedMessageStatus.Unresolved },
+                        { "failureGroups", RavenJToken.FromObject(groups) },
+                        { "attempt", RavenJToken.FromObject(processingAttempt, Serializer)},
```
mikeminutillo, line 124: Is it worth pulling the `failureGroups` and `attempt` definitions out so that we aren't calling `RavenJToken.FromObject` multiple times?

src/ServiceControl/Operations/ErrorPersister.cs
```diff
@@ -80,60 +80,52 @@ await SaveToDb(message.Headers.UniqueId(), processingAttempt, groups)
...
+                        { "uniqueMessageId", uniqueMessageId }
                     }
-                }, JObjectMetadata
-            );
+                },
+                JObjectMetadata).ConfigureAwait(false);
```
mikeminutillo, line 128: Why switch this to `async-await`?

mikeminutillo (commented):
I'm happy with this change but we should do some perf tests on it to see if it has a significant impact.

mikeminutillo:
> I think we should introduce a proper `ProcessingAttemptId`
:+1: :100:

Agreed. I would go further and make that the primary key and store each processing attempt separately. That way we have a forward-only state machine. `Failed` => `Archived` || `Retried`. Anything else we do to show prior attempts or to aggregate that data in any way becomes a display concern.

SzymonPobiega:
> That way we have a forward-only state machine

❤️ Loops in state machines are the root of all evil

tmasternak:
@mikeminutillo after doing some further analysis with Szymon and Sean we decided that moving `ProcessingAttempt` out of the `FailedMessage` is far from trivial:
* There are quite a few indexes that we would need to change and test,
* There are some fields in the `FailedMessage` that require consistency with attempt e.g. `Status` and `Count`. This would mean that we would need to do 2 writes first for the `ProcessingAttempt` and if the document is not duplicate we could update the `FailedMessage` properly,
* We considered the possibility of having `LatestProcessingAttempt` on the `FailedMesssage` but that left us still with 2 writes
* Finally we tried to figure out if we can store the `Attempts` in the `FailedMessage` as a map and have a patch that would update the map based on the insert key (`AttemptedAt`) but it looks that RavenDB does not support that.

In the end, we decided to go with the scripted patch approach. I'm doing performance tests and will update this issue with data.

tmasternak:
### Rabbit Performance Test
Local RabbitMQ instance, 10K failed attempts in each test 

|Attempts per msg| ScriptedPatch [msg/sec)]| Current [msg/sec]|
|---|--|-------|
|2 | 38| 37|
|5| 35| 44|
|10K (all attempts for single doc) | 50 msg/sec | N/A|

SzymonPobiega:
I don't see any performance issues here 🙈

tmasternak:
Performance tests on ASBS transport do not show any performance difference.

mikeminutillo (approved):

mikeminutillo:
I'm happy with those perf results. Most ingested error messages should have a single processing attempt and we don't see an appreciable difference when that holds true. In environments where there a significantly more processing attempts (and the table shows up to 5), going a little slower but potentially storing less data is acceptable. The times when we get large batches of errors with new processing attempts are likely to be after a ServiceControl initiated bulk-retry, at which time the server may be busy managing that process anyway.

ramonsmits:
> > I think we should introduce a proper `ProcessingAttemptId`
> 
> 👍 💯
> 
> Agreed. I would go further and make that the primary key and store each processing attempt separately. That way we have a forward-only state machine. `Failed` => `Archived` || `Retried`. Anything else we do to show prior attempts or to aggregate that data in any way becomes a display concern.

Exactly what I said to @tmasternak, why do we expand a document? The current design isn't good for queurying, paging, etc. and we should have a CQRS type of design in a future version to deal with those indexes.

Not even having deduplication as a key but having duplicate makes it possible to do any kind of deduplication on the query side if needed or alternatively generate a unique key and start storing the attempts separately with that ID and add the ID in the attempts collection.

I didn't' expect any throughput performance issues simply because the operation is IO bound and any javascript CPU increase will not increase IO but potentially only CPU so a better performance test would be comparing CPU usage although I'm not really expecting a significant increase.

ramonsmits:
By the way, microseconds is a hack but physically (yet...) impossible to do multiple retries that start in the same microsecond.

Issue #2137, opened 2020-09-03
Configurable naming of the shared Azure Service Bus topic (bundle-1)


pihalve:
I would like to suggest making it possible to configure the name of the shared topic (bundle-1) when using Service Control with Azure Service Bus transport. This name is configurable in other endpoints, so I think it makes good sense to also have it configurable in Service Control, so it doesn't auto-create a topic called bundle-1 when all other endpoints may have been configured to use e.g. tst-bundle-1 (or whatever name makes sense in the given situation.)

boblangley:
@pihalve Thank you for the request. We will update this issue when the status changes.

SzymonPobiega:
This has been fixed in #2153. We will be releasing a new version of ServiceControl that includes this fix soon.

SzymonPobiega:
@pihalve Let me close this issue now so that we can generate release notes. The fix will be released in 4.13.0 early next week.

SzymonPobiega:
Hi @pihalve . I've just pushed the latest version that contains this fix.

PR #2153, opened 2020-09-14
`TopicName` configuration option support in ASB transport

tmasternak:
Connected to https://github.com/Particular/ServiceControl/issues/2137

### DIFF
--- a/src/ServiceControl.Transports.ASBS/ASBSTransportCustomization.cs
+++ b/src/ServiceControl.Transports.ASBS/ASBSTransportCustomization.cs
@@ -1,5 +1,6 @@
 ﻿namespace ServiceControl.Transports.ASBS
 {
+    using System.Data.Common;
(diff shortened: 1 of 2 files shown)


SzymonPobiega (approved):

mikeminutillo (approved):
The legacy ASB transport has to remove these custom parts from the connection string before providing it to the transport. Since there's already another custom part here and it is not removed, I am assuming that's not necessary for ASBS

tmasternak:
> The legacy ASB transport has to remove these custom parts from the connection string before providing it to the transport. Since there's already another custom part here and it is not removed, I am assuming that's not necessary for ASBS

@mikeminutillo yes the reason is that the older SDK parsing logic assumes that if all parts have been found, all that is left is the accessKey. Which causes problems when users add custom parts at the end. With the newer SDK this is not the case.

Issue #1056, opened 2017-12-19
ServiceControl does not deduplicate audit messages


SzymonPobiega:
### Symptoms

In case of network issues audit messages might get duplicated and, as a result, ServiceInsight shows multiple copies of the same message,

### Who's affected

Endpoints in `SendsAtomicWithReceive` and lower transaction modes. All Outbox-enabled endpoints are affected.

SzymonPobiega:
@Particular/servicecontrol-maintainers what do you think?

WilliamBZA:
I think this can be left as is for now. The performance impact is not worth a release that changes existing behaviour. This must however be addressed in the technical debt work.

SzymonPobiega:
![duplicate message si](https://user-images.githubusercontent.com/163318/34200136-48824786-e570-11e7-9ba0-fd4fa63abed3.png)

@mikeminutillo @WilliamBZA here's the SI view. I generated it by sending a duplicate message to the outbox queue using this behavior:

```
class DuplicateBehavior : Behavior<IDispatchContext>
{
    public override async Task Invoke(IDispatchContext context, Func<Task> next)
    {
        await next();
        await next();
    }
}
```

SzymonPobiega:
Maybe we can do the deduplication by message ID when we do scatter-gather queries?

mikeminutillo:
Given that we have updated Raven and NSB, should we consider putting a deterministic Id in place again? A hash of Message Id, Endpoint, and Time Processed would do.

SzymonPobiega:
@mikeminutillo sounds like a simple enough solution to qualify for MR

SzymonPobiega:
Fixed in #2154

PR #2154, opened 2020-09-14
Audit message deduplication

tmasternak:
Connects to https://github.com/Particular/ServiceControl/issues/1056

### DIFF
--- a/src/ServiceControl.Audit/Auditing/AuditPersister.cs
+++ b/src/ServiceControl.Audit/Auditing/AuditPersister.cs
@@ -157,7 +157,7 @@
 
                 var auditMessage = new ProcessedMessage(context.Headers, new Dictionary<string, object>(metadata))
                 {
-                    Id = $"ProcessedMessages/{context.Headers.UniqueId()}"
+                    Id = $"ProcessedMessages/{context.Headers.ProcessingId()}"
                 };
 
                 foreach (var commandToEmit in commandsToEmit)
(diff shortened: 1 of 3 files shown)


src/ServiceControl.Audit/Auditing/AuditPersister.cs
```diff
@@ -157,8 +157,7 @@ await bodyStorageEnricher.StoreAuditMessageBody(context.Body, context.Headers, m
 
                 var auditMessage = new ProcessedMessage(context.Headers, new Dictionary<string, object>(metadata))
                 {
-                    // We do this so Raven does not spend time assigning a hilo key
-                    Id = $"ProcessedMessages/{Guid.NewGuid()}"
+                    Id = $"ProcessedMessages/{context.Headers.UniqueId()}"
```
SzymonPobiega, line 160: The `UniqueId()` method can fail if headers are missing. I think we should use here similar method but fall back to `Guid.NewGuid()` if the header-based deterministic ID generation fails.

SzymonPobiega (changes requested):

src/ServiceControl.Audit/Auditing/AuditPersister.cs
tmasternak: Can you have a look now?

SzymonPobiega (approved):

mikeminutillo (approved):

SzymonPobiega (approved):

PR #2155, opened 2020-09-16
Make license management consistent between ServiceControl runtime and SCMU

mikeminutillo:
Reduces the number of places that ServiceControl instances and ServiceControl Management look for and write license files to. This reduces the chances that there is a mismatch between where ServiceControl Management and any particular ServiceControl instance find their valid license files.

Technically this is a breaking change. It is possible that this could cause ServiceControl to lose it's license during an upgrade. If that happens, re-importing the license using ServiceControl Management will fix it. 

Risk Mitigations
- ServiceControl Management has been writing the license file to the registry and the common app data folder for some time (since November 2017 and 1.43.0). It is likely that the existing license is in the common app data folder anyway.
- ServiceControl Management will not allow you to upgrade any instances if it cannot find it's license file, so a user will have to re-import their license before upgrading any instances to this version.

Checklist
- [x] ServiceControl Management only looks for licenses on disk
- [x] ServiceControl Management only writes licenses to disk
- [x] ServiceControl only looks for licenses on disk
- [x] ServiceControl Audit only looks for licenses on disk
- [x] ServiceControl Monitoring only looks for licenses on disk

License location is machine-wide Common App Data folder.

### DIFF
--- a/src/ServiceControl.Audit/ServiceControl.Audit.csproj
+++ b/src/ServiceControl.Audit/ServiceControl.Audit.csproj
@@ -3,7 +3,6 @@
   <PropertyGroup>
     <TargetFramework>net462</TargetFramework>
     <OutputType>Exe</OutputType>
-    <DefineConstants>$(DefineConstants);REGISTRYLICENSESOURCE</DefineConstants>
     <ApplicationIcon>Operations.ico</ApplicationIcon>
   </PropertyGroup>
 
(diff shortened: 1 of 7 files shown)


tmasternak (approved):

mikeminutillo:
@TraGicCode I checked [this](https://github.com/TraGicCode/tragiccode-nservicebusservicecontrol/blob/master/manifests/config.pp) and it looks like you're putting the license file into the common app data folder anyway so your automation should be unaffected.

You might want to follow this issue becuase once you get to this version you can stop writing the license into the registry.

SzymonPobiega (approved):

SzymonPobiega:
@mikeminutillo I don't consider it a breaking change because it does not affect *runtime* behavior. It is a config change. I think it is a good one.

PR #2156, opened 2020-09-16
Support processing saga audit messages in the audit instance

SzymonPobiega:
Adds saga audit processing support to the SC audit instance.

The saga audit data is a kind of audit data and logically belongs in the audit instance. What is more, having saga audit instance in the audit database allows scaling out the combined audit processing which enables saga audit usage also in production environments.

The saga audit processing is still possible in the main instance in order to maintain backwards compatibility and allow users to upgrade ServiceControl without the need to update the configuration of all the endpoints.

Corresponding [docs PR](https://github.com/Particular/docs.particular.net/pull/5014)


SzymonPobiega:
I pushed acceptance tests to cover this. Let me know what do you think folks? There is a nasty code duplication for the audit processing and I was thinking about extracting "ServiceControl.SagaAudit" component that would be attached to both main and audit instances. What do you think?

tmasternak:
> I pushed acceptance tests to cover this. Let me know what do you think folks? There is a nasty code duplication for the audit processing and I was thinking about extracting "ServiceControl.SagaAudit" component that would be attached to both main and audit instances. What do you think?

I like the idea of extracting the common parts to a dedicated component. I think that the component should do bulk inserts and in the main instance logic should be changed to pass saga audit messages from handlers to that component.
 
The future step would be to enable configuring the component to work over dedicated queues.

danielmarbach:
What is the impact of putting that index into the audit database? Historically we know that because we process large volumes of messages the indexing pressure is already quite high. Does it really make sense to put SagaAudit into the same bucket or are we falling into the trap of mentally bringing together processed messages audit with saga audit because both have the name audit in there?

SzymonPobiega:
@danielmarbach Yes, in instances with high audit volume adding saga audit might cause problems, but we currently state that saga audit should not be used in production because it will bring down the main SC instance.

With this change user will be able to enable saga audit in production provided they provision enough processing power e.g. one Audit instance for regular audits and another one solely for saga audits. The main benefit is that the saga audit processing will be scalable the same way as audit is now.

src/ServiceControl.Audit.UnitTests/ApprovalFiles/APIApprovals.PublicClr.approved.txt
```diff
@@ -271,6 +271,81 @@ namespace ServiceControl.Audit.Recoverability
         public string PhysicalAddress { get; set; }
     }
 }
+namespace ServiceControl.Audit.SagaAudit
```
mikeminutillo, line 274: Do these need to be public?

mikeminutillo (commented):
This looks good. I would love to see the saga audit stuff pulled into a reusable component but I don't know how easy that will be.

We need to include retention for saga audits here (which will be more duplicated code).

In theory, the index should have negligible impact until there are saga audit documents in the database right? Do we have any way to measure what impact it will have once we do start adding audit documents or is our stance on using saga audit in production good enough? I do like the idea of creating an audit instance specific to sagas.

danielmarbach:
I would like to make it clear that my comment here is not to shut down the initiative here because I fully support the idea of making SagaAudit pluggable. The thing I'm not aligned with is the solution proposal. 

For me analyzing processed messages through audit and saga audit are two orthogonal concerns and I'm worried that by coupling those things together we are in the same area again like when we had recoverability and audits together. Already today we have customers that have several hundreds of gigabyte of data worth of processed messages. While they might have less saga instances the saga data tends to be in some cases quite big and depending on the rate of changes on the saga quite volumnious as well. Bringing that into the audit instance has several consequences:

- Increased storage needs
- Depending on the rate of changes increased indexing needs due to saga changes and thus compeeting on the same indexing resource pool on the same instance
- SagaAudit settings might require independent settings from processed message and then we need to expose different things like audit retention and saga retention. Those different business expiry will again compete on the resources
- If we ever want to scale out saga audits individually we can't really because the scaling needs are now coupled to audits which might not be the same
- If saga data changes reaches a certain threshold that causes problem we can't just nuke it and leave processed messages untouched
- It seems to be counter to the efforts of componentization 

Or are Saga Audit and Processed Messages really the same thing and I'm just not seeing yet?

SzymonPobiega:
> If we ever want to scale out saga audits individually we can't really because the scaling needs are now coupled to audits which might not be the same

We can as you can use dedicated instances for one or the other. The code in the endpoint decides to which instance the audit is sent and it would perfrctly fine to to something like this

```
config.AuditProcessedMessagesTo("audit");
config.AuditSagaStateChanges("audit.saga");
```

> If saga data changes reaches a certain threshold that causes problem we can't just nuke it and leave processed messages untouched

If you put them in the same instance then yes, but there is nothing that forces you.

I would love to have a separate saga audit instance but given the current SC hosting constraints, adding another instance type would be a lot of work. Also, you want to get a decent experience out-of-the-box so we would have to create the saga audit instance by default, just as we do with the audit instance. The fact that "audit instance" can do both saga audits and message audits is not ideal but I think it allows us more freedom in future in creating specialized instance types when the SC hosting model becomes easier to work with.

mikeminutillo:
> Or are Saga Audit and Processed Messages really the same thing and I'm just not seeing yet?

This is a good way of thinking about it. Saga Audit could have been implemented as a header on audited messages. It is done as a separate message, which is probably best due to it's potential size. The Saga Audit data is best utlized _with_ the Audit messages. It is likely to be accessed by the same tools. It should be retained for the same period of time.

tmasternak:
Adding my 2cc. If we want to be super safe we could make audits and saga-audits mutually exclusive. In other words, an instance is either one or the other. So if users start to use the scale-out mode we make sure they scale both separately. In that case, the audit instance input queue is owned by the host and there is one ingestor running.

seanfarmar (approved):
LGTM

SzymonPobiega:
I pushed the change to make saga audits a separate component referenced in both instances. This also adds support for retention and drops an unused index (SagaList)

tmasternak (approved):
I've pushed some minor tweaks. Otherwise this looks 👍

SzymonPobiega:
Thanks @tmasternak . I'll merge when green.

danielmarbach:
If this is merged https://github.com/Particular/ServiceControl/pull/2173 requires a rebase or vise versa

PR #2159, opened 2020-09-17
Added an ‘Update Available’ link when a newer version is available 

seanfarmar:


seanfarmar:
I have lost my last commit with a completed 1 iteration while squashing the branch...
Let me try and restore it...

src/ServiceControl.Config/UI/Shell/VersionCheckerHelper.cs
```diff
@@ -0,0 +1,90 @@
...
+            var handler = new HttpClientHandler
+            {
+                AutomaticDecompression = DecompressionMethods.GZip | DecompressionMethods.Deflate
+            };
+
+            using (var httpClient = new HttpClient(handler))
+            {
+                httpClient.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
+                
+                // TODO: move this to some sort of configuration file/storage?
```
mikeminutillo, line 43: Maybe just shift it to a constant for now. It hasn't changed in five-ish years.

src/ServiceControl.Config/UI/Shell/VersionCheckerHelper.cs
```diff
@@ -0,0 +1,90 @@
...
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Net;
+    using System.Net.Http;
+    using System.Net.Http.Headers;
+    using System.Threading.Tasks;
+
+    public static class VersionCheckerHelper
+    {
+        public static async Task<Release> GetDownloadUrlForNextVersionToUpdate(string currentVersion)
```
mikeminutillo, line 15: I would consider changing this name to something like `GetRecommendedRelease()`.

src/ServiceControl.Config/UI/Shell/VersionCheckerHelper.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            Version current = new Version(currentVersion);
+            List<Release> releases = await GetVersionInformation().ConfigureAwait(false);
+
+            Release topversion = releases.Select(t => (t.Version, t)).Max().t;
```
mikeminutillo, line 20: I did not know that VaueTuple's sorted that way. That's neat :)

The other way to do it would have been `.SortByDescending(t => t.Version).First()`. That may be less efficient under the covers if it has to sort the whole list.

src/ServiceControl.Config/UI/Shell/VersionCheckerHelper.cs
```diff
@@ -0,0 +1,90 @@
...
+            
+            UpgradeInfo upgradeInfo = UpgradeControl.GetUpgradeInfoForTargetVersion(topversion.Version, current);
+            
+            Release recomendedVersion = releases.FirstOrDefault(t => t.Version == upgradeInfo.RecommendedUpgradeVersion);
+
+            if (recomendedVersion != null && recomendedVersion.Version > current)
+                return recomendedVersion;
+            // we have no release available
```
mikeminutillo, line 28: This may not be true. If there's a recommended upgrade version then that version must exist, but it may not exist in the document found at the link provided which is limited to the last X releases. i.e. If you were running version 1.5 today, it would give you a recommended upgrade of 1.48, but that's not in the json file so we'd end up here.

src/ServiceControl.Config/UI/Shell/VersionCheckerHelper.cs
```diff
@@ -0,0 +1,90 @@
...
+    public static class VersionCheckerHelper
+    {
+        public static async Task<Release> GetDownloadUrlForNextVersionToUpdate(string currentVersion)
+        {
+            Version current = new Version(currentVersion);
+            List<Release> releases = await GetVersionInformation().ConfigureAwait(false);
+
+            Release topversion = releases.Select(t => (t.Version, t)).Max().t;
+            
+            UpgradeInfo upgradeInfo = UpgradeControl.GetUpgradeInfoForTargetVersion(topversion.Version, current);
```
mikeminutillo, line 22: I'm not certain that this gives us want we want. Here's why:

The current major is v4. Whenever we release v5, it will show up on the web and this code will find it. A call to `UpgradeControl.GetUpgradeInfoForTargetVersion(5.x, 4.x)` will return a recommended version of 5.x. This is because we don't have minimum/recommended upgrade versions for version 5. We won't have those until we release version 5, by which time, it will be too late because you'll have to upgrade to version 5 to have it.

The `UpgradeControl` class is designed to prevent updating to a major if the current instance version is not compatible. It does not help if you're looking at the next download and trying to figure out if it will be compatible with the current version.

To make this work properly, we'd need to put the restrictions in the release info json on the web. Then we could list upgrade info there and have the system make a recommendation for a specific download using that.

src/ServiceControl.Config/UI/Shell/VersionCheckerHelper.cs
tmasternak: Is it good enough if we just show "New version available" that points to the download? If someone sees that they can decide to download the newest version that will have the update restrictions in place and will guide them when doing the instance level upgrades.

src/ServiceControl.Config/UI/Shell/VersionCheckerHelper.cs
seanfarmar: So what are you suggesting we do?

src/ServiceControl.Config/UI/Shell/VersionCheckerHelper.cs
seanfarmar: Not sure what are we suggesting to do here...

src/ServiceControl.Config/UI/Shell/VersionCheckerHelper.cs
mikeminutillo: @seanfarmar for this enhancement release, I think we just need to check if there is a version that is higher than the currently installed one. I would not use the `UpgradeControl` class. A comparison of `Version`s is good enough for now.

src/ServiceControl.Config/UI/Shell/VersionCheckerHelper.cs
mikeminutillo: I would drop the reference to `UpgradeControl`. If there's an update in the list on the web, then we should tell the user about it.

seanfarmar: I've changed this to select the newest version if it's higher then that current one

src/ServiceControl.Config/UI/Shell/ShellViewModel.cs
```diff
@@ -15,7 +15,7 @@
     using NoInstances;
     using ServiceControlInstaller.Engine.Instances;
 
-    class ShellViewModel : RxConductor<RxScreen>.Collection.OneActive, IHandle<RefreshInstances>
+    class ShellViewModel : RxConductor<RxScreen>.Collection.OneActive, IHandle<RefreshInstances>// , IActivate
```
tmasternak, line 18: Remove ?
> // , IActivate

src/ServiceControl.Config/UI/Shell/ShellViewModel.cs
```diff
@@ -131,6 +143,29 @@ void LoadAppVersion()
...
+            // get the json version file from https://s3.us-east-1.amazonaws.com/platformupdate.particular.net/servicecontrol.txt
+
+            var shortAppVersion = AppVersion.Substring(0, 6);
+            
+            Release availableUpgradeRelease = await VersionCheckerHelper.GetLatestRelease(shortAppVersion).ConfigureAwait(false);
+            
+            if (availableUpgradeRelease.Version.ToString() == shortAppVersion)
+            {
+                UpdateAvailable = false;
+                AvailableUpgradeReleaseLink = string.Empty;
```
tmasternak, line 158: We probably don't need this

src/ServiceControlInstaller.Engine.UnitTests/Configuration/ServiceControl/UpgradeControlDetailsTests.cs
```diff
@@ -41,5 +41,11 @@ public void ValidateVersion2UpgradeData()
             Assert.AreEqual(version2, upgradeInfo.TargetMinimumVersion, "TargetMinimumVersion mismatch");
             Assert.AreEqual(recommendedVersion, upgradeInfo.RecommendedUpgradeVersion, "RecommendedUpgradeVersion mismatch");
         }
+
+        [Test]
+        public void VersionCheckerHelperTests()
```
tmasternak, line 46: Should this be removed?

src/ServiceControl.Config/UI/Shell/VersionCheckerHelper.cs
```diff
@@ -0,0 +1,85 @@
...
+                // TODO: move this to some sort of configuration file/storage?
+                var json = await httpClient.GetStringAsync("https://s3.us-east-1.amazonaws.com/platformupdate.particular.net/servicecontrol.txt").ConfigureAwait(false);
+
+                return JsonConvert.DeserializeObject<List<Release>>(json);
+            }
+        }
+    }
+
+    // TODO: move it to a better place?
+    public class Release
```
tmasternak, line 47: Can this bee an internal class of `VersionCheckerHelper`?

tmasternak (changes requested):

src/ServiceControl.Config/UI/Shell/ShellViewModel.cs
```diff
@@ -131,6 +143,28 @@ void LoadAppVersion()
             }
         }
 
+        private async Task CheckForUpdates()
+        {
+            // Get the lates upgradble version based on the current version
+            // get the json version file from https://s3.us-east-1.amazonaws.com/platformupdate.particular.net/servicecontrol.txt
+
+            var shortAppVersion = AppVersion.Substring(0, 6);
```
mikeminutillo, line 151: This only works for the current version `4.12.1`. It won't work for future versions like `5.0.0`.

```suggestion
            var shortAppVersion = AppVersion.Split('-').First();
```

mikeminutillo (changes requested):

SzymonPobiega:
@seanfarmar ping?

seanfarmar:
All comments (hopefully) addressed @mikeminutillo @SzymonPobiega

mikeminutillo (approved):

tmasternak (approved):

PR #2165, opened 2020-09-21
Auto-migrate ASB Forwarding Topology transport seam to the new ASB seam

tmasternak:
### Description

When an instance gets upgraded to the new ServiceControl version and it's using Deprecated Forwarding Topology it will be auto-migrated to the new ASB seam. Secondly, deprecated transports seams are not available in the UI anymore and new instances using deprecated transports can be created only via PowerShell module.

### DIFF
--- a/src/ServiceControlInstaller.Engine/Instances/MonitoringInstance.cs
+++ b/src/ServiceControlInstaller.Engine/Instances/MonitoringInstance.cs
@@ -290,5 +290,17 @@
                 ReportCard.Errors.Add(ex.Message);
             }
         }
+
+        public void UpgradeTransportSeam()
+        {
+            //Legacy ASB Forwarding Topology should be migrated to the ASBS transport seam
+            if (TransportPackage.Name == TransportNames.AzureServiceBusForwardingTopologyDeprecated)
+            {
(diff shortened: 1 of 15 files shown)


src/ServiceControlInstaller.Engine/Instances/MonitoringInstance.cs
```diff
@@ -290,5 +292,17 @@ public void SetupInstance()
                 ReportCard.Errors.Add(ex.Message);
             }
         }
+
+        public void UpgradeTransportSeam()
+        {
+            //Legacy ASB Forwarding Topology should be migrated to the ASBS transport seam
+            if (TransportPackage.Name == TransportNames.AzureServiceBusForwardingTopologyDeprecated)
```
mikeminutillo, line 299: Would it make sense to put an `AutoMigrateTo` property on `TransportInfo`? That way this code could check for a value and migrate without knowing specific types, and if we want to do this again, it's just metadata.

tmasternak: How about now?

seanfarmar (approved):
LGTM

mikeminutillo (approved):

PR #2184, opened 2020-09-30
[SPIKE] Remove metadata

SzymonPobiega:
This PR attempts to remove the metadata. My goal was to get rid of the `OfType` client side projection but it seems to not be possible for queries that operate on fields of the index that are not part of the `ProcessedMessage` document (unless I am missing something).

For this reason only `GetAllMessages` and `GetByConversationId` queries have been modified to bypass the `OfType` projection and use the `ProcessedMessage` type throughout the whole LINQ expression.

Metadata dictionary has been replaced by explicit properties and a `SearchTerms` bag that only contains pieces of data we actually want to analyze (no `boolean` flags etc.)

The next step would be to actually move all the enrichers code to a single class, possibly with multiple methods. It makes little sense to pretend they are independent "plugins" that can be added or removed.


SzymonPobiega:
I verified that the queries sent to the server are almost exactly the same, regardless if we use `ProcessedMessage` or `MessagesView.Result` as the `T` in the Linq train wreck. The only differences seem to be around order of the fields to retrieve.

### ProcessedMessage

```
from index 'MessagesViewIndex' as x where x.IsSystemMessage = $p0 order by MessageId desc select { ProcessingTime : x.ProcessingTime, SendingEndpoint : x.SendingEndpoint, ReceivingEndpoint : x.ReceivingEndpoint, MessageType : x.MessageType, MessageId : x.MessageId, ConversationId : x.ConversationId, DeliveryTime : x.DeliveryTime, TimeSent : x.TimeSent, CriticalTime : x.CriticalTime, IsSystemMessage : x.IsSystemMessage, Status : x.Status, ProcessedAt : x.ProcessedAt, Headers : Object.keys(x.Headers).map(function(a){return{Key: a,Value:x.Headers[a]};}), MessageIntent : x.MessageIntent, InvokedSagas : x.InvokedSagas, OriginatesFromSaga : x.OriginatesFromSaga, BodyUrl : x.BodyUrl, BodySize : x.BodySize, Id : id(x) } limit $p1, $p2
{"p0":false,"p1":0,"p2":50}
```

#### MessagesView.Result

```
from index 'MessagesViewIndex' as message where message.IsSystemMessage = $p0 order by MessageId desc select { Id : message.UniqueMessageId, MessageId : message.MessageId, MessageType : message.MessageType, SendingEndpoint : message.SendingEndpoint, ReceivingEndpoint : message.ReceivingEndpoint, TimeSent : message.TimeSent, ProcessedAt : message.ProcessedAt, CriticalTime : message.CriticalTime, ProcessingTime : message.ProcessingTime, DeliveryTime : message.DeliveryTime, IsSystemMessage : message.IsSystemMessage, ConversationId : message.ConversationId, Headers : Object.keys(message.Headers).map(function(a){return{Key: a,Value:message.Headers[a]};}), Status : message.Status, MessageIntent : message.MessageIntent, BodyUrl : message.BodyUrl, BodySize : message.BodySize, InvokedSagas : message.InvokedSagas, OriginatesFromSaga : message.OriginatesFromSaga } limit $p1, $p2
{"p0":false,"p1":0,"p2":50}
```

danielmarbach:
@SzymonPobiega can this be closed?

PR #2200, opened 2020-10-12
Add docker files

WilliamBZA:
Adds dockerfiles and the required code changes to ServiceControl


src/ServiceControl/Bootstrapper.cs
```diff
@@ -70,6 +70,11 @@ private void Initialize()
         {
             RecordStartup(loggingSettings, configuration);
 
+            if (!string.IsNullOrWhiteSpace(settings.LicenseFileText))
+            {
+                configuration.License(settings.LicenseFileText);
```
WilliamBZA, line 75: Or is the better approach to add a new license source? Thoughts @yvesgoeleven @HEskandari ?

HEskandari: I think ultimately loading license from env variable is something we need to do. It can be used in other docker container scenarios, like running NServiceBus endpoints in a container. 

However, I'm fine with the suggestion here as a first step. We can raise an cruft issue (or probably a small task, as it is low effort) to get it done in the licensing package.

yvesgoeleven (approved):

HEskandari (approved):

PR #2229, opened 2020-10-19
Check license file for valid setup commands

WilliamBZA:


WilliamBZA:
Tests pass @yvesgoeleven @HEskandari . Ready for review

yvesgoeleven (approved):

src/ServiceControl.LicenseManagement/LicenseManager.cs
```diff
@@ -1,11 +1,11 @@
-﻿namespace ServiceControlInstaller.Engine.LicenseMgmt
+﻿namespace ServiceControl.LicenseManagement
 {
     using System;
     using Particular.Licensing;
 
     public class LicenseManager
     {
-        public static DetectedLicense FindLicense()
+        public static DetectedLicense FindLicense(bool includeTrial = true)
```
HEskandari, line 8: I think I added this, but I don't think this is necessary. It should be:

```suggestion
        public static DetectedLicense FindLicense()
```

src/ServiceControl.LicenseManagement/ServiceControl.LicenseManagement.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net40</TargetFramework>
```
HEskandari, line 4: is this the right target framework?

HEskandari (commented):
Reviewed. Two small comments.

src/ServiceControl.LicenseManagement/ServiceControl.LicenseManagement.csproj
WilliamBZA: Yes, it's used by the installer project which is 40 so this has to match

Issue #2242, opened 2020-10-27
It is not possible to start ServiceControl if the KnownEndpointsIndex is disabled


WilliamBZA:
During the initialization of the RavenBootstrapper, ServiceControl immediately parses over the `KnownEndpointIndex` - even in maintenance mode.

If this index is disabled as part of running out of disk space, then it is not possible to re-enable this index as part of maintenance mode as ServiceControl fails to start in maintenance mode.  The only workaround at that stage is to delete the indexes folder and rebuild all indexes.

Ideally ServiceControl should not do any document processing if it is in maintenance mode. This can be achieved by only invoking `PurgeKnownEndpointsWithTemporaryIdsThatAreDuplicate` if not in maintenance mode.

Issue #2065, opened 2020-06-30
The `IndexDisabledException` can prevent SC from starting in maintenance mode


ramonsmits:
A customer got an `IndexDisabledException` at the following location which prevented SC from starting.
```
at ServiceControl.Infrastructure.RavenDB.RavenBootstrapper.PurgeKnownEndpointsWithTemporaryIdsThatAreDuplicate(IDocumentStore documentStore) in C:\BuildAgent\work\7189a56f9f44affc\src\ServiceControl\Infrastructure\RavenDB\RavenBootstrapper.cs:line 103
at ServiceControl.Infrastructure.RavenDB.RavenBootstrapper.StartRaven(EmbeddableDocumentStore documentStore, Settings settings, MarkerFileService markerFileService, Boolean maintenanceMode) in C:\BuildAgent\work\7189a56f9f44affc\src\ServiceControl\Infrastructure\RavenDB\RavenBootstrapper.cs:line 97
at Particular.ServiceControl.MaintenanceBootstrapper.Run(HostArguments args) in C:\BuildAgent\work\7189a56f9f44affc\src\ServiceControl\MaintenanceBootstrapper.cs:line 20
```

The `PurgeKnownEndpointsWithTemporaryIdsThatAreDuplicate` method could fail but that should not prevent maintenance mode from launching.

## Workaround

Removing the `Indexes` folder will ensure that maintenance mode can be launched again. [It is advised to run to scan for DB consistency by running `esentutl /r RVN /l logs /s system` in the DB folder.](https://docs.particular.net/servicecontrol/db-compaction#using-esentutl-preferred-approach-step-3-administrator-command-prompt)

WilliamBZA:
Fixed in #2243, released in 4.13.2.

PR #2243, opened 2020-10-27
Only purge endpoints when not in maintenance mode

WilliamBZA:
Fixes https://github.com/Particular/ServiceControl/issues/2242

This results in ServiceControl only accessing the `KnownEndpointIndex` index when not in maintenance mode. If the `KnownEndpointIndex` is disabled due to previously running out of disk space, it would be impossible to re-enable that index in maintenance mode as maintenance mode would fail to start.


DavidBoike (approved):

janovesk (approved):

ramonsmits:
I think this resolves this issue I reported a while back https://github.com/Particular/ServiceControl/issues/2065

WilliamBZA:
I agree, it should fix that

PR #2257, opened 2020-11-03
Invert license check for audit

WilliamBZA:
The license check in the audit init containers was inverted.

### DIFF
--- a/src/ServiceControl.Audit/Infrastructure/SetupBootstrapper.cs
+++ b/src/ServiceControl.Audit/Infrastructure/SetupBootstrapper.cs
@@ -101,7 +101,7 @@ namespace ServiceControl.Audit.Infrastructure
                     return false;
                 }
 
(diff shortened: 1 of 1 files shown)


src/ServiceControl.Audit/Infrastructure/SetupBootstrapper.cs
```diff
@@ -101,7 +101,7 @@ bool ValidateLicense(Settings.Settings settings)
                     return false;
                 }
 
-                if (LicenseManager.TryImportLicenseFromText(settings.LicenseFileText, out var importErrorMessage))
+                if (!LicenseManager.TryImportLicenseFromText(settings.LicenseFileText, out var importErrorMessage))
```
danielmarbach, line 104: Good catch, yeah they should be aligned

https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/SetupBootstrapper.cs#L79

danielmarbach (approved):

Issue #2272, opened 2020-11-11
ServiceControl Monitoring does not allow users to specify the MaximumConcurrencyLevel


WilliamBZA:
The `MaximumConcurrencyLevel` setting is ignored for ServiceControl monitoring and defaults to the CPU count * 2 and offers no way for users to change this.

This results in extremely poor performance in situations where performance should not be a problem given that there is no I/O for ServiceControl Monitoring.

PR #2273, opened 2020-11-11
Monitoring concurrency

danielmarbach:
Fixes https://github.com/Particular/ServiceControl/issues/2272


WilliamBZA (approved):

timbussmann (approved):
LGTM (although 32 might seem a bit on the upper end given that we just allow to configure it manually now anyway?)

danielmarbach:
@timbussmann we used the same default we recently introduced in audit which is also on the high load side of things similar to metrics

PR #2266, opened 2020-11-09
Bumps the code signing certificate

janovesk:

PR #2274, opened 2020-11-11
Set the message processing concurrency in the monitoring instance

danielmarbach:
#2273 for release-4.13 branch

- Set the message processing concurrency in the monitoring instance
- Make the MaximumConcurrencyLevel level configurable

### DIFF
--- a/src/ServiceControl.Monitoring/Settings.cs
+++ b/src/ServiceControl.Monitoring/Settings.cs
@@ -46,7 +46,7 @@ namespace ServiceControl.Monitoring
                 HttpPort = reader.Read<string>("Monitoring/HttpPort"),
                 EndpointName = reader.Read<string>("Monitoring/EndpointName"),
                 EndpointUptimeGracePeriod = TimeSpan.Parse(reader.Read("Monitoring/EndpointUptimeGracePeriod", "00:00:40")),
-                MaximumConcurrencyLevel = 10
+                MaximumConcurrencyLevel = reader.Read<int>("Monitoring/MaximumConcurrencyLevel")
             };
             return settings;
         }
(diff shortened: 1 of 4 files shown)


src/ServiceControl.Monitoring/Settings.cs
```diff
@@ -8,16 +8,34 @@ namespace ServiceControl.Monitoring
...
+    using ServiceControl.Monitoring.Infrastructure.Settings;
 
     //TODO: align names with SC and SC.Audit
     public class Settings
     {
+        public Settings()
+        {
+            TryLoadLicenseFromConfig();
+
+            TransportType = SettingsReader<string>.Read("TransportType");
```
timbussmann, line 20: the removed code used `TransportType = reader.Read<string>("Monitoring/TransportType"),` with the monitoring prefix (some other options seem to do that too), is that an intentional change?

src/ServiceControl.Monitoring/Settings.cs
```diff
@@ -59,6 +60,11 @@ internal static string DefaultLogLocation()
             return Path.GetDirectoryName(assemblyLocation);
         }
 
+        void TryLoadLicenseFromConfig()
```
timbussmann, line 63: not sure how this connects to the other changes.

src/ServiceControl.Monitoring/Settings.cs
danielmarbach: Yeah these were changes that came in from 

https://github.com/Particular/ServiceControl/commit/cac7201349f4af9ffffa8f81aa2d19403fb2c007#diff-b616500841627c3849f0e83080ac2d844bbfdf95abd9f255bb7f6129c674fb12

Force pushed

WilliamBZA (approved):

danielmarbach:
Was missing https://github.com/Particular/ServiceControl/pull/2266 to actually get compiling

timbussmann (approved):

PR #2280, opened 2020-11-12
Use digicert timestamp url

WilliamBZA:


danielmarbach:
Not needed anymore. We are switching back to the old cert and see if we can get an EV Code Signing cert for the Apps to avoid smart screen blocking

PR #2281, opened 2020-11-12
Revert "Bumps the code signing certificate"

danielmarbach:
This reverts commit 0e6ef900dcd4b8274a410133e5167ae2dde769b1.


danielmarbach:
This doesn't seem to work and we are not likely to get an EV certificate. So we should just use the new cert

PR #2286, opened 2020-11-13
Adds support for .NET Core linux generated stack traces

janovesk:
This updates the stack trace parser with the latest version from https://github.com/atifaziz/StackTraceParser/blob/master/StackTraceParser.cs to correctly support source file code paths starting with forwards slashes. These kinds of paths are generated by .NET Core on Linux.

### DIFF
--- a/src/ServiceControl/App_Packages/StackTraceParser/StackTraceParser.cs
+++ b/src/ServiceControl/App_Packages/StackTraceParser/StackTraceParser.cs
@@ -43,7 +43,9 @@ namespace ServiceControl
                 ( " + Space + @"+
                     ( # Microsoft .NET stack traces
                     \w+ " + Space + @"+
-                    (?<file> [a-z] \: .+? )
+                    (?<file> ( [a-z] \: # Windows rooted path starting with a drive letter
+                             | / )      # *nix rooted path starting with a forward-slash
+                             .+? )
                     \: \w+ " + Space + @"+
                     (?<line> [0-9]+ ) \p{P}?
                     | # Mono stack traces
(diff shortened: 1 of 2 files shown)


src/ServiceControl/App_Packages/StackTraceParser/StackTraceParser.cs
```diff
@@ -76,6 +78,7 @@ partial class StackTraceParser
             return Parse(text, (idx, len, txt) => txt,
                                (t, m) => new { Type = t, Method = m },
                                (pt, pn) => new KeyValuePair<string, string>(pt, pn),
+                               // ReSharper disable once PossibleMultipleEnumeration
```
danielmarbach, line 81: This should probably go because officially we no longer use Resharper. Otherwise it will trigger Adam analyzer tool

src/ServiceControl/App_Packages/StackTraceParser/StackTraceParser.cs
```diff
@@ -43,7 +43,9 @@ partial class StackTraceParser
                 ( " + Space + @"+
                     ( # Microsoft .NET stack traces
                     \w+ " + Space + @"+
-                    (?<file> [a-z] \: .+? )
+                    (?<file> ( [a-z] \: # Windows rooted path starting with a drive letter
```
danielmarbach, line 46: How about making this also a PR against the original source? This is a source package and should not be modified directly unless the original project is abandoned. Then we would need to detach the source package and have the sources here xopied

danielmarbach (changes requested):

janovesk:
Closing this. Replaced by https://github.com/Particular/ServiceControl/pull/2292

src/ServiceControl/App_Packages/StackTraceParser/StackTraceParser.cs
WilliamBZA: Feedback from @janovesk: this is in an app_packages source file, so we can't change it.

src/ServiceControl/App_Packages/StackTraceParser/StackTraceParser.cs
janovesk: Fix is already in the downstream, but downstream is using the old csproj format, so we can't use in in SC now. 

Just adding this for anybody reading this at a later date.

PR #2288, opened 2020-11-17
Fix to read the correct log level settings

WilliamBZA:


danielmarbach:
How about this? https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/Infrastructure/Settings/LoggingSettings.cs#L28-L32

src/ServiceControl.Audit/Infrastructure/Settings/LoggingSettings.cs
```diff
@@ -25,8 +25,8 @@ public LoggingSettings(string serviceName, LogLevel defaultLevel = null, LogLeve
 
         LogLevel InitializeLevel(string key, LogLevel defaultLevel)
         {
-            var levelText = SettingsReader<string>.Read("ServiceControl.Audit", key);
```
danielmarbach, line 28: Ah so the `key` was used as a default value and `ServiceContro.Audit` as the name 🥇

danielmarbach (approved):

src/ServiceControl.Audit/Infrastructure/Settings/LoggingSettings.cs
WilliamBZA: Yes, this was a real champion refactoring I did that introduced this bug. In all 3 SC instances too.

mikeminutillo (approved):

WilliamBZA:
Closing and re-creating to try get TC to report build status

PR #2290, opened 2020-11-17
Make sure to always DisposeAsync the bulk insert session

danielmarbach:


src/ServiceControl.Audit/Auditing/AuditPersister.cs
```diff
@@ -43,10 +44,11 @@ public async Task<IReadOnlyList<MessageContext>> Persist(List<MessageContext> co
             }
 
             var storedContexts = new List<MessageContext>(contexts.Count);
+            BulkInsertOperation bulkInsert = null;
```
WilliamBZA, line 47: Is it not better to put this in a `using` now?

danielmarbach: Not really. The bulk insert uses very awkward and evil sync over async which I would rather not want on this path

WilliamBZA (approved):

PR #2292, opened 2020-11-18
Adds support for .NET Core linux generated stack traces

janovesk:
This removes the non-working source package reference to stack trace parser and replaces it with a copy of the source code file from https://github.com/atifaziz/StackTraceParser/blob/master/StackTraceParser.cs. 

This is necessary to correctly support source file code paths starting with forwards slashes in Linux generated stack traces. These kinds of paths are generated by .NET Core on Linux.


janovesk:
Replaces https://github.com/Particular/ServiceControl/pull/2286

WilliamBZA (approved):

PR #2301, opened 2020-11-19
Azure Storage Queues v9

SeanFeldman:
Changes included in this PR

- New major version of ASQ (9)
- New Azure Queues SDK (queue length provider)
- New .NET Framework prerequisite (4.7.2)
- Updated AdvancedInstaller (17.6)

~Additional indirect changes _potentially_ affecting Service Control~
- ~ASQ transport is no longer sanitizes endpoint/queue names~


danielmarbach:
Requires advanced installer 17.x to be installed on the agents

src/ServiceControl.Monitoring/Settings.cs
```diff
@@ -120,7 +120,7 @@ static Type LoadTransportCustomizationType(string configuredTransportCustomizati
             {"ServiceControl.Transports.RabbitMQ.DirectRoutingTopologyRabbitMQTransport, ServiceControl.Transports.RabbitMQ", "ServiceControl.Transports.RabbitMQ.RabbitMQDirectRoutingTransportCustomization, ServiceControl.Transports.RabbitMQ"},
 
             //ASQ
-            {"NServiceBus.AzureStorageQueueTransport, NServiceBus.Azure.Transports.WindowsAzureStorageQueues", "ServiceControl.Transports.ASQ.ASQTransportCustomization, ServiceControl.Transports.ASQ"},
```
mikeminutillo, line 123: Shouldn't this be a new entry? In theory we might have legacy installations with either the old or the new values in this diff.

src/ServiceControlInstaller.Engine/Instances/ServiceControlCoreTransports.cs
```diff
@@ -46,7 +46,7 @@ public class ServiceControlCoreTransports
                 AvailableInSCMU = true,
                 Matches = name => name.Equals(TransportNames.AzureStorageQueue, StringComparison.OrdinalIgnoreCase)
                                   || name.Equals("ServiceControl.Transports.ASQ.ASQTransportCustomization, ServiceControl.Transports.ASQ", StringComparison.OrdinalIgnoreCase)
-                                  || name.Equals("NServiceBus.AzureStorageQueueTransport, NServiceBus.Azure.Transports.WindowsAzureStorageQueues", StringComparison.OrdinalIgnoreCase)
+                                  || name.Equals("NServiceBus.AzureStorageQueueTransport, NServiceBus.Transport.AzureStorageQueues", StringComparison.OrdinalIgnoreCase)
```
mikeminutillo, line 49: Same here. This shouldn't change, but a new addition should have been made

mikeminutillo (changes requested):
I assume that the changes to the `.aip` file are correct as I do not know how to interpret that file

src/ServiceControl.Monitoring/Settings.cs
SeanFeldman: The transport is the same transport, what's changed is the package name. Everything else remains the same.

I'm not quite following how this part works in SC. Is this documented somewhere?

mikeminutillo: I don't think it's documented.

This sets up a mapping from the old config value to the real transport type to be used. So it used to be that if the config contained `NServiceBus.AzureStorageQueueTransport, NServiceBus.Azure.Transports.WindowsAzureStorageQueues` then we would use `ServiceControl.Transports.ASQ.ASQTransportCustomization, ServiceControl.Transports.ASQ` instead. 

We want that to still be true, don't we?

SeanFeldman: @mikeminutillo, I'm not sure that's accurate. When a new instance of SC/SC.Audit/SC.Monitoring is installed, it adds the following transport identifier to the configuration file:
```xml
<add key="ServiceControl/TransportType" value="ServiceControl.Transports.ASQ.ASQTransportCustomization, ServiceControl.Transports.ASQ" />
```

Given that the Core transport identifier (`"NServiceBus.AzureStorageQueueTransport, NServiceBus.Transport.AzureStorageQueues"`) has never been used by any old SC* installation, it will never appear in the existing configuration files and therefore does not need to be mapped to the core-provide transport identifier. As a result, the mappings do **not** need to be added. Would it be possible for you to sync with @danielmarbach, who's also involved in this change, to discuss and to an agreement wherever it's not needed or needed?

src/ServiceControlInstaller.Engine/Instances/ServiceControlCoreTransports.cs
SeanFeldman: Same https://github.com/Particular/ServiceControl/pull/2301#discussion_r531116262

src/ServiceControl.Monitoring/Settings.cs
mikeminutillo: You are correct!

There should be no legacy systems using this new transport, so there is no need to add a mapping here at all. This code should be untouched.

src/ServiceControlInstaller.Engine/Instances/ServiceControlCoreTransports.cs
mikeminutillo: Correct. No change should be made here at all.

lailabougria (approved):

SeanFeldman:
@SzymonPobiega, @danielmarbach, the build is green. Please review, approve and merge if there are no blockers. Thanks.

danielmarbach (approved):

SzymonPobiega:
Hello folks

The https://github.com/Particular/PlatformArchitecture/issues/574 will merge this PR to master and then we'll rebase ours and merge to master.

❤️ RavenDB upgrade team

SeanFeldman:
Thank you, @SzymonPobiega. Not skeptical or anything, but before I move on - has this work item been added to https://github.com/Particular/PlatformArchitecture/issues/574's PoA?

PR #2303, opened 2020-11-20
Cherry-pick of Add explicit references to Microsoft.Data packages

WilliamBZA:


mauroservienti (approved):

tmasternak (approved):

danielmarbach:
FYI according to https://github.com/dotnet/SqlClient/issues/211#issuecomment-562635405 only a ref to the SNI package might be needed. Anyway I guess if this works that is also fine because as far as I understand we assume MS packages don't need ranges

https://github.com/Particular/NServiceBus.SqlServer/blob/release-6.2/src/NServiceBus.Transport.SqlServer/NServiceBus.Transport.SqlServer.csproj#L20

PR #2306, opened 2020-11-23
introduce consistent dockerfile naming convention

adamralph:
This came out of a conversation with a customer who wanted to use the ASB (non-legacy) images but couldn't figure out which files to use. When I looked at the files, I couldn't figure it out either, without someone explaining it to me (e.g. the undocumented `asbs` acronym).

This is an attempt to introduce a clear naming convention which removes any ambiguity.

Note that some of the example commands in the top level README.md are currently incorrect. They mix-up `servicecontrolrabbitconventional` and `servicecontrolrabbitdirect`.


README.md
```diff
@@ -28,38 +28,38 @@ How to build and run Docker images
...
 
 ```
-docker build -f .\dockerfile.rabbitmq.conventional.init -t particular/servicecontrolrabbitconventional.init ./../
-docker build -f .\dockerfile.rabbitmq.conventional -t particular/servicecontrolrabbitconventional ./../
-docker build -f .\dockerfile.rabbitmq.conventional.audit.init -t particular/servicecontrolrabbitconventional.audit.init ./../
-docker build -f .\dockerfile.rabbitmq.conventional.audit -t particular/servicecontrolrabbitconventional.audit ./../
-docker build -f .\dockerfile.rabbitmq.conventional.monitoring.init -t particular/servicecontrolrabbitconventional.monitoring.init ./../
-docker build -f .\dockerfile.rabbitmq.conventional.monitoring -t particular/servicecontrolrabbitconventional.monitoring ./../
+docker build -f .\dockerfile.rabbitmq.conventional.main.init        -t particular/servicecontrolrabbitconventional.main.init ./../
```
yvesgoeleven, line 31: @WilliamBZA maybe main should be error?

adamralph: I used these UI elements to guide me, but the naming of these things has always been a bit awkward:

WilliamBZA:
Part of this has been done as part of #2318 

We'll evaluate which other changes make sense too. As it stands now though, the taskforce is not convinced that `.main` is a good addition, but we have no real guidance on it one way or the other.

adamralph:
The changes in #2318 look good. The only thing that still bothers me is that the images for the "main" (for lack of a better name) instance do no appear together, whereas the `audit` and `monitoring` images do:



It's a small thing, but I worry it may be a paper-cut for users.

WilliamBZA:
Yeah, that is troubling. Dockerhub makes this even more confusing as it doesn't even combine the names together at all.

image

adamralph:
> Dockerhub makes this even more confusing as it doesn't even combine the names together at all.

😢 I hope they implement sorting in the future.

I guess there could be some client tools that already do sorting, so it would be nice for images to have a good sort order from the start.

WilliamBZA:
We've taken some of the naming conventions but have not taken others. The filenames here will become less important as the images are now hosted on dockerhub and our build infrastructure expects the names of the files to follow some standards in order to work.

Issue #2169, opened 2020-09-22
Remove deprecated ASB transports


tmasternak:
As part of the enhancement release, deprecated ASB transports were [removed](https://github.com/Particular/ServiceControl/pull/2165) from the UI and are now available only via PowerShell.

This is a follow-up issue to finish the removal in the next major release.

### Plan of Action
- [x] Remove Forwarding and Endpoint-Oriented topologies from the PowerShell
- [ ] Automigrate users on the Endpoint-Oriented topology to ASBS 
   - [ ] Keep the old transports metadata in ServiceControlCoreTransports to support the migration
- [ ] Remove the code base related to the old seam

WilliamBZA:
@mauroservienti I have a concern around removing Endpoint Oriented from powershell.

If we have customers using endpoint oriented and they subscribe to ServiceControl events, they can never upgrade. We have not provided them with a migration plan yet. This PoA seems to be short a couple steps of helping customers migrate to forwarding when subscribing to events.

Very specifically: The powershell changes to remove endpoint oriented cannot be released until the migration work is done.

mauroservienti:
Raised https://github.com/Particular/ServiceControl/pull/2355 to revert changes. @WilliamBZA thanks for pointing out the issue.

DavidBoike:
As part of removing the documentation for the legacy Azure Service Bus transports, we raised https://github.com/Particular/ServiceControl/pull/2923 but was closed when we came to the conclusion that this is not the right time. At present the SCMU UI does not have a capability to prevent an installed instanced from being upgraded, whether that is due to a transport concern (like this one) or do to a version concern, as will be required for the RavenDB upgrade that will change underlying storage from ESENT to Voron.

Because of this missing functionality, the legacy transport cannot be released as a minor, and should probably be released as part of 5.0.0 as suggested by the milestone on this issue, and the "version not upgradeable" feature should be scoped and built intentionally to cover both of these use cases.

DavidBoike:
The Legacy ASB transport was removed in https://github.com/Particular/ServiceControl/pull/3790 as part of ServiceControl 5. Checking for and preventing upgrade of removed transports was accomplished in https://github.com/Particular/ServiceControl/pull/3808 and refactored quite in various PRs after that.

There is also now a documented [migration path off Azure Service Bus Legacy transport](https://docs.particular.net/transports/upgrades/asb-9to9.1), and at the end of that path, you can change the transport to the supported Azure Service Bus transport.

[...]

PR #2307, opened 2020-11-23
Filter out available transports by removing the ones not available in SCMU

mauroservienti:
Related to https://github.com/Particular/ServiceControl/issues/2169

Started from the same assumptions done in https://github.com/Particular/ServiceControl/pull/2165. All the migration thing seems to be handled by the installation engine. I'm not sure what else needs to be done here.


HEskandari (approved):
I think this is good

ramonsmits (approved):

mikeminutillo (approved):
Looks OK to me. I know we used to say that customers that still needed legacy Azure could install using the powershell scripts, but I think we'd be encouraging them to move now anyway.

HEskandari:
@mauroservienti this is reviewed and approved, shall we merge?

Issue #2123, opened 2020-08-21
Installation via Powershell Core / .NET Core broken/unsupported due to dependency on Attribute-based versioning


HEskandari:
### Problem Description

Result of some investigation shows that the new Powershell versions that are .NET Core based do work with most of our cmdlets but there are a few that are broken.

The broken ones are the ones that do the installations (e.g. `New-ServiceControlInstance`) and that is due to how we store release date information as an assembly attribute. The .NET Core runtime seem to have issues reading the values back which stops those cmdlets working properly.

### Suggested Solution

Put the release date information as a text file in the release artifacts (zip files). Read the information from the zip file at installation time.

mauroservienti:
I did an investigation on this today. The installation process checks if the license is valid at installation time, and to do so it needs to extract the release date from the assembly it's trying to install.

The release date is stored in assemblies by the Licensing package. The Licensing package at build time, using an MSBuild task, [writes a `ReleaseDate` attribute](https://github.com/Particular/Operations.Licensing/blob/13608e31ab7ce46e357fd777380b9bc57051ab1b/src/Particular.Licensing/Particular.Licensing.Sources.targets#L16-L18) to the compiled assembly storing the GitVersion `$(GitVersion_CommitDate)`.

Finally, `ServiceControlInstaller.Packaging` project [zips all assemblies](https://github.com/Particular/ServiceControl/blob/a56860be685270d06490342e90007e84480bf00e/src/ServiceControlInstaller.Packaging/ServiceControlInstaller.Packaging.csproj#L38) at build time using an MSBuild task.

At installation time the release date is read by the engine from using reflection from the ServiceControl executable: https://github.com/Particular/ServiceControl/blob/a56860be685270d06490342e90007e84480bf00e/src/ServiceControlInstaller.Engine/FileSystem/ReleaseDateReader.cs#L33-L46

As suggested by @HEskandari one option here is to:

- make so that the build process stores the release date in the zip file using a text file
- tweak the above code to read from the text file in case of failure

Another option that requires fewer changes is to use `Mono.Cecil` to read the assembly definition. I just tested it by reading a custom attribute defined in a .NET Framework 4.8 assembly from a .NET Core 3.1 console application. It's as simple as:

```csharp
var assemblyPath = @"path to the assembly to read";
var assemblyDefinition = Mono.Cecil.AssemblyDefinition.ReadAssembly(assemblyPath);

[...]

PR #2314, opened 2020-11-25
Use Mono.Cecil to read release date

mauroservienti:
Fix https://github.com/Particular/ServiceControl/issues/2123

This PR adds support for `PowerShell Core`. When managing ServiceControl instances using PowerShell Core a .NET Core based infrastructure (PowerShell) needs to load reflection information for a .NET Framework assembly (ServiceControl) and it blows up.

## Background

The installation process checks if the license is valid at installation time, and to do so it needs to extract the release date from the assembly it's trying to install.

The release date is stored in assemblies by the Licensing package. The Licensing package at build time, using an MSBuild task, [writes a `ReleaseDate` attribute](https://github.com/Particular/Operations.Licensing/blob/13608e31ab7ce46e357fd777380b9bc57051ab1b/src/Particular.Licensing/Particular.Licensing.Sources.targets#L16-L18) to the compiled assembly storing the GitVersion `$(GitVersion_CommitDate)`.

Finally, `ServiceControlInstaller.Packaging` project [zips all assemblies](https://github.com/Particular/ServiceControl/blob/a56860be685270d06490342e90007e84480bf00e/src/ServiceControlInstaller.Packaging/ServiceControlInstaller.Packaging.csproj#L38) at build time using an MSBuild task.

At installation time the release date is read by the engine using reflection from the ServiceControl executable: https://github.com/Particular/ServiceControl/blob/a56860be685270d06490342e90007e84480bf00e/src/ServiceControlInstaller.Engine/FileSystem/ReleaseDateReader.cs#L33-L46

One option is to use `Mono.Cecil`, as proposed by this PR, to read the assembly definition.

### DIFF
--- a/src/ServiceControl.Config.Tests/ServiceControl.Config.Tests.csproj
+++ b/src/ServiceControl.Config.Tests/ServiceControl.Config.Tests.csproj
@@ -12,7 +12,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.8.0" />
-    <PackageReference Include="Moq" Version="4.15.1" />
+    <PackageReference Include="Moq" Version="4.15.2" />
     <PackageReference Include="NUnit" Version="3.12.0" />
     <PackageReference Include="NUnit3TestAdapter" Version="3.17.0" />
     <PackageReference Include="Particular.CodeRules" Version="0.7.0" PrivateAssets="All" />
--- a/src/ServiceControl.Transports.SqlServer/ServiceControl.Transports.SqlServer.csproj
+++ b/src/ServiceControl.Transports.SqlServer/ServiceControl.Transports.SqlServer.csproj
@@ -9,6 +9,8 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Data.SqlClient" Version="2.0.1" />
+    <PackageReference Include="Microsoft.Data.SqlClient.SNI" Version="2.1.1" />
     <PackageReference Include="NServiceBus.Transport.SqlServer" Version="6.2.0" />
     <PackageReference Include="Particular.CodeRules" Version="0.7.0" PrivateAssets="All" />
   </ItemGroup>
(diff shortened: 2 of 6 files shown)


HEskandari (approved):
Looks good!

mauroservienti:
@danielmarbach @mikeminutillo @SzymonPobiega @boblangley @WilliamBZA do you see any downside in using this approach?

SzymonPobiega:
Why not loading the assembly for metadata reading only via reflection?

mauroservienti:
Sorry @SzymonPobiega I might have missed some context in the description. This is in order to allow users to use `PowerShell Core`, which means that a .NET Core based infrastructure needs to load reflection information for a .NET Framework assembly and it blows up at runtime when using reflection.

danielmarbach:
What's the benefit of having both code paths in place? Why not always use Mono.Cecil and get rid of the assembly reading? I think you might also need to use the same mono cecil version in the tests as a dependency. PublicApiGenerator has Mono.Cecil as a dependency and there might be a mismatch

danielmarbach:
An alternative could be to use 

https://www.nuget.org/packages/System.Reflection.MetadataLoadContext

see

https://docs.microsoft.com/en-us/dotnet/standard/assembly/inspect-contents-using-metadataloadcontext

mauroservienti:
@danielmarbach thanks for the feedback.

> What's the benefit of having both code paths in place? Why not always use Mono.Cecil and get rid of the assembly reading? 

None, I'm just experimenting and was looking for feedback on the new implementation. In case it's OK I'll remove the old, redundant, code.

> I think you might also need to use the same mono cecil version in the tests as a dependency. PublicApiGenerator has Mono.Cecil as a dependency and there might be a mismatch

Tried that with no luck. Need to further investigate.

> An alternative could be to use
> https://www.nuget.org/packages/System.Reflection.MetadataLoadContext

It works in my isolated test project. It is problematic here because the `ServiceControlInstaller.Engine` project targets `net40` and there is no compatible package. I'm wondering why it's targeting such a low version, though. @HEskandari do you have any idea?

danielmarbach:
I think it is because that package is shared with Powershell

HEskandari:
@mauroservienti The Engine is shared between SCMU and Powershell. ServiceControl.PowerShell is targeting net40, which is why the engine is also on that.

mauroservienti:
Closed in favor of https://github.com/Particular/ServiceControl/pull/2324

PR #2318, opened 2020-11-27
Renamed dockerfiles since they're used in TeamCity script

dvdstelt:
Renamed files to adhere to Docker Hub image names but also to accommodate automating of updating of tags in Octopus. The image names are planned to be added as artifact in a single text-file through which Octopus powershell script can than iterate.

- Added `-windows.dockerfile` to all images
- Renamed `asb.asbs` to `azureservicebus`
- Renamed `asq` to `azurestoragequeues`
- Renamed `sqs` to `amazonsqs`
- Moved `asb.endpoint` and `asb.forwarding` to **backuo** folder, since we don't officially support those anymore and they're migrated to new asb anyway.
- Downgraded os from `ltsc2019` to `ltsc2016`

Naming conventions documented here: https://github.com/Particular/Deployment/pull/147


ramonsmits:
Although not required I would be in favor of having a prefix or extension. . It seems visual studio supports `*.dockerfile` in its editor. In the script we should just trim the name to drop this extension.

WilliamBZA:
This should probably target the release branch, and the builddockerimages.ps1 script needs the filenames update too

dvdstelt:
Updated names of files and images. Added all changes to the description of this PR.

WilliamBZA (approved):

PR #2321, opened 2020-12-01
Document Package Verification Test

mikeminutillo:


docs/packaging.md
```diff
@@ -0,0 +1,42 @@
+# Packaging
+
+Each product is packaged into it's own versioned zip file in the `./zip` folder, and these are included in the installation package. When ServiceControl Management installs a new instance, it uses the files in these zips. Each zip contain two top-level folders
```
SeanFeldman, line 3:
```suggestion
Each product (ServiceControl, ServiceControl.Monitoring and ServiceControl.Audit),  is packaged into its own versioned zip file in the `./zip` folder. These zip files are included in the installation package. When ServiceControl Management installs a new instance, it uses the files in these zips. Each zip contain two top-level folders
```

docs/packaging.md
```diff
@@ -0,0 +1,42 @@
...
+
+- `/Transports` contains a folder for each supported transport
+- `/<product>` contains the instance files specific to the type of instance
+  - `/ServiceControl`
+  - `/ServiceControl.Audit`
+  - `/ServiceControl.Monitoring`
+  
+ When SCMU installs a product it unzips the main instance files, and the transport files into the same folder
```
SeanFeldman, line 11:
```suggestion
When ServiceControl Management Utility (SCMU) installs a product, it unzips the main instance files and the transport files into the same folder.
```

docs/packaging.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Assembly Mismatches
+
+There can be an issue when the main instance folder and the selected transport each have a copy of the same assembly. This can happen when the transport package and the instance each reference the same dependency, but select different versions. At install time, one version or the other will be copied into the instance binary folder and things may break unexpectedly at runtime.
+
+To prevent this, there is a unit test (`DeploymentPackageTests.DuplicatAssemblyFileSizesShouldMatch` that tests the generate zip files to look for duplicated assemblies. If their file sizes match, then the test passes. If the file sizes do not match, then the test will fail with a message like the following.
```
SeanFeldman, line 17:
```suggestion
To prevent this, the unit test `DeploymentPackageTests.DuplicatAssemblyFileSizesShouldMatch` tests the generated zip files for duplicated assemblies. If their file sizes match, then the test passes. If the file sizes do not match, then the test will fail with a message:
```

SeanFeldman (changes requested):

docs/packaging.md
```diff
@@ -0,0 +1,42 @@
...
+
+This is telling you the assemblies that are different from the ones found in the ServiceControl.Audit folder.
+
+### How to resolve
+
+When a mismatch occurs, we need to make sure that the versions of the dependencies match between the transport package and the ServiceControl package. 
+
+There are two mechanisms for doing that:
+
+1. Add an explicit version of the package as a dependency on the transport package that matches the version being used by ServiceControl. This is the preferred option, as Dependabot will pick up when a new version of the dependency has been released and raise a PR for us to test and review.
```
SeanFeldman, line 41: While this might be the case for explicitly referenced packages in two different sources, it doesn't appear to be the case for these two assemblies. Therefore it looks like another edge case. Both, `System.Runtime.CompilerServices.Unsafe.dll` and `System.Threading.Tasks.Extensions.dll` are used by neither the transport project (ServiceControl.Transports.ASQ nor ServiceControl project.

mikeminutillo: So it's a transitive dependency for something else? If there is a NuGet package you can add it in as a direct dependency to lock the version in.

SeanFeldman: I've tried to find what pulls it in and had a hard time.
It feels a bit hacky to just a package to lock the version rather than know what package brings that transitive dependency in.

SeanFeldman: I've [consolidated both transitive dependencies](https://github.com/Particular/ServiceControl/pull/2301/commits/2efe00b2c7fe018f1b2dc3171e4f18524a651880) into a shared build.props file to manage in one spot.

SeanFeldman (approved):

danielmarbach (approved):

PR #2325, opened 2020-12-01
remove suppression of dropped tasks analyser

adamralph:


danielmarbach:
judging based on the brittleness of the domain raise and stuff I think not awaiting the task was deliberate in the code. Maybe it makes more sense as an intermediate step to use the `_` (discard) or introduce `Ignore()` to consume the tasks?
Because what you introduced requires very careful placement of the `await task` which I find very dangerous and not worth the risk it introduces. Long story short I don't think we should change the intended design of ExecuteWhen just to satisfy an analyzer but make sure to consume the tasks that are meant to be fire&forget.

danielmarbach (dismissed):
judging based on the brittleness of the domain raise and stuff I think not awaiting the task was deliberate in the code. Maybe it makes more sense as an intermediate step to use the `_` (discard) or introduce `Ignore()` to consume the tasks?
Because what you introduced requires very careful placement of the `await task` which I find very dangerous and not worth the risk it introduces. Long story short I don't think we should change the intended design of ExecuteWhen just to satisfy an analyzer but make sure to consume the tasks that are meant to be fire&forget.

danielmarbach:
BTW I'm more than happy to help find a better more expressive solution by collaborating on a different PR

src/ServiceControl.AcceptanceTests/AcceptanceTest.cs
```diff
@@ -104,11 +104,11 @@ static void RemoveOtherTransportAssemblies(string name)
             }
         }
 
-        protected void ExecuteWhen(Func<bool> execute, Func<IDomainEvents, Task> action, string instanceName = Settings.DEFAULT_SERVICE_NAME)
+        protected Task ExecuteWhen(Func<bool> execute, Func<IDomainEvents, Task> action, string instanceName = Settings.DEFAULT_SERVICE_NAME)
         {
             var timeout = TimeSpan.FromSeconds(1);
 
-            Task.Run(async () =>
+            return Task.Run(async () =>
```
danielmarbach, line 111: What's the reason the task cannot be discarded here? That would be the most straightforward approach to remove the suppression. Furthermore it would allow the method to remain void returning. In addition to that it would require the caller of that high-level Api to know that the task has to be dropped due to its long idling until condition has been met nature

danielmarbach (approved):

PR #2327, opened 2020-12-02
[WIP] Make the installer file x64 bit

HEskandari:
Update the AIP file to 17.7 format and make the installer 64 bit.


mauroservienti (approved):

HEskandari:
With the installation in the PSModulePath this would not be necessary anymore, as the module will be picked but by both variations of PowerShell.

PR #2330, opened 2020-12-02
remove redundant method

adamralph:
`ServiceControl.MultiInstance.AcceptanceTests.AcceptanceTest.ExecuteWhen`

I was completely confused as to why [this change](https://github.com/Particular/ServiceControl/pull/2325/files#diff-27a55256e4a8beae22709ff4b1e7ea29cfb97c9c88ccbe942b7d601a9e076d13R108-R112) didn't break things until I realised the method is not used. 🙂🗑️

### DIFF
--- a/src/ServiceControl.MultiInstance.AcceptanceTests/AcceptanceTest.cs
+++ b/src/ServiceControl.MultiInstance.AcceptanceTests/AcceptanceTest.cs
@@ -8,8 +8,6 @@ namespace ServiceControl.MultiInstance.AcceptanceTests
     using System.Linq;
     using System.Net;
     using System.Net.Http;
-    using System.Threading;
-    using System.Threading.Tasks;
     using AcceptanceTesting;
     using Newtonsoft.Json;
     using NServiceBus;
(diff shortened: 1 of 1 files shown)


danielmarbach:
During the split time frame this method was introduced as a base infrastructure helper also in the multi instance ATT bar infrastructure to raise domain events directly in the primary. Hence the input is of type dynamic. Given that it hasn't been used since then I'm not opposed to remove it. It just want to clarify that the code we are touching is a base ATT infrastructure that has convenience methods that might not be used but have been added to enable advanced ATT scenarios and not needing to reinvent the wheel over and over. So I think we have to carefully balance the act of removing convenience methods. 

Paging @andreasohlund @mikeminutillo @WilliamBZA for their thoughts. Approving in the meantime since I think the long history of not needing it is enough to justify the removal

danielmarbach (approved):

andreasohlund (approved):

mikeminutillo (approved):
Approved. We can always bring it back if we need to

PR #2332, opened 2020-12-03
Improve audit ingestion to make sure ingestion loop continues even when dispose throws

danielmarbach:
In 4.14. we moved the `DisposeAsync` into a finally block. I assumed that the bulk insertion [follows the framework design guidelines](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1065#dispose-methods) but apparently, [it doesn't](https://github.com/ravendb/ravendb/blob/v3.5/Raven.Client.Lightweight/Document/RemoteBulkInsertOperation.cs#L499). 

Therefore it broke the assumptions that `Persist` cannot throw causing the ingestion loop to halt. This rearranges the ingestion logic to make sure we continue to ingest and move the "marking all messages in the batch as failed into the component to simplify the code"

### DIFF
--- a/src/ServiceControl.Audit/Auditing/AuditIngestionComponent.cs
+++ b/src/ServiceControl.Audit/Auditing/AuditIngestionComponent.cs
@@ -112,6 +112,19 @@
 
                     await ingestor.Ingest(contexts).ConfigureAwait(false);
                 }
+                catch (Exception e) // ingestion must continue
+                {
+                    if (log.IsInfoEnabled)
+                    {
+                        log.Info("Ingesting messages failed", e);
+                    }
+
+                    // let's give up
+                    foreach (var context in contexts)
+                    {
+                        context.GetTaskCompletionSource().TrySetException(e);
+                    }
+                }
                 finally
                 {
                     contexts.Clear();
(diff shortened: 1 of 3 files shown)


src/ServiceControl.Audit/Auditing/AuditIngestionComponent.cs
```diff
@@ -112,6 +112,19 @@ async Task Loop()
 
                     await ingestor.Ingest(contexts).ConfigureAwait(false);
                 }
+                catch (Exception e) // ingestion must continue
```
SzymonPobiega, line 115: Show must go on? ;-)

src/ServiceControl.Audit/Auditing/AuditIngestionComponent.cs
```diff
@@ -112,6 +112,19 @@ async Task Loop()
...
+                {
+                    if (log.IsInfoEnabled)
+                    {
+                        log.Info("Ingesting messages failed", e);
+                    }
+
+                    // let's give up
```
SzymonPobiega, line 122:
```suggestion
                    // signal all message handling tasks to terminate
```

SzymonPobiega (approved):

Issue #2259, opened 2020-11-03
Recoverability doesn't remove all error message headers when retrying


danielmarbach:
The header filter doesn't remove all error message headers because it requires an exact match of certain headers and we have never extended that logic to for example take ExceptionInfo data into account.

https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/Recoverability/Retrying/HeaderFilter.cs#L8-L15

Here is an example:

```
    "Headers": {
        "NServiceBus.MessageId": "caf68027-acce-4260-8442-ac6500e46afc",
        "NServiceBus.MessageIntent": "Publish",
        "NServiceBus.RelatedTo": "2e6c6c9c-c4cb-474e-98b8-ac6500e46a0c",
        "NServiceBus.ConversationId": "37c405c8-e4e4-4873-8eb4-ac6500e46621",
        "NServiceBus.CorrelationId": "6e449174-6f77-4da4-b9c0-ac6500e46621",
        "NServiceBus.OriginatingMachine": "SBMARKSAGA-P1",
        "NServiceBus.OriginatingEndpoint": "SeasNve.Market.PowerSupplyPurchaseOrderService.1.0",
        "$.diagnostics.originating.hostid": "4f8138bdb0421ffe1ceaee86e9145721",
        "NServiceBus.OriginatingSagaId": "9e0d2f01-e903-481a-b272-ac6500e46715",
        "NServiceBus.OriginatingSagaType": "SeasNve.Market.PowerSupplyPurchaseOrderService.PurchaseOrderSaga, SeasNve.Market.PowerSupplyPurchaseOrderService, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null",
        "NServiceBus.ReplyToAddress": "SeasNve.Market.PowerSupplyPurchaseOrderService.1.0@[dbo]@[Market.NServiceBus.Prod]",
        "NServiceBus.ContentType": "application/json",
        "NServiceBus.EnclosedMessageTypes": "SeasNve.Market.PowerSupplyPurchaseOrderService.ApiModels.Events.v1_0.PowerSupplyDebtorBlacklistCheckCompleted, SeasNve.Market.PowerSupplyOrderService.ApiModels, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null",
        "NServiceBus.Version": "7.1.0",
        "NServiceBus.TimeSent": "2020-10-31 13:59:26:479745 Z",
[...]

PR #2339, opened 2020-12-04
Recoverability doesn't remove all error message headers when retrying

seanfarmar:
Added exception, retries and timeout headers (simplest fix)

https://github.com/Particular/ServiceControl/issues/2259


danielmarbach:
I remove the timeout header for now since I'm not 100% sure we should remove it. I pinged the relevant internal channels

src/ServiceControl.UnitTests/Recoverability/HeaderFilterTests.cs
```diff
@@ -0,0 +1,57 @@
...
+                { "NServiceBus.OriginatingEndpoint", "PurchaseOrderService.1.0" },
+                { "$.diagnostics.originating.hostid", "4f8138bdb0421ffe1ceaee86e9145721" },
+                { "NServiceBus.OriginatingSagaId", "9e0d2f01-e903-481a-b272-ac6500e46715" },
+                { "NServiceBus.OriginatingSagaType", "PowerSupplyPurchaseOrderService.PurchaseOrderSaga, PowerSupplyPurchaseOrderService, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null" },
+                { "NServiceBus.ReplyToAddress", "PowerSupplyPurchaseOrderService.1.0@[dbo]@[Market.NServiceBus.Prod]" },
+                { "NServiceBus.ContentType", "application/json" },
+                { "NServiceBus.EnclosedMessageTypes", "PowerSupplyPurchaseOrderService.ApiModels.Events.v1_0.PowerSupplyDebtorBlacklistCheckCompleted, PowerSupplyOrderService.ApiModels, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null" },
+                { "NServiceBus.Version", "7.1.0" },
+                { "NServiceBus.TimeSent", "2020-10-31 13:59:26:479745 Z" },
+                { "NServiceBus.Retries.Timestamp", "2020-10-31 13:56:55:359877 Z" },
```
timbussmann, line 31: according to your comment on https://github.com/Particular/ServiceControl/issues/2259, shouldn't the retries header be removed too?

timbussmann (approved):
LGTM. Wild idea and haven't thought about it further about it, but if we combine all excluded headers into a single string instead of a list, we could do a single string.contains lookup instead of  a nested loop? Might cause some other unwanted matches though, so probably not worth it anyway.

src/ServiceControl.UnitTests/Recoverability/HeaderFilterTests.cs
danielmarbach: It should. Let me double check. Good catch

danielmarbach: Ah. Hehe he. It is. Check the approval file. This is the arrange

danielmarbach:
I thought about the string match too but considered it too dangerous because it could remove things that we don't want to have removed. Given that previously the linq did

```
var headersToRetryWith = headers
                .Where(kv => !KeysToRemoveWhenRetryingAMessage.Contains(kv.Key))
                .ToDictionary(kv => kv.Key, kv => kv.Value);
```

I think we are in similar big O territory as the current solution so I'm not too worried. Thoughts?

danielmarbach:
``` ini

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19042
AMD Ryzen 9 3950X, 1 CPU, 32 logical and 16 physical cores
.NET Core SDK=3.1.403
  [Host]     : .NET Core 3.1.9 (CoreCLR 4.700.20.47201, CoreFX 4.700.20.47203), X64 RyuJIT
  DefaultJob : .NET Core 3.1.9 (CoreCLR 4.700.20.47201, CoreFX 4.700.20.47203), X64 RyuJIT


```
| Method |     Mean |     Error |    StdDev |    StdErr |      Min |       Q1 |   Median |       Q3 |      Max |      Op/s | Ratio |
|------- |---------:|----------:|----------:|----------:|---------:|---------:|---------:|---------:|---------:|----------:|------:|
| Before | 5.133 μs | 0.0358 μs | 0.0335 μs | 0.0087 μs | 5.086 μs | 5.098 μs | 5.136 μs | 5.160 μs | 5.189 μs | 194,801.4 |  1.00 |
|  After | 2.693 μs | 0.0446 μs | 0.0418 μs | 0.0108 μs | 2.636 μs | 2.670 μs | 2.677 μs | 2.706 μs | 2.785 μs | 371,375.6 |  0.52 |

I think that says it all

timbussmann:
> I think we are in similar big O territory as the current solution so I'm not too worried. Thoughts?

yeah I'm totally ok with the proposed solution 👍

PR #2346, opened 2020-12-05
Make sure When_duplicated_messages_are_ingested works even when ordering is not guaranteed

danielmarbach:
The test was assuming that `TrailingMessage` always comes last which is not guaranteed. Even with the acceptance testing transport the file system can have different latency and it is possible that `TrailingMessage` is picked up before `DuplicatedMessage`. 

One could argue the test is still not 100% perfect because technically even when the deduplicate message mechanism is broken it would be possible for the done condition to be true between seeing the `TrailingMessage` and one `DuplicateMessage`. Because this is difficult to even verify I changed the behavior to send out four duplicates concurrently and remove the SC concurrency setting. To make sure the test would capture the dedup mechanism being broken I replaced 

```
        public static string ProcessingId(this IReadOnlyDictionary<string, string> headers)
        {
            var messageId = headers.MessageId();
            var processingEndpointName = headers.ProcessingEndpointName();
            var processingStarted = headers.ProcessingStarted();

            if (messageId == default || processingEndpointName == default || processingStarted == default)
            {
                return Guid.NewGuid().ToString();
            }

            return DeterministicGuid.MakeId(messageId, processingEndpointName, processingStarted).ToString();
        }
```

with

```
        public static string ProcessingId(this IReadOnlyDictionary<string, string> headers)
        {
            return Guid.NewGuid().ToString();
        }
```

and the test reliably failed.

### DIFF
--- a/src/ServiceControl.Audit.AcceptanceTests/Auditing/When_duplicated_messages_are_ingested.cs
+++ b/src/ServiceControl.Audit.AcceptanceTests/Auditing/When_duplicated_messages_are_ingested.cs
@@ -17,11 +17,6 @@
         [Test]
         public async Task Should_deduplicate()
         {
-            SetSettings = settings =>
-            {
-                settings.MaximumConcurrencyLevel = 1;
-            };
-
             var context = await Define<MyContext>()
                 .WithEndpoint<AnEndpoint>(b => b.When(async s =>
                 {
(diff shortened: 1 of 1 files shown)


seanfarmar (approved):
Great catch :-)

src/ServiceControl.Audit.AcceptanceTests/Auditing/When_duplicated_messages_are_ingested.cs
```diff
@@ -65,10 +61,14 @@ public AnEndpoint()
...
             {
-                public override async Task Invoke(IAuditContext context, Func<Task> next)
+                public override Task Invoke(IAuditContext context, Func<Task> next)
                 {
-                    await next();
-                    await next();
+                    if (context.Message.Headers[Headers.EnclosedMessageTypes].Contains(nameof(DuplicatedMessage)))
+                    {
+                        // create a few duplicate audit messages
+                        return Task.WhenAll(next(), next(), next(), next());
```
timbussmann, line 69: Technically, this MIGHT cause issues due to concurrent access on shared structures on the context bags which could not occur in normal execution? I can't think of a specific scenario where this would be a problem though. So not against it.

timbussmann (approved):
An even more "reliable" approach would be to send a follow-up message once the condition is fulfilled (probably sufficient once the duplicate message has been received). This way the the follow-up is enqueued at the end of the queue and helps to ensure that every message has been picked up from the queue before.

PR #2358, opened 2020-12-11
Remove attachment locks since the underlying put static has significantly changed and should no longer need it

danielmarbach:
Introduced in https://github.com/Particular/ServiceControl/pull/1046

It seems put static should no longer need this. At least on my local tests with RabbitMQ this worked perfectly fine

https://github.com/ravendb/ravendb/blob/v3.5/Raven.Database/Actions/AttachmentActions.cs#L90-L163

Seems to have a per body lock in place already on the DB side


danielmarbach:
@SzymonPobiega could you maybe get this a quick spin on the test machine you have running?

danielmarbach:
We are running this now on one of our test machines

danielmarbach:
We have been ingesting over 250 GB of data with RabbitMQ and this works flawlessly

timbussmann (approved):

PR #2359, opened 2020-12-11
Disable chunked bulk inserts

danielmarbach:
image

Given the small batch sizes we have and the complexity the chunking involves I think we should disable it explicitly and set the batch size

Chunking does the following:

It wraps a remote bullk insertion operation and sets the `MaximumDocumentsPerChunk` to four times the batch size of the bulk insertion operation. In addition to that the maximum volume of all documents is set to 8 MB. If either the `MaxChunkVolumeInBytes` or `MaximumDocumentsPerChunk` is reached it flushes the previous remote bulk insertion operation if there was any and then asynchronously disposes of the current one. After that a new remote bulk insertion operation is created. Created a new one requires to allocate another background thread that then tries to asynchronously complete over the queue it manages while the previous one is still disposing and being flushed against the server. 

A remote bulk insertion operation itself though is not limited in size at all so I think reducing the infrastructure complexity and handling inside RavenDB can only benefit us given that we ingest at max `MaxConcurrency` (default `32`) items in a single session providing the option to do junking seems overkill.

### DIFF
--- a/src/ServiceControl.Audit/Auditing/AuditPersister.cs
+++ b/src/ServiceControl.Audit/Auditing/AuditPersister.cs
@@ -50,7 +50,9 @@
                 // deliberately not using the using statement because we dispose async explicitly
                 bulkInsert = store.BulkInsert(options: new BulkInsertOptions
                 {
-                    OverwriteExisting = true
+                    OverwriteExisting = true,
+                    ChunkedBulkInsertOptions = null,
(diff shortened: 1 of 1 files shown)


danielmarbach:
We have been ingesting over 250 GB of data with RabbitMQ and this works flawlessly

WilliamBZA (approved):

seanfarmar (approved):

PR #2361, opened 2020-12-14
Remove tools folder

danielmarbach:
Originally found by 👓 @adamralph 

According to the history the tools folder was created more than 5 years ago to test specific Scenarios around the 1.x version of ServiceControl. At least as long as I was involved with the project I've never needed these tools nor was I really aware that they existed (but also have not missed them in my work). 

How about we remove those since they are not actively maintained or built using CI?


SzymonPobiega (approved):

SzymonPobiega:
Never heard of these ;-) Which means, I might have authored some of them 😂

WilliamBZA (approved):

PR #2367, opened 2020-12-15
Revert trial license installation feedback experience

WilliamBZA:
As part of https://github.com/Particular/ServiceControl/pull/2200 we prevented `--setup` to be executed when using trial licenses.

This was intended only to happen when running docker images, but it also propagated through to Powershell and the SCMU installer. The problem is that the feedback given to users was not that there was no valid license, and SCMU ended up copying files and leaving the instance in a weird state.

This was not the intention, so this PR tweaks that checks the license state to only block if a `IsDocker` environment variable is provided. This variable is set as part of the docker images too.

### DIFF
--- a/src/ServiceControl.LicenseManagement/LicenseManager.cs
+++ b/src/ServiceControl.LicenseManagement/LicenseManager.cs
@@ -35,8 +35,8 @@
                 return false;
             }
 
-            // E.g. when within a docker container
-            if (MustBeNonTrialLicenseForSetup() && license.IsEvaluationLicense)
+            // E.g. Cannot set up a new instance of ServiceControl on a trial license when running in a docker container
+            if (IsRunningInDocker && license.IsEvaluationLicense)
             {
                 errorMessage = "Cannot run ServiceControl in a container with a trial license";
                 return false;
@@ -108,12 +108,7 @@
             return LicenseFileLocationResolver.GetPathFor(Environment.SpecialFolder.CommonApplicationData);
         }
 
-        static bool MustBeNonTrialLicenseForSetup()
-        {
-            var isDockerEnvironmentVariable = Environment.GetEnvironmentVariable("SERVICECONTROL_NO_TRIAL");
-
-            return string.Equals("true", isDockerEnvironmentVariable, StringComparison.InvariantCultureIgnoreCase);
-        }
+        static bool IsRunningInDocker => bool.TryParse(Environment.GetEnvironmentVariable("SERVICECONTROL_RUNNING_IN_DOCKER"), out bool isDocker) && isDocker;
 
         static bool TryDeserializeLicense(string licenseText, out License license)
         {
--- a/src/ServiceControl.LicenseManagement/LicenseManager.cs
+++ b/src/ServiceControl.LicenseManagement/LicenseManager.cs
@@ -35,8 +35,8 @@
                 return false;
             }
 
-            // Not allowed to run setup within a docker container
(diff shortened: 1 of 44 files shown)


ramonsmits:
@WilliamBZA Isn't docker already providing some envvar to probe for?

WilliamBZA:
@ramonsmits not that I am aware of. You can even see a proposal to use a [similar env var in .net core](https://github.com/dotnet/runtime/issues/26433) to achieve a similar result. (Eventually withdrawn as the reasons behind the need were no longer valid)

ramonsmits:
Second pass I think this should not reference docker or containers. Instead of `IsDocker` why not have the key imply behavior like `SERVICECONTROL_NO_TRIAL`?

src/ServiceControl.LicenseManagement/LicenseManager.cs
```diff
@@ -107,6 +108,13 @@ static string GetMachineLevelLicenseLocation()
             return LicenseFileLocationResolver.GetPathFor(Environment.SpecialFolder.CommonApplicationData);
         }
 
+        static bool MustBeNonTrialLicenseForSetup()
+        {
+            var isDockerEnvironmentVariable = Environment.GetEnvironmentVariable("SERVICECONTROL_NO_TRIAL");
+
+            return string.Equals("true", isDockerEnvironmentVariable, StringComparison.InvariantCultureIgnoreCase);
```
danielmarbach, line 115: Is the content relevant or just the existence? If the content is relevant and needs to represent a boolean would it then make sense to use https://docs.microsoft.com/en-us/dotnet/api/system.boolean.tryparse?view=net-5.0 instead?

ramonsmits (approved):

src/ServiceControl.LicenseManagement/LicenseManager.cs
```diff
@@ -107,6 +108,13 @@ static string GetMachineLevelLicenseLocation()
             return LicenseFileLocationResolver.GetPathFor(Environment.SpecialFolder.CommonApplicationData);
         }
 
+        static bool MustBeNonTrialLicenseForSetup()
+        {
+            var isDockerEnvironmentVariable = Environment.GetEnvironmentVariable("SERVICECONTROL_NO_TRIAL");
```
danielmarbach, line 113: Isn't this confusing a bit? What we care about is that SC is running inside containers and the log error message wants to convey the message that you cannot run ServiceControl in a container with a trial license. So I don't understand why `SERVICECONTROL_NO_TRIAL` as a name is used. Because ultimately I could go and say, hmmm let's set it to false and then I can use a trial license. While when the variable is named what information it needs to convey like "SERVICECONTROL_RUNNING_IN_DOCKER" then what requirements we have in order to run with docker are more flexible and users that see that variable are less likely to tweak it. Maybe right now we are saying well we only need to ensure no trial. But maybe in the future there are other things we need to check when running in docker

So consequently if you want to keep the variable named as is you would also need to change the error message to say something like

> Cannot run ServiceControl in the current hosting environment with a trial license"

to convey the fact that it is not just about containers but actually about the license fitting a certain environment or not. Then the current name would make sense again for me. I still lean though towards expressing the intent better and make it more flexible in case we need to detect other things when running in docker

WilliamBZA:
Updated @danielmarbach

HEskandari (approved):
Looks good.

danielmarbach (approved):

Issue #2381, opened 2021-01-04
New-ServiceControlAuditInstance fails after upgrade to 4.15


barrit:
First, I create the Service Control instance. this works well.
Creating the Audit instance however, throws the following error:

`New-ServiceControlAuditInstance : Failed to retrieve release date for new version`

I checked the release notes of 4.15
Is it possible that some of the the changes in [this commit](https://github.com/Particular/ServiceControl/commit/36e55b35e479090c8a93aac6590c3afe7e0786dd#diff-eba1b0cd16ebc1a9437ea217e7a181648f9dcb971f071b7a7b0a35959aaa8809) have to be applied on the [audit instance cmdlets](https://github.com/Particular/ServiceControl/blob/36e55b35e479090c8a93aac6590c3afe7e0786dd/src/ServiceControlInstaller.PowerShell/Cmdlets/AuditInstances/NewServiceControlAuditInstance.cs#L148) as well?

barrit:
Rolled back to 4.14 which does not have the PowerShell Module in a subfolder and everything works fine again.

timbussmann:
I can reproduce and confirm this as not working, definitely seems to be a bug that has been added with 4.15

Cogax:
I do also have this issue after upgrading.
This is the Error:
```
New-ServiceControlAuditInstance : Failed to retrieve release date for new
version
At C:\Users\**\**\ServiceControl\PowerShel
l\ServiceControlManagement.ps1:139 char:12
+     return New-ServiceControlAuditInstance `
+            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [New-ServiceControlAuditInstan
   ce], Exception
    + FullyQualifiedErrorId : ServiceControlInstaller.PowerShell.NewServiceCon
   trolAuditInstance```

mauroservienti:
@barrit @Cogax I raised #2391 to fix the problem. I'm still investigating the reason why the repo ended up in this state. I'll let you know as soon as a new release is available.

mauroservienti:
For the records, here is the reason why it works in versions earlier than `4.15`: https://github.com/Particular/ServiceControl/pull/2391#issuecomment-759995316

ramonsmits:
@barrit We will be releasing 4.15.1 after the weekend.

PR #2391, opened 2021-01-13
Add a reference to Particular.Licensing to make so that ServiceControl.Audit gets a ReleaseDate attribute

mauroservienti:
This PR adds a package reference to the Particular.Licensing package in ServiceControl.Audit project. If the reference is missing the generated exe is not stamped with the ReleaseDateAttribute. The attribute causes the ReleadeDateReader used by unattended installations (e.g. PowerShell) to fail.

Fix https://github.com/Particular/ServiceControl/issues/2381

### DIFF
--- a/src/ServiceControl.Audit/ServiceControl.Audit.csproj
+++ b/src/ServiceControl.Audit/ServiceControl.Audit.csproj
@@ -37,6 +37,7 @@
     <PackageReference Include="Rx-Linq" Version="2.2.5" />
     <PackageReference Include="Microsoft.Owin.Cors" Version="4.1.1" />
     <PackageReference Include="Particular.CodeRules" Version="0.7.0" PrivateAssets="All" />
+    <PackageReference Include="Particular.Licensing.Sources" Version="3.4.0" PrivateAssets="All" />
     <PackageReference Include="System.Threading.Channels" Version="4.7.1" />
   </ItemGroup>
 


ramonsmits (approved):

ramonsmits:
Restarted continuous-integration/acceptancetests-rabbit-conventional as the build failed.

danielmarbach (approved):

danielmarbach:
What introduced the reading of the license attribute in the audit instance? Was it the powershell enhancement release?

WilliamBZA (approved):

mauroservienti:
I understood what the problem is and why the changes we implemented made it surface. Pat pat, good spelunking job 😉 

I debugged what happens when installing a new ServiceControl 4.14.2 (and earlier) Audit instance using PowerShell. The release date check fails in the same way it fails in 4.15.0, `ServiceControl.Audit.exe` doesn't have the `ReleaseDateAttribute`. The difference is that in 4.14.2 (and earlier) the exception is ignored and a weird release date, that makes it always valid, is returned:

image

I'm inclined to go ahead and merge this, and then proceed with all the required steps to backport to `release-4.15` and release a 4.15.1 patch.

@danielmarbach @ramonsmits @WilliamBZA thoughts?

WilliamBZA:
:+1:

danielmarbach:
👍

PR #2427, opened 2021-03-03
Use standard license sources

bording:
This restores ServiceControl's ability to find licenses in the "standard" sources, which was removed as part of #2155.

As of this PR, that means ServiceControl will once again start looking for a license in the following additional locations:

- The application's folder
- LocalApplicationData\ParticularSoftware

It also means that when we we add new standard sources, ServiceControl will stay in sync with those changes.

### DIFF
--- a/src/ServiceControl.Monitoring/Licensing/LicenseManager.cs
+++ b/src/ServiceControl.Monitoring/Licensing/LicenseManager.cs
@@ -1,6 +1,5 @@
 ﻿namespace ServiceControl.Monitoring.Licensing
 {
-    using System;
     using NServiceBus.Logging;
     using Particular.Licensing;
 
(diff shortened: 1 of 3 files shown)


src/ServiceControl.Monitoring/Licensing/LicenseManager.cs
```diff
@@ -12,11 +11,9 @@ public class LicenseManager
         public void Refresh()
         {
             Logger.Debug("Checking License Status");
-            var sources = new LicenseSource[]
-            {
-                new LicenseSourceFilePath(LicenseFileLocationResolver.GetPathFor(Environment.SpecialFolder.CommonApplicationData))
```
mikeminutillo, line 17: As long as this location is included in the `GetStandardLicenseSources()` list ([which it seems to be](https://github.com/Particular/Operations.Licensing/blob/master/src/Particular.Licensing/FindActiveLicense/LicenseSource.cs#L65)) then I am happy.

mikeminutillo (approved):

src/ServiceControl.Monitoring/Licensing/LicenseManager.cs
bording: Yes it is, and I don't see us removing that any time soon.

Issue #1445, opened 2018-10-08
Re-evaluate exposing additional events as part of SC integration


gbiellem:
Hi Particles

Looking at https://github.com/Particular/ServiceControl/tree/develop/src/ServiceControl/Contracts there are a plethora of events that are published however the current ServiceControl Contracts nuget only exposes 5 of them for external integration. 

The current nuget also only supports full framework.  Any chance it can be updated and published as  .net standard rather than full framework?

- [x] Update https://docs.particular.net/samples/servicecontrol/events-subscription/
- [x] Update https://docs.particular.net/servicecontrol/contracts PR: https://github.com/Particular/docs.particular.net/pull/5289
- [x] Show integration into alerting for example on Azure

SzymonPobiega:
Hey @gbiellem 

We didn't intend to publish these events to the outside world. Most of them are now only handled in-memory any way. Do you need any specific ones available?

Regarding the .NET standard support, let me raise a separate issue and it will be handled in the next release. We missed that aspect :-(

SzymonPobiega:
Issue https://github.com/Particular/ServiceControl/issues/1446 raised

gbiellem:
Hi @SzymonPobiega,

What i was specifically after were the MessageFailureResolvedByRetry and MessageFailureResolvedManually events.

SzymonPobiega:
@gbiellem If I remember correctly we were had some concerns related to worst-case volume of these messages (the throughput of the `MessageFailureResolvedByRetry` can be as high as the audit ingestion). But that's something we might want to re-evaluate now that we moved to V7 of NSB and the transports are much faster.

SzymonPobiega:
@gbiellem are you fine with me editing the title to reflect this better?

gbiellem:
@SzymonPobiega  Go for it

gbiellem:
@SzymonPobiega I think you goofed in the rename.  You've already opened an issue for ,Net Core support ( #1446).  I think you meant to rename this to address reevaluating the exposed events.

seanfarmar:
related to this issue: https://github.com/Particular/ServiceControl.Contracts/issues/32

seanfarmar:
Opened a PR for 
- Contracts 
- SC publishing them
- Updated documentation

If we have more time I can look at 
- building an Azure sample 
- Incorporating the new events in https://docs.particular.net/samples/servicecontrol/events-subscription/

seanfarmar:
[...]

PR #2465, opened 2021-03-31
Exposing additional events

seanfarmar:
https://github.com/Particular/ServiceControl/issues/1445

### DIFF
--- a/src/ServiceControl.AcceptanceTests/Recoverability/ExternalIntegration/When_a_failed_message_is_archived.cs
+++ b/src/ServiceControl.AcceptanceTests/Recoverability/ExternalIntegration/When_a_failed_message_is_archived.cs
@@ -8,6 +8,8 @@
     using Contracts;
     using ServiceControl.MessageFailures;
     using Newtonsoft.Json;
+    using TestSupport.EndpointTemplates;
+    using ServiceBus.Management.Infrastructure.Settings;
 
     class When_a_failed_message_is_archived : When_a_message_failed
     {
@@ -32,24 +34,49 @@
                 }))
                 .Do("WaitUntilErrorsContainsFailedMessage", async ctx =>
                 {
-                    return await this.TryGet<FailedMessage>($"/api/errors/{ErrorSender.FailedMessageId}",
+                    return await this.TryGet<FailedMessage>($"/api/errors/{ctx.FailedMessageId}",
                         e => e.Status == FailedMessageStatus.Unresolved);
                 })
                 .Do("WaitForExternalProcessorToSubscribe", ctx => Task.FromResult(ctx.ExternalProcessorSubscribed))
                 .Do("Archive", async ctx =>
                 {
-                    await this.Post<object>($"/api/errors/{ErrorSender.FailedMessageId}/archive");
+                    await this.Post<object>($"/api/errors/{ctx.FailedMessageId}/archive");
                 })
                 .Do("EnsureMessageIsArchived", async ctx =>
                 {
-                    return await this.TryGet<FailedMessage>($"/api/errors/{ErrorSender.FailedMessageId}",
+                    return await this.TryGet<FailedMessage>($"/api/errors/{ctx.FailedMessageId}",
                         e => e.Status == FailedMessageStatus.Archived);
                 })
                 .Done(ctx => ctx.EventDelivered) //Done when sequence is finished
                 .Run();
 
             var deserializedEvent = JsonConvert.DeserializeObject<FailedMessagesArchived>(context.Event);
-            CollectionAssert.Contains(deserializedEvent.FailedMessagesIds, ErrorSender.FailedMessageId);
+            CollectionAssert.Contains(deserializedEvent.FailedMessagesIds, context.FailedMessageId.ToString());
+        }
+
+        public class ExternalProcessor : EndpointConfigurationBuilder
+        {
+            public ExternalProcessor()
+            {
+                EndpointSetup<DefaultServer>(c =>
+                {
+                    var routing = c.ConfigureTransport().Routing();
+                    routing.RouteToEndpoint(typeof(FailedMessagesArchived).Assembly, Settings.DEFAULT_SERVICE_NAME);
+                }, publisherMetadata => { publisherMetadata.RegisterPublisherFor<FailedMessagesArchived>(Settings.DEFAULT_SERVICE_NAME); });
+            }
+
+            public class FailureHandler : IHandleMessages<FailedMessagesArchived>
+            {
+                public Context Context { get; set; }
+
+                public Task Handle(FailedMessagesArchived message, IMessageHandlerContext context)
+                {
+                    var serializedMessage = JsonConvert.SerializeObject(message);
+                    Context.Event = serializedMessage;
+                    Context.EventDelivered = true;
+                    return Task.FromResult(0);
+                }
+            }
         }
     }
 }
--- a/src/ServiceControl.AcceptanceTests/Recoverability/ExternalIntegration/When_a_failed_message_is_archived.cs
+++ b/src/ServiceControl.AcceptanceTests/Recoverability/ExternalIntegration/When_a_failed_message_is_archived.cs
@@ -11,7 +11,7 @@
     using TestSupport.EndpointTemplates;
     using ServiceBus.Management.Infrastructure.Settings;
 
-    class When_a_failed_message_is_archived : When_a_message_failed
+    class When_a_failed_message_is_archived : ExternalIntegrationAcceptanceTest
     {
         [Test]
         public async Task Should_publish_notification()
--- a/src/ServiceControl.AcceptanceTests/Recoverability/ExternalIntegration/When_a_failed_message_is_archived.cs
+++ b/src/ServiceControl.AcceptanceTests/Recoverability/ExternalIntegration/When_a_failed_message_is_archived.cs
@@ -36,7 +36,7 @@
                 }).DoNotFailOnErrorMessages())
                 .WithEndpoint<ExternalProcessor>(b => b.When(async (bus, c) =>
                 {
-                    await bus.Subscribe<FailedMessageArchived>();
+                    await bus.Subscribe<FailedMessagesArchived>();
 
                     if (c.HasNativePubSubSupport)
                     {
@@ -65,8 +65,8 @@
                 .Done(ctx => ctx.EventDelivered) //Done when sequence is finished
                 .Run();
 
-            var deserializedEvent = JsonConvert.DeserializeObject<FailedMessageArchived>(context.Event);
-            Assert.IsTrue(deserializedEvent.FailedMessageId == context.FailedMessageId);
+            var deserializedEvent = JsonConvert.DeserializeObject<FailedMessagesArchived>(context.Event);
+            CollectionAssert.Contains(deserializedEvent.FailedMessagesIds, context.FailedMessageId);
         }
 
         public class Receiver : EndpointConfigurationBuilder
@@ -113,15 +113,15 @@
                 EndpointSetup<DefaultServer>(c =>
                 {
                     var routing = c.ConfigureTransport().Routing();
-                    routing.RouteToEndpoint(typeof(FailedMessageArchived).Assembly, Settings.DEFAULT_SERVICE_NAME);
-                }, publisherMetadata => { publisherMetadata.RegisterPublisherFor<FailedMessageArchived>(Settings.DEFAULT_SERVICE_NAME); });
+                    routing.RouteToEndpoint(typeof(FailedMessagesArchived).Assembly, Settings.DEFAULT_SERVICE_NAME);
+                }, publisherMetadata => { publisherMetadata.RegisterPublisherFor<FailedMessagesArchived>(Settings.DEFAULT_SERVICE_NAME); });
             }
 
-            public class FailureHandler : IHandleMessages<FailedMessageArchived>
+            public class FailureHandler : IHandleMessages<FailedMessagesArchived>
             {
                 public MyContext Context { get; set; }
 
-                public Task Handle(FailedMessageArchived message, IMessageHandlerContext context)
+                public Task Handle(FailedMessagesArchived message, IMessageHandlerContext context)
                 {
                     var serializedMessage = JsonConvert.SerializeObject(message);
                     Context.Event = serializedMessage;
--- a/src/ServiceControl.AcceptanceTests/Recoverability/ExternalIntegration/When_a_failed_message_is_archived.cs
+++ b/src/ServiceControl.AcceptanceTests/Recoverability/ExternalIntegration/When_a_failed_message_is_archived.cs
@@ -1,35 +1,25 @@
 ﻿namespace ServiceControl.AcceptanceTests.Recoverability.ExternalIntegration
 {
-    using System;
-    using System.Collections.Generic;
     using System.Threading.Tasks;
     using AcceptanceTesting;
     using NServiceBus;
     using NServiceBus.AcceptanceTesting;
-    using ServiceBus.Management.Infrastructure.Settings;
     using NUnit.Framework;
     using Contracts;
     using ServiceControl.MessageFailures;
-    using TestSupport;
-    using TestSupport.EndpointTemplates;
     using Newtonsoft.Json;
-    using NServiceBus.Routing;
-    using NServiceBus.Transport;
-    using Conventions = NServiceBus.AcceptanceTesting.Customization.Conventions;
 
-    class When_a_failed_message_is_archived : AcceptanceTest
+    class When_a_failed_message_is_archived : When_a_message_failed
     {
         [Test]
         public async Task Should_publish_notification()
         {
-            CustomConfiguration = config => config.OnEndpointSubscribed<MyContext>((s, ctx) =>
+            CustomConfiguration = config => config.OnEndpointSubscribed<Context>((s, ctx) =>
             {
-                ctx.ExternalProcessorSubscribed = s.SubscriberReturnAddress.Contains("ExternalProcessor");
+                ctx.ExternalProcessorSubscribed = s.SubscriberReturnAddress.Contains(nameof(ExternalProcessor));
             });
 
-            var uniqueMessageId = Infrastructure.DeterministicGuid.MakeId(ErrorSender.MessageId, nameof(ErrorSender)).ToString();
-
-            var context = await Define<MyContext>()
+            var context = await Define<Context>()
                 .WithEndpoint<ErrorSender>(b => b.When(session => Task.CompletedTask).DoNotFailOnErrorMessages())
                 .WithEndpoint<ExternalProcessor>(b => b.When(async (bus, c) =>
                 {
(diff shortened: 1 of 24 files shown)


src/ServiceControl/ExternalIntegrations/FailedMessagesUnArchivedPublisher.cs
```diff
@@ -0,0 +1,35 @@
...
+                MessagesCount = @event.MessagesCount
+            };
+        }
+
+        protected override Task<IEnumerable<object>> PublishEvents(IEnumerable<DispatchContext> contexts, IAsyncDocumentSession session)
+        {
+            return Task.FromResult(contexts.Select(r => (object)new Contracts.FailedMessagesUnArchived
+            {
+                MessagesCount = r.MessagesCount,
+                DocumentIds = r.MessageIds
```
danielmarbach, line 25: Why are we publishing DocumentIds? How is that relevant for someone observing this event that is external to RavenDB? Shouldn't this be called MessagesIds?

src/ServiceControl/MessageFailures/Handlers/UnArchiveMessagesByRangeHandler.cs
```diff
@@ -50,6 +51,7 @@ public async Task Handle(UnArchiveMessagesByRange message, IMessageHandlerContex
 
             await domainEvents.Raise(new FailedMessagesUnArchived
             {
+                DocumentIds = patchedDocumentIds.Select(x => x.Document).ToArray(),
```
danielmarbach, line 54: Is this even correct? Aren't we patching against `FailedMessageViewIndex` and that index is using `FailedMessage` where the `Id` is determined by

```
        public static string MakeDocumentId(string messageUniqueId)
        {
            return $"{CollectionName}/{messageUniqueId}";
        }
```

which then would mean we'd have to do something like


```
DocumentIds = patchedDocumentIds.Select(x => FailedMessage.GetMessageIdFromDocumentId(x.Document)).ToArray(),
```

?

src/ServiceControl/ExternalIntegrations/MessageFailureResolvedByRetryPublisher.cs
```diff
@@ -0,0 +1,35 @@
...
+                FailedMessageId = new Guid(@event.FailedMessageId)
+            };
+        }
+
+        protected override Task<IEnumerable<object>> PublishEvents(IEnumerable<DispatchContext> contexts, IAsyncDocumentSession session)
+        {
+            return Task.FromResult(contexts.Select(r => (object)new Contracts.MessageFailureResolvedByRetry
+            {
+                FailedMessageId = r.FailedMessageId.ToString(),
+                AlternativeFailedMessageIds = r.AlternativeFailedMessageIds
```
danielmarbach, line 25: Isn't this information always empty? We have `RetryHandlerForBackwardsCompatability` and it doesn't set the array. Ah we set it in the recoverability specific enricher in the audit pipeline. 

```
            IEnumerable<string> GetAlternativeUniqueMessageId(IReadOnlyDictionary<string, string> headers)
            {
                var messageId = headers.MessageId();
                if (headers.TryGetValue(Headers.ProcessingEndpoint, out var processingEndpoint))
                {
                    yield return DeterministicGuid.MakeId(messageId, processingEndpoint).ToString();
                }

                if (headers.TryGetValue("NServiceBus.FailedQ", out var failedQ))
                {
                    yield return DeterministicGuid.MakeId(messageId, ExtractQueueNameForLegacyReasons(failedQ)).ToString();
                }

                if (headers.TryGetValue(Headers.ReplyToAddress, out var replyToAddress))
                {
                    yield return DeterministicGuid.MakeId(messageId, ExtractQueueNameForLegacyReasons(replyToAddress)).ToString();
                }
            }
```

How is that information relevant for the end users? Can you elaborate what they would do with this information?

src/ServiceControl/ExternalIntegrations/FailedMessagesUnArchivedPublisher.cs
seanfarmar: Make sense, I'll change the contract

src/ServiceControl/ExternalIntegrations/MessageFailureResolvedByRetryPublisher.cs
seanfarmar: In the case I was looking at the customer wants to use the events to reconcile his backup error collection, so he wants to know if he needs to remove a failed message from his backup. (so if it was resolved, archived or unarchived)

danielmarbach:
I think this lacks also acceptance tests under the corresponding External integration folder. See for example `When_a_message_has_failed_detected`

src/ServiceControl/ExternalIntegrations/MessageFailureResolvedByRetryPublisher.cs
```diff
@@ -0,0 +1,35 @@
...
+    using Contracts.MessageFailures;
+    using Raven.Client;
+
+    class MessageFailureResolvedByRetryPublisher : EventPublisher<MessageFailureResolvedByRetry, MessageFailureResolvedByRetryPublisher.DispatchContext>
+    {
+        protected override DispatchContext CreateDispatchRequest(MessageFailureResolvedByRetry @event)
+        {
+            return new DispatchContext
+            {
+                FailedMessageId = new Guid(@event.FailedMessageId)
```
mikeminutillo, line 16: Was there a reason to convert this to a Guid only to then convert it back to a string when we publish it? Raven is likely going to store it as a json string anyway.

src/ServiceControl/MessageFailures/Handlers/UnArchiveMessagesByRangeHandler.cs
```diff
@@ -50,6 +51,7 @@ public async Task Handle(UnArchiveMessagesByRange message, IMessageHandlerContex
 
             await domainEvents.Raise(new FailedMessagesUnArchived
             {
+                DocumentIds = patchedDocumentIds.Select(x => FailedMessage.GetMessageIdFromDocumentId(x.Document)).ToArray(),
```
mikeminutillo, line 54: Is there a limit to how many of these there can be? We may want to batch them otherwise these messages get very large.

src/ServiceControl/MessageFailures/Handlers/UnArchiveMessagesHandler.cs
```diff
@@ -41,6 +41,7 @@ await session.SaveChangesAsync()
 
             await domainEvents.Raise(new FailedMessagesUnArchived
             {
+                DocumentIds = failedMessages.Select(x => x.UniqueMessageId).ToArray(),
```
mikeminutillo, line 44: We may want to batch these to keep the message sizes low. I am not sure if there are natural limits to how big this collection can be.

mikeminutillo (commented):
Probably also need to expose `FailedMessageGroupArchived`.

Do we know what the impact on performance is?

For Azure Service Bus this will increase the number of operations against the broker, even if there are no subscribers. Should we consider having a way to disable some/all integration events?

src/ServiceControl/ExternalIntegrations/MessageFailureResolvedByRetryPublisher.cs
seanfarmar: Not that I can think of, I'll check if I can leave it as string :-)

src/ServiceControl/MessageFailures/Handlers/UnArchiveMessagesHandler.cs
seanfarmar: What kind of batch size do you think we should do here?
Is there a metric we can use?

src/ServiceControl/MessageFailures/Handlers/UnArchiveMessagesByRangeHandler.cs
seanfarmar: What kind of batch size do you think we should do here?
Is there a metric we can use?

src/ServiceControl.Audit/ServiceControl.Audit.csproj
```diff
@@ -33,7 +33,7 @@
     <PackageReference Include="NServiceBus.NLog" Version="3.0.0" />
     <PackageReference Include="NServiceBus.Raw" Version="3.2.2" />
     <PackageReference Include="Microsoft.Owin.Hosting" Version="4.1.1" />
-    <PackageReference Include="ServiceControl.Contracts" Version="2.0.0" />
+    <PackageReference Include="ServiceControl.Contracts" Version="3.0.0-PullRequest0046.9" />
```
ramonsmits, line 36: Incorrect package

src/ServiceControl/ServiceControl.csproj
```diff
@@ -34,7 +34,7 @@
     <PackageReference Include="NServiceBus.NLog" Version="3.0.0" />
     <PackageReference Include="NServiceBus.Raw" Version="3.2.2" />
     <PackageReference Include="Microsoft.Owin.Hosting" Version="4.1.1" />
-    <PackageReference Include="ServiceControl.Contracts" Version="2.0.0" />
+    <PackageReference Include="ServiceControl.Contracts" Version="3.0.0-PullRequest0046.9" />
```
ramonsmits, line 37: Same here, incorrect package.

src/ServiceControl.UnitTests/ApprovalFiles/APIApprovals.PublicClr.approved.txt
```diff
@@ -1174,6 +1175,7 @@ namespace ServiceControl.Recoverability
     public class FailedMessageGroupArchived : ServiceControl.Infrastructure.SignalR.IUserInterfaceEvent
     {
         public FailedMessageGroupArchived() { }
+        public string[] FailedMessagesIds { get; set; }
```
ramonsmits, line 1178: How many IDs can be in this array? Any issues with transport message sizes? I don't see any bound checks so wondering what the max size is.

As this is text and often van contain GUID that is 36 characters, with quotes and comma its 39. Meaning, 128kb can contain roughly ~3.000 ID's assuming these are GUID's.

ramonsmits (changes requested):

src/ServiceControl.Audit/ServiceControl.Audit.csproj
seanfarmar: It is the new contracts, it will be updated once we release it

src/ServiceControl/ServiceControl.csproj
seanfarmar: It is the new contracts, it will be updated once we release it

src/ServiceControl.UnitTests/ApprovalFiles/APIApprovals.PublicClr.approved.txt
SzymonPobiega: Good catch @ramonsmits . We have added it to a number of contracts. If the number of items was bigger than the max allowed message size the publish would fail and the event would not be delivered. Given that we don't have a good way of checking it I'd say we ignore it for now and document architecture cruft here. 

Alternatively we could split one domain event into multiple on-the-wire events with portions of the data.

seanfarmar: Good point, Do we any pattern/guidance about that? Does that depend on the transport we use?

SzymonPobiega: Unfortunately depends on the transport :\

tmasternak: We could add a new domain event when [creating retry batches](https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/Recoverability/Archiving/ArchiveAllInGroupHandler.cs#L66) and publish an integration event with message ids in that batch.

@seanfarmar do we need `GroupId` and `GroupName` in some of the integration events?

seanfarmar: @tmasternak For the scenario I'm thinking of, we don't because the customer is only tacking single messages...

src/ServiceControl/Recoverability/Archiving/Raven/ArchiveOperation.cs
```diff
@@ -13,6 +14,7 @@ class ArchiveOperation // raven
         public DateTime Started { get; set; }
         public int NumberOfBatches { get; set; }
         public int CurrentBatch { get; set; }
+        public List<string> FailedMessagesIds { get; set; }
```
SzymonPobiega, line 17: I think we don't need this one any more.

src/ServiceControl/ExternalIntegrations/FailedMessageArchivedPublisher.cs
```diff
@@ -0,0 +1,33 @@
...
+        {
+            return Task.FromResult(contexts.Select(r => (object)new Contracts.FailedMessagesArchived
+            {
+                FailedMessagesIds = new[] { r.FailedMessageId.ToString() }
+            }));
+        }
+
+        public class DispatchContext
+        {
+            public Guid FailedMessageId { get; set; }
```
SzymonPobiega, line 30: We can replace this one by `string FailedMessageId`. Both are serialized to string in json so there should be no problem with deserialization and it will save us the conversions in lines 16 and 24.

src/ServiceControl/ExternalIntegrations/FailedMessageGroupArchivedPublisher.cs
```diff
@@ -0,0 +1,33 @@
...
+                FailedMessagesIds = @event.FailedMessagesIds
+            };
+        }
+
+        protected override Task<IEnumerable<object>> PublishEvents(IEnumerable<DispatchContext> contexts, IAsyncDocumentSession session)
+        {
+            return Task.FromResult(contexts.Select(r => (object)new Contracts.FailedMessagesArchived
+            {
+                // cleanup FailedMessages/ publish guids without document collection name
+                FailedMessagesIds = r.FailedMessagesIds.Select(id => id.Replace("FailedMessages/", "")).ToArray()
```
SzymonPobiega, line 24: How about we remove the `FailedMessages` part when publishing the domain event so that we don't need to do it here? I think that would be cleaner and make it align with how `FailedMessageArchivedPublisher` and `FailedMessagesUnArchivedPublisher` work.

SzymonPobiega (dismissed):
Minor suggestions

SzymonPobiega (approved):

src/ServiceControl.AcceptanceTests/Recoverability/ExternalIntegration/When_a_failed_message_is_archived.cs
```diff
@@ -0,0 +1,150 @@
...
+    {
+        [Test]
+        public async Task Should_publish_notification()
+        {
+            CustomConfiguration = config => config.OnEndpointSubscribed<MyContext>((s, ctx) =>
+            {
+                if (s.SubscriberReturnAddress.IndexOf("ExternalProcessor", StringComparison.OrdinalIgnoreCase) >= 0)
+                {
+                    ctx.ExternalProcessorSubscribed = true;
+                }
```
tmasternak, line 28:
```suggestion
                ctx.ExternalProcessorSubscribed = s.SubscriberReturnAddress.Contains("ExternalProcessor");
```

src/ServiceControl.AcceptanceTests/Recoverability/ExternalIntegration/When_a_failed_message_is_resolved_by_retry.cs
```diff
@@ -0,0 +1,158 @@
...
+    {
+        [Test]
+        public async Task Should_publish_notification()
+        {
+            CustomConfiguration = config => config.OnEndpointSubscribed<MyContext>((s, ctx) =>
+            {
+                if (s.SubscriberReturnAddress.IndexOf("ExternalProcessor", StringComparison.OrdinalIgnoreCase) >= 0)
+                {
+                    ctx.ExternalProcessorSubscribed = true;
+                }
```
tmasternak, line 28:
```suggestion
                ctx.ExternalProcessorSubscribed = s.SubscriberReturnAddress.Contains("ExternalProcessor");
```

src/ServiceControl.AcceptanceTests/Recoverability/ExternalIntegration/When_a_failed_message_is_resolved_by_retry.cs
```diff
@@ -0,0 +1,158 @@
...
+                        e => e.Status == FailedMessageStatus.Resolved);
+                })
+                .Done(ctx => ctx.EventDelivered) //Done when sequence is finished
+                .Run();
+
+            var deserializedEvent = JsonConvert.DeserializeObject<MessageFailureResolvedByRetry>(context.Event);
+            Assert.IsTrue(deserializedEvent.FailedMessageId == context.FailedMessageId);
+        }
+
+        public class Receiver : EndpointConfigurationBuilder
```
tmasternak, line 79: Could we move this to a test base class to prevent duplication?

src/ServiceControl/Recoverability/Archiving/ArchiveAllInGroupHandler.cs
```diff
@@ -89,6 +91,15 @@ await documentManager.UpdateArchiveOperation(batchSession, archiveOperation)
                     await batchSession.SaveChangesAsync()
                         .ConfigureAwait(false);
 
+                    if (nextBatch != null)
+                    {
+                        await domainEvents.Raise(new FailedMessageGroupBatchArchived
+                        {
+                            // cleanup FailedMessages/ publish guids without document collection name
```
tmasternak, line 98:
```suggestion
                            // Remove `FailedMessages/` prefix and publish pure GUIDs without Raven collection name
```

tmasternak (changes requested):

src/ServiceControl.AcceptanceTests/Recoverability/ExternalIntegration/When_a_failed_message_is_resolved_by_retry.cs
seanfarmar: @tmasternak you are refactoring this now, so ignoring this for now

src/ServiceControl.AcceptanceTests/Recoverability/ExternalIntegration/When_a_failed_message_is_archived.cs
mikeminutillo: Technically speaking this suggestion could reset the flag to false if a second subscriber exists.

mikeminutillo (commented):
All in all this looks good. My chief concerns would be:

- Publishing the results of bulk opertaions with a list of GUIDs could quickly create very large messages. We should maybe consider batching these at the integration event level.
- Publishing more events may incur more costs in cloud environments. We should look at providing options to disable some/all of these event types.

SzymonPobiega:
> Publishing the results of bulk opertaions with a list of GUIDs could quickly create very large messages. We should maybe consider batching these at the integration event level.

We have changed the code to publish these messages for each archived batch (1000 docs) so we are safe here that we won't cross the threshold even on the most restricted transport (ASQ)

SzymonPobiega:
> We should maybe consider batching these at the integration event level

I think we are not making things significantly worse because the biggest impact on cost is probably `MessageFailed` event.

mikeminutillo:
> > We should maybe consider batching these at the integration event level
> 
> I think we are not making things significantly worse because the biggest impact on cost is probably `MessageFailed` event.

My concern is that for every `MessageFailed` event we are adding at least one more eventual event. So the number of events published will be 2x.

tmasternak:
> My concern is that for every MessageFailed event we are adding at least one more eventual event. So the number of events published will be 2x.

One more event is a situation that was already there before we added new events. Secondly, these events won't get published unless there is an integration event subscriber in the system.

mikeminutillo:
> One more event is a situation that was already there before we added new events. 

I am not following this. Can you explain how this situation was already present?

> Secondly, these events won't get published unless there is an integration event subscriber in the system.

For message-driven pub-sub that is true. For native-pub sub transports like Azure Service Bus (where every operation costs money) we still are going to push the event to the topic, even if there are no subscribers for it.

seanfarmar:
From Slack https://particularsoftware.slack.com/archives/C01SQ08L5S8/p1618993470145600

@mikeminutillo's summary:

To summarize the conversation:
- We already publish 1 event for every failed message
- Now, if you retry, we will publish 1 additional event for every failed message
- If you archive, then we batch those notification events
- If every message is retried once, we double the number of events published
- On most transports, this won't matter much
  - For any with some kind of subscription storage (native or message-driven), we will see that there are no subscribers for these new messages and not bother publishing them
  - For any with native pub-sub, we will publish the message to the broker regardless. This puts more pressure on the broker.
    - For RabbitMQ that's probably OK. Rabbit is often much faster than we are anyway and the customer has already paid the cost of setting up the infra. There is usually not an additional cost per operation.
    - For Azure Service Bus that's probably not OK. Each operation costs money and we may be publishing events that the customer doesn't even care about. For them, they will upgrade to a new version and their costs will go up. They will  have questions about that.

The outcome is that we will introduce a config flag to turn off the publishing of integration events.
- This flag allows those ASB customers to switch off these events if they don't use them.
  - Customers that don't care about any integration events will benefit from this, and we might be able to reduce costs for them
  - There may be a small number of customers that do care about the old integration events but not the new ones. I think we'd like to hear from them before we build anything more complex
- Even non-ASB customers might be interested in this because it takes the pressure off of RavenDB.
  - In high-throughput scenarios where we don't use Integration Events, I would advise turning this feature off.
  - I'd even go so far as to make it off by default, but I'd want to do that in a major with a user experience around turning it back on at runtime

src/ServiceControl.AcceptanceTests/Recoverability/ExternalIntegration/When_a_failed_message_is_archived.cs
```diff
@@ -0,0 +1,82 @@
...
+    using NServiceBus;
+    using NServiceBus.AcceptanceTesting;
+    using NUnit.Framework;
+    using Contracts;
+    using ServiceControl.MessageFailures;
+    using Newtonsoft.Json;
+    using TestSupport.EndpointTemplates;
+    using ServiceBus.Management.Infrastructure.Settings;
+
+    class When_a_failed_message_is_archived : When_a_message_failed
```
danielmarbach, line 14: Is it possible to rename the base class to not contain `When_`? It know this sounds super nitpicky but for my brain a `When_` is always a concrete scenario but here it seems `When_a_message_failed` is just a convenience base class that provides infrastructure to avoid having to repeat things for external integration. If the fact that all those events are based upon message failure how about calling it something like 

- `MessageFailedAcceptanceTests` or
- `MessageFailedIntegrationAcceptanceTests`

src/ServiceControl/ExternalIntegrations/MessageFailureResolvedByRetryPublisher.cs
```diff
@@ -0,0 +1,35 @@
...
+            return Task.FromResult(contexts.Select(r => (object)new Contracts.MessageFailureResolvedByRetry
+            {
+                FailedMessageId = r.FailedMessageId.ToString(),
+                AlternativeFailedMessageIds = r.AlternativeFailedMessageIds
+            }));
+        }
+
+        public class DispatchContext
+        {
+            public Guid FailedMessageId { get; set; }
```
danielmarbach, line 31: Wouldn't it be simpler to just make this a string? Then you wouldn't have to convert to guid and back to string right?

src/ServiceControl/ExternalIntegrations/MessageFailureResolvedManuallyPublisher.cs
```diff
@@ -0,0 +1,33 @@
...
+        {
+            return Task.FromResult(contexts.Select(r => (object)new Contracts.MessageFailureResolvedManually
+            {
+                FailedMessageId = r.FailedMessageId.ToString()
+            }));
+        }
+
+        public class DispatchContext
+        {
+            public Guid FailedMessageId { get; set; }
```
danielmarbach, line 30: Wouldn't it be simpler to just make this a string? Then you wouldn't have to convert to guid and back to string right?

danielmarbach (changes requested):
A few comments

tmasternak:
@danielmarbach can you have a look now? Thx.

danielmarbach (approved):

Issue #1744, opened 2019-02-20
Make EndpointInputQueue a struct


danielmarbach:

danielmarbach:
Done as part of #2467

PR #2467, opened 2021-04-01
Indexing endpoint types converted to structs

tmasternak:
Connected to https://github.com/Particular/ServiceControl/issues/1744

### DIFF
--- a/src/ServiceControl.Monitoring/Infrastructure/EndpointInputQueue.cs
+++ b/src/ServiceControl.Monitoring/Infrastructure/EndpointInputQueue.cs
@@ -15,27 +15,8 @@
 
         public bool Equals(EndpointInputQueue other) => string.Equals(EndpointName, other.EndpointName) && string.Equals(InputQueue, other.InputQueue);
 
-        public override bool Equals(object obj)
-        {
-            if (obj is null)
-            {
-                return false;
-            }
+        public override bool Equals(object obj) => obj is EndpointInputQueue inputQueue && Equals(inputQueue);
 
-            if (obj.GetType() != GetType())
-            {
-                return false;
-            }
-
-            return Equals((EndpointInputQueue)obj);
-        }
-
-        public override int GetHashCode()
-        {
-            unchecked
-            {
-                return ((EndpointName != null ? EndpointName.GetHashCode() : 0) * 397) ^ (InputQueue != null ? InputQueue.GetHashCode() : 0);
-            }
-        }
+        public override int GetHashCode() => (EndpointName, InputQueue).GetHashCode();
     }
 }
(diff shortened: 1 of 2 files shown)


src/ServiceControl.Monitoring/Infrastructure/EndpointInputQueue.cs
```diff
@@ -13,20 +13,7 @@ public EndpointInputQueue(string endpointName, string inputQueue)
...
-            if (ReferenceEquals(this, other))
-            {
-                return true;
-            }
-
-            return string.Equals(EndpointName, other.EndpointName) && string.Equals(InputQueue, other.InputQueue);
-        }
+        public bool Equals(EndpointInputQueue other) => string.Equals(EndpointName, other.EndpointName) && string.Equals(InputQueue, other.InputQueue);
 
         public override bool Equals(object obj)
```
danielmarbach, line 18: Fancy enough this could be simplified with pattern matching

```
public override bool Equals(object obj) => obj is EndpointInputQueue inputQueue && Equals(inputQueue);
```

similar on the other struct.

But now I wonder if we should just use the default value tuple to store things because that would do all of those things for us. So basically replace `EndpointInputQueue` with (string endpoint, string inputQueue) and simply reference System.ValueTuple

danielmarbach: If we stick to those custom tuples the GetHashCode could leverage value tuples like

```
public override int GetHashCode() => (EndpointName, InputQueue).GetHashCode()
```

tmasternak: I like the simplified versions of `Equal` and `GetHashCode`. 

I would keep the named structs instead of tuple especially for `EndpointMessageType` as it would become `(string, string)` and could lead to ending up in the wrong order of arguments when doing an instantiation.

danielmarbach (approved):

src/ServiceControl.Monitoring/Infrastructure/EndpointInputQueue.cs
```diff
@@ -13,47 +13,10 @@ public EndpointInputQueue(string endpointName, string inputQueue)
...
-            return string.Equals(EndpointName, other.EndpointName) && string.Equals(InputQueue, other.InputQueue);
-        }
-
-        public override bool Equals(object obj)
-        {
-            if (obj is null)
-            {
-                return false;
-            }
+        public bool Equals(EndpointInputQueue other) => string.Equals(EndpointName, other.EndpointName) && string.Equals(InputQueue, other.InputQueue);
```
mikeminutillo, line 16: This is going to require an exact case match. So did the original so this isn't a new problem.

mikeminutillo (approved):

PR #2475, opened 2021-04-06
Maintenance mode using CTRL+C / SIGINT instead waiting for keyboard input.

ramonsmits:
When running maintenance mode in Docker this failed because maintenance mode used `Console.ReadKey` while docker runs non-interactive. Now using CTRL+C / SIGINT instead to prevent failure.

image


danielmarbach (approved):

danielmarbach:
@ramonsmits could you quickly retest this in your docker environment. Made small tweaks

ramonsmits:
@danielmarbach validated, see description

Issue #1643, opened 2019-05-17
Starting in maintenance mode requires elevated privileges


mikeminutillo:
Starting an instance in maintenance mode requires elevated privileges due to an incorrect URLACL.

### Who's affected

All customers running a ServiceControl instance with non-elevated account. This only affects running the instance in maintenance mode. Applying the fix is not urgent.

### Symptoms

Attempting to start a ServiceControl instance in maintenance mode fails with a `System.Net.HttpListenerException` exception.

### Analysis

When we install the instance we create a URLACL for the maintenance port that looks like `http://localhost:PORT/`. When we give this address to the embedded RavenDB in maintenance mode and tell it to run it's own web server, it converts into an address that looks like this `http://+:PORT/`. Unfortunately, a URLACL cannot be applied unless the address being requested matches the one in the URLACL exactly. We are creating the wrong URLACL.

mikeminutillo:
Until this is fixed, the workaround is to add a URLACL for the correct address. Something like this:

> netsh http add urlacl url=http://+:PORT/ user=BUILTIN\Users

kbaley:
Relabeled as an improvement as there is a workaround and this doesn't affect the correctness of ServiceControl.

tmasternak:
Done in #2476

PR #2476, opened 2021-04-06
Adding aclurl for all hostnames for the maintenance mode

tmasternak:
Connected to https://github.com/Particular/ServiceControl/issues/1643

### DIFF
--- a/src/ServiceControlInstaller.Engine/Instances/ServiceControlBaseService.cs
+++ b/src/ServiceControlInstaller.Engine/Instances/ServiceControlBaseService.cs
@@ -66,18 +66,6 @@ namespace ServiceControlInstaller.Engine.Instances
             }
         }
 
-        //This is an old aclurl registration for embedded RavenDB instance that includes the hostname.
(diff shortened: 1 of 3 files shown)


SzymonPobiega (approved):

src/ServiceControlInstaller.Engine/Instances/ServiceControlBaseService.cs
```diff
@@ -66,7 +66,10 @@ public string AclUrl
             }
         }
 
-        public string AclMaintenanceUrl
+        //This is an old aclurl registration for embedded RavenDB instance that includes the hostname.
+        //We need that to make sure we can clean-up old registration when removing instances created
+        //using pre 4.17 versions of ServiceControl.
+        public string LegacyAclMaintenanceUrl
```
danielmarbach, line 72: I think it might be better to inline this instead of having a compute only property. The comment could be moved into the RemoveUrlAcl method

danielmarbach (approved):

PR #2478, opened 2021-04-06
RavenDB index health reporter

ramonsmits:
- Check for RavenDB index errors at start.
- Run a custom check every 5 minutes that:
  - Writes a DEBUG log entry with index statistics.
  - Writes WARN/ERROR log entries for indexes where there IndexLag value is above 10.000/100.000
  - Writes ERROR log entries for any index errors reported.


aleksandr-samila (approved):

src/ServiceControl.Infrastructure.RavenDB/RavenHealthReporter.cs
```diff
@@ -0,0 +1,76 @@
...
+                var statistics = documentStore.DatabaseCommands.GetStatistics();
+
+                report.AppendLine("Report:");
+                report.AppendLine($"CountOfStaleIndexesExcludingDisabledAndAbandoned:  {statistics.CountOfStaleIndexesExcludingDisabledAndAbandoned}");
+
+                var indexes = statistics.Indexes.OrderBy(x => x.Name);
+
+                foreach (var indexStats in indexes)
+                {
+                    //var indexLag = unchecked((uint?)indexStats.IndexingLag);
```
HEskandari, line 28: Remove?

HEskandari (approved):

src/ServiceControl.Infrastructure.RavenDB/RavenHealthReporter.cs
ramonsmits: The `IndexLag` value sometimes is negative which I don't understand. How can lag be negative 😃 . The only thing I can image is that the underlying type is a uint so I asked it on the RavenDB forum https://groups.google.com/g/ravendb/c/wKyQqjjWHvc but no response yet.

ramonsmits: https://github.com/ravendb/ravendb/blob/2623ab17c207928110930fe481f78a35a5b0a01a/Raven.Abstractions/Data/IndexStats.cs

```c#
 public void SetLastDocumentEtag(Etag lastDocEtag)
{
    if (lastDocEtag == null)
        return;

    IndexingLag = (int) (lastDocEtag.Changes - LastIndexedEtag.Changes);

    if (lastDocEtag.Restarts != LastIndexedEtag.Restarts)
    {
        IndexingLag *= -1;
    }
}
```

Not sure what a negative value means but it is NOT unsigned. Going to make it absolute instead.

mikeminutillo:
Is it worth considering doing this as a custom check? We have other platform health checks that use the custom check pattern to report to users.

ramonsmits:
I'll convert it into a custom check.

src/ServiceControl.Audit/Infrastructure/CheckRavenDBIndexLag.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        static int CheckAndReportIndexesWithTooMuchIndexLag(DatabaseStatistics statistics, IndexStats[] indexes)
+        {
+            int indexCountWithTooMuchLag = 0;
+
+            foreach (var indexStats in indexes)
+            {
+                // IndexingLag is the number of documents that the index is behind, it is not a time unit.
+                var indexLag = indexStats.IndexingLag.GetValueOrDefault();
+                indexLag = Math.Abs(indexLag);
```
SzymonPobiega, line 47: Why `Abs`? Can `indexLag` be negative?

SzymonPobiega:
I am hesitant about waiting for non stale indexes at start or throwing until we are sure that there are no customer who do have stale indexes and working ServiceControl. I would fear that we are going to blow up perfectly fine instance of ServiceControl.

ramonsmits:
> I am hesitant about waiting for non stale indexes at start or throwing until we are sure that there are no customer who do have stale indexes and working ServiceControl. I would fear that we are going to blow up perfectly fine instance of ServiceControl.

A perfectly fine instance wouldn't have a lot of work to get non-stale indexes at start. At most a few seconds if the instance was underload when it gracefully exited.

This guards again the problem that when the SC instance exits ungracefully (kill, crash, etc.) and has index issues where indexes get rebuild by RavenDB at start start will be delayed until that point.

src/ServiceControl.Audit/Infrastructure/CheckRavenDBIndexLag.cs
ramonsmits: @SzymonPobiega Yes, see:

- https://groups.google.com/g/ravendb/c/wKyQqjjWHvc
- https://github.com/ravendb/ravendb/blob/3.5.10-patch-35307/Raven.Abstractions/Data/IndexStats.cs#L154-L165

But reading that code again I think it the `Abs` should be removed.

tmasternak:
@ramonsmits one of the arguments for not delaying the start was to prevent the situation when our clients are already in the situation when the indexes are failing or significantly lagging behind.

That said these clients will first face the delay when upgrading to the new minor of the Service Control. I would assume that this is a moment when they actually have sometimes put set aside to have a closer look at this. Secondly, if we put the bare minimum alerting on the lagging indexes they should be covered on an ongoing basis.

ramonsmits:
@tmasternak Then lets only keep the reporter?

aleksandr-samila:
As for me, I like the idea with a flag(-s)

> ... Finally, I would put in place a setting flag that would enable skipping the checks when set... 

or vice versa doesn't matter at all - but we add a possibility.

tmasternak:
> Then let's only keep the reporter?

@ramonsmits and remove the startup delay?

ramonsmits:
@tmasternak @SzymonPobiega I've removed the wait for indexes to become non-stale at start. Still throwing on index errors.

src/ServiceControl.Infrastructure.RavenDB/Extensions.cs
```diff
@@ -16,5 +19,21 @@ public static void Query<TState>(this DocumentDatabase db, string index, IndexQu
                 onItem(doc, state);
             }
         }
+
+        public static void ThrowWhenIndexErrors(this EmbeddableDocumentStore documentStore)
```
tmasternak, line 23: Is this still used?

tmasternak (approved):

src/ServiceControl.Audit.UnitTests/ApprovalFiles/APIApprovals.CustomCheckDetails.approved.txt
```diff
@@ -1,3 +1,4 @@
 ServiceControl Health: Audit Message Ingestion Process
 ServiceControl.Audit Health: Audit Message Ingestion
+Storage: CheckRavenDBIndexLag
```
ramonsmits, line 3: @tmasternak @SzymonPobiega do we have any naming guidelines for these?

mikeminutillo: I don't believe so. They should be human readable.

src/ServiceControl.Audit/Infrastructure/CheckRavenDBIndexLag.cs
```diff
@@ -0,0 +1,90 @@
...
+            report.AppendLine("Internal RavenDB index health report:");
+
+            foreach (var indexStats in indexes)
+            {
+                // IndexingLag is the number of documents that the index is behind, it is not a time unit.
+                var indexLag = indexStats.IndexingLag.GetValueOrDefault();
+                indexLag = Math.Abs(indexLag);
+                report.AppendLine($"- Index [{indexStats.Name,-44}] Stale: {statistics.StaleIndexes.Contains(indexStats.Name),-5}, Lag: {indexLag,9:n0}, Valid: {indexStats.IsInvalidIndex,-5}, LastIndexed: {indexStats.LastIndexedTimestamp:u}, LastIndexing: {indexStats.LastIndexingTime:u}");
+            }
+            _log.Debug(report.ToString());
```
mikeminutillo, line 81: Given that this logs to debug, should we terminate early if debug logging is not enabled?

src/ServiceControl.Audit/Infrastructure/CheckRavenDBIndexLag.cs
```diff
@@ -0,0 +1,90 @@
...
+
+                if (indexLag > IndexLagThresholdError)
+                {
+                    indexCountWithTooMuchLag++;
+                    _log.Error($"Index [{indexStats.Name}] IndexingLag {indexLag:n0} is above {IndexLagThresholdError:n0}, consider starting the instance in maintenance mode to ensure index can recover.");
+                }
+                else if (indexLag > IndexLagThresholdWarning)
+                {
+                    indexCountWithTooMuchLag++;
+                    _log.Warn($"Index [{indexStats.Name}] IndexingLag {indexLag:n0} is above {IndexLagThresholdWarning:n0}, please start the instance in maintenance mode to ensure index can recover.");
```
mikeminutillo, line 57: This message and the one above seem to be around the wrong way. If you're above the error threshold we tell you to _consider_ starting in maintenance mode. If you're between warning and error we _tell you_ to start in maintenance mode.

It feels like this should be reversed:

- Warning: Maybe think about maintenance mode to let things catch up
- Error: Run in maintenance mode to let things catch up

Given that this is an audit instance, should we also link to guidance about scaling out audit, or will the docs link from the custom check do that?

src/ServiceControl.Audit/Infrastructure/CheckRavenDBIndexLag.cs
```diff
@@ -0,0 +1,90 @@
...
+                }
+            }
+
+            foreach (var indexError in statistics.Errors)
```
mikeminutillo, line 61: This feels out of place in this method. It doesn't impact the return value at all. Maybe pull it into a separate  method and call it from `PerformCheck`?

src/ServiceControl/Operations/CheckRavenDBIndexLag.cs
```diff
@@ -0,0 +1,90 @@
...
+    using System.Linq;
+    using System.Text;
+    using System.Threading.Tasks;
+    using NServiceBus.CustomChecks;
+    using NServiceBus.Logging;
+    using Raven.Client;
+    using Raven.Abstractions.Data;
+    using ServiceBus.Management.Infrastructure.Settings;
+
+    class CheckRavenDBIndexLag : CustomCheck
```
mikeminutillo, line 13: I think all of the same comments applied to the other custom check likely apply to this one. :)

mikeminutillo (commented):
Overall, I love this! Great job.

HEskandari (approved):

src/ServiceControl/Operations/CheckRavenDBIndexLag.cs
```diff
@@ -0,0 +1,90 @@
...
+    {
+        public CheckRavenDBIndexLag(IDocumentStore store, LoggingSettings settings)
+            : base(nameof(CheckRavenDBIndexLag), "Storage", TimeSpan.FromMinutes(5))
```
danielmarbach, line 16: You need a unique name otherwise you can't really distinguish those on the UI. I would also prefer the name hard-coded as a string so that when the name of the class changes the public visible name that is exposed on the SP page doesn't

src/ServiceControl.Audit/Infrastructure/CheckRavenDBIndexLag.cs
```diff
@@ -0,0 +1,90 @@
...
+    using NServiceBus.CustomChecks;
+    using NServiceBus.Logging;
+    using Raven.Client;
+    using Raven.Abstractions.Data;
+    using ServiceControl.Audit.Infrastructure.Settings;
+
+    class CheckRavenDBIndexLag : CustomCheck
+    {
+        public CheckRavenDBIndexLag(IDocumentStore store, LoggingSettings settings)
+            : base(nameof(CheckRavenDBIndexLag), "Storage", TimeSpan.FromMinutes(5))
```
danielmarbach, line 16: You need a unique name otherwise you can't really distinguish those on the UI. I would also prefer the name hard-coded as a string so that when the name of the class changes the public visible name that is exposed on the SP page doesn't

src/ServiceControl.Audit.UnitTests/ApprovalFiles/APIApprovals.CustomCheckDetails.approved.txt
danielmarbach: How about

Storage index: Number of audit documents behind

src/ServiceControl.UnitTests/ApprovalFiles/APIApprovals.CustomCheckDetails.approved.txt
```diff
@@ -3,4 +3,5 @@ Health: ServiceControl Remotes
 ServiceControl Health: Audit Message Ingestion
 ServiceControl Health: Error Message Ingestion
 ServiceControl Health: Error Message Ingestion Process
+Storage: CheckRavenDBIndexLag
```
danielmarbach, line 6: How about

Storage index: Number of error messages behind

src/ServiceControl.Audit.UnitTests/ApprovalFiles/APIApprovals.CustomCheckDetails.approved.txt
danielmarbach: and now I wonder why we don't put them all into ServiceControl/.Audit Health category

src/ServiceControl.Infrastructure.RavenDB/Extensions.cs
danielmarbach, line 23: Strictly speaking, I think we should move this part into a dedicated PR since the custom checks provides value on its own and for me this one is the most controversial change. But if other reviewers are fine with this change I'm OK to leave it here

src/ServiceControl.Infrastructure.RavenDB/Extensions.cs
```diff
@@ -16,5 +19,21 @@ public static void Query<TState>(this DocumentDatabase db, string index, IndexQu
...
+        {
+            var statistics = documentStore.DatabaseCommands.GetStatistics();
+
+            if (statistics.Errors.Length > 0)
+            {
+                var text = new StringBuilder();
+                text.AppendLine("Detected RavenDB index errors, please start maintenance mode and resolve the following issues:");
```
danielmarbach, line 30: I know we normally don't do it but I wonder if this would be a good case to actually have a link to the documentation in the text

src/ServiceControl.Infrastructure.RavenDB/Extensions.cs
danielmarbach: I think these changes are orthogonal and not really about reporting due the the blocking nature of throwing an exception. And w if we move this change out we can merge the custom check things and have a focused discussion on the potential edge cases about this change

src/ServiceControl.Infrastructure.RavenDB/Extensions.cs
mikeminutillo: We do have one in the custom check message, so there is already precedent. I think we just link to the ServiceControl troubleshooting page which feels relatively safe to do.

src/ServiceControl/Operations/CheckRavenDBIndexLag.cs
ramonsmits: True, but how would you be able to distinguish between audit instances?

ramonsmits:
@mikeminutillo @danielmarbach thanks for your great feedback. I've applied some refactorings. The Index error check at start is removed. Only reporting remains. I'll cherry pick the index error check on startup into a seperate PR.

src/ServiceControl.Audit/Infrastructure/CheckRavenDBIndexLag.cs
```diff
@@ -0,0 +1,100 @@
...
+                if (indexLag > IndexLagThresholdError)
+                {
+                    indexCountWithTooMuchLag++;
+                    _log.Error($"Index [{indexStats.Name}] IndexingLag {indexLag:n0} is above error threshold ({IndexLagThresholdError:n0}). Launch in maintenance mode to let indexes catch up.");
+                }
+                else if (indexLag > IndexLagThresholdWarning)
+                {
+                    indexCountWithTooMuchLag++;
+                    _log.Warn($"Index [{indexStats.Name}] IndexingLag {indexLag:n0} is above warning threshold ({IndexLagThresholdWarning:n0}). Launch in maintenance mode to let indexes catch up.");
+                }
```
mikeminutillo, line 60: A minor thought, but if debug logging is enabled, would we consider logging the IndexLag for any index that is OK? I'm not sure if that would ever be useful.

mikeminutillo (approved):

danielmarbach (approved):

PR #2479, opened 2021-04-06
Lower MaximumConcurrencyLevel default to 2

ramonsmits:
MaximumConcurrencyLevel default to 2 to prevent indexes become too stale under load.

During load testing it was observed that ingestion rate can be high (hundreds of message per second) but that indexes become too stale. Some never achieve to 'keep up' with the ingestion rate. 

This behavior was more evident the faster the storage becomes as this resulted in the CPU to max out. Likely due to indexing which results in indexing operation to be too slow.

Taking 2 as the default should result in async indexes to "keep up". During load testing a continuous ingestion rate of more than 50 messages per second was still possible.


SzymonPobiega:
Not sure about that. Our current baseline was 100 msg/s. If this PR essentially halves the already bad throughput I think we should do something in order to get the indexing performance better (e.g. drop the full text indexing?) rather than slowing down the ingestion.

ramonsmits:
Closing, decided that having faster defaults is preferred and that reporting problems with index lag would be a better strategy for most deployments.

PR #2485, opened 2021-04-07
Check for RavenDB index errors and delay start until no stale indexes

ramonsmits:


src/ServiceControl.Infrastructure.RavenDB/Extensions.cs
```diff
@@ -16,5 +18,40 @@ public static void Query<TState>(this DocumentDatabase db, string index, IndexQu
...
+            var interval = TimeSpan.FromMinutes(1);
+            var next = DateTime.MinValue;
+            int staleIndexes;
+
+            while ((staleIndexes = documentStore.DatabaseCommands.GetStatistics().StaleIndexes.Length) > 0)
+            {
+                var now = DateTime.UtcNow;
+                if (next < now)
+                {
+                    Console.WriteLine($"Stale indexes detected ({staleIndexes}), will delay starting until all indexes are non-stale. DO NOT KILL THIS PROCESS! Operation can run for a very long time!");
```
HEskandari, line 49: Use the logger instead?

src/ServiceControl.Infrastructure.RavenDB/Extensions.cs
```diff
@@ -16,5 +18,40 @@ public static void Query<TState>(this DocumentDatabase db, string index, IndexQu
...
+
+            if (statistics.Errors.Length > 0)
+            {
+                var text = new StringBuilder();
+                text.AppendLine("Detected RavenDB index errors, please start maintenance mode and resolve the following issues:");
+                foreach (var indexError in statistics.Errors)
+                {
+                    text.AppendLine($"Index [{indexError.IndexName}] error: {indexError.Error} (Action: {indexError.Action},  Doc: {indexError.Document}, At: {indexError.Timestamp})");
+                }
+                throw new Exception(text.ToString());
```
HEskandari, line 34: Does this actually terminate the process or catch/display the error and continue? Maybe an `Environment.Exit(1)` is an alternative?

ramonsmits: @HEskandari Didn't test this yet but as this happens as a blocking call during bootstrap I assumed it would fail to start.

src/ServiceControl.Infrastructure.RavenDB/Extensions.cs
ramonsmits: Yeah, good comment! I ported this from a standalone spike I created before copy pasting/

HEskandari (approved):

SzymonPobiega:
Folks, please if possible delay merging this PR until we're done with the enhancement release because we are working on the same area (SC performance) and we fear that we might step into each other's toes.

We also think it is a potentially very dangerous change for some customers as we don't know how many people have SC instances with stale indexes running in the wild and now showing severe enough symptoms that would justify raising a support case.

ramonsmits:
@SzymonPobiega A healthy instance should have no issues. This should not delay a healthy startup much.

I'll also tweak the message that stale indexes are "normal".

src/ServiceControl.Infrastructure.RavenDB/Extensions.cs
```diff
@@ -16,5 +19,48 @@ public static void Query<TState>(this DocumentDatabase db, string index, IndexQu
...
+
+            if (statistics.Errors.Length > 0)
+            {
+                var text = new StringBuilder();
+                text.AppendLine("Detected RavenDB index errors, please start maintenance mode and resolve the following issues:");
+                foreach (var indexError in statistics.Errors)
+                {
+                    text.AppendLine($"- Index [{indexError.IndexName}] error: {indexError.Error} (Action: {indexError.Action},  Doc: {indexError.Document}, At: {indexError.Timestamp})");
+                }
+                throw new Exception(text.ToString());
```
aleksandr-samila, line 35: should we consider creating a specific exception? Something like "RavenDBHasStaleIndexes"?

ramonsmits:
Closing this in favor of https://github.com/Particular/ServiceControl/pull/2478 which contains the changes from this PR.

PR #2488, opened 2021-04-08
Recoverability: For non-binary below LOH error messages bypass body store

danielmarbach:
## Throughput

73.72048436 | avg
-- | --
74 | median
80 | P95
114.82 | P99

Compared to results of

https://github.com/Particular/ServiceControl/pull/2477

58.3953252 | avg
-- | --
60 | median
66 | P95
74 | P99

### DIFF
--- a/src/ServiceControl.Audit.UnitTests/BodyStorage/BodyStorageEnricherTests.cs
+++ b/src/ServiceControl.Audit.UnitTests/BodyStorage/BodyStorageEnricherTests.cs
@@ -170,10 +170,10 @@ namespace ServiceControl.UnitTests.BodyStorage
         {
             public int StoredBodySize { get; set; }
 
-            public Task<string> Store(string bodyId, string contentType, int bodySize, Stream bodyStream)
+            public Task Store(string bodyId, string contentType, int bodySize, Stream bodyStream)
             {
                 StoredBodySize = bodySize;
-                return Task.FromResult(default(string));
+                return Task.CompletedTask;
             }
 
             public Task<StreamResult> TryFetch(string bodyId)
--- a/src/ServiceControl.Audit.UnitTests/BodyStorage/BodyStorageEnricherTests.cs
+++ b/src/ServiceControl.Audit.UnitTests/BodyStorage/BodyStorageEnricherTests.cs
@@ -27,10 +27,15 @@ namespace ServiceControl.UnitTests.BodyStorage
             var body = Encoding.UTF8.GetBytes(new string('a', maxBodySizeToStore + 1));
             var metadata = new Dictionary<string, object>();
 
-            await enricher.StoreAuditMessageBody(body, new Dictionary<string, string>(), metadata);
+            var headers = new Dictionary<string, string>
+            {
+                { Headers.MessageId, "someid" }
+            };
+            await enricher.StoreAuditMessageBody(body, headers, metadata);
 
             Assert.AreEqual(0, fakeStorage.StoredBodySize, "Body should be removed if above threshold");
             Assert.IsFalse(metadata.ContainsKey("Body"));
+            Assert.AreEqual("/messages/someid/body", metadata["BodyUrl"]);
         }
 
         [Test]
@@ -46,12 +51,17 @@ namespace ServiceControl.UnitTests.BodyStorage
             var enricher = new BodyStorageFeature.BodyStorageEnricher(fakeStorage, settings);
             var body = Encoding.UTF8.GetBytes(new string('a', maxBodySizeToStore + 1));
             var metadata = new Dictionary<string, object>();
-            var headers = new Dictionary<string, string> { { Headers.ContentType, "application/binary" } };
+            var headers = new Dictionary<string, string>
+            {
+                { Headers.ContentType, "application/binary" },
+                { Headers.MessageId, "someid" }
+            };
 
             await enricher.StoreAuditMessageBody(body, headers, metadata);
 
             Assert.AreEqual(0, fakeStorage.StoredBodySize, "Body should be removed if above threshold");
             Assert.IsFalse(metadata.ContainsKey("Body"));
+            Assert.AreEqual("/messages/someid/body", metadata["BodyUrl"]);
         }
 
         [Test]
@@ -69,10 +79,15 @@ namespace ServiceControl.UnitTests.BodyStorage
             var body = Encoding.UTF8.GetBytes(new string('a', expectedBodySize));
             var metadata = new Dictionary<string, object>();
 
-            await enricher.StoreAuditMessageBody(body, new Dictionary<string, string>(), metadata);
+            var headers = new Dictionary<string, string>
+            {
+                { Headers.MessageId, "someid" }
+            };
+            await enricher.StoreAuditMessageBody(body, headers, metadata);
 
             Assert.AreEqual(body, metadata["Body"], "Body should be stored if below threshold");
             Assert.AreEqual(0, fakeStorage.StoredBodySize);
+            Assert.AreEqual("/messages/someid/body", metadata["BodyUrl"]);
         }
 
         [Test]
@@ -90,10 +105,15 @@ namespace ServiceControl.UnitTests.BodyStorage
             var body = Encoding.UTF8.GetBytes(new string('a', expectedBodySize));
             var metadata = new Dictionary<string, object>();
 
-            await enricher.StoreAuditMessageBody(body, new Dictionary<string, string>(), metadata);
+            var headers = new Dictionary<string, string>
+            {
+                { Headers.MessageId, "someid" }
+            };
+            await enricher.StoreAuditMessageBody(body, headers, metadata);
 
             Assert.AreEqual(expectedBodySize, fakeStorage.StoredBodySize, "Body should be stored if below threshold");
             Assert.IsFalse(metadata.ContainsKey("Body"));
+            Assert.AreEqual("/messages/someid/body", metadata["BodyUrl"]);
         }
 
         [Test]
@@ -110,12 +130,17 @@ namespace ServiceControl.UnitTests.BodyStorage
             var expectedBodySize = BodyStorageFeature.BodyStorageEnricher.LargeObjectHeapThreshold + 1;
             var body = Encoding.UTF8.GetBytes(new string('a', expectedBodySize));
             var metadata = new Dictionary<string, object>();
-            var headers = new Dictionary<string, string> { { Headers.ContentType, "application/binary" } };
+            var headers = new Dictionary<string, string>
+            {
+                { Headers.ContentType, "application/binary" },
+                { Headers.MessageId, "someid" }
+            };
 
             await enricher.StoreAuditMessageBody(body, headers, metadata);
 
             Assert.AreEqual(expectedBodySize, fakeStorage.StoredBodySize, "Body should be stored if below threshold");
             Assert.IsFalse(metadata.ContainsKey("Body"));
+            Assert.AreEqual("/messages/someid/body", metadata["BodyUrl"]);
         }
 
         [Test]
@@ -133,10 +158,12 @@ namespace ServiceControl.UnitTests.BodyStorage
             var body = Encoding.UTF8.GetBytes(new string('a', expectedBodySize));
             var metadata = new Dictionary<string, object>();
 
-            await enricher.StoreAuditMessageBody(body, new Dictionary<string, string>(), metadata);
+            var headers = new Dictionary<string, string> { { Headers.MessageId, "someid" } };
+            await enricher.StoreAuditMessageBody(body, headers, metadata);
 
             Assert.AreEqual(expectedBodySize, fakeStorage.StoredBodySize, "Body should be stored if below threshold");
             Assert.IsFalse(metadata.ContainsKey("Body"));
+            Assert.AreEqual("/messages/someid/body", metadata["BodyUrl"]);
         }
 
         class FakeBodyStorage : IBodyStorage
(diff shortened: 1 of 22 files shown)


seanfarmar (approved):

SzymonPobiega (approved):

danielmarbach:
Damn I forgot that the staging mechanism needs a change as well

tmasternak (approved):

src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSender.cs
```diff
@@ -30,23 +37,51 @@ public virtual async Task HandleMessage(MessageContext message, IDispatchMessage
 
             if (outgoingHeaders.TryGetValue("ServiceControl.Retry.Attempt.MessageId", out var attemptMessageId))
             {
-                var result = await bodyStorage.TryFetch(attemptMessageId)
-                    .ConfigureAwait(false);
-                if (result.HasResult)
+                if (outgoingHeaders.Remove("ServiceControl.Retry.HasAttachment"))
                 {
-                    using (result.Stream)
+                    var result = await bodyStorage.TryFetch(attemptMessageId)
```
mikeminutillo, line 42: It might be worth refactoring out each method here to get some clarity on what's going on

src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSender.cs
```diff
@@ -30,23 +37,51 @@ public virtual async Task HandleMessage(MessageContext message, IDispatchMessage
...
+                    {
+                        string resultBody = ((loadResult["$values"] as RavenJArray)?.SingleOrDefault() as RavenJObject)?.ToObject<MessagesBodyTransformer.Result>()?.Body;
+                        if (resultBody != null)
+                        {
+                            body = Encoding.UTF8.GetBytes(resultBody);
+                        }
+
+                        if (Log.IsDebugEnabled)
+                        {
+                            Log.DebugFormat("{0}: Body size: {1} bytes retrieved from index", messageId, body.LongLength);
```
mikeminutillo, line 78: This might need to be moved into the if-statement above. We might end up with

```
MESSAGE-ID: Body size: 0 bytes retrieved from index
```

when in fact, we did not get a value from the index. In any case, we might want to put some more logging in here to catch all of the cases that might come up because they will be a pain to diagnose in support cases.

src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSender.cs
```diff
@@ -2,22 +2,29 @@ namespace ServiceControl.Recoverability
...
     using NServiceBus.Routing;
     using NServiceBus.Transport;
     using Operations.BodyStorage;
+    using Raven.Client;
+    using Raven.Json.Linq;
 
     class ReturnToSender
     {
-        public ReturnToSender(IBodyStorage bodyStorage)
+        public ReturnToSender(IBodyStorage bodyStorage, IDocumentStore documentStore)
```
mikeminutillo, line 19: Would it make sense to hide some of the implementation details behind `IBodyStorage`? Or to create a new abstration that encompasses both of these approaches. There's a lot of body retrieval code in the return to sender now.

danielmarbach: I'm happy to address this once we have some good data that shows this approach is worth doing.

danielmarbach: we also already have similar yet different logic inside the `GetBodyByIdApi` controller

src/ServiceControl/Operations/BodyStorage/BodyStorageFeature.cs
```diff
@@ -58,36 +58,39 @@ static string GetContentType(IReadOnlyDictionary<string, string> headers, string
...
 
-            async Task StoreBody(byte[] body, IReadOnlyDictionary<string, string> headers, IDictionary<string, object> metadata, int bodySize, string contentType)
+            async ValueTask StoreBody(byte[] body, IReadOnlyDictionary<string, string> headers, IDictionary<string, object> metadata, int bodySize, string contentType)
             {
                 var bodyId = headers.MessageId();
+                var bodyUrl = string.Format(BodyUrlFormatString, bodyId);
                 var isBinary = contentType.Contains("binary");
                 var avoidsLargeObjectHeap = bodySize < LargeObjectHeapThreshold;
 
-                var bodyUrl = await StoreBodyInBodyStorage(body, bodyId, contentType, bodySize)
```
ramonsmits, line 67: I saw this last weekend and wondering why this was happening. Based on the changes is seems "dual" storing was a bug?

src/ServiceControl/Operations/BodyStorage/BodyStorageFeature.cs
```diff
@@ -58,36 +58,39 @@ static string GetContentType(IReadOnlyDictionary<string, string> headers, string
                 return contentType;
             }
 
-            async Task StoreBody(byte[] body, IReadOnlyDictionary<string, string> headers, IDictionary<string, object> metadata, int bodySize, string contentType)
+            async ValueTask StoreBody(byte[] body, IReadOnlyDictionary<string, string> headers, IDictionary<string, object> metadata, int bodySize, string contentType)
             {
                 var bodyId = headers.MessageId();
+                var bodyUrl = string.Format(BodyUrlFormatString, bodyId);
                 var isBinary = contentType.Contains("binary");
```
ramonsmits, line 65: I also saw this... plenty of other content types that are binary that do not have "binary" in the content type.

```suggestion
                var isBinary = !contentType.Contains("text") && !contentType.Contains("xml")  && !contentType.Contains("json")
```

What do you think?

danielmarbach: That would have to be an OR condition right? Plus I think it is dangerous to introduce this now since we could potentially be breaking things.

ramonsmits: @danielmarbach Not text, and not xml, and not JSON, then its binary. Could also invert the whole and do
```suggestion
                var isBinary = !(contentType.Contains("text") || contentType.Contains("xml")  || contentType.Contains("json"));
```
I'm not sure why its there. I assume that we store non binaries bodies in the message for full-text-search reasons? Storing binary messages in the message does not make sense for full-text-search. 

But.. that is just a guess.

src/ServiceControl/Operations/BodyStorage/BodyStorageFeature.cs
mikeminutillo: I wasn't a bug but it was suboptimal. We store the body in each place for different reasons. The attachment storage was there from the beginning for every message because that gave us a binary blob we could use to recreate the message when returning to the source queue. The in-document storage is there purely to support the full text indexing and was never really intended to be used for retries.

src/ServiceControl/Operations/BodyStorage/BodyStorageFeature.cs
mikeminutillo: That is correct. The in-document storage was only ever to support full text indexing. If the tokenizer can't read it, then indexing it is pointless. In reality we only have a very small handful of officially supported serializers. We could turn this into configuration and have a list of "indexable" content types, defaulted to a known list. That might also make it easier to turn off full-text indexing. Clear out your list of indexable content types and we'll stop putting message bodies in the metadata.

src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSender.cs
danielmarbach: Mike for now I refactored it into two methods. I'm wondering how much more it should be abstracted since the GetBodyByIdApi uses a different mechanism to fetch the body. I could try to see if we can use the same ID based querying with transformer there as a quick spike

danielmarbach: The API needs to query by by the message id of the last attempt for some reasons and uses 

```
               var message = await session.Query<MessagesViewIndex.SortAndFilterOptions, MessagesViewIndex>()
                    .Statistics(out var stats)
                    .TransformWith<MessagesBodyTransformer, MessagesBodyTransformer.Result>()
                    .FirstOrDefaultAsync(f => f.MessageId == messageId)
                    .ConfigureAwait(false);
```

I don't think we can consolidate that

src/ServiceControl.Audit.UnitTests/BodyStorage/BodyStorageEnricherTests.cs
```diff
@@ -27,10 +27,15 @@ public async Task Should_remove_body_when_above_threshold()
             var body = Encoding.UTF8.GetBytes(new string('a', maxBodySizeToStore + 1));
             var metadata = new Dictionary<string, object>();
 
-            await enricher.StoreAuditMessageBody(body, new Dictionary<string, string>(), metadata);
+            var headers = new Dictionary<string, string>
+            {
+                { Headers.MessageId, "someid" }
```
mikeminutillo, line 32: Is this an indication that this header has become required? Or is it just for the validation of the url?

src/ServiceControl/Operations/ErrorIngestionComponent.cs
```diff
@@ -24,8 +27,12 @@ class ErrorIngestionComponent
         Channel<MessageContext> channel;
         ErrorIngestor ingestor;
         ErrorPersister persister;
+        Counter receivedMeter;
```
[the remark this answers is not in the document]
danielmarbach: Yes for testing it on the system we did a deliberate rebase and added it

src/ServiceControl.UnitTests/ApprovalFiles/APIApprovals.PublicClr.approved.txt
```diff
@@ -169,6 +169,7 @@ namespace ServiceControl.CompositeViews.Messages
     public class MessagesBodyTransformer : Raven.Client.Indexes.AbstractTransformerCreationTask<ServiceControl.CompositeViews.Messages.MessagesViewTransformer.Result>
     {
         public MessagesBodyTransformer() { }
+        public static string Name { get; }
```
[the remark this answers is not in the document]
danielmarbach: I think I used the same visibility as the other indexes that have a name

src/ServiceControl.Audit.UnitTests/BodyStorage/BodyStorageEnricherTests.cs
danielmarbach: Just for the validation of the URL. We never did before validate the URL so I added this extension to the test

src/ServiceControl.UnitTests/BodyStorage/BodyStorageEnricherTests.cs
```diff
@@ -5,34 +5,70 @@ namespace ServiceControl.UnitTests.BodyStorage
...
     {
         [Test]
-        public async Task Should_store_body_regardless_of_the_body_size()
+        public async Task Should_store_body_in_storage_when_binary()
         {
             var fakeStorage = new FakeBodyStorage();
-            // previously the max body storage default was larger than 100 KB
+
+            var enricher = new BodyStorageFeature.BodyStorageEnricher(fakeStorage);
+            var expectedBodySize = BodyStorageFeature.BodyStorageEnricher.LargeObjectHeapThreshold + 1;
```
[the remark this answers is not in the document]
danielmarbach: Good idea and I think there should be another test for binary below the large object heap I can add this later this afternoon or feel free to do the changes

src/ServiceControl.UnitTests/BodyStorage/BodyStorageEnricherTests.cs
```diff
@@ -5,34 +5,70 @@ namespace ServiceControl.UnitTests.BodyStorage
...
+        public async Task Should_store_body_in_metadata_when_below_large_object_heap_and_not_binary()
+        {
+            var fakeStorage = new FakeBodyStorage();
+
+            var enricher = new BodyStorageFeature.BodyStorageEnricher(fakeStorage);
+            var expectedBodySize = BodyStorageFeature.BodyStorageEnricher.LargeObjectHeapThreshold - 1;
+            var body = Encoding.UTF8.GetBytes(new string('a', expectedBodySize));
+            var metadata = new Dictionary<string, object>();
+
+            await enricher.StoreErrorMessageBody(body, new Dictionary<string, string>(), metadata);
```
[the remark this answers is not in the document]
danielmarbach: I think the idea always was to even index as long as we have some textual representation available and just rely on the raven analyzers. I do agree it is an orthogonal concern and we have to think through the implications of potentially breaking things.

src/ServiceControl.UnitTests/BodyStorage/BodyStorageEnricherTests.cs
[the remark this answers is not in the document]
SzymonPobiega: I'll take care of this

src/ServiceControl.UnitTests/BodyStorage/BodyStorageEnricherTests.cs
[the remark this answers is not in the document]
SzymonPobiega: Yeah. Agree we should use allow-lists rather than deny-lists but that's a separate concern.

src/ServiceControl/Operations/BodyStorage/BodyStorageFeature.cs
SzymonPobiega: This is separate concern from this PR. We can address this is a different PR.

PR #2496, opened 2021-04-12
Bring the rudimentary metrics support from the Raven 5 branch

SzymonPobiega:
I copied the code from the Raven 5 branch where we added a very simple support for metrics for the purposes of measuring increase (or decrease) in performance.

### DIFF
--- a/src/ServiceControl.Audit/Auditing/AuditIngestionComponent.cs
+++ b/src/ServiceControl.Audit/Auditing/AuditIngestionComponent.cs
@@ -2,6 +2,7 @@
 {
     using System;
     using System.Collections.Generic;
+    using System.Diagnostics;
     using System.Threading;
     using System.Threading.Channels;
     using System.Threading.Tasks;
@@ -12,10 +13,12 @@
     using NServiceBus.Logging;
     using NServiceBus.Transport;
     using Raven.Client;
+    using ServiceControl.Infrastructure.Metrics;
 
     class AuditIngestionComponent
     {
         static ILog log = LogManager.GetLogger<AuditIngestionComponent>();
+        static readonly long FrequencyInMilliseconds = Stopwatch.Frequency / 1000;
 
         ImportFailedAudits failedImporter;
         Watchdog watchdog;
@@ -24,8 +27,12 @@
         AuditIngestor ingestor;
         Task ingestionWorker;
         Settings settings;
+        Counter receivedMeter;
+        Meter batchSizeMeter;
+        Meter batchDurationMeter;
 
         public AuditIngestionComponent(
+            Metrics metrics,
             Settings settings,
             IDocumentStore documentStore,
             RawEndpointFactory rawEndpointFactory,
@@ -35,9 +42,18 @@
             AuditIngestionCustomCheck.State ingestionState
         )
         {
+            receivedMeter = metrics.GetCounter("Audit ingestion - received");
+            batchSizeMeter = metrics.GetMeter("Audit ingestion - batch size");
+            var ingestedAuditMeter = metrics.GetCounter("Audit ingestion - ingested audit");
+            var ingestedSagaAuditMeter = metrics.GetCounter("Audit ingestion - ingested saga audit");
+            var auditBulkInsertDurationMeter = metrics.GetMeter("Audit ingestion - audit bulk insert duration", FrequencyInMilliseconds);
+            var sagaAuditBulkInsertDurationMeter = metrics.GetMeter("Audit ingestion - saga audit bulk insert duration", FrequencyInMilliseconds);
+            var bulkInsertCommitDurationMeter = metrics.GetMeter("Audit ingestion - bulk insert commit duration", FrequencyInMilliseconds);
+            batchDurationMeter = metrics.GetMeter("Audit ingestion - batch processing duration", FrequencyInMilliseconds);
+
             this.settings = settings;
             var errorHandlingPolicy = new AuditIngestionFaultPolicy(documentStore, loggingSettings, FailedMessageFactory, OnCriticalError);
-            auditPersister = new AuditPersister(documentStore, bodyStorageEnricher, enrichers);
+            auditPersister = new AuditPersister(documentStore, bodyStorageEnricher, enrichers, ingestedAuditMeter, ingestedSagaAuditMeter, auditBulkInsertDurationMeter, sagaAuditBulkInsertDurationMeter, bulkInsertCommitDurationMeter);
             ingestor = new AuditIngestor(auditPersister, settings);
 
             var ingestion = new AuditIngestion(
@@ -45,6 +61,8 @@
                 {
                     var taskCompletionSource = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
                     messageContext.SetTaskCompletionSource(taskCompletionSource);
+
+                    receivedMeter.Mark();
 
                     await channel.Writer.WriteAsync(messageContext).ConfigureAwait(false);
                     await taskCompletionSource.Task.ConfigureAwait(false);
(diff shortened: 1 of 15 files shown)


seanfarmar (approved):

src/ServiceControl.Audit/Infrastructure/Bootstrapper.cs
```diff
@@ -257,6 +266,8 @@ void RecordStartup(LoggingSettings loggingSettings, EndpointConfiguration endpoi
         TransportSettings transportSettings;
         TransportCustomization transportCustomization;
         static HttpClient httpClient;
+        static ILog metricsLog = LogManager.GetLogger("Metrics");
```
tmasternak, line 269: Could we use some non-default logger or add application settings to make sure that the clients don't get performance metrics printed to the logs by default?

tmasternak (changes requested):

src/ServiceControl.Audit/Infrastructure/Bootstrapper.cs
mikeminutillo: Would that be covered by having a configuration value to turn metrics on (defaults to off)? Given that we have fallback systems to get config values from environment variables, it would allow us to turn it on by default in our development environments.

SzymonPobiega: I added a config setting `PrintMetrics` for this

tmasternak (approved):

SzymonPobiega:
@danielmarbach @mikeminutillo please take a look?

mikeminutillo (approved):

src/ServiceControl.Infrastructure.Metrics/MetricsReporter.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        public void Start()
+        {
+            tokenSource = new CancellationTokenSource();
+            task = Task.Run(async () =>
+            {
+                while (!tokenSource.IsCancellationRequested)
+                {
+                    Print();
+                    await Task.Delay(interval).ConfigureAwait(false);
```
danielmarbach, line 30: should probably accept the cancellation token and try catch for shutdown?

src/ServiceControl.Infrastructure.Metrics/MeterValues.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace ServiceControl.Infrastructure.Metrics
+{
+    public readonly struct MeterValues
+    {
+        public MeterValues(string name, float current, float average15, float average60, float average300)
```
danielmarbach, line 5: The [general guidance for structs](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/choosing-between-class-and-struct) is no more than 16 bytes. This one is 24 bytes. I think it might still be ok though, but we would need to check whether it is copied or not.

src/ServiceControl.Infrastructure.Metrics/Counter.cs
```diff
@@ -0,0 +1,113 @@
...
+            do
+            {
+                initialValue = location;
+                if (initialValue >= comparison)
+                {
+                    previous = -1;
+                    return false;
+                }
+            }
+            while (Interlocked.CompareExchange(ref location, newValue, initialValue) != initialValue);
```
danielmarbach, line 108: Are we confident this always settles the value within a reasonable amount of time? Otherwise, this would require a Spin potentially to not churn CPU

src/ServiceControl.Infrastructure.Metrics/Counter.cs
```diff
@@ -0,0 +1,113 @@
...
+            movingAverageEpochs = new long[300];
+            epoch = DateTime.Now.Minute;
+        }
+
+        public void Mark()
+        {
+            var ticks = Stopwatch.GetTimestamp();
+            var currentEpoch = ticks / Stopwatch.Frequency;
+            var bucketTierIndex = currentEpoch % 2;
+
```
danielmarbach, line 29: 🔥

src/ServiceControl.Infrastructure.Metrics/Meter.cs
```diff
@@ -0,0 +1,138 @@
...
+        {
+            return new Measurement(this);
+        }
+
+        public void Mark(long value)
+        {
+            var ticks = Stopwatch.GetTimestamp();
+            var currentEpoch = ticks / Stopwatch.Frequency;
+            var bucketTierIndex = currentEpoch % 2;
+
```
danielmarbach, line 40: 🔥

src/ServiceControl.Infrastructure.Metrics/Meter.cs
```diff
@@ -0,0 +1,138 @@
...
+            do
+            {
+                initialValue = location;
+                if (initialValue >= comparison)
+                {
+                    previous = -1;
+                    return false;
+                }
+            }
+            while (Interlocked.CompareExchange(ref location, newValue, initialValue) != initialValue);
```
danielmarbach, line 133: Same comment as before

src/ServiceControl.Audit/Auditing/AuditIngestionComponent.cs
```diff
@@ -12,10 +13,12 @@
     using NServiceBus.Logging;
     using NServiceBus.Transport;
     using Raven.Client;
+    using ServiceControl.Infrastructure.Metrics;
 
     class AuditIngestionComponent
     {
         static ILog log = LogManager.GetLogger<AuditIngestionComponent>();
+        static readonly long FrequencyInMilliseconds = Stopwatch.Frequency / 1000;
```
danielmarbach, line 21: Isn't the stopwatch frequency dependent on the resolution of the operating system?

https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch.frequency?view=net-5.0#remarks

so under high-resolution mode this value might differ right?

src/ServiceControl.Audit/Infrastructure/Bootstrapper.cs
```diff
@@ -83,6 +84,10 @@ void Initialize()
             var rawEndpointFactory = new RawEndpointFactory(settings, transportSettings, transportCustomization);
             containerBuilder.RegisterInstance(rawEndpointFactory).AsSelf();
 
+            var metrics = new Metrics();
```
danielmarbach, line 87: Is there any reason why we decided to have our own metrics instead of something like

https://docs.microsoft.com/en-us/dotnet/core/diagnostics/event-counters?

this should also be available on .NET Standard if I'm not mistaken

src/ServiceControl.Audit/Auditing/AuditPersister.cs
```diff
@@ -82,17 +90,29 @@ public async Task<IReadOnlyList<MessageContext>> Persist(List<MessageContext> co
                         {
                             Logger.Debug($"Adding audit message for bulk storage");
                         }
-                        await bulkInsert.StoreAsync(processedMessage).ConfigureAwait(false);
+
+                        using (auditBulkInsertDurationMeter.Measure())
```
danielmarbach, line 94: Given that we mostly only need this for internal purposes right now would it make sense to make this completely opt out from the hot path when not enabled? I think given the risk of CPU churn due to the metrics epoch refresh logic I would be in favor of that

src/ServiceControl.Infrastructure.Metrics/MetricsReporter.cs
SzymonPobiega: There is no cancellation token to pass (from Bootstrapper)

src/ServiceControl.Audit/Auditing/AuditIngestionComponent.cs
SzymonPobiega: Will still be fine. According to the docs it `Gets the frequency of the timer as the number of ticks per second. This field is read-only.` So if I want to get the number of ticks per millisecond, then I just divide by 1000.

src/ServiceControl.Audit/Infrastructure/Bootstrapper.cs
SzymonPobiega: I found the answer. When we started the Raven 5 branch SC was on `net462` so using EventCounters was not an option.

src/ServiceControl.Infrastructure.Metrics/MetricsReporter.cs
danielmarbach: But you have `tokenSource` available which has a token right?

Issue #1758, opened 2019-09-06
Not all headers are visible in SC HTTP API


ramonsmits:
Not all headers are visible in SC HTTP API. These seems to be some filtering applied for unknown reasons.

Missing headers after comparison:

- `$.diagnostics.originating.hostid`
- `$.diagnostics.hostid`
- `$.diagnostics.hostdisplayname`

## Document in SC RavenDB
```json
{
    "UniqueMessageId": "7d43e148-b940-ad3d-1eb4-222cc2e95cf7",
    "MessageMetadata": {
        "SearchableMessageType": "Message1",
        "IsRetried": false,
        "MessageIntent": 1,
        "IsSystemMessage": false,
        "SendingEndpoint": {
            "$type": "ServiceControl.Audit.Monitoring.EndpointDetails, ServiceControl.Audit",
            "Name": "Samples.Azure.ServiceBus.Endpoint1",
            "HostId": "7460bf52-c24a-a831-6c21-00565eacdd97",
            "Host": "ZOLDER"
        },
        "ConversationId": "fb1cf3fb-f6e2-4fe1-b7a0-aac000c8e99d",
        "ReceivingEndpoint": {
            "$type": "ServiceControl.Audit.Monitoring.EndpointDetails, ServiceControl.Audit",
            "Name": "Samples.Azure.ServiceBus.Endpoint2",
            "HostId": "41c0fe1e-45d5-0ae9-b333-e37ddba3d4bf",
            "Host": "ZOLDER"
        },
        "DeliveryTime": "00:00:00.0680880",
        "ContentType": "application/json",
        "MessageType": "Message1",
        "Body": "﻿{\"Property\":\"Hello from Endpoint1\"}",
        "ContentLength": 38,
        "CriticalTime": "00:00:00.0800840",
        "MessageId": "aaafb1b1-dc5d-4ad5-b6f4-aac000c8e99d",
        "ProcessingTime": "00:00:00.0119960",
        "TimeSent": "2019-09-06T12:11:30.0158360Z",
        "BodyUrl": "/messages/aaafb1b1-dc5d-4ad5-b6f4-aac000c8e99d/body"
    },
    "Headers": {
        "NServiceBus.ReplyToAddress": "Samples.Azure.ServiceBus.Endpoint1@default",
        "NServiceBus.MessageId": "aaafb1b1-dc5d-4ad5-b6f4-aac000c8e99d",
        "NServiceBus.MessageIntent": "Send",
[...]

PR #2500, opened 2021-04-13
Check if diagnostic headers are present

seanfarmar:
@ramonsmits and I investigated this as  part of this issue https://github.com/Particular/ServiceControl/issues/1758

From the test it is clear that the headers are dropped when reading from raven client

### DIFF
--- a/src/ServiceControl.UnitTests/CompositeViews/MessagesViewTests.cs
+++ b/src/ServiceControl.UnitTests/CompositeViews/MessagesViewTests.cs
@@ -270,10 +270,31 @@
                         { "OriginatesFromSaga", ""}
                     },
                     Headers = {
-                        { "$.diagnostics.originating.hostid", "41c0fe1e45d50ae9b333e37ddba3d4bf" },
-                        { "$.diagnostics.hostdisplayname", "ZOLDER" },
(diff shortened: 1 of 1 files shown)


mikeminutillo (commented):
It appears as if the $ prefix is used to denote metadata in json and the serializer is stripping it out. I tried to add this to the InMemoryStoreBuilder but it had no effect:

```csharp
store.Conventions.CustomizeJsonSerializer = serializer =>
{
  serializer.MetadataPropertyHandling = MetadataPropertyHandling.Ignore;
};
```

Does this mean that the data is stripped out when we store it in RavenDB or that it isn't returned when we query it? Is there any way for us to retrieve this data? We may want to consider renaming those headers. I don't have any historical context for why they were named with a `$` symbol in the first place and if they aren't working with the Raven API then we likely don't have anyone or anything using them.

seanfarmar:
@mikeminutillo The data is stored in Raven ok, it's only on the read part that it is filtered, maybe it is done on the Raven client?

Might be a good idea to check what is the impact of renaming these headers...

tmasternak:
@seanfarmar I think we can close this as it's been [confirmed](https://github.com/Particular/ServiceControl/issues/1758#issuecomment-847087646) that it was RavenDB removing the headers and we have a release that fixes that.

PR #2501, opened 2021-04-13
Email alerting for custom checks

tmasternak:
This PR provides email notifications for essential custom check status changes in ServiceControl. When a custom check notification is triggered this causes a `SendLocal` operation with details of the email to be sent. When the SMTP server is not available email notifications are dropped if they are success notifications. If there are some failure notifications at least one of them will be sent and we will do our best to append an additional warning line to the notification prompting users to look into ServicePulse for a full list of notifications.

This approach has been chosen as it makes sure that if there are any failure notifications at least one of them will reach the user and at the same time it prevents notification messages piling up in ServiceControl input queue.


src/ServiceControl/Infrastructure/Mail/CustomChecksMailNotification.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        public async Task Handle(CustomCheckFailed domainEvent)
+        {
+            var settings = await LoadSettings().ConfigureAwait(false);
+            if (settings == null || !settings.AlertingEnabled)
+            {
+                return;
+            }
+
+            var smtpClient = CreateSmtpClient(settings);
```
ramonsmits, line 30: Forget `using`

src/ServiceControl/Infrastructure/Mail/CustomChecksMailNotification.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            var mailMessage = new MailMessage(settings.From, settings.To, "Health check failed",
```
ramonsmits, line 32: Forget `using`

src/ServiceControl/Infrastructure/Mail/CustomChecksMailNotification.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        public async Task Handle(CustomCheckSucceeded domainEvent)
+        {
+            var settings = await LoadSettings().ConfigureAwait(false);
+            if (settings == null || !settings.AlertingEnabled)
+            {
+                return;
+            }
+
+            var smtpClient = CreateSmtpClient(settings);
```
ramonsmits, line 63: Forget `using`

src/ServiceControl/Infrastructure/Mail/CustomChecksMailNotification.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            var mailMessage = new MailMessage(settings.From, settings.To, "Health check succeeded",
```
ramonsmits, line 65: Forget `using`

src/ServiceControl/Infrastructure/Mail/CustomChecksMailNotification.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            if (emailDropFolder != null)
+            {
+                return new SmtpClient
+                {
+                    PickupDirectoryLocation = emailDropFolder,
+                    DeliveryMethod = SmtpDeliveryMethod.SpecifiedPickupDirectory
+                };
+            }
+            return new SmtpClient(settings.SmtpServer, settings.SmtpPort ?? 25)
```
ramonsmits, line 48: WHy not use mailsettings in app.config? IMHO this should not be done in code:

```xml
<system.net>
  <mailSettings>
    <smtp from="mail@domain.com">
      <network host="myserver.com" port="25" defaultCredentials="true" />
    </smtp>
  </mailSettings>
</system.net>
```

This way you can support authentication methods.

ramonsmits (changes requested):

src/ServiceControl/Infrastructure/Mail/CustomChecksMailNotification.cs
SzymonPobiega: Good question. We assumed all the config is going to be in RavenDB and available via ServicePulse. @tmasternak what do you think about the configuration-based approach?

tmasternak: I would keep this in the UI. If users want to test if notifications are working they can trigger that from the UI but would not be able to see the settings.

src/ServiceControl/Infrastructure/Mail/CustomChecksMailNotification.cs
SzymonPobiega: Fixed now.

src/ServiceControl/Infrastructure/Mail/CustomChecksMailNotification.cs
SzymonPobiega: Fixed now.

src/ServiceControl/Infrastructure/Mail/CustomChecksMailNotification.cs
SzymonPobiega: Fixed now.

src/ServiceControl/Alerting/Mail/CustomChecksMailNotification.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            var settings = await LoadSettings().ConfigureAwait(false);
+            if (settings == null || !settings.Email.Enabled)
+            {
+                return;
+            }
+
+            await EmailSender.Send(
+                    settings.Email,
+                "Health check failed",
```
mikeminutillo, line 31: Should these mention ServiceControl or is it enough that it will come from a particular address?

Should we put details about the ServiceControl instance here to allow customers with many instances to distinguish them? Alternatively we could ask for a name in the notification settings.

src/ServiceControl/Alerting/Mail/CustomChecksMailNotification.cs
```diff
@@ -0,0 +1,61 @@
+﻿namespace ServiceControl.Alerting.Mail
+{
+    using System.Threading.Tasks;
+    using Alerting;
+    using Contracts.CustomChecks;
+    using Infrastructure.DomainEvents;
+    using Raven.Client;
+    using ServiceBus.Management.Infrastructure.Settings;
+
+    class CustomChecksMailNotification : IDomainHandler<CustomCheckFailed>, IDomainHandler<CustomCheckSucceeded>
```
mikeminutillo, line 10: We might want to be a little defensive in these handlers. Domain event handlers are not isolated from each other and an exception in one will bubble up to where the domain event was raised. If the SMTP server is down, this will disrupt the business transaction that persists the custom check state, and may prevent the event from being published to the transport.

mikeminutillo: They are also run synchronously so the transaction will be delayed if the email server is unreachable.

src/ServiceControl/Alerting/Api/NotificationsController.cs
```diff
@@ -0,0 +1,129 @@
...
+            }
+        }
+
+        [Route("notifications/email/test")]
+        [HttpPost]
+        public async Task<HttpResponseMessage> SendTestEmail()
+        {
+            using (var session = store.OpenAsyncSession())
+            {
+                var settings = await session.LoadAsync<NotificationsSettings>(NotificationsSettings.SingleDocumentId).ConfigureAwait(false);
```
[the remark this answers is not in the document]
tmasternak: done

src/ServiceControl/Alerting/Mail/CustomChecksMailNotification.cs
tmasternak: I've added the name of the service control instance to the subject, and service name and address to the email body.

src/ServiceControl/Alerting/Mail/CustomChecksMailNotification.cs
tmasternak: I've added exception handling and logging around SMTP calls. The send operation is now scheduled as an asynchronous async operation that will not delay the execution of other domain event handlers

src/ServiceControl/Alerting/Api/NotificationsController.cs
```diff
@@ -0,0 +1,129 @@
...
+        [HttpPost]
+        public async Task<HttpResponseMessage> UpdateSettings(UpdateAlertingSettingsRequest request)
+        {
+            using (var session = store.OpenAsyncSession())
+            {
+                var settings = await LoadSettings(session).ConfigureAwait(false);
+
+                var emailSettings = settings.Email;
+
+                emailSettings.SmtpServer = request.SmtpServer;
```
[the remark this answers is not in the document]
tmasternak: I'm not sure what we could do here. Currently, Service Pulse has "Test" button to trigger a test email notification which should mitigate problem of entering invalid options at least to some degree.

src/ServiceControl/Alerting/Mail/CustomChecksMailNotification.cs
tmasternak: After more consideration, we went with [this](https://github.com/Particular/ServiceControl/pull/2501#issue-614492658) design. I think it's a good enough compromise.

seanfarmar (approved):

src/ServiceControl/Notifications/Email/SendEmailNotificationHandler.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            using (var session = store.OpenAsyncSession())
+            {
+                notifications = await session.LoadAsync<NotificationsSettings>(NotificationsSettings.SingleDocumentId)
+                    .ConfigureAwait(false);
+            }
+
+            if (notifications == null || !notifications.Email.Enabled)
+            {
+                log.Info("Skipping email sending. Notifications turned-off.");
```
mikeminutillo, line 36: I was going to argue that this should be logged at debug, but this is the case where we have in-flight messages for sending emails and then we turn notifications off, right? if so, that makes sense to log at this level.

src/ServiceControl/Notifications/Email/SendEmailNotificationHandler.cs
```diff
@@ -0,0 +1,115 @@
...
+                {
+                    throttlingState.ThrottlingOn();
+
+                    await Task.Delay(throttlingDelay).ConfigureAwait(false);
+
+                    throttlingState.ThrottlingOff();
+
+                    throttlingState.RetriedMessageId = context.MessageId;
+
+                    throw new EmailNotificationException(e);
```
mikeminutillo, line 76: If I am reading this right, when an exception occurs, we turn throttling on which causes all pending operations to simply stop processing, consuming their message. Any new requests that come in during that time will also get dropped without any attempt to send an email. After a while we turn throttling off but then throw an exception, forcing this message to be retried. Why is this email notification more important than all of the ones we dropped? Why not just drop this one?

tmasternak: First, thanks for a great write-up on what is happening in terms of throttling.

The reason for this behavior is the following. We want to make sure that a user gets at least one failure notification even if the SMTP is down - that's why we throw `EmailNotificationException` and always do immediate retires for this exception. At the same we want to prevent messages from piling up in ServiceControl input queue - that's why we drop all messages, except one for which an exception was thrown, when in throttling mode.

The main UX argument for this behavior is that all the notifications are important enough that the users are very likely to look into ServicePulse when they any. At that point, they will see all the other failures for which we potentially dropped the notification. To make it a bit more visible to the user email body gets an additional line that prompts to look into ServicePulse for other notifications when there were some SMTP connectivity issues..

src/ServiceControl/Notifications/Email/CustomChecksMailNotification.cs
```diff
@@ -0,0 +1,80 @@
...
+            {
+                serviceControlHealthCustomCheckIds = settings.NotificationsFilter.Split(new[] { '#' }, StringSplitOptions.RemoveEmptyEntries);
+            }
+        }
+
+        public Task Handle(CustomCheckFailed domainEvent)
+        {
+            if (IsHealthCheck(domainEvent.CustomCheckId))
+            {
+                return messageSession.SendLocal(new SendEmailNotification
```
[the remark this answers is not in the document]
tmasternak: Agreed.

src/ServiceControl/Notifications/Email/CustomChecksMailNotification.cs
```diff
@@ -0,0 +1,80 @@
...
+    using Infrastructure.DomainEvents;
+    using NServiceBus;
+    using ServiceBus.Management.Infrastructure.Settings;
+
+    class CustomChecksMailNotification : IDomainHandler<CustomCheckFailed>, IDomainHandler<CustomCheckSucceeded>
+    {
+        readonly IMessageSession messageSession;
+        readonly string instanceName;
+        string instanceAddress;
+        string[] serviceControlHealthCustomCheckIds = {
```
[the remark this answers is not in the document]
tmasternak: These are good points. What I was planning to do is to have a dedicated PR that aligns Custom Check Categories and does the filtering based on that value. I'll make sure that we do escape for the separator in the configuration so that there are no disallowed values.

That said I propose to have that in a dedicated PR.

src/ServiceControl/Infrastructure/RavenDB/RavenBootstrapper.cs
```diff
@@ -110,6 +110,8 @@ public void StartRaven(EmbeddableDocumentStore documentStore, Settings settings,
                 // Only purge endpoints when not in maintenance mode
                 PurgeKnownEndpointsWithTemporaryIdsThatAreDuplicate(documentStore);
             }
+
+            settings.InitializeStore?.Invoke(documentStore);
```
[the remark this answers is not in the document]
tmasternak: Agreed. I'll cherry-pick the changes to this PR.

SzymonPobiega (approved):

src/ServiceControl.AcceptanceTests/Monitoring/CustomChecks/When_email_notifications_are_enabled.cs
```diff
@@ -0,0 +1,104 @@
...
+    using Notifications;
+    using NServiceBus;
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.CustomChecks;
+    using NUnit.Framework;
+    using ServiceBus.Management.Infrastructure.Settings;
+    using TestSupport.EndpointTemplates;
+
+    [TestFixture]
+    [RunOnAllTransports]
```
danielmarbach, line 17: Is it worth running this on all transports?

src/ServiceControl.AcceptanceTests/Monitoring/CustomChecks/When_email_notifications_are_enabled.cs
```diff
@@ -0,0 +1,104 @@
...
+                .WithEndpoint<EndpointWithFailingCustomCheck>()
+                .Done(c =>
+                {
+                    emails = Directory.EnumerateFiles(emailDropPath).ToArray();
+                    return emails.Length > 0;
+                })
+                .Run();
+
+
+
```
danielmarbach, line 66: 🔥

src/ServiceControl.AcceptanceTests/Monitoring/CustomChecks/When_email_notifications_are_enabled.cs
```diff
@@ -0,0 +1,104 @@
...
+            Assert.True(emails.Length > 0);
```
danielmarbach, line 67: Maybe use collection asserts here or use a string message to indicate what was wrong should this fail?

src/ServiceControl.UnitTests/ApprovalFiles/APIApprovals.PublicClr.approved.txt
```diff
@@ -70,8 +71,10 @@ namespace ServiceBus.Management.Infrastructure.Settings
         public string Hostname { get; }
         public int HttpDefaultConnectionLimit { get; set; }
         public bool IngestErrorMessages { get; set; }
+        public System.Action<Raven.Client.IDocumentStore> InitializeStore { get; set; }
```
danielmarbach, line 74: This one should be dropped now that we have the async one

src/ServiceControl/Infrastructure/RavenDB/RavenBootstrapper.cs
danielmarbach, line 114: This one should be dropped now that we have the async one

src/ServiceControl/Infrastructure/Settings/Settings.cs
```diff
@@ -46,18 +47,26 @@ public Settings(string serviceName = null)
...
+        public string NotificationsFilter { get; set; }
+
         public bool AllowMessageEditing { get; set; }
 
         public Func<string, Dictionary<string, string>, byte[], Func<Task>, Task> OnMessage { get; set; } = (messageId, headers, body, next) => next();
+        public Func<IDocumentStore, Task> StoreInitializer { get; set; } = _ => Task.CompletedTask;
 
         public bool RunInMemory { get; set; }
 
+        public Action<IDocumentStore> InitializeStore { get; set; }
```
danielmarbach, line 66: This one should be dropped now that we have the async one

src/ServiceControl/Notifications/Api/NotificationsController.cs
```diff
@@ -0,0 +1,135 @@
...
+            {
+                var settings = await LoadSettings(session).ConfigureAwait(false);
+
+                return new JsonResult<EmailNotifications>(
+                    settings.Email,
+                    new JsonSerializerSettings
+                    {
+                        NullValueHandling = NullValueHandling.Include,
+                        ContractResolver = new UnderscoreMappingResolver()
+                    },
```
danielmarbach, line 38: Don't we have some default json settings that we can reuse? Or make it at least static on the controller?

src/ServiceControl/Notifications/Api/NotificationsController.cs
```diff
@@ -0,0 +1,135 @@
...
+                            settings.Email,
+                            $"[{instanceName}] health check notification check successful",
+                            $"[{instanceName}] health check notification check successful.")
+                        .ConfigureAwait(false);
+                }
+                catch (Exception e)
+                {
+                    return new HttpResponseMessage(HttpStatusCode.InternalServerError)
+                    {
+                        Content = new StringContent($"{e.Message} {e.InnerException?.Message}"),
```
danielmarbach, line 106: Is it ever a good idea to use the exception message and expose that to the user?

src/ServiceControl/Notifications/Email/CustomChecksMailNotification.cs
```diff
@@ -0,0 +1,80 @@
...
+        public CustomChecksMailNotification(IMessageSession messageSession, Settings settings)
+        {
+            this.messageSession = messageSession;
+
+            instanceName = settings.ServiceName;
+            instanceAddress = settings.ApiUrl;
+
+            if (string.IsNullOrWhiteSpace(settings.NotificationsFilter) == false)
+            {
+                serviceControlHealthCustomCheckIds = settings.NotificationsFilter.Split(new[] { '#' }, StringSplitOptions.RemoveEmptyEntries);
```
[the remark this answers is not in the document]
danielmarbach: Could be a static array

src/ServiceControl/Notifications/Email/EmailSender.cs
```diff
@@ -0,0 +1,58 @@
...
+
+    class EmailSender
+    {
+        public static async Task Send(EmailNotifications settings, string subject, string body, string emailDropFolder = null)
+        {
+            try
+            {
+                using (var client = CreateSmtpClient(settings, emailDropFolder))
+                {
+                    using (var mailMessage = new MailMessage(settings.From, settings.To, subject, body))
```
danielmarbach, line 17: using could be on the same level as outer to reduce nesting for better readability

src/ServiceControl/Notifications/Email/EmailSender.cs
```diff
@@ -0,0 +1,58 @@
...
+                log.Warn("Failure sending email.", e);
+                throw;
+            }
+        }
+
+        static SmtpClient CreateSmtpClient(EmailNotifications settings, string emailDropFolder)
+        {
+            if (emailDropFolder != null)
+            {
+                return new SmtpClient
```
danielmarbach, line 34: What do you think about this?

https://docs.microsoft.com/en-us/dotnet/api/system.net.mail.smtpclient?redirectedfrom=MSDN&view=net-5.0#remarks

image

src/ServiceControl/Notifications/Email/EmailSender.cs
```diff
@@ -0,0 +1,58 @@
...
+                {
+                    PickupDirectoryLocation = emailDropFolder,
+                    DeliveryMethod = SmtpDeliveryMethod.SpecifiedPickupDirectory
+                };
+            }
+
+            var smtpClient = new SmtpClient(settings.SmtpServer, settings.SmtpPort ?? 25)
```
danielmarbach, line 41: Why not have a client that is created when needed, reused until the settings change and disposed when they do or SC is shutdown?

>The connection established by the current instance of the SmtpClient class to the SMTP server may be re-used if an application wishes to send multiple messages to the same SMTP server. This is particularly useful when authentication or encryption are used establish a connection to the SMTP server. The process of authenticating and establishing a TLS session can be expensive operations. A requirement to re-establish a connection for each message when sending a large quantity of email to the same SMTP server could have a significant impact on performance. There are a number of high-volume email applications that send email status updates, newsletter distributions, or email alerts. Also many email client applications support an off-line mode where users can compose many email messages that are sent later when a connection to the SMTP server is established. It is typical for an email client to send all SMTP messages to a specific SMTP server (provided by the Internet service provider) that then forwards this email to other SMTP servers.

> The SmtpClient class implementation pools SMTP connections so that it can avoid the overhead of re-establishing a connection for every message to the same server. An application may re-use the same SmtpClient object to send many different emails to the same SMTP server and to many different SMTP servers.

https://docs.microsoft.com/en-us/dotnet/api/system.net.mail.smtpclient

src/ServiceControl/Notifications/Email/SendEmailNotificationHandler.cs
```diff
@@ -0,0 +1,115 @@
...
+            this.throttlingState = throttlingState;
+
+            emailDropFolder = settings.EmailDropFolder;
+        }
+
+        public async Task Handle(SendEmailNotification message, IMessageHandlerContext context)
+        {
+            NotificationsSettings notifications;
+
+            using (var session = store.OpenAsyncSession())
```
danielmarbach, line 28: would this be a good case for https://ravendb.net/docs/article-page/3.5/csharp/client-api/how-to/setup-aggressive-caching ? Especially because `SaveChangesAsync` will invalidate the cache https://ravendb.net/docs/article-page/3.5/csharp/client-api/how-to/setup-aggressive-caching#invalidate-on-savechanges

src/ServiceControl/Notifications/Email/SendEmailNotificationHandler.cs
```diff
@@ -0,0 +1,115 @@
...
+                await EmailSender.Send(notifications.Email, message.Subject, message.Body, emailDropFolder)
+                    .ConfigureAwait(false);
+            }
+            catch (Exception e) when (!(e is OperationCanceledException))
+            {
+                if (message.IsFailure)
+                {
+                    throttlingState.ThrottlingOn();
+
+                    await Task.Delay(throttlingDelay).ConfigureAwait(false);
```
danielmarbach, line 70: Wouldn't this require a cancellation token for shutdown scenarios to make sure we don't hang in there? Also aren't we risking running into message timeout scenarios? For example, the default visibility timeout for SQS is 30 seconds.

src/ServiceControl/Notifications/Email/SendEmailNotificationHandler.cs
```diff
@@ -0,0 +1,115 @@
...
+            {
+                log.Info("Skipping email sending. Notifications turned-off.");
+                return;
+            }
+
+            var hasSemaphore = false;
+
+            try
+            {
+                while (hasSemaphore == false)
```
danielmarbach, line 44: Aren't we risking running into message timeout scenarios? For example, the default visibility timeout for SQS is 30 seconds.

src/ServiceControl/Notifications/Email/EmailSender.cs
SzymonPobiega: @tmasternak raised it in the beginning but I called yagni on this based on the github link from that warning https://github.com/dotnet/platform-compat/blob/master/docs/DE0005.md. It says it is OK to use it for occasional emails from tooling.

src/ServiceControl/Notifications/Api/NotificationsController.cs
SzymonPobiega: We have a media type registration in `Startup.cs`:

```C#
var jsonMediaTypeFormatter = config.Formatters.JsonFormatter;
                jsonMediaTypeFormatter.SerializerSettings = JsonNetSerializerSettings.CreateDefault();
                jsonMediaTypeFormatter.SupportedMediaTypes.Add(new MediaTypeHeaderValue("application/vnd.particular.1+json"));
```
@tmasternak isn't that enough?

tmasternak: No, it's now enough as the default serializer does not align with the ServicePulse conventions we use for naming variables.

src/ServiceControl/Notifications/Api/NotificationsController.cs
tmasternak: These are details that are shown to the user testing SMTP connectivity and include pretty descriptive information like "Could not reach server" or "Server certificate is not valid" etc.

src/ServiceControl/Notifications/Email/EmailSender.cs
tmasternak: done

src/ServiceControl/Notifications/Email/EmailSender.cs
tmasternak: As discussed, this is meant to be rarely executed. Secondly, based on test it looks that there is some caching going on behind the scenes as 2nd and follow-up connections are considerably quicker than the 1st one.

src/ServiceControl/Notifications/Email/SendEmailNotificationHandler.cs
tmasternak: good points. I've added cancellation token support and decreased the delay to 30 sec

src/ServiceControl/Notifications/Email/SendEmailNotificationHandler.cs
tmasternak: done

src/ServiceControl/Notifications/Email/CustomChecksMailNotification.cs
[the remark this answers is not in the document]
tmasternak: I've added better parsing so that no character is restricted. '#' is a delimiter but when doubled `##` is treated as non-delimiting `#`. I've added tests to cover that.

tmasternak:
@mikeminutillo I think I addressed all the comments.

tmasternak:
@mikeminutillo I think that I've covered all the comments.

mikeminutillo (approved):

PR #2507, opened 2021-04-19
How ServiceControl Retries Works improvements

ramonsmits:


docs/bulk-retries-design.md
```diff
@@ -18,22 +18,33 @@ When ServiceControl restarts, the in-memory representations are rebuilt by aggre
...
+When ServiceControl starts up it will attempt to adopt (what does adopt mean?) any batches that it finds in this status and move them to `Staging`. This can only happen if the ServiceControl process stops during the above process. Any documents that were already marked will be added to the batch. Any documents that had not yet been marked are ignored and will have to be retried again by the user. (<- what does this mean?) When ServiceControl starts up, it will generate a `Session ID` GUID. This GUID is stamped onto each new batch as it is created. This is how ServiceControl can tell if a batch is from a previous session and adopt it. Only Batches with a non-current session Id will be adopted by the orphan batch process.
```
mikeminutillo, line 25: > what does adopt mean?

We may be missing a concept here. When ServiceControl starts it will look for documents marked with a retry batch id, without a matching retry batch document. i.e. We started making the batch but we never completed it. We call retry batches in this state "orphaned" because the process that was assembling them has been lost.

When we "adopt" and orphaned batch, we take over the processing of it. We find all of the messages with the same missing retry batch id and create a retry batch with that id for them. This new retry batch document will then get picked up and processed like any other.

> Any documents that had not yet been marked are ignored and will have to be retried again by the user. (<- what does this mean?)

When you issue a bulk retry we start a background process that looks  for all of the messages that match your criteria and we assign them a retry batch id (in lots of 1000). If you kill the process before it is done, the details of what you were trying to retry are gone. Any documents we had already marked for retry will be retried. Any that were not marked for retry will require manual intervention to retry

docs/bulk-retries-design.md
```diff
@@ -18,22 +18,33 @@ When ServiceControl restarts, the in-memory representations are rebuilt by aggre
...
+
+TODO: Not very clear what "adoption" does and what unmarked message are being ignored means.
 
 ### Staging
 
 When a batch enters this state, it means that failed messages belonging to the batch are being added to a special `staging` queue. This queue is used during `Forwarding` to ensure that messages are sent back to their original destination transactionally (using the transports recieve transaction).
 
 NOTE: Although multiple batches may be in `Staging` or `Forwarding` at a time, these are processed by a single thread ensuring that the rest of the process is serialized. This is important to ensure that only one batch at a time has access to the `staging` queue. Batches in `Staging` will only be processed if there are no batches in `Forwarding`.
 
+TODO: What value does the staging queue have when transaction mode is ReceiveOnly?
```
mikeminutillo, line 37: The staging queue typically hass a transaction mode of `SendsAtomicWithReceive` on any transport that supports it. For those that don't (ASQ, RabbitMQ, SQS) it doesn't provide a lot of value. Perhaps we could consider introducing an alternate retry mechanism for those.

docs/bulk-retries-design.md
```diff
@@ -18,22 +18,33 @@ When ServiceControl restarts, the in-memory representations are rebuilt by aggre
...
+
+TODO: What happens might forwarding a specific batch always keep failing?
```
mikeminutillo, line 39: If a message fails to be forwarded it is removed from the batch. If the batch contains no messages, then it is marked as complete.

docs/bulk-retries-design.md
```diff
@@ -18,22 +18,33 @@ When ServiceControl restarts, the in-memory representations are rebuilt by aggre
...
+
 When a batch is selected for staging a new `Staging Id` is generated for it and the list of failed messages belonging to the batch is loaded. At this time, if a message had been snagged by another batch, it gets filtered out.
 
+TODO: Aren't messages selected during the Marking stage?
+
 Each message, one at a time, is dispatched to the `staging` queue. As each message is dispatched:
-1. It's Raven document is updated to reflect that it has entered `RetryIssued` mode.
+
+1. It's Raven document (which document, FailedMessageRetry?) is updated to reflect that it has entered `RetryIssued` mode.
```
mikeminutillo, line 47: No. The `Failedmessage` document.

docs/bulk-retries-design.md
```diff
@@ -18,22 +18,33 @@ When ServiceControl restarts, the in-memory representations are rebuilt by aggre
...
 
 NOTE: Although multiple batches may be in `Staging` or `Forwarding` at a time, these are processed by a single thread ensuring that the rest of the process is serialized. This is important to ensure that only one batch at a time has access to the `staging` queue. Batches in `Staging` will only be processed if there are no batches in `Forwarding`.
 
+TODO: What value does the staging queue have when transaction mode is ReceiveOnly?
+
+TODO: What happens might forwarding a specific batch always keep failing?
+
 When a batch is selected for staging a new `Staging Id` is generated for it and the list of failed messages belonging to the batch is loaded. At this time, if a message had been snagged by another batch, it gets filtered out.
 
+TODO: Aren't messages selected during the Marking stage?
```
mikeminutillo, line 43: Yes. They are.

I don't have the context of why this was introduced. It would require deeper analysis.

docs/bulk-retries-design.md
```diff
@@ -53,21 +65,26 @@ Because each message send is done in the context of a Satellite Receive operatio
 
 If there is a message in the `Forwarding` status when ServiceControl starts, then we don't know how many messages there are still in the staging queue to send. To counter this we turn on the satellite in Non-Counting mode. The idea for this is that the satellite will run until the queue is empty. Unfortunately there is nothing built into the Transport abstraction that allows us to query this so SC assumes that if it does not see any new messages from the `staging` queue within 45 seconds then it is empty.
 
+TODO: What happend when forwarding fails? Transport error, network error? What happens in the edge case then forwarding completes and we crash? Restart and finds an empty stating queue and times out?
```
mikeminutillo, line 68: This is one benefit of the staging queue. It makes it more likely that if we can reach the broker to receive the message, then we can reach the broker to send it on.

> What happend when forwarding fails? Transport error, network error?

If forwarding fails for a message, then we count it, mark it as unresolved, and remove it from the batch. You should see warnings in the log that look like this:

```
Failed to send UNIQUE-MESSAGE-ID message to DESTINATION for retry. Attempting to revert message status to unresoved so it can be tried again.
```
The error information will be attached.

> What happens in the edge case then forwarding completes and we crash? Restart and finds an empty stating queue and times out?

Yes.

docs/bulk-retries-design.md
```diff
@@ -18,22 +18,33 @@ When ServiceControl restarts, the in-memory representations are rebuilt by aggre
...
+When ServiceControl starts up it will attempt to adopt (what does adopt mean?) any batches that it finds in this status and move them to `Staging`. This can only happen if the ServiceControl process stops during the above process. Any documents that were already marked will be added to the batch. Any documents that had not yet been marked are ignored and will have to be retried again by the user. (<- what does this mean?) When ServiceControl starts up, it will generate a `Session ID` GUID. This GUID is stamped onto each new batch as it is created. This is how ServiceControl can tell if a batch is from a previous session and adopt it. Only Batches with a non-current session Id will be adopted by the orphan batch process.
+
+TODO: How is selected which message is part of a retry batch? If that done before the retry batch is created (try to add 1,000 messages, if some are part of another retry batch just skip? Or b, the retry batch itself tried to create FailedMessageRetry documents until either no more messages or it reaches 1,000? Based on line 23 I think the last? )
+
+TODO: Not very clear what "adoption" does and what unmarked message are being ignored means.
 
 ### Staging
```
ramonsmits, line 32:
```suggestion

Staging ensures transactional exactly-once processing. Transports that do not support `SendsAtomicWithReceive` strictly would not need staging to prevent more-than-once delivery.
```

docs/bulk-retries-design.md
ramonsmits: @mikeminutillo I've created https://github.com/Particular/ServiceControl/issues/2558

docs/bulk-retries-design.md
```diff
@@ -18,22 +18,33 @@ When ServiceControl restarts, the in-memory representations are rebuilt by aggre
...
+When ServiceControl starts up it will attempt to adopt (what does adopt mean?) any batches that it finds in this status and move them to `Staging`. This can only happen if the ServiceControl process stops during the above process. Any documents that were already marked will be added to the batch. Any documents that had not yet been marked are ignored and will have to be retried again by the user. (<- what does this mean?) When ServiceControl starts up, it will generate a `Session ID` GUID. This GUID is stamped onto each new batch as it is created. This is how ServiceControl can tell if a batch is from a previous session and adopt it. Only Batches with a non-current session Id will be adopted by the orphan batch process.
+
```
ramonsmits, line 26:
```suggestion
## Recovering retry batches

When ServiceControl starts it will look for documents marked with a retry batch id, without a matching retry batch document (i.e. We started making the batch but we never completed it). Retry batches in this state as called "orphaned" because the process that was assembling them has been lost.

When we "adopt" and orphaned batch, we take over the processing of it. We find all of the messages with the same missing retry batch id and create a retry batch with that id for them. This new retry batch document will then get picked up and processed like any other.

File: docs/bulk-retries-design.md (hunk 1 above)
Comment (ramonsmits) [inline docs/bulk-retries-design.md:25]:
```suggestion
When ServiceControl starts up it will attempt to adopt any batches that it finds in this status and move them to `Staging`. This can only happen if the ServiceControl process stops during the above process. Any documents that were already marked will be added to the batch. Any documents that had not yet been marked are ignored and will have to be retried again by the user.

When a bulk retry is issued a background process is started that looks for all of the messages that match its criteria and assign them to a retry batch id (in lots of 1,000). If the process is killed before it is done, the details of what you were trying to retry are gone. Any documents that had already marked for retry will be retried. Any that were not marked for retry will require manual intervention to retry

When ServiceControl starts up, it will generate a `Session ID` GUID. This GUID is stamped onto each new batch as it is created. This is how ServiceControl can tell if a batch is from a previous session and adopt it. Only Batches with a non-current session Id will be adopted by the orphan batch process.
```

docs/bulk-retries-design.md
```diff
@@ -18,22 +18,33 @@ When ServiceControl restarts, the in-memory representations are rebuilt by aggre
...
+TODO: How is selected which message is part of a retry batch? If that done before the retry batch is created (try to add 1,000 messages, if some are part of another retry batch just skip? Or b, the retry batch itself tried to create FailedMessageRetry documents until either no more messages or it reaches 1,000? Based on line 23 I think the last? )
+
+TODO: Not very clear what "adoption" does and what unmarked message are being ignored means.
```
ramonsmits, line 37:
```suggestion
```

docs/bulk-retries-design.md
ramonsmits: @WilliamBZA might you know?

WilliamBZA: Probably something to do with not having a cutoff period when group retries were introduced. I'm not actually sure.

I do think the entire marking process needs an overhaul.

docs/bulk-retries-design.md
```diff
@@ -18,22 +18,38 @@ When ServiceControl restarts, the in-memory representations are rebuilt by aggre
...
 When a batch enters this state, it means that failed messages belonging to the batch are being added to a special `staging` queue. This queue is used during `Forwarding` to ensure that messages are sent back to their original destination transactionally (using the transports recieve transaction).
 
 NOTE: Although multiple batches may be in `Staging` or `Forwarding` at a time, these are processed by a single thread ensuring that the rest of the process is serialized. This is important to ensure that only one batch at a time has access to the `staging` queue. Batches in `Staging` will only be processed if there are no batches in `Forwarding`.
 
+If a message fails to be forwarded it is removed from the batch. If the batch contains no messages, then it is marked as complete.
+
 When a batch is selected for staging a new `Staging Id` is generated for it and the list of failed messages belonging to the batch is loaded. At this time, if a message had been snagged by another batch, it gets filtered out.
 
+TODO: Aren't messages selected during the Marking stage?
+
```
ramonsmits, line 49:
```suggestion
```

docs/bulk-retries-design.md
WilliamBZA: I do not know

ramonsmits:
@WilliamBZA @mikeminutillo can either of you approve so the changes can get merged?

docs/bulk-retries-design.md
```diff
@@ -18,22 +18,36 @@ When ServiceControl restarts, the in-memory representations are rebuilt by aggre
...
+When ServiceControl starts up it will attempt to adopt any batches that it finds in this status and move them to `Staging`. This can only happen if the ServiceControl process stops during the above process. Any documents that were already marked will be added to the batch. Any documents that had not yet been marked are ignored and will have to be retried again by the user.
+
+When a bulk retry is issued a background process is started that looks for all of the messages that match its criteria and assign them to a retry batch id (in lots of 1,000). If the process is killed before it is done, the details of what you were trying to retry are gone. Any documents that had already marked for retry will be retried. Any that were not marked for retry will require manual intervention to retry
+
+When ServiceControl starts up, it will generate a `Session ID` GUID. This GUID is stamped onto each new batch as it is created. This is how ServiceControl can tell if a batch is from a previous session and adopt it. Only Batches with a non-current session Id will be adopted by the orphan batch process.
+## Recovering retry batches
```
WilliamBZA, line 30:
```suggestion

## Recovering retry batches
```

docs/bulk-retries-design.md
```diff
@@ -18,22 +18,36 @@ When ServiceControl restarts, the in-memory representations are rebuilt by aggre
...
+
+When ServiceControl starts it will look for documents marked with a retry batch id, without a matching retry batch document (i.e. We started making the batch but we never completed it). Retry batches in this state as called "orphaned" because the process that was assembling them has been lost.
+
+When we "adopt" and orphaned batch, we take over the processing of it. We find all of the messages with the same missing retry batch id and create a retry batch with that id for them. This new retry batch document will then get picked up and processed like any other.
```
WilliamBZA, line 34:
```suggestion
When an instance of ServiceControl "adopts" an orphaned batch, it takes over the processing of it. That instance will find all of the messages with the same missing retry batch id and create a retry batch with that id for them. This new retry batch document will then get picked up and processed like any other.
```

docs/bulk-retries-design.md
```diff
@@ -18,22 +18,36 @@ When ServiceControl restarts, the in-memory representations are rebuilt by aggre
...
+TODO: How is selected which message is part of a retry batch? If that done before the retry batch is created (try to add 1,000 messages, if some are part of another retry batch just skip? Or b, the retry batch itself tried to create FailedMessageRetry documents until either no more messages or it reaches 1,000? Based on line 23 I think the last? )
```
WilliamBZA, line 35:
```suggestion
```

docs/bulk-retries-design.md
```diff
@@ -18,22 +18,36 @@ When ServiceControl restarts, the in-memory representations are rebuilt by aggre
...
 
 ### Staging
 
+Staging ensures transactional exactly-once processing. Transports that do not support `SendsAtomicWithReceive` strictly would not need staging to prevent more-than-once delivery.
```
WilliamBZA, line 39:
```suggestion
Staging reduces the chance of sending a message for retry more than once. Transports that do not support `SendsAtomicWithReceive` strictly would not need staging to prevent more-than-once delivery.
```

docs/bulk-retries-design.md
```diff
@@ -18,22 +18,36 @@ When ServiceControl restarts, the in-memory representations are rebuilt by aggre
...
 NOTE: Although multiple batches may be in `Staging` or `Forwarding` at a time, these are processed by a single thread ensuring that the rest of the process is serialized. This is important to ensure that only one batch at a time has access to the `staging` queue. Batches in `Staging` will only be processed if there are no batches in `Forwarding`.
 
+If a message fails to be forwarded it is removed from the batch. If the batch contains no messages, then it is marked as complete.
+
 When a batch is selected for staging a new `Staging Id` is generated for it and the list of failed messages belonging to the batch is loaded. At this time, if a message had been snagged by another batch, it gets filtered out.
 
 Each message, one at a time, is dispatched to the `staging` queue. As each message is dispatched:
-1. It's Raven document is updated to reflect that it has entered `RetryIssued` mode.
+
+1. It's Raven `Failedmessage` document is updated to reflect that it has entered `RetryIssued` mode.
```
WilliamBZA, line 50:
```suggestion
1. The corresponding `Failedmessage` document is updated to reflect that it has entered `RetryIssued` mode.
```

docs/bulk-retries-design.md
```diff
@@ -53,21 +68,30 @@ Because each message send is done in the context of a Satellite Receive operatio
...
+> Failed to send UNIQUE-MESSAGE-ID message to DESTINATION for retry. Attempting to revert message status to unresoved so it can be tried again.
+
+Might SC run in the edge case that forwarding completes and immediately crash then after restart it finds an empty staging queue and times out.
+
 ### Done
+
 There is no status to indicate that a batch is `Done`. When the `Forwarding` status is completed, the batch is deleted as it is no longer relevant. Note that each message that was retried as a part of the batch still have a corresponding `FailedMessageRetry/{messageId}` document. This will prevent the message from being retried again.
 
-Once all batches for a retry operation complete, we add two enties into a retry history document. An "unacknowledged" entry is kept until a user acknowledges the completion of the operation, while the other is used to show users a historic retry operations.
+Once all batches for a retry operation complete, we add two enties into a retry history document. An "unacknowledged" entry is kept until a user acknowledges the completion of the operation, while the other is used to show users the historic retry operations.
```
WilliamBZA, line 81:
```suggestion
Once all batches for a retry operation complete, we add two entries into a retry history document. An "unacknowledged" entry is kept until a user acknowledges the completion of the operation, while the other is used to show users the historic retry operations.
```

WilliamBZA (changes requested):

docs/bulk-retries-design.md
```diff
@@ -53,21 +68,30 @@ Because each message send is done in the context of a Satellite Receive operatio
 
 If there is a message in the `Forwarding` status when ServiceControl starts, then we don't know how many messages there are still in the staging queue to send. To counter this we turn on the satellite in Non-Counting mode. The idea for this is that the satellite will run until the queue is empty. Unfortunately there is nothing built into the Transport abstraction that allows us to query this so SC assumes that if it does not see any new messages from the `staging` queue within 45 seconds then it is empty.
 
+If forwarding a specific message fail, then we count it, mark it as unresolved, and remove it from the batch. You should see warnings in the log that look like this with the error information attached:
+
+> Failed to send UNIQUE-MESSAGE-ID message to DESTINATION for retry. Attempting to revert message status to unresoved so it can be tried again.
+
+Might SC run in the edge case that forwarding completes and immediately crash then after restart it finds an empty staging queue and times out.
```
mikeminutillo, line 75:
```suggestion
If forwarding completes and ServiceControl immediately crashes then after restart it finds an empty staging queue and times out.
```

mikeminutillo (approved):

WilliamBZA (approved):

PR #2508, opened 2021-04-20
Review suggestions for metrics

SzymonPobiega:
- do not execute Mark is disabled
 - removed some extra new lines

### DIFF
--- a/src/ServiceControl.Infrastructure.Metrics/Counter.cs
+++ b/src/ServiceControl.Infrastructure.Metrics/Counter.cs
@@ -7,14 +7,16 @@
     public class Counter
     {
         readonly string name;
+        readonly Func<bool> enabled;
         readonly int[] eventsPerSecond;
         readonly int[] movingAverage;
         readonly long[] movingAverageEpochs;
         long epoch;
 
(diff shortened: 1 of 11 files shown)


src/ServiceControl.Infrastructure.Metrics/Counter.cs
```diff
@@ -7,14 +7,16 @@
...
     {
         readonly string name;
+        readonly Func<bool> enabled;
         readonly int[] eventsPerSecond;
         readonly int[] movingAverage;
         readonly long[] movingAverageEpochs;
         long epoch;
 
-        public Counter(string name)
+        internal Counter(string name, Func<bool> enabled)
```
danielmarbach, line 16: Do we support hot reloading of configuration? If No I'd say it would be better to just make this a bool. Otherwise we will always have a closure allocation due to the indirection

SzymonPobiega: 👍 I fixed it.

danielmarbach (approved):
Other than the failing approvals all good

PR #2511, opened 2021-04-21
Custom check to check for RavenDB index errors and reports errors as log entry and custom check failure

ramonsmits:


tmasternak (approved):

danielmarbach:
@ramonsmits and I had a chat and agreed we will be turning this check into a custom check for now to make it less invasive for now until we get a better understanding of when indexes go into the error state

mikeminutillo (approved):

tmasternak (approved):

src/ServiceControl.Audit/Infrastructure/CheckRavenDBIndexErrors.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            _store = store;
+        }
+
+        public override Task<CheckResult> PerformCheck()
+        {
+            var statistics = _store.DatabaseCommands.GetStatistics();
+            var indexes = statistics.Indexes.OrderBy(x => x.Name).ToArray();
+
+            if (statistics.Errors.Length <= 0)
```
SzymonPobiega, line 24:
```suggestion
            if (statistics.Errors.Length = 0)
```

Nitpicking but it feels odd if length is negative

src/ServiceControl/Operations/CheckRavenDBIndexErrors.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            _store = store;
+        }
+
+        public override Task<CheckResult> PerformCheck()
+        {
+            var statistics = _store.DatabaseCommands.GetStatistics();
+            var indexes = statistics.Indexes.OrderBy(x => x.Name).ToArray();
+
+            if (statistics.Errors.Length <= 0)
```
SzymonPobiega, line 24:
```suggestion
            if (statistics.Errors.Length = 0)
```

Nitpicking

SzymonPobiega (approved):
Two minor comments :)

src/ServiceControl/Operations/CheckRavenDBIndexErrors.cs
danielmarbach: But then the code should at least compile ;) We are not writing javascript here ;)

danielmarbach (approved):

PR #2516, opened 2021-04-26
Ability to toggle full text indexing

danielmarbach:
- [x] Additional tests for mixed data
- [x] Additional tests for recoverability search by body

### DIFF
--- a/src/ServiceControl.AcceptanceTests/Recoverability/MessageFailures/When_failed_message_searched_by_body_content.cs
+++ b/src/ServiceControl.AcceptanceTests/Recoverability/MessageFailures/When_failed_message_searched_by_body_content.cs
@@ -1,0 +1,135 @@
+﻿namespace ServiceControl.AcceptanceTests.Recoverability.MessageFailures
+{
+    using System;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using CompositeViews.Messages;
+    using NServiceBus;
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.AcceptanceTesting.Customization;
+    using NServiceBus.Settings;
+    using NUnit.Framework;
+    using TestSupport.EndpointTemplates;
+
+    class When_failed_message_searched_by_body_content : AcceptanceTest
+    {
+        [Test]
+        public async Task Should_be_found_when_fulltext_search_enabled()
+        {
+            // setting it even if it is the default
+            SetSettings = settings => settings.EnableFullTextSearchOnBodies = true;
+
+            var searchString = "42";
+
+            var context = await Define<MyContext>()
+                .WithEndpoint<Sender>(b => b.When((bus, c) => bus.Send(new MyMessage
+                {
+                    Something = "Somewhere in the body is the answer to all of the questions. 42"
+                })))
+                .WithEndpoint<Receiver>(b => b.DoNotFailOnErrorMessages())
+                .Done(async c =>
+                {
+                    if (c.MessageId != null && await this.TryGetMany<MessagesView>($"/api/messages/search/{c.MessageId}"))
+                    {
+                        c.MessageIngested = true;
+                    }
+
+                    if (!c.MessageIngested)
+                    {
+                        return false;
+                    }
+                    c.MessageFound = await this.TryGetMany<MessagesView>($"/api/messages/search/{searchString}");
+                    return true;
+                })
+                .Run();
+
+            Assert.IsTrue(context.MessageFound);
+        }
+
+        [Test]
+        public async Task Should_not_be_found_when_fulltext_search_disabled()
+        {
+            SetSettings = settings => settings.EnableFullTextSearchOnBodies = false;
+
+            var searchString = "42";
+
+            var context = await Define<MyContext>()
+                .WithEndpoint<Sender>(b => b.When((bus, c) => bus.Send(new MyMessage
+                {
+                    Something = "Somewhere in the body is the answer to all of the questions. 42"
+                })))
+                .WithEndpoint<Receiver>(b => b.DoNotFailOnErrorMessages())
+                .Done(async c =>
+                {
+                    if (c.MessageId != null && await this.TryGetMany<MessagesView>($"/api/messages/search/{c.MessageId}"))
+                    {
+                        c.MessageIngested = true;
+                    }
+
+                    if (!c.MessageIngested)
+                    {
+                        return false;
+                    }
+
+                    c.MessageFound = await this.TryGetMany<MessagesView>($"/api/messages/search/{searchString}");
+                    return true;
+                })
+                .Run();
+
+            Assert.IsTrue(context.MessageIngested);
+            Assert.IsFalse(context.MessageFound);
+        }
+
+        public class Sender : EndpointConfigurationBuilder
+        {
+            public Sender()
+            {
+                EndpointSetup<DefaultServer>(c =>
+                {
+                    var routing = c.ConfigureTransport().Routing();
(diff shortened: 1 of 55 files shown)


SzymonPobiega (approved):

src/ServiceControl.AcceptanceTests/Recoverability/MessageFailures/When_failed_message_searched_by_body_content.cs
```diff
@@ -0,0 +1,135 @@
...
+
+    class When_failed_message_searched_by_body_content : AcceptanceTest
+    {
+        [Test]
+        public async Task Should_be_found_when_fulltext_search_enabled()
+        {
+            // setting it even if it is the default
+            SetSettings = settings => settings.EnableFullTextSearchOnBodies = true;
+
+            var searchString = "42";
```
mikeminutillo, line 22: In theory, it's possible that the string `42` may appear in a guid and therefore this test (and the one below) could randomly fail. Maybe changing it `forty-two` :)

src/ServiceControl.Audit.AcceptanceTests/Auditing/When_processed_message_searched_by_body_content.cs
```diff
@@ -0,0 +1,129 @@
...
+
+    class When_processed_message_searched_by_body_content : AcceptanceTest
+    {
+        [Test]
+        public async Task Should_be_found_when_fulltext_search_enabled()
+        {
+            // setting it even if it is the default
+            SetSettings = settings => settings.EnableFullTextSearchOnBodies = true;
+
+            var searchString = "42";
```
mikeminutillo, line 21: Same potential issue as the test against the error instance. It's possible that any Guid could contain the string `42` and cause this test to be wonky

mikeminutillo:
A couple of other random thoughts that came from reviewing https://github.com/Particular/ServiceControl/pull/2501

- Both of these PRs add some kind of document store initialize property to enable tests but the mechanism for calling it is in different places. We need to make sure it doesn't get called twice.
- Both PRs add a new user-visible setting. They should probably get added to the startup diagnostics. [This is how the other PR does ie](https://github.com/Particular/ServiceControl/pull/2501/files?file-filters%5B%5D=.cs#diff-b862e8242aa6c06cfffc9b029ce1214b3fbb055a9a3ff6b2f1a7eb64c64fb6f4R255)

src/ServiceControl/CompositeViews/Messages/MessagesBodyTransformer.cs
```diff
@@ -1,21 +1,25 @@
...
     {
         public MessagesBodyTransformer()
         {
             TransformResults = messages => from message in messages
-                                           let metadata =
-                                               message.ProcessingAttempts != null
-                                                   ? message.ProcessingAttempts.Last().MessageMetadata
-                                                   : message.MessageMetadata
+                                           let metadata = message.ProcessingAttempts != null
+                                               ? message.ProcessingAttempts.Last().MessageMetadata
```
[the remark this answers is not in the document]
danielmarbach: Yep we do that all the time like this so I sticked to it

src/ServiceControl.AcceptanceTests/Recoverability/MessageFailures/When_failed_message_searched_by_body_content.cs
danielmarbach: Addressed

src/ServiceControl.Audit.AcceptanceTests/Auditing/When_processed_message_searched_by_body_content.cs
danielmarbach: Addressed

src/ServiceControl.UnitTests/BodyStorage/BodyStorageEnricherTests.cs
```diff
@@ -19,48 +21,83 @@ public class BodyStorageEnricherTests
...
+
+            await enricher.StoreErrorMessageBody(body, attempt);
 
             Assert.AreEqual(body, metadata["Body"], "Body should be stored if below threshold");
+            Assert.IsNull(attempt.Body, "Body property was set but shouldn't have been");
+            Assert.AreEqual(0, fakeStorage.StoredBodySize);
+        }
+
+        [Test]
+        public async Task Should_store_body_in_metadata_when_full_text_disabled_and_not_binary_and_below_LOH_threshold()
```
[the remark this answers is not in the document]
danielmarbach: Addressed

danielmarbach:
When upgrading

image

When upgrading and splitting

image

and if you select NO then the config value is populated to the split settings

image

which you can override

danielmarbach:
Edit view

image

the only drawback it has is that it is also shown on older instances that are not yet upgrade. Right now we have no concept of conditionally enable certain settings on the UI

src/ServiceControl.Audit.AcceptanceTests/Auditing/When_processed_message_searched_by_body_content.cs
```diff
@@ -0,0 +1,127 @@
...
+        public async Task Should_not_be_found_when_fulltext_search_disabled()
+        {
+            SetSettings = settings => settings.EnableFullTextSearchOnBodies = false;
+
+            var searchString = "forty-two";
+
+            var context = await Define<MyContext>()
+                .WithEndpoint<Sender>(b => b.When((bus, c) => bus.Send(new MyMessage
+                {
+                    Something = "Somewhere in the body is the answer to all of the questions. forty-two"
```
SzymonPobiega, line 58: I love it ❤️

src/ServiceControl.UnitTests/BodyStorage/BodyStorageEnricherTests.cs
```diff
@@ -19,48 +21,83 @@ public class BodyStorageEnricherTests
         public async Task Should_store_body_in_storage_when_binary_and_below_LOH_threshold()
```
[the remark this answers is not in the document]
SzymonPobiega: The On/Off applies only to below threshold, non-binary so we are actually introducing here only one new state.

SzymonPobiega (approved):

mikeminutillo:
I added a little more detail to the upgrade message as customers may not know if they want full text indexing or not.

image

This makes it clearer that the feature is related to the message search function in ServiceInsight, and that it is having a performance impact on the instance.

I think we should consider reversing the question too, so rather than

> Do you want to continue to support full text indexing for messages?

We should swap it to:

> Do you want to disable full text indexing for messages?

danielmarbach:
@mikeminutillo I like the idea of swapping the message 👍

mikeminutillo:
I think there's an issue with the powershell flag as it stands. At the moment, the flag is optional and defaults to true, so it will enable full text indexing if no value is supplied, _even if a prior upgrade had disabled it_. I think we should switch it to `DisableFullTextSearchOnBodies`. If it's set, then we can change the config value to `false` and turn off the feature. If it is not set, then we do nothing, and the instance should get the default value of `true`.

danielmarbach:
@mike or how about `FullTextSearchOnBodies`?

danielmarbach:
Or maybe make the default a nullable bool?

mikeminutillo:
I think that the upgrade is about transitioning to the new feature (which is the ability to turn off full text indexing) and not about being able to control which features are on/off. Other features with a toggle do not appear in the upgrade questsions. We typically assume that the prior value was correct. In this case, the options are:

a) Turn it off
b) (The default) Leave it as it was (which is on)

tmasternak (approved):

PR #2517, opened 2021-04-26
Use monotonic ID for processed messages

SzymonPobiega:


src/ServiceControl.Audit/Auditing/AuditPersister.cs
```diff
@@ -260,12 +260,19 @@ async Task ProcessAuditMessage(MessageContext context)
                     enricher.Enrich(enricherContext);
                 }
 
+                var processingStartedTicks =
```
danielmarbach, line 263: This would be such a nice use case of [`String.Create`](https://docs.microsoft.com/en-us/dotnet/api/system.string.create?view=net-5.0) which is not available on .NET Framework :(

danielmarbach (approved):
LGTM. I think this should not clash with https://github.com/Particular/ServiceControl/pull/2516 since we only ever retrieve the processed message documents by other values and never by the actual document ID right?

SzymonPobiega:
@mikeminutillo @tmasternak ?

seanfarmar (approved):

mikeminutillo (approved):
As long as no tests (or worse products) are trying to hand-craft IDs using knowledge of the algorith used, this should be OK. I believe we looked for that when we made this change on a previous PR.

PR #2540, opened 2021-05-12
lock analyzers to 0.9.0 while on NSB 7.x

adamralph:
and remove redundant NSB0002 entires, which are only relevant for NSB 8.x


andreasohlund (approved):

HEskandari (approved):
LGTM

PR #2541, opened 2021-05-12
Remove the version and key token from the message type header when publishing integration events

tmasternak:
Resolves consumers getting the following error:

```txt
NServiceBus.MessageDeserializationException: An error occurred while attempting to extract logical messages from incoming physical message a55afff1-019b-475b-9595-ad2500fb4775
 ---> System.IO.FileLoadException: Could not load file or assembly 'ServiceControl.Contracts, Version=3.0.0.0, Culture=neutral, PublicKeyToken=6f2d506f609d9f4d'. A strongly-named assembly is required. (0x80131044)
File name: 'ServiceControl.Contracts, Version=3.0.0.0, Culture=neutral, PublicKeyToken=6f2d506f609d9f4d'
 ---> System.IO.FileLoadException: A strongly-named assembly is required. (0x80131044)
```


src/ServiceControl/ExternalIntegrations/FallbackToOldContractTypes.cs
```diff
@@ -0,0 +1,30 @@
...
+        public override Task Invoke(IOutgoingPhysicalMessageContext context, Func<Task> next)
+        {
+
+            if (context.Headers.TryGetValue(EnclosedTypesHeaderName, out var qualifiedTypeName))
+            {
+                if (qualifiedTypeName.EndsWith(NewAssemblyIdentifier))
+                {
+                    var typeName = qualifiedTypeName.Split(',')[0];
+
+                    context.Headers[EnclosedTypesHeaderName] = $"{typeName}, {OldAssemblyIdentifier}";
```
SzymonPobiega, line 18: How about we always just remove the version number?

SzymonPobiega (changes requested):

src/ServiceControl/ExternalIntegrations/FallbackToOldContractTypes.cs
tmasternak: I wanted to make sure that the state is exactly as it was before we switched to ServiceControlContracts 3.0.

SzymonPobiega: I think by removing the version number we would make the contracts less restricted so that would be a non-breaking change for the existing customers.

ramonsmits: @tmasternak The version number can be removed, I've done that plenty of times (i.e. https://gist.github.com/ramonsmits/0ead15453e3bc1843cdc7f97718b5244)

I've never seen that break deserialization.

src/ServiceControl/ExternalIntegrations/RemoveVersionInformationBehavior.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    using System;
+    using System.Threading.Tasks;
+    using NServiceBus.Pipeline;
+
+    public class RemoveVersionInformationBehavior : Behavior<IOutgoingPhysicalMessageContext>
+    {
+        public override Task Invoke(IOutgoingPhysicalMessageContext context, Func<Task> next)
+        {
+
```
ramonsmits, line 11:
```suggestion
```

src/ServiceControl/ExternalIntegrations/RemoveVersionInformationBehavior.cs
```diff
@@ -0,0 +1,30 @@
...
+            if (context.Headers.TryGetValue(EnclosedTypesHeaderName, out var qualifiedTypeName)
+                && qualifiedTypeName.EndsWith(ContractsAssemblyIdentifier, StringComparison.OrdinalIgnoreCase))
```
ramonsmits, line 13: Wouldn't a type check based on assembly match be more efficient to a string compare? Especially because the string is quite lengthy.

ramonsmits:
@tmasternak I've updated the PR description with the error details.

src/ServiceControl/ExternalIntegrations/RemoveVersionInformationBehavior.cs
SzymonPobiega: We don't have access to the type in this behavior. The header is added in the connector between logical and physical outgoing stage. At the same point the type information is lost.

SzymonPobiega (approved):

src/ServiceControl/ExternalIntegrations/RemoveVersionInformationBehavior.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    using System;
+    using System.Threading.Tasks;
+    using NServiceBus.Pipeline;
+
+    public class RemoveVersionInformationBehavior : Behavior<IOutgoingPhysicalMessageContext>
+    {
+        public override Task Invoke(IOutgoingPhysicalMessageContext context, Func<Task> next)
+        {
+            if (context.Headers.TryGetValue(EnclosedTypesHeaderName, out var qualifiedTypeName)
```
mikeminutillo, line 11: Not that this is likely to change any time soon, but there's an NSB constant for this: `Headers.EnclosedMessageTypes`

mikeminutillo (changes requested):
Technically speaking, an outgoing message might have multiple message types (if it is part of an inheritance hierarchy).

See https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl.Audit/Infrastructure/FullTypeNameOnlyBehavior.cs for a similar implementation of the same idea (from the Audit split)

seanfarmar (approved):

tmasternak (commented):
LGTM

SzymonPobiega:
@mikeminutillo I'll leave it for you to take a look tomorrow. If you like it, please hit the green button. I'll then backport it to the release branch and push to the users.

mikeminutillo (approved):

PR #2574, opened 2021-06-10
Business components

tmasternak:


src/ServiceControl/Recoverability/RecoverabilityComponent.cs
```diff
@@ -0,0 +1,347 @@
...
+                context.CreateQueue(settings.ErrorQueue);
+            }
+
+            if (settings.ForwardErrorMessages && settings.ErrorLogQueue != null)
+            {
+                context.CreateQueue(settings.ErrorLogQueue);
+            }
+
+            context.AddIndexAssembly(typeof(RavenBootstrapper).Assembly);
+            context.AddIndexAssembly(typeof(SagaSnapshot).Assembly);
```
mikeminutillo, line 129: Should this be in the SagaAudit Component?

src/ServiceControl/Infrastructure/DomainEvents/DomainEventsServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,32 @@
+﻿// unset
```
mikeminutillo, line 1: These appear to be an artifact of the Resharper. Should we clean them all up?

mikeminutillo (approved):
Pretty minor comments

src/ServiceControl/SagaAudit/SagaAuditComponent.cs
```diff
@@ -0,0 +1,28 @@
...
+    using Particular.ServiceControl;
+    using ServiceBus.Management.Infrastructure.Settings;
+
+    class SagaAuditComponent : ServiceControlComponent
+    {
+        public override void Configure(Settings settings, IHostBuilder hostBuilder)
+        {
+            hostBuilder.ConfigureServices(collection =>
+            {
+                collection.AddSingleton<SagaRelationshipsEnricher>();
```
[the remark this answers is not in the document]
mikeminutillo: I went to add a Multi-Instance Acceptance Test for this and I believe something is off here. I don't think that the SagaAudit plugin will add headers to a message that ends up in the error queue. [Here](https://github.com/Particular/NServiceBus.SagaAudit/blob/master/src/NServiceBus.SagaAudit/AuditInvokedSagaBehavior.cs) is the code that adds the headers. It does so after the handlers have been run. If a handler fails, then the headers are never added. Even if the message fails later in the pipeline, then we would likely be returning it to the input queue and re-receiving it, which would effectively wipe the headers out. 

I've left it there for now until we can confirm that this is all true.

src/ServiceControl/ServiceControlComponent.cs
```diff
@@ -0,0 +1,12 @@
+
+namespace Particular.ServiceControl
+{
+    using Microsoft.Extensions.Hosting;
+    using ServiceBus.Management.Infrastructure.Settings;
+
+    abstract class ServiceControlComponent
+    {
+        public abstract void Configure(Settings settings, IHostBuilder hostBuilder);
+        public abstract void Setup(Settings settings, IComponentSetupContext context);
```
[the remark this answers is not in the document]
mikeminutillo: I left this for now. We can revisit it later.

src/ServiceControl/SetupBootstrapper.cs
```diff
@@ -114,9 +110,37 @@ static TransportSettings MapSettings(Settings settings)
             return transportSettings;
         }
 
-        readonly Settings settings;
+        static void SetTransportSpecificFlags(SettingsHolder settings, string poisonQueue)
```
mikeminutillo, line 113: Can/should we move all of these into the transport specific packages? i.e. Couldn't TransportCustomization include a `CreateQueues` and then have a virtual method to update the settings?

src/ServiceControl/Properties/launchSettings.json
```diff
@@ -0,0 +1,8 @@
+{
```
mikeminutillo, line 1: Should this be  in here?

PR #2576, opened 2021-06-14
Pass message session to the failed audits component

SzymonPobiega:


SzymonPobiega:
I found this one on 4.18 but I am pretty sure it is much older than that. Anyway, we need someone to review it ;-) I'll raise the KTLO issue for it.

WilliamBZA (approved):
Should we try add a test to catch this in the future?

SzymonPobiega:
@WilliamBZA I'll see if we can do it in the new version that we work on. This is not trivial because the acceptance tests currently don't exercise commands, they start on the lever below.

mikeminutillo (approved):

tmasternak (approved):

mikeminutillo:
@SzymonPobiega is this still needed now that 4.19 is out?

SzymonPobiega:
@mikeminutillo you are right, no longer needed because it is fixed in the latest.

Issue #2579, opened 2021-06-17
Switch ServiceControl to Microsoft Generic Host


mikeminutillo:
Switch ServiceControl instances to be hosted with [`Microsoft.Extensions.Hosting`](https://www.nuget.org/packages/Microsoft.Extensions.Hosting).

Currently ServiceControl contains a custom Windows Service Host which is used to configure and start an NServiceBus endpoint. All other features of ServiceControl are built inside of this endpoint. In implementing this issue we changed all of the ServiceControl instances to use the [Microsoft generic host](https://www.nuget.org/packages/Microsoft.Extensions.Hosting), and moved many of the ServiceControl features to be implemented as [Hosted Services](https://docs.microsoft.com/en-us/dotnet/architecture/microservices/multi-container-microservice-net-applications/background-tasks-with-ihostedservice) on top of that host.

**This is an internal refactoring only. There should be no functional difference to any of the ServiceControl instances with this change.**

Switching to the microsoft generic host makes it easier for us to control when parts of the application are started and stopped, and reduces a lot of the complexity associated with the startup and shutdown processes. It will also enable us to more easily integrate with 3rd party libraries or other Microsoft abstractions to provide logging,  IoC containers, configuration, web api hosting, etc. as they move to the Microsoft hosting abstractions.

PR #2580, opened 2021-06-17
Draft coding and design guidelines

mikeminutillo:
This document contains a number of coding and design guidelines that we have been following as we switch ServiceControl to using the Microsoft generic host https://github.com/Particular/ServiceControl/issues/2579


SzymonPobiega (approved):
This is pure gold! ❤️ ❤️ ❤️

docs/coding-and-design-guidelines.md
```diff
@@ -0,0 +1,50 @@
...
+### Exceptions
+
+There is one exception to the preference for Microsoft abstractions
+
+- **Use NServiceBus Logging abstractions over Microsoft or NLog abstractions**: The existing ServiceControl code uses the static `LogManager` classes to get access to the logging infrastructure and all new code should follow this pattern. In the future we are likely to switch to the Microsoft abstraction but until then we want to maintain consistency.
+
+
+## Prefer explicit container registration
+
+Where possible, we prefer explicit container registration for services as opposed to convention-bassed registration. This allows us to better see which classes belong to which ServiceControl components or which part of the ServiceControl infrastructure. It also allows us more explicit control over what services are available within the container which helps to reduce inappropriate cross-component access. In the future we may be able to make this more explicit by moving ServiceControl components into their own assemblies and keeping non-shared services internal.
```
WilliamBZA, line 29:
```suggestion
Where possible, we prefer explicit container registration for services as opposed to convention-based registration. This allows us to better see which classes belong to which ServiceControl components or which part of the ServiceControl infrastructure. It also allows us more explicit control over what services are available within the container which helps to reduce inappropriate cross-component access. In the future we may be able to make this more explicit by moving ServiceControl components into their own assemblies and keeping non-shared services internal.
```

docs/coding-and-design-guidelines.md
```diff
@@ -0,0 +1,50 @@
...
+  - When registering components from within an NServiceBus Feature, it still  makes sense to use IConfigureComponents but we should consider whether it makes sense to move the code out of NServiceBus features
+  - When relying on Autofac specific features, it makes sense to rely on IContainerBuilder. If a feature can be implemented in a way to avoid this, it should be preferred. We may in the future decide to remove our dependency on Autofac but that decision has not been made.
+- **Prefer IServiceProvider over ILifetimeService over IContainer**: As above, we'd rather rely on the generic Microsoft DI abstraction where possible and only fall back to NServiceBus or Autofac abstractions where strictly necessary.
+  - If relying on a specific feature of Autofac, then `ILifetimeService` should be used.
+  - `IContainer` should never be used. It is functionally equivalent to `IServiceProvider`
+
+### Exceptions
+
+There is one exception to the preference for Microsoft abstractions
```
WilliamBZA, line 22: This doesn't sound like an exception to the preference, but rather a practical "it will take time to migrate to the preferred abstraction"

WilliamBZA (approved):

docs/coding-and-design-guidelines.md
andreasohlund: Agreed, perhaps say "until NServiceBus core removes its static logging"?

Side note: this is tracked here https://github.com/Particular/PlatformArchitecture/blob/master/problems/nservicebus-core.md#custom-logging-abstraction

kbaley (approved):

docs/coding-and-design-guidelines.md
```diff
@@ -0,0 +1,50 @@
...
+- **Prefer Hosted Services over NServiceBus FeatureStartupTasks**: ServiceControl includes a lot of background tasks and these are both abstractions for building background tasks. `FeatureStartupTask` implementations are tied to the lifecycle of an endpoint. In general, we prefer to use the `IHostedService` abstraction which is tied to the lifecycle of the host application. NOTE: Hosted Services are started in the order that they are registered which allows us to better control the startup sequence. They are shut down in the reverse order.
+- **Prefer IServiceCollection over IConfigureComponents or IContainerBuilder**: Where possible we'd rather use the generic Microsoft DI abstractions over the NServiceBus one (IConfigureComponents) or the Autofac one (IContainerBuilder).
+  - When registering components from within an NServiceBus Feature, it still  makes sense to use IConfigureComponents but we should consider whether it makes sense to move the code out of NServiceBus features
```
mauroservienti, line 14: Does it make sense to add a note here to mention that this will be “solved” when SC will be migrated to v8, since v8 uses IServiceCollection by default?

docs/coding-and-design-guidelines.md
```diff
@@ -0,0 +1,50 @@
...
+  - When relying on Autofac specific features, it makes sense to rely on IContainerBuilder. If a feature can be implemented in a way to avoid this, it should be preferred. We may in the future decide to remove our dependency on Autofac but that decision has not been made.
+- **Prefer IServiceProvider over ILifetimeService over IContainer**: As above, we'd rather rely on the generic Microsoft DI abstraction where possible and only fall back to NServiceBus or Autofac abstractions where strictly necessary.
```
mauroservienti, line 16: Same as above.

mauroservienti (approved):

docs/coding-and-design-guidelines.md
adamralph, line 16:
```suggestion
- **Prefer `IServiceProvider` over `ILifetimeService` and `IContainer`**: As above, we prefer to use the Microsoft DI abstraction where possible and only fall back `ILifetimeService` where strictly necessary.
```

docs/coding-and-design-guidelines.md
```diff
@@ -0,0 +1,50 @@
...
+  - If relying on a specific feature of Autofac, then `ILifetimeService` should be used.
```
adamralph, line 17:
```suggestion
  - When relying on a Autofac-specific feature, `ILifetimeService` should be used.
```

docs/coding-and-design-guidelines.md
adamralph, line 29:
```suggestion
Where possible, we prefer explicit container registration for services instead of convention-based registration. This provides better visibility of which classes belong to which ServiceControl components or to which part of the ServiceControl infrastructure. It also gives us more explicit control over which services are available within the container, which helps to reduce inappropriate cross-component access. In the future we may be able to make this more explicit by moving ServiceControl components into their own assemblies and keeping non-shared services internal.
```

docs/coding-and-design-guidelines.md
```diff
@@ -0,0 +1,50 @@
...
+- **Use NServiceBus Logging abstractions over Microsoft or NLog abstractions**: The existing ServiceControl code uses the static `LogManager` classes to get access to the logging infrastructure and all new code should follow this pattern. In the future we are likely to switch to the Microsoft abstraction but until then we want to maintain consistency.
+
+
+## Prefer explicit container registration
+
+Where possible, we prefer explicit container registration for services as opposed to convention-bassed registration. This allows us to better see which classes belong to which ServiceControl components or which part of the ServiceControl infrastructure. It also allows us more explicit control over what services are available within the container which helps to reduce inappropriate cross-component access. In the future we may be able to make this more explicit by moving ServiceControl components into their own assemblies and keeping non-shared services internal.
+
+### Exceptions
+
+There are a few things that are still registered via convention. Note that these are all registered via Autofac and not via the Microsoft DI abstractions.
```
adamralph, line 33:
```suggestion
There are a few things that are still registered using convention. Note that these are all registered using Autofac and not the Microsoft DI abstractions.
```

docs/coding-and-design-guidelines.md
```diff
@@ -0,0 +1,50 @@
...
+
+- **API Controllers**
+- **Scatter-Gather API components**
+
+Additionally, because NServiceBus does a type-scan at startup it will automatically register any implementations of `Feature` and `IHandleMessage<>`. We have chosen to leave this alone as we would be fighting with NServiceBus in order to turn this off. 
+
+
+## Avoid property injection
+
+Although the Autofac container can be configured to allow property injection, we prefer to avoid it. There is no way to specify property injection using the generic microsoft DI abstractions, nor does the default Service Provider support it. Where possible, use constructor injection instead.
```
adamralph, line 43:
```suggestion
Although the Autofac container can be configured to allow property injection, we prefer to avoid it. There is no way to specify property injection using the Microsoft DI abstractions, and the default `IServiceProvider` implementation does not support it. Where possible, use constructor injection instead.
```

docs/coding-and-design-guidelines.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Exceptions
+
+There are a few places where property injection is still used. Note that these are all registered via Autofac and not via the Microsoft DI abstractions.
```
adamralph, line 47:
```suggestion
There are a few places where property injection is still used. These are all registered using Autofac, and not the Microsoft DI abstractions.
```

docs/coding-and-design-guidelines.md
mikeminutillo: We didn't want to make the decision that we are definitely moving away from NServiceBus logging, especially as it has not been deprecated for NServiceBus version 8. With that in mind, it is an exception to the preference. In this one case we do not prefer the Microsoft abstration.

@WilliamBZA @andreasohlund Do you think we need to make a stronger statement about it now?

WilliamBZA: I'm not terribly concerned about it. If we haven't made a clear decision then that's fine too.

tmasternak (approved):

PR #2586, opened 2021-06-25
SagaAudit enricher removed from the error ingestion

tmasternak:
SagaAudit enricher for error messages has been in ServiceControl at least since [2015](https://github.com/Particular/ServiceControl/commit/20cd88323b3babb2ccd725a5899456675b5daa0c#diff-1776ca4a1f88dd295fcbefcd1ab9bea5542e4e0cdfc8489ab218260605646e27R23). That said it looks there is no scenario possible in version 5, 6 or 7 of Core that would result in failing message having `InvokedSagas` header added either by Core or the SagaAudit plugin.
One [old bug report issue](https://github.com/Particular/ServiceControl/issues/977) seems to indicate that this is possible but after closer examination, the stack trace points to audit ingestor and not error ingestor.

### DIFF
--- a/src/ServiceControl/SagaAudit/SagaAuditComponent.cs
+++ b/src/ServiceControl/SagaAudit/SagaAuditComponent.cs
@@ -1,8 +1,6 @@
 ﻿namespace ServiceControl.SagaAudit
 {
-    using Microsoft.Extensions.DependencyInjection;
     using Microsoft.Extensions.Hosting;
-    using Operations;
     using Particular.ServiceControl;
     using ServiceBus.Management.Infrastructure.Settings;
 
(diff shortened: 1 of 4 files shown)


SzymonPobiega (approved):

src/ServiceControl/SagaAudit/SagaAuditComponent.cs
```diff
@@ -1,29 +1,18 @@
 ﻿namespace ServiceControl.SagaAudit
 {
-    using Microsoft.Extensions.DependencyInjection;
     using Microsoft.Extensions.Hosting;
-    using Operations;
     using Particular.ServiceControl;
     using ServiceBus.Management.Infrastructure.Settings;
 
     class SagaAuditComponent : ServiceControlComponent
```
timbussmann, line 7: what is this still used for?

timbussmann (approved):

src/ServiceControl/SagaAudit/SagaAuditComponent.cs
tmasternak: We need this to call
> context.AddIndexAssembly(typeof(SagaSnapshot).Assembly);

This ensures RavenDB index setup for saga snapshots.

PR #2589, opened 2021-06-30
Handle retry succeeded notifications

tmasternak:


src/ServiceControl/Operations/RetryConfirmationProcessor.cs
```diff
@@ -0,0 +1,83 @@
...
+            var storedContexts = new List<MessageContext>(contexts.Count);
+            var allCommands = new List<ICommandData>(contexts.Count);
+
+            foreach (var context in contexts)
+            {
+                try
+                {
+                    var commands = ProcessOne(context);
+                    allCommands.AddRange(commands);
+                    storedContexts.Add(context);
```
timbussmann, line 36: why do we need to return a new collection that would contain the same elements as the `context` parameter?

src/ServiceControl.Audit/Recoverability/FailedMessagesFeature.cs
```diff
@@ -28,21 +30,36 @@ public void Enrich(AuditEnricherContext context)
...
                 context.Metadata.Add("IsRetried", hasBeenRetried);
 
-                if (!hasBeenRetried)
+                if (!hasBeenRetried || isAckHandled)
                 {
                     return;
                 }
 
-                context.AddForSend(new MarkMessageFailureResolvedByRetry
+                if (hasAckQueue && isNewRetry)
```
tmasternak, line 45: Do we need `isNewRetry`? If the message has `ack-queue` header is had to be generated from a version of SC that is based on the `UniqueMessageId`. Right?

src/ServiceControl/CompositeViews/Messages/ScatterGatherApiMessageView.cs
```diff
@@ -31,9 +32,17 @@ protected override IList<MessagesView> ProcessResults(HttpRequestMessage request
                     }
                 }
 
-                combined.AddRange(messagesViews);
+                //HINT: De-duplicate the results as some messages might be present in multiple instances (e.g. when they initially failed and later were successfully processed)
+                foreach (MessagesView messagesView in messagesViews)
+                {
+                    if (!deduplicated.ContainsKey(messagesView.MessageId))
+                    {
+                        deduplicated.Add(messagesView.Id, messagesView);
```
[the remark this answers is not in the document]
SzymonPobiega: The `ScatterGatherApi` `Execute` guarantees that the first results batch is going to be fetched from local main instance store. But I'll add a comment that mentions this.

src/ServiceControl/Operations/RetryConfirmationProcessor.cs
SzymonPobiega: You mean the `storedContexts`? We need to return it because it may contain less elements than the input parameter if one of the messages failed to be processed. In that case there will be no commands associated with it and also we won't be forwarding that message to the error.log queue nor publishing integration events.

src/ServiceControl/CompositeViews/Messages/ScatterGatherApiMessageView.cs
```diff
@@ -29,11 +30,18 @@ protected override IList<MessagesView> ProcessResults(HttpRequestMessage request
...
                     }
-                }
 
-                combined.AddRange(messagesViews);
+                    //HINT: De-duplicate the results as some messages might be present in multiple instances (e.g. when they initially failed and later were successfully processed)
+                    //The Execute method guarantees that the first item in the results collection comes from the main SC instance so the data fetched from failed messages has
+                    //precedence over the data from the audit instances.
+                    if (!deduplicated.ContainsKey(result.MessageId))
+                    {
+                        deduplicated.Add(result.Id, result);
```
tmasternak, line 39: Should we be using `MessageId` here as well?

src/ServiceControl/Operations/RetryConfirmationProcessor.cs
timbussmann: but it doesn't really look like the returned elements is smaller than the incoming one for the retry confirmation processor? There isn't really anything in this part of the code path that is expected to go wrong?

timbussmann (commented):
what happens in the scenario of "new SC", "new endpoint" and "old audit instance"?

src/ServiceControl.Audit/Auditing/AuditPersister.cs
```diff
@@ -64,7 +65,7 @@ public async Task<IReadOnlyList<MessageContext>> Persist(List<MessageContext> co
                 var inserts = new List<Task>(contexts.Count);
                 foreach (var context in contexts)
                 {
-                    inserts.Add(ProcessMessage(context));
+                    inserts.Add(ProcessMessage(context, dispatcher));
```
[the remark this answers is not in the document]
SzymonPobiega: I looked at the code path and it seems it would be complex. The `IMessageSession` comes from the host that starts NSB but `IDispatchMessages` comes from the raw endpoint which lifecycle is managed by `AuditIngestion` class.

There is probably a lot of accidental complexity in `AuditImporter`, `AuditIngestion`, `AuditIngestionComponent`, `AuditIngestor` and `AuditPersister` classes but I think dealing with it would require a separate PR.

src/ServiceControl.MultiInstance.AcceptanceTests/Recoverability/When_a_message_retry_audit_from_old_sc_is_sent_to_audit_instance.cs
```diff
@@ -0,0 +1,125 @@
...
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.Pipeline;
+    using NServiceBus.Settings;
+    using NUnit.Framework;
+    using ServiceControl.Infrastructure;
+    using TestSupport;
+    using TestSupport.EndpointTemplates;
+
+    [RunOnAllTransports]
+    class When_a_message_retry_audit_from_old_sc_is_sent_to_audit_instance : AcceptanceTest
```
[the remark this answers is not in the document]
SzymonPobiega: This test removes the `AcknowledgementQueue` header so it should cover two scenarios:
 - Old SC sending a retry message to a new endpoint
 - New SC sending a retry message to an old endpoint
 
 In both of these cases the endpoint logic is not executed and we end up going through the audit path.

src/ServiceControl.MultiInstance.AcceptanceTests/Recoverability/When_a_message_retry_audit_from_old_sc_is_sent_to_audit_instance.cs
```diff
@@ -0,0 +1,125 @@
...
+                    var result = await GetFailedMessage(c);
+                    failure = result;
+                    if (!result)
+                    {
+                        return false;
+                    }
+
+                    if (failure.Status == FailedMessageStatus.Unresolved)
+                    {
+                        await IssueRetry(c, () => this.Post<object>($"/api/errors/{failure.UniqueMessageId}/retry", null, null, ServiceControlInstanceName));
```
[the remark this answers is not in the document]
SzymonPobiega: Yep, you are right. This test needs some love.

src/ServiceControl/Operations/RetryConfirmationProcessor.cs
SzymonPobiega: Ah yes. Sorryg, was thinking about failed messages.

src/ServiceControl.MultiInstance.AcceptanceTests/Recoverability/When_a_message_retry_audit_from_old_sc_is_sent_to_audit_instance.cs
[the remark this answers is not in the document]
SzymonPobiega: Actually I checked and the `IssueRetry` method sets a flag and then checks it to ensure the retry is issued only once so we are good.

src/ServiceControl.MultiInstance.AcceptanceTests/Recoverability/When_a_message_retry_audit_from_old_sc_is_sent_to_audit_instance.cs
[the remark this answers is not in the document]
timbussmann: discussed this with Szymon on the sync. We think these test cases could actually be done as audit acceptance tests instead of multi-instance tests, as we don't really need the full setup to test the audit behavior. The only need for a multi-instance test is to verify that an enabled audit instance does not mess up with the new retry confirmation path (which only requires the main instance to work), e.g. by messing up query data (through duplicates).

src/ServiceControl/Operations/RetryConfirmationProcessor.cs
SzymonPobiega: We discussed that and we concluded that we can remove the complexity for the retry confirmation path.

src/ServiceControl.Audit/Auditing/AuditIngestionComponent.cs
```diff
@@ -57,16 +57,21 @@ AuditIngestionCustomCheck.State ingestionState
...
 
-                    receivedMeter.Mark();
+                         await channel.Writer.WriteAsync(messageContext).ConfigureAwait(false);
+                         await taskCompletionSource.Task.ConfigureAwait(false);
+                     }
 
-                    await channel.Writer.WriteAsync(messageContext).ConfigureAwait(false);
-                    await taskCompletionSource.Task.ConfigureAwait(false);
-                },
+                     return settings.OnMessage(messageContext.MessageId, messageContext.Headers, messageContext.Body, Process);
```
timbussmann, line 73: I don't remember the details but I thought there were some delegate caching issues when using method groups. Might be worth to double check. (https://github.com/dotnet/roslyn/issues/5835)

src/ServiceControl/Operations/ErrorIngestionComponent.cs
```diff
@@ -44,28 +42,37 @@ class ErrorIngestionComponent
...
+            var errorProcessor = new ErrorProcessor(bodyStorageEnricher, enrichers, failedMessageEnrichers, domainEvents, ingestedMeter);
+            var retryConfirmationProcessor = new RetryConfirmationProcessor(domainEvents);
 
-            var ingestion = new ErrorIngestion(async messageContext =>
-                {
-                    var taskCompletionSource = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
-                    messageContext.SetTaskCompletionSource(taskCompletionSource);
+            ingestor = new ErrorIngestor(errorProcessor, retryConfirmationProcessor, documentStore, bulkInsertDurationMeter, settings.ForwardErrorMessages, settings.ErrorLogQueue);
+
+            async Task Process(MessageContext messageContext)
```
timbussmann, line 57: maybe move to the end of the function to be less interrupting when reading through the method

src/ServiceControl/Operations/RetryConfirmationProcessor.cs
```diff
@@ -0,0 +1,66 @@
...
+        const string RetryUniqueMessageIdHeader = "ServiceControl.Retry.UniqueMessageId";
+
+        public RetryConfirmationProcessor(IDomainEvents domainEvents)
+        {
+            this.domainEvents = domainEvents;
+        }
+
+        public IReadOnlyCollection<ICommandData> Process(List<MessageContext> contexts)
+        {
+            var allCommands = new List<ICommandData>(contexts.Count);
```
timbussmann, line 25: that list would be too small by default given we return two commands per context?

src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSender.cs
```diff
@@ -27,6 +29,7 @@ public virtual async Task HandleMessage(MessageContext message, IDispatchMessage
             var outgoingHeaders = new Dictionary<string, string>(message.Headers);
 
             outgoingHeaders.Remove("ServiceControl.Retry.StagingId");
+            outgoingHeaders["ServiceControl.Retry.AcknowledgementQueue"] = settings.ErrorQueue;
```
timbussmann, line 32: will that be the correct address for remote endpoints? e.g. when using MSMQ?

timbussmann (approved):

src/ServiceControl.Audit.AcceptanceTests/Recoverability/When_a_successful_retry_is_detected.cs
```diff
@@ -1,49 +1,72 @@
...
 
-                    options.SetHeader(uniqueMessageIdHeaderName, failedMessageId);
+                    options.SetHeader("ServiceControl.Retry.UniqueMessageId", failedMessageId);
+                    options.SetHeader("ServiceControl.Retry.AcknowledgementQueue", Conventions.EndpointNamingConvention(typeof(AcknowledgementSpy)));
                     options.RouteToThisEndpoint();
                     return s.Send(new MyMessage(), options);
                 }))
-                .Done(async c =>
+                .WithEndpoint<AcknowledgementSpy>()
+                .Done(c => c.AcknowledgementSent)
```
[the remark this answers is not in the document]
tmasternak: This test has been removed in favour of direct enriched tests

src/ServiceControl/Operations/ErrorIngestionComponent.cs
tmasternak: The message filtering logic has been refactored and there is no longer a local function defined.

src/ServiceControl.Audit/Auditing/AuditIngestionComponent.cs
tmasternak: The message filtering logic has been refactored and we no longer use a method group.

src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSender.cs
SzymonPobiega: Great catch @timbussmann ❤️  If we released it as-was it would have been a disaster. We fixed it by doing some `TransportInfrastructure` hackery and I've checked in debug that now it sets proper queue name i.e. `error@LAPTOP-PHIOSCQO`

src/ServiceControl/Operations/RetryConfirmationProcessor.cs
SzymonPobiega: :+1: Fixed now

PR #2603, opened 2021-08-02
Remove machine name when using logical queues

WilliamBZA:


ramonsmits:
@WilliamBZA Closing this as https://github.com/Particular/ServiceControl/pull/2602 is already merged and this seems like a duplicate.

ramonsmits (approved):
As https://github.com/Particular/ServiceControl/pull/2602 is already merged that was for release 4.20 this is to update master and is the same fix

PR #2641, opened 2021-10-15
Update ServiceControl transport to ASBS transport 2.x

mauroservienti:


src/ServiceControl.Transports.ASBS/ServiceControl.Transports.ASBS.csproj
```diff
@@ -10,7 +10,7 @@
 
   <ItemGroup>
     <PackageReference Include="NServiceBus.CustomChecks" Version="3.0.1" />
-    <PackageReference Include="NServiceBus.Transport.AzureServiceBus" Version="1.9.0" />
+    <PackageReference Include="NServiceBus.Transport.AzureServiceBus" Version="2.0.0" />
```
andreasohlund, line 13: 2.0.1

mikeminutillo:
There is [this set of transitive dependencies](https://github.com/Particular/ServiceControl/blob/master/src/TransitiveDependencies.Build.props) that get copied into all of the transports and hosts. 

When transports and hosts use the same dependency they must all reference the exact same version or we get some very difficult to pin down bugs. Sometimes this can be achieved by specifying a concrete dependency version in the ServiceControl transport project itself, where this would normally be a transitive dependency.

mauroservienti:
Thanks, @mikeminutillo. I added those to the ServiceControl transport project in https://github.com/Particular/ServiceControl/pull/2641/commits/aba94ed7fdfd430ea25d93192811b41645710b0b. The build continues to fail with the same error, and it seems to me the results of the tests didn't change. The tests log also mention the following dependencies as problematic:

- System.Runtime.CompilerServices.Unsafe
- System.Text.Encodings.Web
- System.Text.Json

I already added System.Runtime.CompilerServices.Unsafe as a dependency. The other two, however, are used nowhere in the repo, except for System.Text.Json that's used in ASQ.

src/ServiceControl.Audit/ServiceControl.Audit.csproj
```diff
@@ -48,6 +48,7 @@
     <PackageReference Include="System.Memory" Version="4.5.4" />
     <PackageReference Include="System.Buffers" Version="4.5.1" />
     <PackageReference Update="System.Runtime.CompilerServices.Unsafe" Version="4.7.1" />
+    <PackageReference Include="System.Text.Encodings.Web" Version="4.7.2" />
```
mikeminutillo, line 51: If might make sense to move this to [here](https://github.com/Particular/ServiceControl/blob/06754557fe54623b74bdf4e647620f3c8eb5f53f/src/TransitiveDependencies.Build.props). The same for the Monitoring and Primary projects. They all reference this

mikeminutillo (approved):

Issue #2655, opened 2021-10-26
ServiceControl Audit instance throws System.UnauthorizedAccessException after upgrading to 4.19.0


Asopus:
## Symptoms

When starting ServiceControl Audit instance, after creation or upgrade, the startup fails. The following exception can be found in the logs:

```
Application: ServiceControl.Audit.exe
Framework Version: v4.0.30319
Description: The process was terminated due to an unhandled exception.
Exception Info: System.UnauthorizedAccessException
   at System.IO.__Error.WinIOError(Int32, System.String)
   at System.IO.Directory.InternalCreateDirectory(System.String, System.String, System.Object, Boolean)
   at System.IO.Directory.InternalCreateDirectoryHelper(System.String, Boolean)
   at Raven.Database.Server.AssemblyExtractor.Extract(System.Reflection.Assembly, System.Collections.Generic.IEnumerable`1<System.Collections.Generic.KeyValuePair`2<System.String,AssemblyToExtract>>, System.String)
   at Raven.Database.Server.AssemblyExtractor.ExtractEmbeddedAssemblies(Raven.Database.Config.InMemoryRavenConfiguration)
   at Owin.AppBuilderExtensions.UseRavenDB(Owin.IAppBuilder, Raven.Database.Server.RavenDBOptions)
   at Raven.Database.Server.OwinHttpServer.<.ctor>b__4_0(Owin.IAppBuilder)
   at Microsoft.Owin.Hosting.Engine.HostingEngine.Start(Microsoft.Owin.Hosting.Engine.StartContext)
   at Raven.Database.Embedded.OwinEmbeddedHost.Configure(System.Action`1<Owin.IAppBuilder>, Microsoft.Owin.Hosting.StartOptions)
   at Raven.Database.Server.OwinHttpServer..ctor(Raven.Database.Config.InMemoryRavenConfiguration, Raven.Database.DocumentDatabase, Boolean, System.Action`1<Raven.Database.Server.RavenDBOptions>)
   at Raven.Server.RavenDbServer.Initialize(System.Action`1<Raven.Database.Server.RavenDBOptions>)
   at Raven.Database.Client.EmbeddedDocumentStore.Initialize()
   at Raven.Client.Embedded.EmbeddableDocumentStore.Initialize()
   at ServiceControl.Audit.Infrastructure.RavenDB.EmbeddedRavenDbHostedService+<StartAsync>d__4.MoveNext()
[...]

PR #2657, opened 2021-10-27
Run RavenDbHostedService when setting up audit instance

tmasternak:
Connected to https://github.com/Particular/ServiceControl/issues/2655

### DIFF
--- a/src/ServiceControl.Audit/Infrastructure/RavenDB/EmbeddedRavenDbHostedService.cs
+++ b/src/ServiceControl.Audit/Infrastructure/RavenDB/EmbeddedRavenDbHostedService.cs
@@ -27,7 +27,9 @@
             this.dataMigrations = dataMigrations;
         }
 
-        public async Task StartAsync(CancellationToken cancellationToken)
+        public Task StartAsync(CancellationToken cancellationToken) => SetupDatabase();
+
+        public async Task SetupDatabase()
         {
             Logger.Info("Database initialization starting");
             documentStore.Initialize();
--- a/src/ServiceControl.Audit/Infrastructure/RavenDB/EmbeddedRavenDbHostedService.cs
+++ b/src/ServiceControl.Audit/Infrastructure/RavenDB/EmbeddedRavenDbHostedService.cs
@@ -16,9 +16,14 @@
         readonly IEnumerable<IDataMigration> dataMigrations;
 
         public EmbeddedRavenDbHostedService(IDocumentStore documentStore, IOptions<RavenStartup> ravenStartup, IEnumerable<IDataMigration> dataMigrations)
+        : this(documentStore, ravenStartup.Value, dataMigrations)
+        {
+        }
+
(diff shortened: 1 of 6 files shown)


src/ServiceControl.Audit/Infrastructure/SetupBootstrapper.cs
```diff
@@ -102,6 +103,17 @@ public async Task Run(string username)
...
+
+
+                var ravenOptions = new RavenStartup();
+                foreach (var indexAssembly in RavenBootstrapper.IndexAssemblies)
+                {
+                    ravenOptions.AddIndexAssembly(indexAssembly);
+                }
+
+                var service = new EmbeddedRavenDbHostedService(documentStore, ravenOptions, new IDataMigration[0]);
+                await service.StartAsync(CancellationToken.None).ConfigureAwait(false);
```
kentdr, line 115: What was the value in making this class a hosted service as opposed to just having an async method to do initialization/index creation?

src/ServiceControl.Audit/Infrastructure/RavenDB/EmbeddedRavenDbHostedService.cs
```diff
@@ -16,9 +16,14 @@ class EmbeddedRavenDbHostedService : IHostedService
...
         public EmbeddedRavenDbHostedService(IDocumentStore documentStore, IOptions<RavenStartup> ravenStartup, IEnumerable<IDataMigration> dataMigrations)
+        : this(documentStore, ravenStartup.Value, dataMigrations)
+        {
+        }
+
+        public EmbeddedRavenDbHostedService(IDocumentStore documentStore, RavenStartup ravenStartup, IEnumerable<IDataMigration> dataMigrations)
         {
             this.documentStore = documentStore;
-            this.ravenStartup = ravenStartup.Value;
+            this.ravenStartup = ravenStartup;
```
kentdr, line 26: I noticed below that the IDocumentStore instance is disposed of when this service stops, but this service didn't initialize the store.  Should the disposal be handled by the DI container or the calling code instead?

kentdr (approved):

src/ServiceControl.Audit/Infrastructure/SetupBootstrapper.cs
mikeminutillo: The hosted service already existed. All we are doing is calling into it. This is following the same pattern as the primary instance, which was already doing this. I like the idea of pulling the initialization/index creation out into a separate method. We don't really need to call stop because the document store is inside of a using statement.

src/ServiceControl.Audit/Infrastructure/RavenDB/EmbeddedRavenDbHostedService.cs
mikeminutillo: It kind of does. An `IDcoumentStore` instance is unusable before you call `.Initialize()` on it, which we do in our startup. Everything before that is effectively configuration.

We struggled with the right way to do this when we introduced the hosted service and this was the only way we found that works.

PR #2672, opened 2021-11-02
Audit retention check

kentdr:


src/ServiceControl.UnitTests/ApprovalFiles/APIApprovals.CustomCheckDetails.approved.txt
```diff
@@ -5,4 +5,5 @@ ServiceControl Health: Error Database Index Errors
 ServiceControl Health: Error Database Index Lag
 ServiceControl Health: Error Message Ingestion
 ServiceControl Health: Error Message Ingestion Process
+ServiceControl Health: Saga audit data retention check
```
mikeminutillo, line 8: I would take the word "check" out. These are all checks

src/ServiceControl/CustomChecks/InternalCustomChecks/InternalCustomChecks.cs
```diff
@@ -7,6 +7,7 @@
     using NServiceBus.CustomChecks;
     using NServiceBus.Hosting;
     using Operations;
+    using SagaAudit;
```
mikeminutillo, line 10: I don't think this is  still needed

```suggestion
```

src/ServiceControl.MultiInstance.AcceptanceTests/SagaAudit/When_sending_saga_audit_to_main_instance.cs
```diff
@@ -22,10 +26,17 @@ public void ConfigSetup()
...
+            MimicSettingsOfAnUpgradedOldMainInstanceThatHasNeverHadAuditRetentionSetting(appSettings);
             config.Save();
             ConfigurationManager.RefreshSection("appSettings");
+
+
         }
 
+        static void MimicSettingsOfAnUpgradedOldMainInstanceThatHasNeverHadAuditRetentionSetting(
+            AppSettingsSection appSettings) =>
+            appSettings.Settings.Remove("ServiceControl/AuditRetentionPeriod");
```
mikeminutillo, line 38: We might want to store this so that we can restore it after the test. The config file is shared by all of the tests in the same assembly. I _think_ none of the other tests are relying on it being there but it might be a nasty surprise for someone trying to write a test in the future. It will work when run in isolation but will fail when it's run after this (which may not even be deterministic).

src/ServiceControl.MultiInstance.AcceptanceTests/SagaAudit/When_sending_saga_audit_to_main_instance.cs
```diff
@@ -66,6 +77,67 @@ public async Task Saga_history_can_be_fetched_from_main_instance()
...
+                    if (!c.SagaId.HasValue)
+                    {
+                        return false;
+                    }
+
+                    var sagaAudiDatatInMainInstanceIsAvailableForQuery =
+                        await this.TryGet<SagaHistory>($"/api/sagas/{c.SagaId}", instanceName: ServiceControlInstanceName);
+                    return sagaAudiDatatInMainInstanceIsAvailableForQuery;
+                })
+                .Run();
```
mikeminutillo, line 136: Each time we run a scenario using this API the test infra will spin up a ServiceControl instances and a ServiceControl Audit instance, run the tests and then tear everything down. I'm surprised that multiple executions seem to carry the same database as I thought that the databases were all in-memory for the tests. I would consider combining this `Done` with the one above and making it a single execution. It will be slightly less readable but with some comments it should be OK.

src/ServiceControl/SagaAudit/AuditRetentionCustomCheck.cs
```diff
@@ -0,0 +1,45 @@
...
+    using ServiceBus.Management.Infrastructure.Settings;
+
+    class AuditRetentionCustomCheck : CustomCheck
+
+    {
+        readonly IDocumentStore _documentStore;
+        readonly bool _auditRetentionPeriodIsSet;
+
+        public AuditRetentionCustomCheck(IDocumentStore documentStore, Settings settings, TimeSpan? repeatAfter = null)
+            : base("Saga audit data retention check", "ServiceControl Health", repeatAfter.HasValue ? repeatAfter : TimeSpan.FromHours(24))
```
mikeminutillo, line 16: I think the longest duration we have on another check is 1 hour. It means that a new instance may take a day to report the problem if the startup invocation happens before the first saga audit message is ingested.

I don't necessarily think it's an issue to only run every 24 hours, but I would consider dropping it down. Some points to consider:
- This check is not particularly taxing. It's a query on an existing index
- Custom Check events are only fired on state changes. Even if this fails once per hour forever, from an external perspective only a single event will be emitted.

src/ServiceControl/SagaAudit/AuditRetentionCustomCheck.cs
```diff
@@ -0,0 +1,45 @@
...
+        public override async Task<CheckResult> PerformCheck()
+        {
+            if (_auditRetentionPeriodIsSet)
+            {
+                return CheckResult.Pass;
+            }
+
+            if (await DetectSagaAuditData().ConfigureAwait(false))
+            {
+                return CheckResult.Failed("Saga snapshot data detected without an audit retention period configured. If saga audit data is allowed to accumulate, it can result in degraded performance.");
```
mikeminutillo, line 31: It might be worth linking to docs somewhere (and raising a PR in docs to address this) to explain to the user what to do if they see this message. I think a couple of the other custom checks point to https://docs.particular.net/search?q=servicecontrol+troubleshooting

mikeminutillo:
Good stuff. Just a few minor comments but overall very happy.

mikeminutillo (commented):
Good stuff. Just a few minor comments but overall very happy.

src/ServiceControl.MultiInstance.AcceptanceTests/SagaAudit/When_sending_saga_audit_to_main_instance.cs
mikeminutillo: I think maybe because we re-use the same scenario, it all gets executed twice. i.e. We set up everything, have the endpoint send an audited saga message, store it in the database, retrieve it from the API, and then tear everything down. Then we do all of that again when we run the scenario again above.

src/ServiceControl.UnitTests/ApprovalFiles/APIApprovals.CustomCheckDetails.approved.txt
tmasternak: And maybe make in capital letters as well?

src/ServiceControl.MultiInstance.AcceptanceTests/SagaAudit/When_sending_saga_audit_to_main_instance.cs
```diff
@@ -66,6 +83,57 @@ public async Task Saga_history_can_be_fetched_from_main_instance()
...
+            await Define<MyContext>()
+                .WithEndpoint<SagaEndpoint>(b => b.When((bus, c) => bus.SendLocal(new MessageInitiatingSaga { Id = "Id" })))
+                .Done(async c =>
+                {
+                    if (!c.SagaId.HasValue)
+                    {
+                        return false;
+                    }
+
+                    if (sagaAudiDataInMainInstanceIsAvailableForQuery?.HasResult == false)
```
mikeminutillo, line 118: The left-hand side of this expression will be `null`, so the condition will never be true. The test still works because the event log item below still gets generated and that's all we assert on once the scenario finishes. You could get rid of this block altogether or turn it into `!= true`.

cquirosj: My bad. I wanted to make it a bit faster by not querying /api/eventlogitems/ every time but didn't get the logic right. Just pushed a fix. It saves 2 seconds.

tmasternak (approved):
Once we have the doco PR opened this can be merged

src/ServiceControl/SagaAudit/AuditRetentionCustomCheck.cs
cquirosj: Here is the doc PR. https://github.com/Particular/docs.particular.net/pull/5546

mikeminutillo (approved):

PR #2669, opened 2021-11-02
Spike package to centralize Particular Service Platform connection

mikeminutillo:
This is a package that allows setting up all plugins to connect an NServiceBus endpoint to the Particular Service Platform.

In practice it looks like this:

```csharp
var json = File.ReadAllText(pathToPlatformConnection);
var servicePlatformConnection = ServicePlatformConnectionConfiguration.Parse(json);
endpointConfiguration.ConnectToServicePlatform(servicePlatformConnection);
```

The file contains json and looks like this:

```json
{
  "AuditQueue": "myAuditQueue",
  "ErrorQueue": "myErrorQueue",
  "Heartbeats": {
    "HeartbeatQueue": "serviceControlHeartbeatsQueue", 
    "Frequency": "00:00:30",
    "TimeToLive": "00:02:00"
  },
  "CustomChecks": {
    "CustomChecksQueue": "serviceControlCustomChecksQueue",
    "TimeToLive": "00:02:00"
  },
  "SagaAudit": {
    "SagaAuditQueue": "serviceControlSagaAuditQueue"
  },
  "Metrics": {
    "MetricsQueue": "serviceControlMetricsQueue",
    "Interval": "00:00:10",
    "TimeToBeReceived": "00:05:00",
    "InstanceId": "MyInstanceId"
  }
}
```

Alternatively, the connection can be configured manually:

```csharp
var servicePlatformConnection = new ServicePlatformConnectionConfiguration
{
  AuditQueue = "myAuditQueue",
  ErrorQueue = "myErrorQueue",
  Heartbeats = {
    HeartbeatQueue = "serviceControlHeartbeatsQueue", 
    Frequency = TimeSpan.FromSeconds(30),
    TimeToLive = TimeSpan.FromMinutes(2)
  },
  CustomChecks = {
    CustomChecksQueue = "serviceControlCustomChecksQueue",
    TimeToLive = TimeSpan.FromMinutes(2)
  },
  SagaAudit = {
    SagaAuditQueue = "serviceControlSagaAuditQueue"
  },
  Metrics = {
    MetricsQueue = "serviceControlMetricsQueue",
    Interval = TimeSpan.FromSeconds(10),
    TimeToBeReceived = TimeSpan.FromMinutes(5),
    InstanceId = "MyInstanceId"
  }
};

endpointConfiguration.ConnectToServicePlatform(servicePlatformConnection);
```

And you can combine the two approaches by loading from Json and then manually setting values to override them.

[...]

PR #2674, opened 2021-11-04
Expose connection info via API

mikeminutillo:
A companion to https://github.com/Particular/ServiceControl/pull/2669

This creates a new api at `/api/connection` which exposes the info needed for the platform connection package.

```json
{
  "Heartbeats": {
    "Enabled": true,
    "HeartbeatsQueue": "Particular.ServiceControl",
    "Frequency": "00:00:10",
    "TimeToLive": "00:00:40"
  },
  "CustomChecks": {
    "Enabled": true,
    "CustomChecksQueue": "Particular.ServiceControl"
  },
  "ErrorQueue": "error",
  "MessageAudit": {
    "Enabled": true,
    "AuditQueue": "audit"
  },
  "SagaAudit": {
    "Enabled": true,
    "SagaAuditQueue": "audit"
  }
}
```

NOTE: The `MessageAudit` and `SagaAudit` entries come from the audit instance. If you have multiple audit instances then subsequent ones will be rendered as `MessageAudit1`, `MessageAudit2`, and so on. The order that this happens is not defined.

### DIFF
--- a/src/ServiceControl.Audit/Connection/ConnectionController.cs
+++ b/src/ServiceControl.Audit/Connection/ConnectionController.cs
@@ -1,0 +1,26 @@
+namespace ServiceControl.Audit.Connection
+{
+    using System.Net;
+    using System.Net.Http;
+    using System.Threading.Tasks;
+    using System.Web.Http;
+    using Infrastructure.Settings;
+
+    public class ConnectionController : ApiController
+    {
+        readonly Settings settings;
+
+        public ConnectionController(Settings settings) => this.settings = settings;
+
+        [Route("connection")]
+        [HttpGet]
+        public Task<HttpResponseMessage> GetConnectionDetails() => Task.FromResult(Request.CreateResponse(HttpStatusCode.OK, new
+        {
+            settings.AuditQueue,
+            SagaAudit = new
+            {
+                SagaAuditQueue = settings.AuditQueue
+            }
+        }));
+    }
+}
--- a/src/ServiceControl.Audit/Connection/ConnectionController.cs
+++ b/src/ServiceControl.Audit/Connection/ConnectionController.cs
@@ -1,26 +1,27 @@
 namespace ServiceControl.Audit.Connection
 {
-    using System.Net;
-    using System.Net.Http;
-    using System.Threading.Tasks;
     using System.Web.Http;
     using Infrastructure.Settings;
+    using Newtonsoft.Json;
 
     public class ConnectionController : ApiController
     {
         readonly Settings settings;
+        readonly JsonSerializerSettings jsonSerializerSettings = new JsonSerializerSettings();
 
         public ConnectionController(Settings settings) => this.settings = settings;
 
         [Route("connection")]
         [HttpGet]
-        public Task<HttpResponseMessage> GetConnectionDetails() => Task.FromResult(Request.CreateResponse(HttpStatusCode.OK, new
-        {
-            settings.AuditQueue,
-            SagaAudit = new
+        public IHttpActionResult GetConnectionDetails() =>
+            Json(new
             {
-                SagaAuditQueue = settings.AuditQueue
-            }
-        }));
+                settings.AuditQueue,
+                SagaAudit = new
+                {
+                    SagaAuditQueue = settings.AuditQueue
+                }
+            },
+            jsonSerializerSettings);
     }
 }
--- a/src/ServiceControl.Audit/Connection/ConnectionController.cs
+++ b/src/ServiceControl.Audit/Connection/ConnectionController.cs
@@ -16,7 +16,10 @@ namespace ServiceControl.Audit.Connection
         public IHttpActionResult GetConnectionDetails() =>
             Json(new
             {
-                settings.AuditQueue,
+                MessageAudit = new
+                {
+                    settings.AuditQueue
+                },
                 SagaAudit = new
                 {
                     SagaAuditQueue = settings.AuditQueue
--- a/src/ServiceControl.Audit/Connection/ConnectionController.cs
+++ b/src/ServiceControl.Audit/Connection/ConnectionController.cs
@@ -16,17 +16,31 @@ namespace ServiceControl.Audit.Connection
         public IHttpActionResult GetConnectionDetails() =>
             Json(new
             {
-                MessageAudit = new
+                MessageAudit = new MessageAuditConnectionDetails
                 {
                     Enabled = true,
-                    settings.AuditQueue
+                    AuditQueue = settings.AuditQueue
                 },
-                SagaAudit = new
+                SagaAudit = new SagaAuditConnectionDetails
                 {
                     Enabled = true,
                     SagaAuditQueue = settings.AuditQueue
                 }
             },
             jsonSerializerSettings);
+    }
+
+    // HINT: This should match the type in the PlatformConnector package
+    class MessageAuditConnectionDetails
+    {
+        public bool Enabled { get; set; }
+        public string AuditQueue { get; set; }
+    }
+
+    // HINT: This should match the type in the PlatformConnector package
+    class SagaAuditConnectionDetails
+    {
+        public bool Enabled { get; set; }
+        public string SagaAuditQueue { get; set; }
     }
 }
--- a/src/ServiceControl.Audit/Connection/ConnectionController.cs
+++ b/src/ServiceControl.Audit/Connection/ConnectionController.cs
@@ -18,10 +18,12 @@ namespace ServiceControl.Audit.Connection
             {
                 MessageAudit = new
                 {
+                    Enabled = true,
                     settings.AuditQueue
                 },
                 SagaAudit = new
                 {
+                    Enabled = true,
                     SagaAuditQueue = settings.AuditQueue
                 }
             },
--- a/src/ServiceControl.MultiInstance.AcceptanceTests/Connection/PlatformConnectionTests.cs
+++ b/src/ServiceControl.MultiInstance.AcceptanceTests/Connection/PlatformConnectionTests.cs
@@ -53,32 +53,39 @@
 
             if (!string.IsNullOrWhiteSpace(TransportIntegration.ConnectionString))
             {
-                var builder = new DbConnectionStringBuilder { ConnectionString = TransportIntegration.ConnectionString };
-
-                // SQS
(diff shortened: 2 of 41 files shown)


mikeminutillo:
The acceptance tests are failing because we change the folder structure on the build server. The compile step finds all of the AcceptanceTest assemblies and exports them as artifacts. When the tests run in a later step, they don't have access to the ApprovalFiles. Not only that, but because we move the assemblies to a new location, the approver tries to write to a folder that it doesn't have permission to.

mikeminutillo:
We resolved the issue with the approval tests inside of an acceptance testing project on the build server by:
- Moving the acceptance test DLL to a location at least 3 deep within the build artifacts (Particular.Approvals is assuming this)
- Copying the ApprovalFiles folder into the correct location within the build artifacts

We also resolved some of the issues with the approval tests that work locally on a developer machine that do not work the same way on the build server. i.e. SQL database name is different, SQS Prefix

mikeminutillo:
The SQS test is still broken because the local dev environment connection string includes a queue prefix and the build server connection string does not. We will need to update the build server connection string to include a queue prefix.

src/ServiceControl.Audit/Connection/ConnectionController.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        readonly Settings settings;
+        readonly JsonSerializerSettings jsonSerializerSettings = new JsonSerializerSettings();
+
+        public ConnectionController(Settings settings) => this.settings = settings;
+
+        [Route("connection")]
+        [HttpGet]
+        public IHttpActionResult GetConnectionDetails() =>
+            Json(new
```
tmasternak, line 17: Should we have a separate type for the result and approvals to ensure we detect any breaking changes?

src/ServiceControl.MultiInstance.AcceptanceTests/Connection/PlatformConnectionTests.cs
```diff
@@ -0,0 +1,140 @@
...
+                            result = result.Replace(
+                                $"[{databaseAsString}]",
+                                "[DATABASE]"
+                            );
+                        }
+                    }
+                }
+                catch
+                {
+                    // NOTE: Learning Transport has a connection string in an invalid format
```
tmasternak, line 88: Should we detect LearningTransport explicitly and exit `Scrub` early in such a case?

src/ServiceControl/Connection/PlatformConnectionBuilder.cs
```diff
@@ -0,0 +1,26 @@
...
+        readonly IProvidePlatformConnectionDetails[] platformConnectionProviders;
+
+        public PlatformConnectionBuilder(IProvidePlatformConnectionDetails[] platformConnectionProviders)
+            => this.platformConnectionProviders = platformConnectionProviders;
+
+        public async Task<PlatformConnectionDetails> BuildPlatformConnection()
+        {
+            var connectionDetails = new PlatformConnectionDetails();
+
+            // TODO: Handle errors in providers (including timeouts?)
```
tmasternak, line 17: This is [done already](https://github.com/Particular/ServiceControl/pull/2674/files#diff-8b54e2ea88c21bb9db165c40c2a65eb0f58d8fa36322c86da684a211522214bcR33) I think.

src/ServiceControl/Connection/PlatformConnectionDetails.cs
```diff
@@ -0,0 +1,29 @@
...
+        readonly ConcurrentDictionary<string, object> values = new ConcurrentDictionary<string, object>();
+
+        public void Add(string key, object value)
+        {
+            if (values.TryAdd(key, value))
+            {
+                return;
+            }
+
+            // Add a numeric suffix to duplicated keys
```
tmasternak, line 17:
```suggestion
            // Add a numeric suffix to duplicated keys e.g. can happen when there are two audit instances configured
```

src/ServiceControl/Connection/PlatformConnectionDetails.cs
```diff
@@ -0,0 +1,29 @@
...
+            var suffix = 0;
+            do
+            {
+                suffix++;
+            } while (!values.TryAdd($"{key}{suffix}", value));
```
tmasternak, line 22: Should we do this only when the key values a different e.g. if there are two audit instances but have the same audit queue name we should not require any user intervention.

src/ServiceControl.UnitTests/ApprovalFiles/APIApprovals.PublicClr.approved.txt
```diff
@@ -247,6 +247,32 @@ namespace ServiceControl.CompositeViews.Messages
...
+        System.Threading.Tasks.Task<ServiceControl.Connection.PlatformConnectionDetails> BuildPlatformConnection();
+    }
+    public class PlatformConnectionDetails
+    {
+        public PlatformConnectionDetails() { }
+        public System.Collections.Concurrent.ConcurrentBag<string> Errors { get; }
+        public void Add(string key, object value) { }
+        public System.Collections.Generic.IDictionary<string, object> ToDictionary() { }
+    }
+    public class PlatformConnectionQueryStatus
```
tmasternak, line 270: This is not used anymore I think.

src/ServiceControl/Connection/ConnectionController.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        public ConnectionController(IPlatformConnectionBuilder connectionBuilder) => this.connectionBuilder = connectionBuilder;
+
+        [Route("connection")]
+        [HttpGet]
+        public async Task<IHttpActionResult> GetConnectionDetails()
+        {
+            var connectionDetails = await connectionBuilder.BuildPlatformConnection().ConfigureAwait(false);
+
+            return Json(
```
tmasternak, line 20: Should we have a dedicated type for this response?

src/ServiceControl/Connection/RemotePlatformConnectionDetailsProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+            using (var client = httpClientFactory())
+            {
+                try
+                {
+                    var result = await client.GetStringAsync($"{remote.ApiUri}/connection")
+                        .ConfigureAwait(false);
+                    var dictionary = JsonConvert.DeserializeObject<Dictionary<string, dynamic>>(result);
+                    if (dictionary == null)
+                    {
+                        Log.Warn($"Unexpected response from ${remote.ApiUri}/connection: ${result}");
```
tmasternak, line 40:
```suggestion
                        Log.Warn($"Unexpected response from {remote.ApiUri}/connection: ${result}");
```

src/ServiceControl/Connection/RemotePlatformConnectionDetailsProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+                    }
+
+                    foreach (var kvp in dictionary)
+                    {
+                        connection.Add(kvp.Key, kvp.Value);
+                    }
+                }
+                catch (Exception ex)
+                {
+                    var message = $"Unable to get connection details from ${remote.ApiUri}/connection";
```
tmasternak, line 51:
```suggestion
                    var message = $"Unable to get connection details from ServiceControl Audit instance at {remote.ApiUri}/connection.";
```

src/ServiceControl/HostingComponent.cs
```diff
@@ -1,14 +1,19 @@
...
     class HostingComponent : ServiceControlComponent
     {
-        public override void Configure(Settings settings, IHostBuilder hostBuilder)
-        {
-        }
+        public override void Configure(Settings settings, IHostBuilder hostBuilder) =>
+            hostBuilder.ConfigureServices(services =>
+            {
+                services.AddPlatformConnectionProvider<RemotePlatformConnectionDetailsProvider>();
+                services.AddTransient<IPlatformConnectionBuilder, PlatformConnectionBuilder>();
```
tmasternak, line 15: Why transient?

src/ServiceControl/Monitoring/HeartbeatsPlatformConnectionDetailsProvider.cs
```diff
@@ -0,0 +1,41 @@
...
+            this.settings = settings;
+            instanceMainQueue = endpointSettings.LocalAddress();
+        }
+
+        public Task ProvideConnectionDetails(PlatformConnectionDetails connection)
+        {
+            // NOTE: The default grace period is 40s and the default frequency is 10s.
+            // In a low-latency environment, an endpoint would need to miss more than 4 heartbeats to be considered down
+            var frequency = TimeSpan.FromTicks(settings.HeartbeatGracePeriod.Ticks / 4);
+            // TODO: Figure out if we should even provide this. The default in the plugin is 4xFrequency
```
tmasternak, line 26: I think we decided that we should keep this. Right?

tmasternak (changes requested):

src/ServiceControl.Audit/Connection/ConnectionController.cs
mikeminutillo: I added concrete types. There is an approval test for each transport in the multi-instance tests

src/ServiceControl/Connection/PlatformConnectionDetails.cs
mikeminutillo: The ones which are likely to be duplicates (Message and Saga Audit) are deserialized as a dynamic json object which makes it challenging to determine equality. I think this is an optimization that can come later.

If they aren't equal, then we wouldn't be able to deduplicate, we'll show both, and the customer will have to hand-edit the file to get their desired results.

If they are equal, they we could in theory deduplicate but if we don't then the endpoint will still get details that work.

src/ServiceControl.MultiInstance.AcceptanceTests/Connection/PlatformConnectionTests.cs
mikeminutillo: Moved the scrubber to the transport integration classes so each one can provide only the scrubbing needed for it's own idiosyncracies

src/ServiceControl/Connection/PlatformConnectionDetails.cs
tmasternak: Makes sense.

tmasternak (approved):
image

PR #2694, opened 2021-11-13
SPIKE: Release action use net runtime 4.8-windowsservercore-ltsc2019 (cached image)

DavidBoike:


DavidBoike:
Changing each dockerfile from using the base image tag `4.7.2-windowsservercore-ltsc2016` to `4.8-windowsservercore-ltsc2019` improved performance of the release workflow drastically because the GitHub Action runners already have `4.8-windowsservercore-ltsc2019` preloaded and don't have to download a 13.3GB image before getting started.

This is a spike, so I'm closing, but these were the results:

### With `4.7.2-windowsservercore-ltsc2016`

image

### With `4.8-windowsservercore-ltsc2019`

image

PR #2768, opened 2022-01-18
Use Microsoft.Data.SqlClient for Sql Connection String Validation

### LINKED ISSUE #2767: ServiceControl Installer fails for Sql connection string with Azure Active Directory authentication
The Sql Transport uses Microsoft.Data.SqlClient for interacting with Sql Server databases. This is good and necessary to be able to connect to Azure Sql using Azure Active Directory credentials. This works fine having installed a Service Control instance with the Sql Transport if one is able to work around this problem with installation of the instance.

For example, the following connection string can be used successfully in a context with an Azure Active Directory Managed Identity to authenticate. If an already installed Service Control instance (or other type of endpoint) is using this connection string in an Azure context, like a virtual machine, with a managed identity, it works flawlessly.

```
Server=sometransportdatabaseserver.database.windows.net;Database=sometransportdatabase;Authentication=Active Directory Managed Identity;
```

Installing such an instance, though, fails via both the Service Control Management graphical tool and the PowerShell modules. This is because the ServiceControlInstallerEngine library validates connection strings by attempting to connect to the database specified. It does so using System.Data.SqlClient in the `ConnectionStringValidator` class, which does not support the use of Azure Active Directory authentication types in the connection string.

```
Connection String is invalid - Invalid value for key 'authentication'.
```

To make a seamless experience when using Sql for transport, it's necessary for such connection strings to successfully validate.

I want to script the setup of these instances with the PowerShell modules as a part of a broader Terraform environment setup and this issue is standing in my way.

raelyard:
To install instances with ServiceControl Management and/or the PowerShell modules using an Azure Active Directory identity to configure the connection string for the Sql Transport.

If it's not desirable to add this dependency in this place, there are probably other ways it could be managed, but if connecting to the database from this library is the desired method of validation, this is the shortest path.

This doesn't seem it should impact any of the existing tests, but getting the tests running on my machine seems it would take setting up dependencies in a way that was going to take more cycles than I wanted in the moment. I didn't add any test for this here. It's not obvious to me what that would like like with trying to test Azure in the contexts the tests run, so will leave that for something to talk about here if it's necessary/desirable.

This resolves #2767

### DIFF
--- a/src/ServiceControlInstaller.Engine/ServiceControlInstaller.Engine.csproj
+++ b/src/ServiceControlInstaller.Engine/ServiceControlInstaller.Engine.csproj
@@ -13,6 +13,7 @@
 
   <ItemGroup>
     <PackageReference Include="DotNetZip" Version="1.16.0" />
+    <PackageReference Include="Microsoft.Data.SqlClient" Version="2.1.2" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
     <PackageReference Include="System.Reflection.MetadataLoadContext" Version="6.0.0" />
     <PackageReference Update="System.Runtime.CompilerServices.Unsafe" Version="6.0.0" />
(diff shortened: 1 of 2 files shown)


kentdr (approved):

kentdr:
Thanks for the submission @raelyard!

PR #2775, opened 2022-01-31
Remove the IsDebugEnabled check from audit expiration

WilliamBZA:
`IsDebugEnabled` appears to be incorrectly set to `false` possibly due to caching of the value in [core](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Logging/DefaultLoggerFactory.cs#L31-L35). This makes it impossible to get insight into the expiration cleanup process.

This PR removes the `IsDebugEnabled` check from the expiration process. This process is not on the hot path for ingestion so skipping the check here should not incur a major performance penalty.

### DIFF
--- a/src/ServiceControl.Audit/Auditing/BodyStorage/BodyStorageFeature.cs
+++ b/src/ServiceControl.Audit/Auditing/BodyStorage/BodyStorageFeature.cs
@@ -1,6 +1,5 @@
 ﻿namespace ServiceControl.Audit.Auditing.BodyStorage
 {
-    using System;
     using System.Collections.Generic;
     using System.Text;
     using System.Threading.Tasks;
--- a/src/ServiceControl.Audit/Infrastructure/RavenDB/EmbeddedRavenDbHostedService.cs
+++ b/src/ServiceControl.Audit/Infrastructure/RavenDB/EmbeddedRavenDbHostedService.cs
@@ -57,6 +57,5 @@
         }
 
         static readonly ILog Logger = LogManager.GetLogger(typeof(EmbeddedRavenDbHostedService));
-
     }
 }
--- a/src/ServiceControl.SagaAudit/SagaHistoryCleaner.cs
+++ b/src/ServiceControl.SagaAudit/SagaHistoryCleaner.cs
@@ -77,33 +77,21 @@
 
             var deletionCount = Chunker.ExecuteInChunks(items.Count, (itemsForBatch, db, s, e) =>
             {
-                if (logger.IsDebugEnabled)
-                {
-                    logger.Debug($"Batching deletion of {s}-{e} saga history documents.");
-                }
+                logger.Debug($"Batching deletion of {s}-{e} saga history documents.");
 
                 var results = db.Batch(itemsForBatch.GetRange(s, e - s + 1), CancellationToken.None);
-                if (logger.IsDebugEnabled)
-                {
-                    logger.Debug($"Batching deletion of {s}-{e} saga history documents completed.");
-                }
+                logger.Debug($"Batching deletion of {s}-{e} saga history documents completed.");
 
                 return results.Count(x => x.Deleted == true);
             }, items, database, cancellationToken);
 
             if (deletionCount == 0)
             {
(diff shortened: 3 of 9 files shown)


src/ServiceControl.SagaAudit/SagaHistoryCleaner.cs
```diff
@@ -77,33 +77,21 @@ public static void Clean(int deletionBatchSize, DocumentDatabase database, DateT
 
             var deletionCount = Chunker.ExecuteInChunks(items.Count, (itemsForBatch, db, s, e) =>
             {
-                if (logger.IsDebugEnabled)
-                {
-                    logger.Debug($"Batching deletion of {s}-{e} saga history documents.");
-                }
+                logger.Debug($"Batching deletion of {s}-{e} saga history documents.");
```
danielmarbach, line 80: This is a more general comment for all changes in this PR, where we still use interpolation. For consistency, would it make sense to also use the format-based overloads, like you did in some other changes in this PR?

src/ServiceControl/Infrastructure/RavenDB/Expiration/ExpiredDocumentsCleaner.cs
```diff
@@ -4,8 +4,8 @@
     using System.Threading;
     using System.Threading.Tasks;
     using BackgroundTasks;
+    using NServiceBus.Logging;
     using Raven.Abstractions;
-    using Raven.Abstractions.Logging;
```
danielmarbach, line 8: Ah this was probably still around because this used to be a bundle...

src/ServiceControl.Audit/Auditing/BodyStorage/BodyStorageFeature.cs
```diff
@@ -1,6 +1,5 @@
 ﻿namespace ServiceControl.Audit.Auditing.BodyStorage
 {
-    using System;
```
danielmarbach, line 3: I guess this change is not strictly necessary. I'm good though to take it in

src/ServiceControl.Audit/Infrastructure/RavenDB/EmbeddedRavenDbHostedService.cs
```diff
@@ -57,6 +57,5 @@ public Task StopAsync(CancellationToken cancellationToken)
         }
 
         static readonly ILog Logger = LogManager.GetLogger(typeof(EmbeddedRavenDbHostedService));
-
```
danielmarbach, line 60: I guess this change is not strictly necessary. I'm good though to take it in

danielmarbach (approved):

src/ServiceControl.Audit/Auditing/BodyStorage/BodyStorageFeature.cs
WilliamBZA: Tidying up as I go along

src/ServiceControl.Audit/Infrastructure/RavenDB/EmbeddedRavenDbHostedService.cs
WilliamBZA: Tidying up as I go along

src/ServiceControl.SagaAudit/SagaHistoryCleaner.cs
WilliamBZA: I've actually gone the other way and have gone for interpolation rather than using the format overloads. My reasons:

* It's easier to read what the value when compared to looking at the index number and comparing to the args array
* If we in the future support structured logging, we may be able to pull more meaning from the logs.

danielmarbach: I'm not an expert on this. If I read this right, we would have to use the template-based approach outlined in 
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-6.0#log-message-template
docs.microsoft.comdocs.microsoft.com

Which means "not using interpolation".

PR #2782, opened 2022-02-04
Revert "Bump System.Reflection.MetadataLoadContext from 5.0.1 to 6.0.…

### LINKED ISSUE #2780: Powershell cmdlet sc-upgrade fails but SCMU servicecontrol upgrade works
I've Installed 4.21.4 and tried to upgrade an existing servicecontrol instance via the sc-upgrade powershell cmdlet.  

This fails with the error:

```
sc-upgrade : Failed to retrieve release date for new version
```
Updating the same instance via SCMU works fine

raelyard:
This reverts commit 2b3a950a0bcdac95c6209d09c261654400f09ef6 because
it's the cause of #2780.

This resolved #2780


WilliamBZA:
Thanks @raelyard. Rather than revert though, we want to address the binding redirect issues.  Replaced by https://github.com/Particular/ServiceControl/pull/2784

PR #2783, opened 2022-02-07
ReAdd IsDebugEnabled check for logging

WilliamBZA:
Re-adds the `IsDebugEnabled` check that was removed in #2775.

The explicit call to `builder.SetMinimumLevel(loggingSettings.ToHostLogLevel());` sets the logging level correctly between the instances of `NServiceBus.Extensions.Hosting.DeferredLogger` and `NServiceBus.Extensions.Hosting.Logger` which means the fix for #2775 is no longer required.

### DIFF
--- a/src/ServiceControl.Audit/Infrastructure/Bootstrapper.cs
+++ b/src/ServiceControl.Audit/Infrastructure/Bootstrapper.cs
@@ -59,6 +59,7 @@ namespace ServiceControl.Audit.Infrastructure
                     builder.ClearProviders();
                     //HINT: configuration used by NLog comes from LoggingConfigurator.cs
                     builder.AddNLog();
+                    builder.SetMinimumLevel(loggingSettings.ToHostLogLevel());
                 })
                 .UseServiceProviderFactory(new AutofacServiceProviderFactory())
                 .ConfigureServices(services =>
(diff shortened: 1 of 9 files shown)


ramonsmits (approved):
+1 for restoring the `if(logger.IsDebugEnabled` on the log entries that use string interpolation. 

Regarding `builder.SetMinimumLevel(loggingSettings.ToHostLogLevel())`, that is required as AFAIK the NLog configuration is independant on the logging extensions. For example, NLog verbosity could be set to Trace, and logging extensions to Info meaning `logger.Debug` would not log anything.

Reason this is different for the `DeferredLogger` is that only at the start `IsDebugEnabled` likely returns true until the endpoint is started if NLog was configured to only log Info or more severe. The items temporarily stored in the deferredlogger instance will be flushed but none of them will be written.

Maybe @danielmarbach can comment on my deferred logger assumptions.

I'm not really sure how the NServiceBus hosting extensions work. I do think there is a code smell, the deferred logger instanced shouldn't be needed due the way how start tasks currently work it is required but IMHO should be improved. Currently you could for example not cache `IsDebugEnabled` in a `readonly bool `IsDebugEnabled = logger.IsDebugEnabled` to JIT compile all these `if`'s from the runtime code.

Small comment on the PR, these 2 changes are independant (IsDebugEnabled vs init) and maybe consider 2 PR's?

PR #2855, opened 2022-05-06
Turn casts into polymorphic dispatch

mikeminutillo:
Whenever I see code checking types and casting, I want to turn it into a polymorphic dispatch.


andreasohlund:
I like it, should we remove the properties as well since we only use them for testing?

andreasohlund:
But then we can’t test it as easily so lets keep them?

PR #2863, opened 2022-05-16
Remove usage of custom NSB Features for enricher configuration in Auditing

timbussmann:


src/ServiceControl.Audit/Infrastructure/RavenDB/RavenHostBuilderExtensions.cs
```diff
@@ -17,6 +19,8 @@ static class RavenHostBuilderExtensions
 
                 serviceCollection.AddSingleton<IDocumentStore>(embeddedDocumentStore);
                 serviceCollection.AddHostedService<EmbeddedRavenDbHostedService>();
+
+                serviceCollection.AddSingleton<IBodyStorage, RavenAttachmentsBodyStorage>();
```
SzymonPobiega, line 23: Technically the previous code allowed customization of the body storage by dropping a dll into the bin folder that contains a feature that registers a body storage. But that would depend on the feature activation order so I think it is fine to lock this faulty extension point up (for now).

👍

SzymonPobiega (approved):
LGTM

src/ServiceControl.Audit/Infrastructure/RavenDB/RavenHostBuilderExtensions.cs
timbussmann: yeah it removes a potential "magic" extension point but if we start to make use of that, I think we should make it explicit (especially if there might be a chance that we won't use NSB internally anymore)

cquirosj (approved):

src/ServiceControl.Audit/Auditing/AuditIngestionComponent.cs
```diff
@@ -35,11 +39,12 @@ class AuditIngestionComponent
...
             IDocumentStore documentStore,
+            IBodyStorage bodyStorage,
             RawEndpointFactory rawEndpointFactory,
             LoggingSettings loggingSettings,
-            BodyStorageFeature.BodyStorageEnricher bodyStorageEnricher,
-            IEnrichImportedAuditMessages[] enrichers,
-            AuditIngestionCustomCheck.State ingestionState
+            AuditIngestionCustomCheck.State ingestionState,
+            EndpointInstanceMonitoring endpointInstanceMonitoring,
+            IEnumerable<IEnrichImportedAuditMessages> auditEnrichers // allows extending message enrichers with custom enrichers registered in the DI container
```
timbussmann, line 47: I had to bring this back in since this "feature" is being used by the acceptance tests. Not excited to keep this in just for the acceptance tests but than can be solved separately.

src/ServiceControl.Audit.UnitTests/ApprovalFiles/APIApprovals.PublicClr.approved.txt
```diff
@@ -5,11 +5,6 @@
 [assembly: System.Runtime.Versioning.TargetFramework(".NETFramework,Version=v4.7.2", FrameworkDisplayName=".NET Framework 4.7.2")]
 namespace ServiceControl.Audit.Auditing
 {
-    public class DefaultEnrichers : NServiceBus.Features.Feature
```
timbussmann, line 8: I couldn't find any reference to this feature

PR #2864, opened 2022-05-16
Fix embedded resource path for help text file

timbussmann:
Currently, using `--help` on the audit instance's executable returns no help text as the path to the embedded resource isn't correct  as the text file is located in a different folder than the file in the primary instance.

### DIFF
--- a/src/ServiceControl.Audit/Infrastructure/Hosting/HostArguments.cs
+++ b/src/ServiceControl.Audit/Infrastructure/Hosting/HostArguments.cs
@@ -149,7 +149,7 @@ namespace ServiceControl.Audit.Infrastructure.Hosting
             using (
                 var stream =
                     Assembly.GetCallingAssembly()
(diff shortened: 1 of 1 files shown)


src/ServiceControl.Audit/Infrastructure/Hosting/HostArguments.cs
```diff
@@ -149,7 +149,7 @@ public void PrintUsage()
             using (
                 var stream =
                     Assembly.GetCallingAssembly()
-                        .GetManifestResourceStream("ServiceControl.Hosting.Help.txt"))
+                        .GetManifestResourceStream("ServiceControl.Audit.Infrastructure.Hosting.Help.txt"))
             {
                 if (stream != null)
```
ramonsmits, line 154: I think this should be an assert or an invalidoperationexception? It is expected that it is not null

SzymonPobiega (approved):

PR #2871, opened 2022-05-19
Removing NSB and Autofac from audit setup bootstraper

cquirosj:
Removing NSB from setup bootstraper.

### DIFF
--- a/src/ServiceControl.Audit/Infrastructure/SetupBootstrapper.cs
+++ b/src/ServiceControl.Audit/Infrastructure/SetupBootstrapper.cs
@@ -2,13 +2,9 @@ namespace ServiceControl.Audit.Infrastructure
 {
     using System.Collections.Generic;
     using System.Threading.Tasks;
-    using Auditing.BodyStorage;
-    using Auditing.BodyStorage.RavenAttachments;
-    using Autofac;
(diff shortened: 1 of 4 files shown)


src/ServiceControl.Audit/Infrastructure/SetupBootstrapper.cs
```diff
@@ -61,52 +56,12 @@ public async Task Run(string username)
...
-
-            var configuration = new EndpointConfiguration(settings.ServiceName);
-            var assemblyScanner = configuration.AssemblyScanner();
-            assemblyScanner.ExcludeAssemblies("ServiceControl.Plugin");
-            if (excludeAssemblies != null)
-            {
-                assemblyScanner.ExcludeAssemblies(excludeAssemblies);
-            }
-
-            configuration.EnableInstallers(username);
```
timbussmann, line 73: doesn't that mean that we're no longer running the [CreateEventSource](https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl.Audit/Infrastructure/Installers/CreateEventSource.cs) installer?

timbussmann:
@cquirosj I think this PR is no longer needed and can be closed, right?

PR #2876, opened 2022-05-23
Move audit queue ingestion logic into hosted service

timbussmann:
Remove the logic from the `AuditIngestionComponent` to separate the "queue ingestion logic" and the "audit processing logic". Once the failed import processing logic is extracted as well, the "audit processing logic" can more easily be registered via DI and be reused by whatever ingestion mechanism.


cquirosj (approved):

src/ServiceControl.Audit/Auditing/AuditIngestor.cs
```diff
@@ -59,16 +102,7 @@ await Forward(stored, settings.AuditLogQueue)
             }
         }
 
-        public async Task Initialize(IDispatchMessages dispatcher)
-        {
-            this.dispatcher = dispatcher;
-            if (settings.ForwardAuditMessages)
```
SzymonPobiega, line 65: @timbussmann this check is now removed after the `Initialize` method has been moved to `AuditIngestion`.

PR #2877, opened 2022-05-23
Split ingestion transport from SC transport

timbussmann:
The `AuditPersister` used both the full NServiceBus endpoint's `IMessageSession` as well as the raw transport's `IDispatchMessages` to send messages. Since there is no need to float the raw transport dispatcher to the persisted, it can use the dispatcher from the NServiceBus endpoint instead to further split the transport usage from the "queue ingestion logic", using the raw transport, and the "audit processing logic", using the NServiceBus endpoint (which could technically use a different transport).

Since the `IDispatchMessages` cannot be resolved directly via DI until the NServiceBus endpoint has been fully started (a limit that NSB handles with a wrapper class for the `IMessageSession`), the dispatcher needs to be resolved at "runtime".

### DIFF
--- a/src/ServiceControl.Audit/Auditing/AuditIngestionComponent.cs
+++ b/src/ServiceControl.Audit/Auditing/AuditIngestionComponent.cs
@@ -1,5 +1,6 @@
 ﻿namespace ServiceControl.Audit.Auditing
 {
+    using System;
     using System.Collections.Generic;
     using System.Diagnostics;
     using System.Linq;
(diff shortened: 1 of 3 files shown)


timbussmann:
Thinking a bit more about this, this PR should maybe not be merged:
The only usage of the dispatcher is to send the retry-confirmation message to recoverability in case the origin endpoint is using an older NServiceBus version which can't send the retry-ack message directly to SC. In that case, Auditing "emulates" the endpoint behavior by sending the confirmation message on the endpoint's behalf. This all happens on the "user transport" since that's where the retry-ack message would also be sent by a regular endpoint. In this case, it would make sense to keep that operation on the "user transport" and not move it to the "SC backplane transport" (conceptually). From this perspective, the current implementation still separates these concepts but it's harder to understand that from the genering `AddForSend`-APIs which don't contain this type of information.

SzymonPobiega:
@timbussmann I think you are right. I missed the fact that this "send on behalf of the endpoint" goes to the error queue and not to the main input queue.

PR #2884, opened 2022-05-30
Remove custom Autofac constructor finder

timbussmann:
Removes the `AllConstructorFinder` for Autofac that is used to allow resolving the public controller classes that have private actors. If the goal is to move away from Autofac, we should avoid these DI customizations as other DI containers (e.g. `ServiceProvider`) won't support this and need the constructors to be public.

To prevent the rippling effect of making all controller dependencies public for now, a custom controller resolver type, inspired by the default implementation, allows resolving controllers that aren't public classes instead.

### DIFF
--- a/src/ServiceControl.Audit/Infrastructure/OWIN/Startup.cs
+++ b/src/ServiceControl.Audit/Infrastructure/OWIN/Startup.cs
@@ -2,9 +2,11 @@
 {
     using System;
     using System.Collections.Generic;
+    using System.Linq;
     using System.Net.Http.Headers;
     using System.Reflection;
     using System.Web.Http;
+    using System.Web.Http.Controllers;
     using System.Web.Http.Dependencies;
     using System.Web.Http.Dispatcher;
     using Autofac;
@@ -33,8 +35,8 @@
 
                 config.DependencyResolver = new ExternallyOwnedContainerDependencyResolver(new AutofacWebApiDependencyResolver(container));
                 config.Services.Replace(typeof(IAssembliesResolver), new OnlyExecutingAssemblyResolver(additionalAssembly));
+                config.Services.Replace(typeof(IHttpControllerTypeResolver), new InternalControllerTypeResolver());
                 config.MapHttpAttributeRoutes();
-
                 map.UseCors(Cors.AuditCorsOptions);
 
                 config.MessageHandlers.Add(new XParticularVersionHttpHandler());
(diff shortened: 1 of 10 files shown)


cquirosj:
> public controller classes that have private `actors`.

Did you mean ctors? autocompletion I guess. 

___

What is the motivation/driver for having controllers not being public? I am sure there is a good reason, just looking to know more about it. I wonder if we could introduce a test that fails if it detects a public controller is added in the future, as living documentation to explain the design decision.

cquirosj (approved):

timbussmann:
> Did you mean ctors? autocompletion I guess.

lol yes. Fixed.

src/ServiceControl.Audit/Infrastructure/OWIN/Startup.cs
```diff
@@ -89,4 +91,40 @@ public override ICollection<Assembly> GetAssemblies()
...
+                    controllerTypes.AddRange(source.Where(t => t != null && t.IsClass && !t.IsAbstract && typeof(IHttpController).IsAssignableFrom(t) && HasValidControllerName(t)));
+                }
+            }
+            return controllerTypes;
+        }
+
+        internal static bool HasValidControllerName(Type controllerType)
+        {
+            string controllerSuffix = DefaultHttpControllerSelector.ControllerSuffix;
+            return controllerType.Name.Length > controllerSuffix.Length && controllerType.Name.EndsWith(controllerSuffix, StringComparison.OrdinalIgnoreCase);
```
mikeminutillo, line 127: We should probably have a unit test that looks for controllers that don't have this suffix because they'll just quietly not get picked up and it might be frustrating to figure out why.

src/ServiceControl.Audit/Infrastructure/WebApi/WebApiHostBuilderExtensions.cs
```diff
@@ -41,20 +41,8 @@ static void RegisterInternalWebApiControllers(ContainerBuilder containerBuilder)
 
             foreach (var controllerType in controllerTypes)
             {
-                containerBuilder.RegisterType(controllerType).FindConstructorsWith(new AllConstructorFinder());
+                containerBuilder.RegisterType(controllerType);
```
mikeminutillo, line 44: I wonder if this is needed anymore. Will the ASP.NET infra already add these controllers now? I'm guessing that's the next step.

mikeminutillo (approved):

src/ServiceControl.Audit/Infrastructure/WebApi/WebApiHostBuilderExtensions.cs
cquirosj: Not sure if this is what you are referring to @mikeminutillo: according to this [documentation](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.applicationparts.applicationpartattribute?view=aspnetcore-6.0) as long as a library references MVC, an instance of ApplicationPartAttribute will be auto generated and the library will be treated as an application part

timbussmann: I'll check if we can just remove the explicit registration

src/ServiceControl.Audit/Infrastructure/OWIN/Startup.cs
timbussmann: not opposed to that, but shouldn't we expect that every of the exposed APIs is somehow being used as part of acceptance tests anyway?

src/ServiceControl.Audit/Infrastructure/WebApi/WebApiHostBuilderExtensions.cs
timbussmann: If I remove these registrations, the web server throws 500 errors because of missing empty actors. I'm not sure whether this is a limitation of older versions of web API, OWIN, Autofac or something else. I'll add a note to our POA to see if would still be needed without Autofac.

mikeminutillo: No worries. I just figured that most ASP.NET applications don't have to manually register controllers individually so we must be doing something funky. Maybe we need an `.AddControllers()` somewhere or something.

PR #2888, opened 2022-06-08
INeedToInstall usage removal

cquirosj:
First small step towards removing usage of NSB during setup


src/ServiceControl.Audit/Infrastructure/CreateEventSource.cs
```diff
@@ -1,12 +1,11 @@
-﻿namespace ServiceControl.Audit.Infrastructure.Installers
+﻿namespace ServiceControl.Audit.Infrastructure
 {
     using System.Diagnostics;
     using System.Threading.Tasks;
-    using NServiceBus.Installation;
 
-    public class CreateEventSource : INeedToInstallSomething
+    class EventSource
```
timbussmann, line 6: should the filename be adjusted too?

src/ServiceControl.Audit/Infrastructure/CreateEventSource.cs
```diff
@@ -1,12 +1,11 @@
...
     {
-        public Task Install(string identity)
+        public Task Create()
```
timbussmann, line 8: any reason to not make this static (and `void`)?

SzymonPobiega (approved):

src/ServiceControl.Audit/Infrastructure/CreateEventSource.cs
cquirosj: No special reason. I had it static at first but with the task. Then I thought about asking if we have any preference for the static part. For the void part, it makes sense to make it so.

PR #2890, opened 2022-06-09
Retrive release date from correct attribute

timbussmann:
With the [switch to MinVer](https://github.com/Particular/ServiceControl/pull/2827) and the latest versions of the licensing code, the SC specific code to fetch the release date in the installer hasn't been updated and caused an error when trying to add new instances via the SCMU with the latest version.

The release date seems to no longer be set on the `ReleaseDateAttribute` but added as a `AssemblyMetadataAttriubute` instead. The retrieval of the date is a bit more complicated than the default implementation available from the licensing package since the installer loads the assembly only into the `MetadataLoadContext` which doesn't allow `GetCustomAttributes` APIs.

### DIFF
--- a/src/ServiceControlInstaller.Engine/FileSystem/ReleaseDateReader.cs
+++ b/src/ServiceControlInstaller.Engine/FileSystem/ReleaseDateReader.cs
@@ -2,9 +2,7 @@
 {
     using System;
     using System.Collections.Generic;
-    using System.Globalization;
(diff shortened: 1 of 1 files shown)


bording:
I have opened #2891 as an alternative approach to fixing this that should also ensure that things don't get out of sync again.

timbussmann:
closing this in favour of #2891

PR #2897, opened 2022-06-23
-m parameter causes incorrect build output for Particular.PlatformSample

DavidBoike:
Using the -m parameter (ostensibly to try to get faster builds by building in parallel?) results in the Particular.PlatformSample.ServiceControl package not having any of the ServiceControl assemblies needed to function correctly. Removing the parameter forces the right behavior.


awright18 (approved):

bording:
I've opened #2898 to make this change unnecessary.

PR #2906, opened 2022-07-04
SQL-based Heartbeats component

tmasternak:


src/ServiceControl/Monitoring/HeartbeatMonitoringHostedService.cs
```diff
@@ -10,7 +10,7 @@
 
     class HeartbeatMonitoringHostedService : IHostedService
     {
-        public HeartbeatMonitoringHostedService(EndpointInstanceMonitoring monitor, MonitoringDataStore persistence, IAsyncTimer scheduler, Settings settings)
+        public HeartbeatMonitoringHostedService(EndpointInstanceMonitoring monitor, IMonitoringDataStore persistence, IAsyncTimer scheduler, Settings settings)
```
HEskandari, line 13: To be able to switch between different implementations, do we need to create (and use here) a Factory instead?

src/ServiceControl/Monitoring/SqlDbMonitoringDataStore.cs
```diff
@@ -0,0 +1,126 @@
...
+            using (var connection = new SqlConnection(connectionString))
+            {
+                await connection.OpenAsync().ConfigureAwait(false);
+
+                await connection.ExecuteAsync(
+                    @"IF EXISTS(SELECT * FROM [KnownEndpoints] WHERE Id = @Id)
+                            UPDATE [KnownEndpoints] SET [Monitored] = @Monitored WHERE [Id] = @Id
+                          ELSE
+                            INSERT INTO [KnownEndpoints](Id, HostId, Host, HostDisplayName, Monitored) 
+                            VALUES(@Id, @HostId, @Host, @HostDisplayName, 1)",
```
HEskandari, line 58: Should the last parameter be just `1`? or `@Monitored` instead?

HEskandari (commented):
Had a first look, and made a couple of comments. How do we test this thing?

src/ServiceControl/Monitoring/SqlDbMonitoringDataStore.cs
```diff
@@ -0,0 +1,126 @@
...
+
+            var id = DeterministicGuid.MakeId(endpoint.Name, endpoint.HostId.ToString());
+
+            using (var connection = new SqlConnection(connectionString))
+            {
+                await connection.OpenAsync().ConfigureAwait(false);
+
+                await connection.ExecuteAsync(
+                    @"INSERT INTO [KnownEndpoints](Id, HostId, Host, HostDisplayName, Monitored) 
+                          VALUES(@Id, @HostId, @Host, @HostDisplayName, @Monitored)",
```
jpalac, line 31: Should we be checking here if it already exists before inserting?

src/ServiceControl/Monitoring/SqlDbMonitoringDataStore.cs
```diff
@@ -0,0 +1,126 @@
...
+
+        public SqlDbMonitoringDataStore(EndpointInstanceMonitoring monitoring)
+        {
+            this.monitoring = monitoring;
+        }
+
+        public async Task CreateIfNotExists(EndpointDetails endpoint)
+        {
+            var connectionString = Environment.GetEnvironmentVariable("SQLServerConnectionString");
+
```
jpalac, line 22: I'm not sure what the standard practice is with this type of thing, but perhaps the sql connection string should be passed in the constructor, the same way the `RavenDbMonitoringDataStore` gets the `IDocumentStore` passed in? If we store the connection string on `Settings` then it will be available at the time of setting up the `HeartbeatMonitoringComponent`

src/ServiceControl/Monitoring/HeartbeatMonitoringComponent.cs
```diff
@@ -21,7 +21,7 @@ public override void Configure(Settings settings, IHostBuilder hostBuilder)
                 collection.AddSingleton<IDataMigration, PurgeKnownEndpointsWithTemporaryIdsThatAreDuplicateDataMigration>();
                 collection.AddHostedService<HeartbeatMonitoringHostedService>();
                 collection.AddSingleton<EndpointInstanceMonitoring>();
-                collection.AddSingleton<MonitoringDataStore>();
+                collection.AddSingleton<IMonitoringDataStore, SqlDbMonitoringDataStore>();
```
jpalac, line 24: Is this perhaps where we could have a setting on `Settings` to specify the data store for heartbeats? Based on that setting we could then change what we load into the collection.

src/ServiceControl/Infrastructure/Settings/Settings.cs
```diff
@@ -205,6 +207,10 @@ public int ExpirationProcessBatchSize
 
         public bool ExposeApi { get; set; } = true;
 
+        public bool SqlStorageEnabled { get; set; }
```
jpalac, line 210: I feel this should be an `enum` as we will have more than just SqlStorage (for now there'll be Raven, Sql and InMemory)

src/ServiceControl/Infrastructure/Settings/Settings.cs
```diff
@@ -526,6 +526,14 @@ int GetDataSpaceRemainingThreshold()
             return threshold;
         }
 
+
+        DataStoreType GetDataStoreType()
+        {
+            var value = SettingsReader<string>.Read("DataStoreType", "DataStoreType.RavenDb");
```
HEskandari, line 532: This default wouldn't parse. It needs to be the string value of the enum, e.g. `RavenDb`.

```suggestion
            var value = SettingsReader<string>.Read("DataStoreType", "RavenDb");
```

HEskandari: Also, the other "ConnectionString" property, uses the ConnectionStrings section of the config file as a fallback. Maybe we should be doing the same? See `Settings.GetConnectionString()` method.

src/ServiceControl/Monitoring/IMonitoringDataStore.cs
```diff
@@ -11,5 +11,6 @@ interface IMonitoringDataStore
         Task UpdateEndpointMonitoring(EndpointDetails endpoint, bool isMonitored);
         Task WarmupMonitoringFromPersistence(EndpointInstanceMonitoring endpointInstanceMonitoring);
         Task Delete(Guid endpointId);
+        Task<KnownEndpoint[]> GetAllKnownEndpoints();
```
HEskandari, line 14: I think this is more aligned with other APIs:

```suggestion
        Task<IReadonlyList<KnownEndpoint>> GetAllKnownEndpoints();
```

HEskandari (commented):
Looks good, some minor comments.

src/ServiceControl.AcceptanceTests/AcceptanceTest.cs
```diff
@@ -89,6 +104,33 @@ public void Teardown()
...
+BEGIN
+    SELECT @SQL = 'DROP TABLE ' + QUOTENAME(@schema) + '.' + QUOTENAME(RTRIM(@name))
+    EXEC (@SQL)
+    PRINT 'Dropped Table: ' + @schema + '.' + @name
+	SELECT  @name = NULL
+    SELECT TOP 1 @name = sys.objects.name, @schema = sys.schemas.name FROM sys.objects INNER JOIN sys.schemas ON sys.objects.schema_id = sys.schemas.schema_id WHERE [type] = 'U' ORDER BY sys.objects.name
+END
+", connection);
+
+                command.ExecuteNonQuery();
```
HEskandari, line 130: To delete all the tables it should be an easy 2 liner:

```csharp
            using (var connection = new SqlConnection(sqlDbConnectionString))
            {
                var dropConstraints = "EXEC sp_msforeachtable 'ALTER TABLE ? NOCHECK CONSTRAINT all'";
                var dropTables = "EXEC sp_msforeachtable 'DROP TABLE ?'";

                await connection.OpenAsync().ConfigureAwait(false);
                await connection.ExecuteAsync(dropConstraints).ConfigureAwait(false);
                await connection.ExecuteAsync(dropTables).ConfigureAwait(false);
            }
```

src/ServiceControl.Persistence.RavenDb/ServiceControl.Persistence.RavenDb.csproj
```diff
@@ -0,0 +1,22 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net472</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="NServiceBus.Raw" Version="3.2.5" />
```
mikeminutillo, line 8: Is this reference needed? I'm not sure why the persistence has a dependency on NSB Raw

src/ServiceControl.Persistence.SqlServer/ServiceControl.Persistence.SqlServer.csproj
```diff
@@ -0,0 +1,24 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net472</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Dapper" Version="2.0.123" />
+    <PackageReference Include="NServiceBus.Raw" Version="3.2.5" />
```
mikeminutillo, line 9: Same as RavenDb. Why is NSB.Raw needed?

src/ServiceControl.PersistenceTests/ServiceControl.PersistenceTests.csproj
```diff
@@ -0,0 +1,34 @@
...
+    <ProjectReference Include="..\ServiceControl.Persistence.RavenDb\ServiceControl.Persistence.RavenDb.csproj" />
+    <ProjectReference Include="..\ServiceControl.Persistence.SqlServer\ServiceControl.Persistence.SqlServer.csproj" />
+    <ProjectReference Include="..\ServiceControl.Transports.Learning\ServiceControl.Transports.Learning.csproj" />
+    <ProjectReference Include="..\ServiceControl\ServiceControl.csproj" />
+    
+  </ItemGroup>
+
+  <ItemGroup>
+    <None Update="app.config">
+      <CopyToOutputDirectory>Never</CopyToOutputDirectory>
```
mikeminutillo, line 30: Do we need this file?

src/ServiceControl/CustomChecks/CustomCheckResultProcessor.cs
```diff
@@ -0,0 +1,69 @@
...
+        {
+            this.domainEvents = domainEvents;
+            this.store = store;
+        }
+
+        public async Task ProcessResult(CustomCheckDetail checkDetail)
+        {
+            try
+            {
+                var statusChange = await store.UpdateCustomCheckStatus(checkDetail).ConfigureAwait(false);
```
mikeminutillo, line 23: Rather than return whether or not the status changed, we could just return the original status. That way the logic to determine whether or not it has changed could be container here.

src/ServiceControl/ExternalIntegrations/ExternalIntegrationsHostBuilderExtensions.cs
```diff
@@ -4,7 +4,7 @@ namespace ServiceControl.ExternalIntegrations
     using Microsoft.Extensions.DependencyInjection;
     using Microsoft.Extensions.Hosting;
 
-    static class ExternalIntegrationsHostBuilderExtensions
+    static class PersistenceHostBuilderExtensions
```
mikeminutillo, line 7: Why did this change? Should the filename be changed to match?

src/ServiceControl/InternalsVisibleTo.cs
```diff
@@ -3,4 +3,8 @@
 [assembly: InternalsVisibleTo("ServiceControl.UnitTests")]
 [assembly: InternalsVisibleTo("ServiceControl.AcceptanceTests")]
 [assembly: InternalsVisibleTo("ServiceControl.AcceptanceTesting")]
+[assembly: InternalsVisibleTo("ServiceControl.PersistenceTests")]
 [assembly: InternalsVisibleTo("ServiceControl.MultiInstance.AcceptanceTests")]
+[assembly: InternalsVisibleTo("ServiceControl.Persistence.InMemory")]
+[assembly: InternalsVisibleTo("ServiceControl.Persistence.RavenDb")]
+[assembly: InternalsVisibleTo("ServiceControl.Persistence.SqlServer")]
```
mikeminutillo, line 10: Should we just make the things these rely upon public instead?

mikeminutillo (approved):

src/ServiceControl.Persistence.InMemory/InMemoryCustomCheckDataStore.cs
```diff
@@ -0,0 +1,57 @@
...
+    class InMemoryCustomCheckDataStore : ICustomChecksDataStore
+    {
+        public Task Setup() => Task.CompletedTask;
+
+        public Task<CheckStateChange> UpdateCustomCheckStatus(CustomCheckDetail detail)
+        {
+            if (storage.ContainsKey(detail.CustomCheckId))
+            {
+                var storedCheck = storage[detail.CustomCheckId];
+                if (storedCheck.HasFailed == detail.HasFailed)
```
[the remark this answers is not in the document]
jpalac: If the stored check has failed but the new detail has not, then it won't go into that if statement and hence return the changed status. Or am I missing something?

src/ServiceControl.PersistenceTests/ServiceControl.PersistenceTests.csproj
jpalac: It's used in testing for parameters that are not set in the environment, for example the transport is taken from there, otherwise it wants to use msmq.

src/ServiceControl/ExternalIntegrations/ExternalIntegrationsHostBuilderExtensions.cs
jpalac: Must have been an auto rename change when I copied the file and renamed the copied version - I have changed it back.

src/ServiceControl/InternalsVisibleTo.cs
jpalac: I think I tried that originally but then the classes it was using also had to be changed, and then the classes those were using - I was going down a rabbit hole so I aborted.

src/ServiceControl.Persistence.SqlServer/ServiceControl.Persistence.SqlServer.csproj
jpalac: removed

src/ServiceControl.Persistence.RavenDb/ServiceControl.Persistence.RavenDb.csproj
jpalac: removed

src/ServiceControl.Persistence.InMemory/InMemoryCustomCheckDataStore.cs
```diff
@@ -0,0 +1,57 @@
...
+    using System.Threading.Tasks;
+    using CompositeViews.Messages;
+    using Contracts.CustomChecks;
+    using Infrastructure;
+    using ServiceControl.CustomChecks;
+    using ServiceControl.Persistence;
+
+    class InMemoryCustomCheckDataStore : ICustomChecksDataStore
+    {
+        public Task Setup() => Task.CompletedTask;
```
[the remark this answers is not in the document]
jpalac: removed

src/ServiceControl.Persistence.InMemory/InMemoryCustomCheckDataStore.cs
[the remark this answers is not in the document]
mikeminutillo: I don't know how many hundreds of times I have read this and not properly understood what it was doing. You are right. The other implementations use the term Failed as a enumerated value. This is doing a direct boolean comparison so it doesn't need to test all cases individually.

It doesn't actually update the state on the stored check though, right? There should be a `storedCheck.HasFailed = detail.HasFailed` somewhere.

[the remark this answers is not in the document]
jpalac: Yes it doesn't update the state - I think we brought that up a while back on slack:
image

src/ServiceControl.AcceptanceTests/AcceptanceTest.cs
tmasternak: changed

src/ServiceControl.Persistence.InMemory/InMemoryCustomCheckDataStore.cs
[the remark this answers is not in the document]
HEskandari: This is addressed now in the latest version.

PR #2911, opened 2022-07-05
Integration event publish throughput increased 8x

ramonsmits:
Integration event publish throughput increased by:

1. Reducing the Task.Delay from 1000 to 100 (not sure why its there, not commented as why the delay is required). This already reduces 900ms until the next cycle.
2. In the publishing itself publishing of a batch of events was done sequentially. Publishing over ASBS took 4 to 5 seconds and doing the publishes in parallel reduces it to roughly 1 second.

In total 4.5-5.0 seconds gained which increased throughput 8x from 20/s to 160/s.

TODO:

- [ ] Make interval configurable so it can be overriden

### DIFF
--- a/src/ServiceControl/ExternalIntegrations/EventDispatcherHostedService.cs
+++ b/src/ServiceControl/ExternalIntegrations/EventDispatcherHostedService.cs
@@ -2,6 +2,7 @@
 {
     using System;
     using System.Collections.Generic;
+    using System.Diagnostics;
     using System.Linq;
     using System.Reactive.Linq;
     using System.Threading;
@@ -103,7 +104,7 @@
                 if (more && !cancellationToken.IsCancellationRequested)
                 {
                     //if there is more events to dispatch we sleep for a bit and then we go again
-                    await Task.Delay(1000, CancellationToken.None).ConfigureAwait(false);
+                    await Task.Delay(100, CancellationToken.None).ConfigureAwait(false);
                 }
             }
             while (!cancellationToken.IsCancellationRequested && more);
@@ -140,6 +141,7 @@
                     eventsToBePublished.AddRange(events);
                 }
 
+                var tasks = new List<Task>();
                 foreach (var eventToBePublished in eventsToBePublished)
                 {
                     if (Logger.IsDebugEnabled)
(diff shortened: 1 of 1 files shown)


mikeminutillo (approved):
This was written before the transport could accept multiple transport operations. I wonder if it would be better to go much lower level and hand the entire batch to the transport at once. That's a much more invasive change and this is a good next step.

danielmarbach:
I think the original intent was to not put too much pressure on the subscribers from a publishing perspective as well as not trying to fetch changes too aggressively. This is why an interval of 1000 ms was used. Why not decouple the two things and actually use the transport's capability to pass multiple operations as suggested. 

> That's a much more invasive change and this is a good next step.

Is that really true? We are already using the batching capabilities in other areas of ServiceControl. At this stage we also don't know the impact of the shorter delay. In what sense is it more invasive? Are you talking about the complexity of the code?

I think the counter argument I could see for not doing the batched dispatch is that not all transports properly handle it thus the "workaround" of doing parallel sends might end up being "faster" on certain transports like ASBS. But then again, we are optimizing for the scenario of "full queue of stuff to dispatch". Why is that an important scenario to optimize for for the external events integration?

mikeminutillo:
It's a little more invasive in that we'd need to bypass the pipeline to do bulk dispatch. That means we need to re-implement the relevant parts of message driven pub-sub in order to continue to support MSMQ. To be fair we already control the subscription persistence so maybe this isn't a big issue. It may be worth spiking it out.

If we had a bulk-dispatch (or stand alone outbox) API we could use that here.

An alternative would be to turn this into a recurrent (scheduled) message so we'd have an incoming message to capture all of the publishes for dispatch. That comes with all of the challenges around using messages for scheduling.

If the parallel send is faster than a bulk dispatch, that's just an indication that the transport should be updated to handle batches better.

tmasternak (approved):

andreasohlund:
What is the status on this one? (complete or close?)

tmasternak:
@andreasohlund based on my understanding of the discussion my understanding is that we don't have a clear answer to the following:

> What is the maximum acceptable load  (in msg/sec) we want to put on the integration endpoints? 

The discussion also shows that we are probably not aligned about the answer. As a result, I would propose closing this PR.

ramonsmits:
> @andreasohlund based on my understanding of the discussion my understanding is that we don't have a clear answer to the following:
> 
> > What is the maximum acceptable load  (in msg/sec) we want to put on the integration endpoints?
> 
> The discussion also shows that we are probably not aligned about the answer. As a result, I would propose closing this PR.

That load should be controlled by the consumer. IMHO if you connect to the fire hoze you should take preventive measures like rate limiting or just limit the amount of concurrency.

Meaning, if you have a 250msg/s IN, this can result in 250msg/s OUT... that is sort of expected. IMHO its a very bad that SC currently already acts as a buffer instead of an actual queue.


I would then at least want to propose to merge the PR but then with the previous delay of 1000 seconds.

ramonsmits:
In fact, its very similar when you would have a consumer for `error.log` or `audit.log`. Those aren't creation rate limited either.

PR #2915, opened 2022-07-07
SQL support for custom checks

HEskandari:


src/ServiceControl/CustomChecks/Web/StatisticsResult.cs
```diff
@@ -0,0 +1,11 @@
+namespace ServiceControl.CustomChecks
+{
+    using System.Collections.Generic;
+
+    public class StatisticsResult
```
mikeminutillo: This pattern might come up a lot. I wonder if this should be something like:

```csharp
class QueryResult<T>
{
  public IList<T> Results { get; set; }
  public int TotalResults { get; set; }
  public string ETag { get; set; }
}
```

We do already have something like this for scatter-gather.

We may want to hold off on doing this if we only have one instance at the moment.

src/ServiceControl/CustomChecks/SqlDbCustomCheckDataStore.cs
```diff
@@ -0,0 +1,212 @@
...
+                        CustomCheckId = detail.CustomCheckId,
+                        Category = detail.Category,
+                        SucceededAt = detail.ReportedAt,
+                        OriginatingEndpoint = detail.OriginatingEndpoint
+                    }).ConfigureAwait(false);
+                }
+            }
+        }
+
+        public async Task<StatisticsResult> GetStats(HttpRequestMessage request, string status = null)
```
mikeminutillo, line 98: Would it be possible to push the `PagingInfo` abstraction to this call (and ideally to the interface)? That would prevent the Http abstraction from pushing down this far.

src/ServiceControl/CustomChecks/SqlDbCustomCheckDataStore.cs
```diff
@@ -0,0 +1,212 @@
...
+        public SqlDbCustomCheckDataStore(string connectionString, IDomainEvents domainEvents, EndpointDetailsMapper mapper)
+        {
+            this.connectionString = connectionString;
+            this.domainEvents = domainEvents;
+            this.mapper = mapper;
+        }
+
+        public async Task UpdateCustomCheckStatus(CustomCheckDetail detail)
+        {
+            var publish = false;
```
mikeminutillo, line 23: I wonder if this flag (and the details of the custom check) should be returned to the caller. That way all of the publishing logic can happen outside of the storage.

src/ServiceControl/CustomChecks/Web/StatisticsResult.cs
HEskandari: I refactored to use the existing QueryResult<T> class.

src/ServiceControl/CustomChecks/SqlDbCustomCheckDataStore.cs
HEskandari: I have changed the interface so it doesn't rely on `HttpRequestMessage` anymore.

src/ServiceControl/CustomChecks/SqlDbCustomCheckDataStore.cs
HEskandari: I have extracted the custom check status notification from the DataStore. It should address this comment, please review the commit [1ff24d8](https://github.com/Particular/ServiceControl/pull/2915/commits/1ff24d8439a8e192583f5bc97755faa809c5a44b).

mikeminutillo (approved):

PR #2918, opened 2022-07-07
Improve performance by logging asynchronously

ramonsmits:
Previously all logging is sync, meaning a target that is slow or blocked like for example the console impacts performance. By wrapping all targets in an `AsyncTargetWrapper` logging is async which increases the performance a little.

The `AsyncTargetWrapper` is lossy when its internal buffer overflows but in general that can only happen if the logging level is set to Verbose.

### DIFF
--- a/src/ServiceControl/LoggingConfigurator.cs
+++ b/src/ServiceControl/LoggingConfigurator.cs
@@ -5,6 +5,7 @@ namespace Particular.ServiceControl
     using NLog.Config;
     using NLog.Layouts;
     using NLog.Targets;
+    using NLog.Targets.Wrappers;
     using NServiceBus;
     using NServiceBus.Logging;
     using ServiceBus.Management.Infrastructure.Settings;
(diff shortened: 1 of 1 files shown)


danielmarbach:
Is there any downside of doing this for the Console Target, for example will there be an impact an readability when running in console mode?

Is it fair to assume the actual order does of entries doesn't change because they are inserted into some kind of concurrent buffer that is read from by a single thread and written to the target?

danielmarbach (approved):

ramonsmits:
> Notice it is good idea to [Remember to Flush](https://github.com/NLog/NLog/wiki/Tutorial#5-remember-to-flush) when enabling NLog async-writer-threads (or using NLog targets that depend on network-communication).

I assumed we would be invoking `Shutdown` already when making this change. Thanks for reminding as [we mention that in our own guidance for endpoints](https://docs.particular.net/nservicebus/hosting/critical-errors#custom-handling-implementation-concerns),

ramonsmits:
@snakefoot By the way, I noticed that `NLog.LogManager.AutoShutdown` = `true` which according to the code comments does:

> Gets or sets a value indicating whether to automatically call `LogFactory.Shutdown` on `AppDomain.Unload` or `AppDomain.ProcessExit`.

Any reason to still add it in the main method with a try..finally ? As I see your involvement in https://github.com/NLog/NLog/pull/3578 and https://github.com/NLog/NLog/commit/1d1433421e86ce5297fae009737db58543434733

snakefoot:
The `NLog.LogManager.AutoShutdown` was introduced to actually disable the default behavior, because it doesn't work well on .NET Core on Linux. The events triggers before the application is fully shutdown, thus all Logging disappear before exit of main-function.

NLog.Extensions.Logging v5 and NLog.Web.AspNetCore v5 now assigns `NLog.LogManager.AutoShutdown = false`, and expects users to explictly call `NLog.LogManager.Shutdown()`.

PR #2921, opened 2022-07-08
Extract endpoint detection

mikeminutillo:
_I recommend reviewing this PR one commit at a time_

We detect endpoints by looking in the headers of failed messages ingested from the error queue. These get added to the in-memory monitoring one at a time, and any that we're seeing for the first time get added to bulk database operation for persistence as `KnownEndpoint` documents.

This process is quite difficult to follow and very entangled with RavenDB which makes it challenging to switch to a different persistence.

This PR changes things so that all detected endpoints get handed off to the in-memory monitoring which raises a domain event for new endpoints, and they are persisted all in a batch. This matches the way that all other `KnownEndpoint` persistence is handled and allows us to swap isolate heartbeats from RavenDB. It provides cleaner separation between the ingestion pipeline and the monitoring components as well.

This should be safe to do as the number of endpoints detected this way should always be very low. It may be that the first batch discovers a lot of new endpoints but subsequent batches likely contain messages processed and sent by endpoints that we're already aware of.

### DIFF
--- a/src/ServiceControl/Monitoring/EndpointInstanceMonitoring.cs
+++ b/src/ServiceControl/Monitoring/EndpointInstanceMonitoring.cs
@@ -9,6 +9,7 @@ namespace ServiceControl.Monitoring
     using Contracts.EndpointControl;
     using Contracts.HeartbeatMonitoring;
     using Contracts.Operations;
+    using Infrastructure;
     using Infrastructure.DomainEvents;
 
     class EndpointInstanceMonitoring
@@ -35,13 +36,6 @@ namespace ServiceControl.Monitoring
             var stats = GetStats();
 
             await Update(stats).ConfigureAwait(false);
-        }
-
-        public bool IsNewInstance(EndpointDetails newEndpointDetails)
-        {
-            var endpointInstanceId = newEndpointDetails.ToInstanceId();
-
-            return endpoints.TryAdd(endpointInstanceId.UniqueId, new EndpointInstanceMonitor(endpointInstanceId, false, domainEvents));
         }
 
         public async Task EndpointDetected(EndpointDetails newEndpointDetails)
@@ -73,6 +67,26 @@ namespace ServiceControl.Monitoring
         {
             var endpointInstanceId = new EndpointInstanceId(endpointDetails.Name, endpointDetails.Host, endpointDetails.HostId);
             endpoints.GetOrAdd(endpointInstanceId.UniqueId, id => new EndpointInstanceMonitor(endpointInstanceId, monitored, domainEvents));
+        }
+
+        public IEnumerable<KnownEndpoint> DetectEndpointsFromBulkIngestion(IEnumerable<EndpointDetails> observedEndpoints)
+        {
+            foreach (var observedEndpoint in observedEndpoints)
+            {
+                var endpointInstanceId = observedEndpoint.ToInstanceId();
+
+                if (endpoints.TryAdd(endpointInstanceId.UniqueId,
+                        id => new EndpointInstanceMonitor(id, false, domainEvents)))
+                {
+                    yield return new KnownEndpoint
+                    {
+                        Id = DeterministicGuid.MakeId(observedEndpoint.Name, observedEndpoint.HostId.ToString()),
+                        EndpointDetails = observedEndpoint,
+                        HostDisplayName = observedEndpoint.Host,
+                        Monitored = false
+                    };
+                }
+            }
         }
 
         async Task Update(EndpointMonitoringStats stats)
--- a/src/ServiceControl/Monitoring/EndpointInstanceMonitoring.cs
+++ b/src/ServiceControl/Monitoring/EndpointInstanceMonitoring.cs
@@ -9,7 +9,7 @@ namespace ServiceControl.Monitoring
     using Contracts.EndpointControl;
     using Contracts.HeartbeatMonitoring;
     using Contracts.Operations;
-    using Infrastructure;
+    using EndpointControl.Contracts;
     using Infrastructure.DomainEvents;
 
     class EndpointInstanceMonitoring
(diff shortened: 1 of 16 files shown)


src/ServiceControl/Monitoring/EndpointInstanceMonitoring.cs
```diff
@@ -75,6 +69,30 @@ public void DetectEndpointFromPersistentStore(EndpointDetails endpointDetails, b
             endpoints.GetOrAdd(endpointInstanceId.UniqueId, id => new EndpointInstanceMonitor(endpointInstanceId, monitored, domainEvents));
         }
 
+        public async Task DetectEndpointsFromBulkIngestion(IEnumerable<EndpointDetails> observedEndpoints)
+        {
+            var newEndpoints = new Lazy<List<EndpointDetails>>();
```
danielmarbach, line 74: The lazy is not really helping here. In case you would like to save the list allocation, if would be better to only initialize the collection when you need it. 

Here are some guidelines from my talk

- Use `Array.Empty<T>` to represent empty arrays
- Use `Enumerable.Empty<T>` to represent empty enumerables
- When the size of the items to be added to the collection are known upfront, initialize the collection with the correct count to prevent the collection from growing and thus allocating more and reshuffling things internally
- Use the concrete collection type instead of interfaces or abstract types. For example, when enumerating through a `List<T>` with `foreach` it uses a non-allocating `List<T>.Enumerator` struct. But when it is used through for example `IEnumerable<T>` that struct is boxed to `IEnumerator<T>` in the foreach and thus allocates. Of course for public APIs it is sometimes better to use the most generic type to be as broadly applicable as possible. Mileage may vary.
- With more modern CSharp versions that have good pattern matching support, it is sometimes possible to do a quick type check and based on the underlying collection type get access to the count without having to use `Count()`. With .NET 6, and later, it is also possible to use [`Enumerable.TryGetNonEnumeratedCount`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.trygetnonenumeratedcount) which internally does collection type checking to get the count without enumerating.
- Wait with instantiating collections until you really need them.

This thing is a bit of a tricky beast though. Because you are exposing on the even an array but because you don't  know how many endpoints you really observed you need to track them in a list and then expose it as an array. So the best option would be in case you want to safe allocations to forget the nice things about "immutability", highest collection type possible etc. and then do the following:

- Instead of IEnumerable<EndpointDetails> pass in at least `IReadOnlyCollection` because that exposes a count
- Lazy initialize the `List<EndpointDetails>` with the count when you need it to prevent it from growing at runtime. Another option is to optimize for the path that you mostly don't get new endpoints or just one (if that is true) and then work with a local variable or a List of one item
- Change the domain event type to be List or `IReadOnlyList` to avoid copying again

src/ServiceControl/Monitoring/InMemoryMonitoringDataStore.cs
```diff
@@ -71,6 +71,28 @@ public Task Delete(Guid endpointId)
             return Task.CompletedTask;
         }
 
+        public Task BulkCreate(EndpointDetails[] newEndpoints)
+        {
+            foreach (var endpoint in newEndpoints)
+            {
+                var id = DeterministicGuid.MakeId(endpoint.Name, endpoint.HostId.ToString());
+
+                if (!endpoints.Any(a => a.Id == id))
```
danielmarbach, line 80: Are we still assuming this stuff is never executed concurrently because we only have one batch going on and never concurrently interleaved batches? So a list is safe?

If we always end up doing `Any` on the list, would it be simpler to have a dictionary? That would also be the less allocation heavy variant because the Any will always closure allocate due to the access to the local `id`

src/ServiceControl/Monitoring/MonitorEndpointsFoundInFailedMessageHeaders.cs
```diff
@@ -0,0 +1,44 @@
...
+    using Monitoring;
+    using NServiceBus.Transport;
+    using Operations;
+    using ServiceControl.Contracts.Operations;
+
+    class MonitorEndpointsFoundInFailedMessageHeaders : IErrorMessageBatchPlugin
+    {
+        public Task AfterProcessing(List<MessageContext> batch)
+        {
+            var observedEndpoints = new HashSet<EndpointDetails>();
```
danielmarbach, line 15: Would it be better to have a singleton comparer that uses the unique key of the details instead of relying on the default comparer?

src/ServiceControl/Monitoring/MonitorEndpointsFoundInFailedMessageHeaders.cs
```diff
@@ -0,0 +1,44 @@
...
+            foreach (var context in batch)
+            {
+                var errorEnricherContext = context.Extensions.Get<ErrorEnricherContext>();
+
+                void RecordKnownEndpoint(string key)
```
danielmarbach, line 20: It's funny with local functions and heap allocations. There is a lot of mad science going and edge cases where it does or does not heap allocate

> Depending on their use, local functions can avoid heap allocations that are always necessary for lambda expressions. If a local function is never converted to a delegate, and none of the variables captured by the local function are captured by other lambdas or local functions that are converted to delegates, the compiler can avoid heap allocations.

I generally tend to avoid them or directly mark them as static

https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/local-functions#heap-allocations

src/ServiceControl/Monitoring/InMemoryMonitoringDataStore.cs
```diff
@@ -71,6 +71,28 @@ public Task Delete(Guid endpointId)
             return Task.CompletedTask;
         }
 
+        public Task BulkCreate(EndpointDetails[] newEndpoints)
+        {
+            foreach (var endpoint in newEndpoints)
+            {
+                var id = DeterministicGuid.MakeId(endpoint.Name, endpoint.HostId.ToString());
```
danielmarbach, line 78: We seem to be doing this all the time. Why not have this encoded as a proper ID in the endpoint details? then we can do it once and use it across the board

src/ServiceControl/Monitoring/RavenDbMonitoringDataStore.cs
```diff
@@ -119,6 +119,27 @@ public async Task Delete(Guid endpointId)
             }
         }
 
+        public async Task BulkCreate(EndpointDetails[] endpoints)
+        {
+            using (var session = store.OpenAsyncSession())
```
danielmarbach, line 124: If I understood correctly the assumption here is that we will never create a lot of endpoints and therefore moving it out from bulk session is OK. 

We got some great insights from the calls with Hibernating Rhinos talking about RavenDB 5. Especially this part

> When moving to RavenDB 5 we should move away from BulkInserts. BulkInserts are faster in RavenDB 3.5 because RavenDB acquired a global lock when doing session operations. With bulk inserts, that problem is less exacerbated. The suggestion was to have a pattern of interleaved operations feeding from the transport and shuffling to RavenDB. Since naturally our batch sizes are quite small bulk inserts with RavenDB are not worth it. Instead we should for example split the "concurrency" in half and then have a few RavenDB sessions that store multiple operations and then call SaveChanges. More or less the following pattern

this would indicate we will have to move away from doing BulkInsert and actually have one document session wrapping everything that happens during ingestion. That is also a good pattern in general with SQL Server. Wouldn't it make more sense under that light to introduce a unit of work pattern that opens a storage session and saves it at the end and have all "plugins" rely on that being around via DI or some explicit mechanism like we treat the synchronized storage in core (I would prefer that)? Then we write today code that is more future proof because

a) we can move the bulk session handling into the unit of work as well
b) for the changes here just create another session alongside with it

once we move to v5 we can just delete the bulk session handling and rely on the session only. I think this pattern of starting to have individual plugins doing individual connection handling can be dangerous on the long run because it is easy to create all sorts of plugins that do "micro" transactions against the underlying storage that causes the ingestion rate to drop significantly.

src/ServiceControl/Monitoring/SqlDbMonitoringDataStore.cs
```diff
@@ -119,6 +119,34 @@ public async Task Delete(Guid endpointId)
             }
         }
 
+        // NOTE: Temp implementation. Need to find a bulk insert
+        public async Task BulkCreate(EndpointDetails[] endpoints)
+        {
+            using (var connection = new SqlConnection(connectionString))
```
danielmarbach, line 125: Same comment as above

src/ServiceControl/Monitoring/RavenDbMonitoringDataStore.cs
SzymonPobiega: Sounds reasonable

mikeminutillo:
Closed in favor of #2927

PR #2925, opened 2022-07-14
Update RabbitMQ.Transport to 7.0.0

kentdr:
This PR upgrades the ServiceControl RabbitMQ transport to the latest RabbitMQ transport version 7.0.0.  7.0.0 adds support for quorum queues and, therefore, ServiceControl needs to be able to support the additional topology options.  This was accomplished by adding more options to the transport dropdown in ServiceControl Management. 

Additionally, adding more topology choices imposed a name change to the existing transport customization classes which requires a migration to the new class names during the upgrade process. 

*Note:* Upgrading to RabbitMQ.Transport 7.0.0 requires customers are on RabbitMQ broker version 3.10.0 or higher. Also, the `quorum_queue` and `stream_queue` feature flags m just be enabled.  If any of these requirements are not met, ServiceControl will not start.


src/_connection.txt
```diff
@@ -1,3 +0,0 @@
-RabbitMQ
```
tmasternak, line 1: I think this should stay. It's been in the repository for some time to make the local setup more convenient. https://github.com/Particular/ServiceControl/blob/master/docs/running-acceptance-tests.md#connection-file

src/ServiceControlInstaller.Engine/Instances/ServiceControlCoreTransports.cs
```diff
@@ -65,29 +65,79 @@ public class ServiceControlCoreTransports
...
+                                  || name.Equals("ServiceControl.Transports.RabbitMQ.QuorumConventialRoutingTopologyRabbitMQTransport, ServiceControl.Transports.RabbitMQ", StringComparison.OrdinalIgnoreCase)
+            },
+            new TransportInfo
+            {
+                Name = TransportNames.RabbitMQDirectRoutingTopologyDeprecated,
+                TypeName = "ServiceControl.Transports.RabbitMQ.RabbitMQDirectRoutingTransportCustomization, ServiceControl.Transports.RabbitMQ",
+                ZipName = "RabbitMQ",
+                SampleConnectionString = "host=<HOSTNAME>;username=<USERNAME>;password=<PASSWORD>;DisableRemoteCertificateValidation=<true|false(default)>;UseExternalAuthMechanism=<true|false(default)>",
+                AvailableInSCMU = false,
+                AutoMigrateTo = TransportNames.RabbitMQClassicDirectRoutingTopology,
```
tmasternak, line 113: This will put users in a potentially hard-to-rollback situation if they are on a broker version lover than 3.10.0. They will not be able to connect to the broker and will have to change to transport configuration by hand in the config file. Maybe doing something similar to what we do in the installation engine could be useful https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl.Config/Commands/UpgradeServiceControlInstanceCommand.cs#L70

kentdr: @tmasternak we opted to pop up a dialog informing them that there are minimum broker requirements and giving them a chance to cancel the upgrade.  The reality is, at some point, the broker requirements will have to be there in order to move forward with the platform at all, we didn't think prolonging it gained us anything that justified the extra complexity in ServiceControl.

bording (approved):

PR #2927, opened 2022-07-18
Unit of work abstraction for ingestion

mikeminutillo:
Alternative to #2921 

This one works by adding a Unit of Work abstraction for failed error ingestion. The Raven implementation works the same as it did before. It collects everying into patch commands and then submits them all to the database in a batch.

There is also an in-memory and SQL implementation which only cover the monitoring parts of the batch. This requires allowing both the Raven and the alternative to run at the same time. When the alternative doesn't provide an implementation, it falls back to the Raven implementation.


src/ServiceControl/Operations/UnitOfWork/InMemory/InMemoryIngestionUnitOfWorkFactory.cs
```diff
@@ -0,0 +1,17 @@
...
+
+    class InMemoryIngestionUnitOfWorkFactory : IIngestionUnitOfWorkFactory
+    {
+        IMonitoringDataStore dataStore;
+
+        public InMemoryIngestionUnitOfWorkFactory(IMonitoringDataStore dataStore) => this.dataStore = dataStore;
+
+        public Task<IIngestionUnitOfWork> StartNew()
+            => Task.FromResult(
+                (IIngestionUnitOfWork)new InMemoryIngestionUnitOfWork(dataStore)
```
HEskandari, line 14: Is this cast necessary since the class implements the interface and it is implied?

src/ServiceControl/Operations/UnitOfWork/InMemory/InMemoryIngestionUnitOfWork.cs
```diff
@@ -0,0 +1,36 @@
...
+        }
+
+        public IMonitoringIngestionUnitOfWork Monitoring { get; }
+        public IRecoverabilityIngestionUnitOfWork Recoverability { get; }
+
+        public async Task Complete()
+        {
+            foreach (var endpoint in knownEndpoints)
+            {
+                await dataStore.CreateIfNotExists(endpoint.EndpointDetails)
```
HEskandari, line 25: Not related to this change, but this `CreateIfNotExists` is very database-centric and does not sound correct. Should it be named just 'store' instead?

```suggestion
                await dataStore.CreateIfNotExists(endpoint.EndpointDetails)
```

HEskandari (approved):
Some small comments, but generally looks good.

tmasternak (approved):

jpalac (approved):

src/ServiceControl/Operations/UnitOfWork/InMemory/InMemoryIngestionUnitOfWorkFactory.cs
mikeminutillo: I get a compiler warning about not being able to cast `Task<InMemoryIngestionUnitOfWork>` to `Task<IIngestionUnitOfWork>` without it.

mikeminutillo: I turned this into `Task.FromResult<IIngestionUnitOfWork>(new InMemoryIngestionUnitOfWork(dataStore))` which seems neater

src/ServiceControl/Operations/UnitOfWork/SqlDb/SqlIngestionUnitOfWorkFactory.cs
```diff
@@ -0,0 +1,22 @@
...
+        readonly string connectionString;
+
+        public SqlIngestionUnitOfWorkFactory(string connectionString)
+            => this.connectionString = connectionString;
+
+        public async Task<IIngestionUnitOfWork> StartNew()
+        {
+            var connection = new SqlConnection(connectionString);
+
+            await connection.OpenAsync().ConfigureAwait(false);
```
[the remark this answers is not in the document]
mikeminutillo: Pushed a basic implementation. Do we need an explicit rollback? If you dispose a SQL Transaction without committing it, it should rollback anyway.

src/ServiceControl/Operations/UnitOfWork/InMemory/InMemoryIngestionUnitOfWork.cs
mikeminutillo: In this case it definitely should not update an existing instance. We have `CreateIfNotExists` and `CreateOrUpdate`.

src/ServiceControl.Persistence.RavenDb/UnitOfWork/RavenDbIngestionUnitOfWork.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        public RavenDbIngestionUnitOfWork(IDocumentStore store)
+        {
+            this.store = store;
+            commands = new ConcurrentBag<ICommandData>();
+            Monitoring = new RavenDbMonitoringIngestionUnitOfWork(this);
+            Recoverability = new RavenDbRecoverabilityIngestionUnitOfWork(this);
+        }
+
+        internal void AddCommand(ICommandData command) => commands.Add(command);
```
danielmarbach, line 22: Why not public?

src/ServiceControl.Persistence.InMemory/UnitOfWork/InMemoryIngestionUnitOfWork.cs
```diff
@@ -0,0 +1,37 @@
...
+            foreach (var endpoint in knownEndpoints)
+            {
+                await dataStore.CreateIfNotExists(endpoint.EndpointDetails)
+                    .ConfigureAwait(false);
+            }
+        }
+
+        internal void AddEndpoint(KnownEndpoint knownEndpoint) => knownEndpoints.Add(knownEndpoint);
+
+        public void Dispose()
```
danielmarbach, line 33: I would naturally seal things because otherwise this class would trigger an analyzer warning (if we had them in place ;)) since it doesn't implement the disposable pattern. Alternative make the disposable pattern opt-in and not part of the interface

danielmarbach (commented):
A few nits. In general I like this

src/ServiceControl/Operations/UnitOfWork/SqlDb/SqlIngestionUnitOfWorkFactory.cs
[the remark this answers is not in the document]
tmasternak: Looks good to me.

src/ServiceControl.Persistence.RavenDb/InternalsVisibleTo.cs
```diff
@@ -0,0 +1,3 @@
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("ServiceControl.UnitTests")]
```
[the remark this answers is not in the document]
mikeminutillo: We just copied what the other test project was doing. We might change them all in a separate PR.

src/ServiceControl.Persistence.InMemory/UnitOfWork/InMemoryIngestionUnitOfWork.cs
mikeminutillo: We moved the disposable implementation into a base class and had that implement the full disposing pattern. That allows the implementations to only override it if they have something to dispose.

PR #2974, opened 2022-08-02
Add warning when installing new RabbitMQ instances

kentdr:
When installing new instances (including Monitoring instances) that use RabbitMQ, a warning dialog will pop up requiring confirmation that the RabbitMQ transport 7.0 prerequisites have been met (version 3.10+, and quorum queues and stream queue feature flags enabled).

### DIFF
--- a/src/ServiceControl.Config/UI/InstanceAdd/MonitoringAddAttachment.cs
+++ b/src/ServiceControl.Config/UI/InstanceAdd/MonitoringAddAttachment.cs
@@ -68,6 +68,19 @@
                 ServiceAccountPwd = viewModel.Password
             };
 
+            if ((instanceMetadata.TransportPackage.Name == TransportNames.RabbitMQClassicConventionalRoutingTopology ||
+                 instanceMetadata.TransportPackage.Name == TransportNames.RabbitMQQuorumConventionalRoutingTopology ||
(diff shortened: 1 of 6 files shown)


src/ServiceControl.Config/UI/InstanceAdd/MonitoringAddAttachment.cs
```diff
@@ -68,6 +68,19 @@ async Task Add()
                 ServiceAccountPwd = viewModel.Password
             };
 
+            if ((instanceMetadata.TransportPackage.Name == TransportNames.RabbitMQClassicConventionalRoutingTopology ||
+                 instanceMetadata.TransportPackage.Name == TransportNames.RabbitMQQuorumConventionalRoutingTopology ||
+                 instanceMetadata.TransportPackage.Name == TransportNames.RabbitMQClassicDirectRoutingTopology ||
+                 instanceMetadata.TransportPackage.Name == TransportNames.RabbitMQQuorumDirectRoutingTopology) &&
```
HEskandari, line 74: You can probably encapsulate this as an extension method on the `TransportInfo` class and reuse it.

HEskandari (approved):
Made a small note, but I think generally looks good.

bording (approved):

PR #2989, opened 2022-08-17
Upgrade ServiceControl Management to Autofac 6.4

mikeminutillo:


src/ServiceControl.Config.Tests/AutoAttachmentModuleTests.cs
```diff
@@ -0,0 +1,91 @@
...
+    using Framework.Modules;
+    using NUnit.Framework;
+
+    [TestFixture]
+    public class AutoAttachmentModuleTests
+    {
+        [Test]
+        public void AttachmentsAreAttached()
+        {
+            AssemblySource.Instance.Add(typeof(AutoAttachmentModuleTests).Assembly);
```
HEskandari, line 15: This being static, should it be done in a pre-test step? (and potentially cleaned up after?)

mikeminutillo: Good call. Moved this to setup method and added teardown method to remove it afterwards

HEskandari (approved):

PR #3002, opened 2022-08-23
Switch Raven implementation to use Unit of Work pattern

mikeminutillo:
Implements a Unit of Work pattern for Audit Ingestion

### DIFF
--- a/src/ServiceControl.Audit.Persistence.RavenDb/UnitOfWork/RavenDbAuditUnitOfWorkFactory.cs
+++ b/src/ServiceControl.Audit.Persistence.RavenDb/UnitOfWork/RavenDbAuditUnitOfWorkFactory.cs
@@ -1,0 +1,23 @@
+﻿namespace ServiceControl.Audit.Persistence.RavenDb.UnitOfWork
+{
+    using Persistence.UnitOfWork;
+    using Raven.Abstractions.Data;
+    using Raven.Client;
+
+    class RavenDbAuditIngestionUnitOfWorkFactory : IAuditIngestionUnitOfWorkFactory
+    {
+        IDocumentStore store;
+
+        public RavenDbAuditIngestionUnitOfWorkFactory(IDocumentStore store) => this.store = store;
+
+        public IAuditIngestionUnitOfWork StartNew(int batchSize)
+            => new RavenDbAuditIngestionUnitOfWork(
+                store.BulkInsert(
+                    options: new BulkInsertOptions
+                    {
+                        OverwriteExisting = true,
+                        ChunkedBulkInsertOptions = null,
+                        BatchSize = batchSize
+                    }));
+    }
+}
--- a/src/ServiceControl.Audit/Auditing/AuditPersister.cs
+++ b/src/ServiceControl.Audit/Auditing/AuditPersister.cs
@@ -14,21 +14,19 @@
     using NServiceBus.Extensibility;
     using NServiceBus.Logging;
     using NServiceBus.Transport;
-    using Raven.Abstractions.Data;
-    using Raven.Client;
-    using Raven.Client.Document;
+    using Persistence.UnitOfWork;
     using ServiceControl.Infrastructure.Metrics;
     using ServiceControl.SagaAudit;
     using JsonSerializer = Newtonsoft.Json.JsonSerializer;
 
     class AuditPersister
     {
-        public AuditPersister(IDocumentStore store, BodyStorageEnricher bodyStorageEnricher,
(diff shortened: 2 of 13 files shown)


src/ServiceControl.Audit.Persistence.RavenDb/UnitOfWork/RavenDbAuditUnitOfWorkFactory.cs
```diff
@@ -0,0 +1,23 @@
+﻿namespace ServiceControl.Audit.Persistence.RavenDb.UnitOfWork
+{
+    using Persistence.UnitOfWork;
+    using Raven.Abstractions.Data;
+    using Raven.Client;
+
+    class RavenDbAuditIngestionUnitOfWorkFactory : IAuditIngestionUnitOfWorkFactory
+    {
+        IDocumentStore store;
```
HEskandari, line 9: by convention, we mostly move the fields to the bottom, so there's less noise as you open the file.

src/ServiceControl.Audit.Persistence.RavenDb/UnitOfWork/RavenDbAuditUnitOfWorkFactory.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        public RavenDbAuditIngestionUnitOfWorkFactory(IDocumentStore store) => this.store = store;
+
+        public IAuditIngestionUnitOfWork StartNew(int batchSize)
+            => new RavenDbAuditIngestionUnitOfWork(
+                store.BulkInsert(
+                    options: new BulkInsertOptions
+                    {
+                        OverwriteExisting = true,
+                        ChunkedBulkInsertOptions = null,
```
HEskandari, line 19: Is this needed to be initiated as `null`?

src/ServiceControl.Audit/Infrastructure/WebApi/Cors.cs
```diff
@@ -40,13 +40,21 @@ static CorsPolicy GetDefaultPolicy()
...
             return policy;
         }
 
+        static void AddRange(IList<string> list, params string[] items)
+        {
+            foreach (var item in items)
+            {
+                list.Add(item);
+            }
+        }
```
HEskandari, line 56: Is this the extension method you mentioned? Can it be extracted as an extension method? I suspect there'll be other use-cases for it (maybe)? If not, that's probably fine.

src/ServiceControl.Audit/Persistence/IFailedAuditStorage.cs
```diff
@@ -0,0 +1,16 @@
...
+    using System;
+    using System.Threading.Tasks;
+    using Auditing;
+
+    interface IFailedAuditStorage
+    {
+        Task Store(dynamic failure);
+
+        Task ProcessFailedMessages(
+            Func<FailedTransportMessage, Func<CancellationToken, Task>, CancellationToken, Task> onMessage,
```
HEskandari, line 13: I wonder if it'd be better to have delegates defined for these and not rely on implicit `Func`. Especially in this case because we have `Func<T, Func<U, Task>, U, Task>`.

HEskandari (commented):
Although raised a few comments, it generally looks good and I'm happy if you want to go ahead and merge it.

src/ServiceControl.Audit/Auditing/AuditPersister.cs
```diff
@@ -50,16 +48,12 @@ public async Task<IReadOnlyList<MessageContext>> Persist(List<MessageContext> co
             }
 
             var storedContexts = new List<MessageContext>(contexts.Count);
-            BulkInsertOperation bulkInsert = null;
+            IAuditIngestionUnitOfWork unitOfWork = null;
```
jpalac, line 51: should we be using `default `instead of `null`?

jpalac (commented):
Looks good. In terms of the indexes, I have moved some other indexes to the persistence project (like `ExpiryKnownEndpointsIndex`). I think they should be ok to move since RavenDb startup code is all running from there - should just pick them all up.  Direct usage is from the persistence project too.
If they are moved then we can change this code in `RavenBootstrapper` to not have the last assembly:
```
public static Assembly[] IndexAssemblies =
        {
            typeof(RavenBootstrapper).Assembly, typeof(SagaSnapshot).Assembly, typeof(Auditing.FailedAuditImportIndex).Assembly
        };
```

HEskandari (approved):

src/ServiceControl.Audit/Infrastructure/WebApi/Cors.cs
mikeminutillo: I didn't want to cause conflict with the existing extension method in places where we _do_ need to import the RavenDB namespace.

PR #3009, opened 2022-08-29
Add support for stable message contact generation to audit load generator

andreasohlund:
This pulls in the improvements made by the previous analysis task force

### DIFF
--- a/src/ServiceControl.LoadTests.AuditGenerator/Program.cs
+++ b/src/ServiceControl.LoadTests.AuditGenerator/Program.cs
@@ -1,115 +1,397 @@
 ﻿namespace ServiceControl.LoadTests.AuditGenerator
 {
     using System;
+    using System.Collections.Concurrent;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Text;
     using System.Threading;
     using System.Threading.Tasks;
-    using Messages;
-    using Metrics;
     using NServiceBus;
+    using NServiceBus.Extensibility;
+    using NServiceBus.Features;
+    using NServiceBus.Routing;
     using NServiceBus.Support;
+    using NServiceBus.Transport;
+    using ServiceControl.Infrastructure.Metrics;
     using Transports;
 
     class Program
     {
-        const string ConfigRoot = "AuditGenerator";
+        public static readonly string HostId = Guid.NewGuid().ToString();
 
-        static void Main()
+        public static IDispatchMessages Dispatcher { get; set; }
+
+        static async Task Main(string[] commandLineArgs)
         {
-            Start().GetAwaiter().GetResult();
-        }
+            bodySize = int.Parse(commandLineArgs[0]);
+            var transportCustomizationName = commandLineArgs.Length > 1
+                ? commandLineArgs[1]
+                : QueueLengthProviderMsmqTransportCustomizationName;
 
-        static async Task Start()
-        {
-            Metric.Config.WithReporting(r =>
+            var connectionString = commandLineArgs.Length > 2 ? commandLineArgs[2] : null;
+
+            var transportCustomization = (TransportCustomization)Activator.CreateInstance(Type.GetType(transportCustomizationName, true));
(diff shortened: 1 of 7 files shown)


src/ServiceControl.LoadTests.AuditGenerator/Program.cs
```diff
@@ -1,115 +1,397 @@
...
+            var commands = new (string, Func<CancellationToken, string[], Task>)[]
+            {
+                ("f|Fill the sender queue. Syntax: f <number of messages> <number of tasks> <destination>",
+                    (ct, args) => Fill(args, Dispatcher)),
+                ("s|Start sending messages to the queue. Syntax: s <number of tasks> <destination>",
+                    (ct, args) => FullSpeedSend(args, Dispatcher, ct)),
+                ("t|Throttled sending that keeps the receiver queue size at n. Syntax: t <number of msgs> <destination>",
+                    (ct, args) => ConstantQueueLengthSend(args, Dispatcher, queueLengthProvider, ct)),
+                ("c|Constant-throughput sending. Syntax: c <number of msgs per second> <destination>",
+                    (ct, args) => ConstantThroughputSend(args, Dispatcher, ct))
```
mauroservienti, line 62: One thing to keep in mind is that we had various degrees of success with these options in the previous TF. They felt pretty unstable in their behavior. We fixed one (I don't recall which one) and used that only. @danielmarbach, do you remember which option we used to load test RavenDB 5?

danielmarbach: Throttled I think. The constant one overloads things

mikeminutillo: I think we can remove them.

- **f** might be useful to do a quick smoke test but we can just run **t** for a while to do the same job.
- **s** will pretty much always overwhelm the system
- **c** if we set it too high then the SC instance will get overwhelmed and then it's functionally not different to **s**. If we set it too low then the SC instance will spend a bunch of time idle. This one _might_ be useful if ServiceControl ever does any sort of adaptive processing (i.e. turning off features to dedicate more resources to ingestion if the queue length gets high enough). Even then, we'd probably be more interested in simulating waves of peak processing.

andreasohlund: Decided to keep c and t

andreasohlund:
This one should be good for final review now

mikeminutillo (approved):
This looks good. A couple of minor comments but nothing that would stop this from being merged

src/ServiceControl.LoadTests.AuditGenerator/Program.cs
```diff
@@ -1,115 +1,397 @@
...
+                {
+                    return Task.CompletedTask;
+                }
+            }
+        }
 
-            var config = new EndpointConfiguration(endpointName);
-            config.AssemblyScanner().ExcludeAssemblies("ServiceControl");
-            config.UseSerialization<NewtonsoftJsonSerializer>();
+        static void GenerateLayouts()
```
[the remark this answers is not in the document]
andreasohlund: agreed, POA item added to consider this in a separate PR

src/ServiceControl.LoadTests.AuditGenerator/Program.cs
```diff
@@ -1,115 +1,397 @@
...
+                        MessageBaseLayouts.Add((intermediateLayout, propertyPositionPlaceHolder));
+                        break;
+                    }
+                }
+            }
+        }
+
+        public static string RandomString(int length, Random random)
+        {
+            const string chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789";
```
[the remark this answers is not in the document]
andreasohlund: I guess it could become a problem for future features in SC that parses the payloads in some way

@danielmarbach do U think it would have a negative impact if we just removed the `0123456789` part?

PR #3013, opened 2022-09-01
Fix NRE when service registry is not created

HEskandari:


jpalac (approved):

src/ServiceControlInstaller.Engine/Instances/ServiceControlBaseService.cs
```diff
@@ -202,8 +202,12 @@ protected string DefaultLogPath()
         {
             // The default Logging folder in ServiceControl uses the env vae"%LocalApplicationData%".  Since this is env user specific we'll determine it based on profile path instead.
             // This only works for a user that has already logged in, which is fine for existing instances
-            var userAccountName = UserAccount.ParseAccountName(Service.Account);
+            if (Service.Account == null)
+            {
+                return null;
+            }
 
+            var userAccountName = UserAccount.ParseAccountName(Service.Account);
```
mikeminutillo, line 210: ParseAccountName explicitly handles `null` as input. It should convert null into the `LocalSystem` account.

mikeminutillo:
Why is the Service Registry not created?

src/ServiceControlInstaller.Engine/Instances/ServiceControlBaseService.cs
jpalac: But I believe it throws an `ArgumentNullException` if the argument passed into the `Contains` call is null

jpalac: You're right - I just tested passing in null into `ParseAccountName` and that returned the system account.
So I think we can get rid of that null check.

HEskandari:
@mikeminutillo that's the million-dollar question. From my traces (and it is hard to debug this as it is a race condition) the UI refresh happens first before registry changes are there.

src/ServiceControlInstaller.Engine/Services/WindowsServiceController.cs
```diff
@@ -35,7 +35,7 @@ public string Account
                 var account = (string)ReadValue("ObjectName");
                 if (account == null)
                 {
-                    throw new NullReferenceException("ServiceAccount entry not found in the registry");
+                    return null;
```
jpalac, line 38: If account is null and we want to return null, you don't need the null check at all - just return account

jpalac (commented):
I'm finding that if I don't have Service Pulse running then the error (ie getting a null value from the registry) is not happening - does that even make sense?

jpalac:
You also get the error if you have the main and audit instance, then delete the main one and click Refresh on the UI.

abparticular:
> Why is the Service Registry not created?

The problem is more that it no longer exists.  There is an "old" instance of InstanceDetailsViewModel in memory that is tied to the deleted SC installation that is still handling Refresh messages.

I'm trying to figure out the life cycle of InstanceDetailsViewModel objects to see if there is some way to have it ignore Refresh messages after it is no longer valid.

jpalac:
Perhaps in the `DeleteServiceControlInstanceCommand` before it finishes it could publish an `InstanceDeleted` event which would contain the instance name. Instances subscribe to that event and if it's theirs (ie matches their instance name) then they set a flag on the viewodel (ie Deleted = true) and then when reload/refresh is called that flag could be checked first.

abparticular (approved):

jpalac (approved):

mikeminutillo (approved):

PR #3017, opened 2022-09-01
Adding a raven5 persistence to Audit instance

WojcikMike:
So far I copied the other persistence and I am updating the references, namespace and such


src/ServiceControl.Audit/Auditing/FailedAuditImport.cs
```diff
@@ -1,10 +1,8 @@
 ﻿namespace ServiceControl.Audit.Auditing
 {
-    using System;
-
     public class FailedAuditImport
     {
-        public Guid Id { get; set; }
+        public string Id { get; set; }
```
mikeminutillo, line 5: RavenDb 5 generates non-guid ids by default. This is how it was [handled in the prior upgrade branch](https://github.com/Particular/ServiceControl/blob/update_raven_to_5/src/ServiceControl.Audit/Auditing/FailedAuditImport.cs). We need to verify that RavenDb 3.5 handles this change.

src/ServiceControl.Audit.Persistence.Tests/AuditTests.cs
```diff
@@ -2,26 +2,32 @@
...
         {
-            var metadata = new Dictionary<string, object>();
+            var metadata = new Dictionary<string, object>
+            {
+                { "MessageIntent", MessageIntentEnum.Send },
+                { "CriticalTime", TimeSpan.FromSeconds(5) },
+                { "ProcessingTime", TimeSpan.FromSeconds(1) },
+                { "DeliveryTime", TimeSpan.FromSeconds(4) },
+                { "IsSystemMessage", false },
+                { "ContentLength", 25 }
```
mikeminutillo, line 24: Note that all of these values are required for the RavenDb 5 implementation or it will not deserialize

src/ServiceControlInstaller.Engine/Instances/ServiceControlAuditNewInstance.cs
```diff
@@ -0,0 +1,96 @@
...
+    {
+        public static ServiceControlAuditNewInstance CreateWithDefaultPersistence()
+        {
+            return CreateWithDefaultPersistence(Assembly.GetExecutingAssembly().Location);
+        }
+
+        public static ServiceControlAuditNewInstance CreateWithDefaultPersistence(string deploymentCachePath)
+        {
+            var zipInfo = ServiceControlAuditZipInfo.Find(deploymentCachePath);
+            var persistenceManifest = ServiceControlAuditPersisters.LoadAllManifests(zipInfo.FilePath)
```
mikeminutillo, line 25: We seem to frequently load these manifests. I wonder if we should introduce an `AuditInstanceFactory` or something from the Deploy Cache Path that can load this stuff once and cache it.

src/ServiceControlInstaller.Engine/Instances/ServiceControlAuditNewInstance.cs
```diff
@@ -0,0 +1,96 @@
...
+                .Single(manifest => manifest.Name == "RavenDb5");
```
mikeminutillo, line 26: We have two "default" persistences. One for existing instances that don't have a persistence configured (Raven3.5) and one for brand new instances (Raven5). This info is scattered around in a few different locations. Maybe we can consolidate that somehow.

mikeminutillo (commented):
Review of the installer and packaging changes

andreasohlund (approved):

PR #3025, opened 2022-09-09
Default audit retention to same as SCMU

andreasohlund:
Since this allows us to remove the app.configs from the atts


src/ServiceControl.Audit/Auditing/AuditIngestionFaultPolicy.cs
```diff
@@ -68,7 +68,7 @@ await DoLogging(exception, failure)
 
         async Task DoLogging(Exception exception, dynamic failure)
         {
-            var id = Guid.NewGuid();
+            var id = Guid.NewGuid().ToString();
```
andreasohlund, line 71: @mikeminutillo I had to do this to get the tests to pass, looks related to https://github.com/Particular/ServiceControl/pull/3025/files#diff-f4bfa645e79dc51b6b0a1e892230e1a31eae7909c51149d6da22748c71e3d355R21

mikeminutillo (approved):

PR #3026, opened 2022-09-09
Adjust storage packaging to allow multiple audit stores

andreasohlund:


andreasohlund:
@mauroservienti adding you in the loop since you kindly volunteered ;)

PR #3074, opened 2022-09-28
Fix ATT execution when custom db name is defined per for each test

mauroservienti:


andreasohlund:
I know what causes the need for the hack, I’ll refactor and push a fix for that tomorrow so we can remove it

src/ServiceControl.Audit.AcceptanceTests/TestSupport/ServiceControlComponentRunner.cs
```diff
@@ -162,6 +166,13 @@ async Task InitializeServiceControl(ScenarioContext context)
 
                 var runtimePersistenceSettings = new PersistenceSettings(settings.AuditRetentionPeriod, settings.EnableFullTextSearchOnBodies, settings.MaxBodySizeToStore);
 
+                if (databaseName != null)
+                {
+                    // we want to preserve the database name across execution, otherwise when SC is started after setup
```
WojcikMike, line 171: This is needed only for ATT right?

andreasohlund: Correct

andreasohlund (approved):

PR #3075, opened 2022-09-29
Separate setup phase from runtime phase

andreasohlund:


andreasohlund:
I've now made the resolution of IDocumentStore happen via a provider so given that the persistence hosted service will start first we should be all good.

Please review

src/ServiceControl.Audit/Persistence/PersistenceLifecyleManager.cs
```diff
@@ -0,0 +1,22 @@
+﻿namespace ServiceControl.Audit.Persistence
+{
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Extensions.Hosting;
+
+    class PersistenceLifecyleManager : IHostedService
```
WojcikMike, line 7: I am not a big fan of that name. We have PersistenceLifecycle and PersistentLifecycleManager. How about we call this PersistentLifecycleHostedService? (most hosted services have that in their names in SC codebase)

WojcikMike (approved):
Only one comment abotu class name, but the rest looks good!

src/ServiceControl.Audit/Persistence/PersistenceLifecyleManager.cs
andreasohlund: Agreed, renamed to `PersistenceLifecycleHostedService`

src/ServiceControl.Audit.Persistence/IPersistenceConfiguration.cs
```diff
@@ -1,9 +1,12 @@
 ﻿namespace ServiceControl.Audit.Persistence
 {
+    using System.Threading;
+    using System.Threading.Tasks;
     using Microsoft.Extensions.DependencyInjection;
 
     public interface IPersistenceConfiguration
     {
-        void ConfigureServices(IServiceCollection serviceCollection, PersistenceSettings settings);
+        IPersistenceLifecycle ConfigureServices(IServiceCollection serviceCollection, PersistenceSettings settings);
```
mikeminutillo, line 9: It feels like this needs another name now because it's doing more than just configure the services.

mikeminutillo (approved):
Overall I like it. I wonder if the Setup call was moved to the Persistence Lifecycle interface if it might simplify a few things. I will try that in a separate PR.

src/ServiceControl.Audit.Persistence.RavenDb5/RavenDbPersistenceConfiguration.cs
```diff
@@ -1,74 +1,35 @@
...
-            serviceCollection.AddSingleton(settings);
-            serviceCollection.AddSingleton<IRavenDbSessionProvider, RavenDbSessionProvider>();
-            serviceCollection.AddSingleton<IAuditDataStore, RavenDbAuditDataStore>();
-            serviceCollection.AddSingleton<IAuditIngestionUnitOfWorkFactory, RavenDbAuditIngestionUnitOfWorkFactory>();
-            serviceCollection.AddSingleton<IFailedAuditStorage, RavenDbFailedAuditStorage>();
-            serviceCollection.AddSingleton<IRavenDbDocumentStoreProvider>(_ => persistenceLifeCycle);
+            var dataBaseConfiguration = GetDatabaseConfiguration(settings);
+            var expirationProcessTimerInSeconds = GetExpirationProcessTimerInSeconds(settings);
+            var databaseSetup = new DatabaseSetup(expirationProcessTimerInSeconds, settings.EnableFullTextSearchOnBodies, dataBaseConfiguration);
+            var lifecycle = CreateLifecycle(settings, dataBaseConfiguration);
```
mauroservienti, line 14: I understand what's going on. However, something seems to be off with the naming. We're creating the lifecycle here and not in `CreateLifecyle` where instead we're only configuring DI.

Shouldn't the lifecycle creation be moved to the `RavenDb5Persistence` class?

src/ServiceControl.Audit.Persistence/IPersistence.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace ServiceControl.Audit.Persistence
+{
+    using Microsoft.Extensions.DependencyInjection;
+
+    public interface IPersistence
+    {
+        IPersistenceLifecycle CreateLifecycle(IServiceCollection serviceCollection);
```
mauroservienti, line 7: `CreateLifecycle` sounds a bit off as a name because one doesn't understand why it gets an `IServiceCollection` as an argument. `Configure` seems to better align with MS APIs accepting a service collection.

andreasohlund: agreed, will come up with a better name

src/ServiceControl.Audit.Persistence.RavenDb5/RavenDbPersistenceConfiguration.cs
andreasohlund: Agreed, I have now moved it

PR #3078, opened 2022-10-01
Adding another missing Path.GetDirectoryName

WojcikMike:
Another example of that. When creating Audit instance it was passing file path and not directory path


mauroservienti (approved):
Are this and the previous one signs that we're missing some tests?

andreasohlund (approved):

PR #3084, opened 2022-10-04
Fix multi instance tests

andreasohlund:


src/ServiceControl.MultiInstance.AcceptanceTests/TestSupport/ServiceControlComponentRunner.cs
```diff
@@ -271,8 +272,14 @@ async Task InitializeServiceControlAudit(ScenarioContext context, int instancePo
                 typeof(ServiceControlComponentRunner).Assembly.GetName().Name
             };
 
+            ConfigurationManager.AppSettings.Set("ServiceControl.Audit/PersistenceType", typeof(Audit.Persistence.RavenDb.RavenDbPersistenceConfiguration).AssemblyQualifiedName);
```
mauroservienti, line 275: Why not in-memory for the multi-instance tests?

andreasohlund: I tried, but we have some tests that assume raven35 is in use.

I agree that we should use InMemory but I wanted to get the tests green as a first step

PR #3097, opened 2022-10-06
Remove the ravendb5 use embedded setting since its no longer used

andreasohlund:


src/ServiceControl.Audit.Persistence.RavenDb5/RavenDbPersistenceConfiguration.cs
```diff
@@ -97,11 +96,11 @@ static int GetExpirationProcessTimerInSeconds(PersistenceSettings settings)
...
-        internal const string ConnectionStringKey = "ServiceControl/Audit/RavenDb5/ConnectionString";
-        internal const string HostNameKey = "ServiceControl.Audit/HostName";
-        internal const string DatabaseMaintenancePortKey = "ServiceControl.Audit/DatabaseMaintenancePort";
-        internal const string ExpirationProcessTimerInSecondsKey = "ServiceControl.Audit/ExpirationProcessTimerInSeconds";
+        public const string DatabaseNameKey = "ServiceControl/Audit/RavenDb5/DatabaseName";
+        public const string DatabasePathKey = "ServiceControl.Audit/DbPath";
+        public const string ConnectionStringKey = "ServiceControl/Audit/RavenDb5/ConnectionString";
+        public const string HostNameKey = "ServiceControl.Audit/HostName";
+        public const string DatabaseMaintenancePortKey = "ServiceControl.Audit/DatabaseMaintenancePort";
+        public const string ExpirationProcessTimerInSecondsKey = "ServiceControl.Audit/ExpirationProcessTimerInSeconds";
```
mauroservienti, line 104: Is the reason for these to be public to use them in tests? If that’s the case can we use an `InternarlsVisibleTo` attribute?

andreasohlund: The acceptance tests use them so I was unsure if that was a good idea?

I think this is good enough for now since we don't expose this as a public API

PR #3108, opened 2022-10-11
Prevent capitalization of certain boolean config values

andreasohlund:
Some boolean app settings are being capitalized and some are not. This is causing developers who write configuration management code (puppet, chef, ansible, etc) to have to ensure they properly capitalize or not capitalize settings which is rather annoying.


WojcikMike (approved):

andreasohlund:
@TraGicCode realized that you had hoped for some hacktoberfest points but we couldn't come up with a way to get CI running so we could merge. 

Since then we realized that we could have run the other PR to see that all tests passed and then force merge yours. Sorry about this!

andreasohlund:
Commented on the wrong pr @TraGicCode but you get it :)

PR #3140, opened 2022-10-21
Display persister name in SCMU

andreasohlund:
image


andreasohlund:
See screenshot in description

mauroservienti (approved):

src/ServiceControlInstaller.Engine/Instances/PersistenceManifest.cs
```diff
@@ -8,6 +8,8 @@ public class PersistenceManifest
 
         public string Name { get; set; }
 
+        public string DisplayName { get; set; }
```
andreasohlund, line 11: @mikeminutillo fyi I added a display name to enable blank spaces in the name

PR #3144, opened 2022-10-24
Show persistence package name in powershell

mikeminutillo:
Adds a `PersistencePackageName` property to the powershell output for audit instances.

image

This mirrors the transport package name property that was already present.

### DIFF
--- a/src/ServiceControlInstaller.PowerShell/Cmdlets/AuditInstances/PsAuditInstance.cs
+++ b/src/ServiceControlInstaller.PowerShell/Cmdlets/AuditInstances/PsAuditInstance.cs
@@ -18,6 +18,8 @@
         public string TransportPackageName { get; set; }
         public string ConnectionString { get; set; }
 
+        public string PersistencePackageName { get; set; }
+
         public string AuditQueue { get; set; }
         public string AuditLogQueue { get; set; }
         public bool ForwardAuditMessages { get; set; }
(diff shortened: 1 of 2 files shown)


src/ServiceControlInstaller.PowerShell/Cmdlets/AuditInstances/PsAuditInstance.cs
```diff
@@ -53,6 +55,7 @@ public static PsAuditInstance FromInstance(ServiceControlAuditInstance instance)
                 AuditRetentionPeriod = instance.AuditRetentionPeriod,
                 ServiceControlQueueAddress = instance.ServiceControlQueueAddress,
                 EnableFullTextSearchOnBodies = instance.EnableFullTextSearchOnBodies,
+                PersistencePackageName = instance.PersistenceManifest?.Name
```
andreasohlund, line 58: Should we use the display name instead?

andreasohlund (approved):

src/ServiceControlInstaller.PowerShell/Cmdlets/AuditInstances/PsAuditInstance.cs
mikeminutillo: That's not what the transport one is doing. This might get used for programatic purposes. i.e. Get the persistence of an instance and create a new instance with the same persistence.

andreasohlund: Fair enough

mikeminutillo: I stand corrected, the transport package name is a display name. I will change this to match.

mikeminutillo:
Closed in favor of #3146

PR #3146, opened 2022-10-24
Show persistence package name in powershell

mikeminutillo:
Adds a `PersistencePackageName` property to the powershell output for audit instances.

image

This mirrors the transport package name property that was already present.


andreasohlund (approved):

andreasohlund:
Nitpick but I don't like the "Name" suffix, but I guess it aligns with the transport name but I don't like that either :)

I would go with "Transport" and "Persistence"

PR #3153, opened 2022-10-27
Update installer to support RavenDb5

mikeminutillo:
Adds VS 2015-2019 C++ Redistributable (x64) as a prerequisite.


andreasohlund:
Is it possible to add .NET as a prereq as well?

PR #3160, opened 2022-11-01
Include RavenDB 3.5 persister in main instance docker images

andreasohlund:
I realized that when we adjusted how we package the main instance we broke the docker images. This should fix that.


andreasohlund:
@mikeminutillo I think we do have a problem with check , see this PR as an example 
image

andreasohlund:
@mauroservienti would you be able to test this one?

mauroservienti:
Yes, I will

mikeminutillo (approved):

mauroservienti:
Tested ✅

PR #3183, opened 2022-11-08
Read persistence settings from all sources

andreasohlund:
This makes sure that docker images can use environment variables to set values


WojcikMike (approved):

mikeminutillo (approved):

mauroservienti (approved):
Smoke-tested ✅

Issue #2623, opened 2021-09-03
ServiceControl queue cannot be configured


dvdstelt:
In the ServiceControl Management utility, the `error` and `audit` queues can be configured, but there's also a queue for the ServiceControl error instance itself. The name of this queue is an exact copy of the instance name and cannot be configured. However this is the exact queue name that endpoint plugins are required to configure to be able to send information to ServiceControl. Customers should be able to change the name of the queue, unrelated to the name of the Windows service. It is not clear to customers that the two are the same.

Since documentation is more and more talking about `ServiceControl Error instance` perhaps we can by default rename the error instance to `ConfiguredName.Errors` (for example `Particular.ServiceControl.Errors`) just like we do with audit instance to `Particular.ServiceControl.Audit` and then copy the original name of `Particular.ServiceControl` and name that the servicecontrol input queue by default. And finally give customers the option to change that default name.

PR #3192, opened 2022-11-10
Allow the internal queues to be configurable

WilliamBZA:
Fixes https://github.com/Particular/ServiceControl/issues/2623

### DIFF
--- a/src/ServiceControl.Audit/Infrastructure/Settings/Settings.cs
+++ b/src/ServiceControl.Audit/Infrastructure/Settings/Settings.cs
@@ -20,6 +20,9 @@
                 ServiceName = DEFAULT_SERVICE_NAME;
             }
 
+            // Overwrite the service name if it is specified in ENVVAR, reg, or config file
+            ServiceName = SettingsReader<string>.Read("InternalQueueName", ServiceName);
+
             {
                 // order matters
                 AuditQueue = GetAuditQueue();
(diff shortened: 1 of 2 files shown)


src/ServiceControl.Audit/Infrastructure/Settings/Settings.cs
```diff
@@ -20,6 +20,9 @@ public Settings(string serviceName = null)
                 ServiceName = DEFAULT_SERVICE_NAME;
             }
 
+            // Overwrite the service name if it is specified in ENVVAR, reg, or config file
+            ServiceName = SettingsReader<string>.Read("InternalQueueName", ServiceName);
+
             {
```
awright18, line 26: I know this block was here previously but it struck me as odd. I wasn't aware that you could have random blocks in C#. Does it make sense to leave this? I'm ok either way, just new to me.

src/ServiceControl/Infrastructure/Settings/Settings.cs
```diff
@@ -24,6 +24,9 @@ public Settings(string serviceName = null)
                 ServiceName = DEFAULT_SERVICE_NAME;
             }
 
+            // Overwrite the service name if it is specified in ENVVAR, reg, or config file
+            ServiceName = SettingsReader<string>.Read("InternalQueueName", ServiceName);
+
             {
```
awright18, line 30: Same here if we remove the block { } above let's remove it here, if not, then lets not.

awright18 (approved):

Issue #2960, opened 2020-12-03
Validation errors do not show when creating new SC instances


timbussmann:
## Relevant information

Validation errors when trying to create new SC instances are hard to find. It's not obvious that those are visible as tooltip when hovering over the warning:

![image](https://user-images.githubusercontent.com/3524870/101065996-23fbc080-3596-11eb-833d-5efe547b0b45.png)

This was especially hard to figure out because I first got a validation error directly on the input field when I didn't specify a host. After providing the host and clicking "Add" again, I got no feedback on the failing input fields and I figured out an issue about conflicting port numbers myself before I realized by accident that the validation errors are shown as a tooltip.

Issue #2956, opened 2021-09-01
Users cannot install error instance without audit instance


dvdstelt:
When using the ServiceControl Management Utility, users are unable to install an error instance without an audit instance, even though these days they are separated. Even worse, after deleting the audit instance, you cannot easily add a new one without also creating an error instance.

Also, instead of naming it "ServiceControl instance", this should probably be changed to "ServiceControl error instance" to make it clear what this instance is for. It will usually only be installed for a first time or on a development machine, so it should not bother existing users that much.

PR #3206, opened 2022-11-15
Allow installing of single instance type from SCMU

WilliamBZA:
Fixes https://github.com/Particular/ServiceControl/issues/2956
Fixes https://github.com/Particular/ServiceControl/issues/2960

### DIFF
--- a/src/ServiceControl.Audit/Infrastructure/NServiceBusFactory.cs
+++ b/src/ServiceControl.Audit/Infrastructure/NServiceBusFactory.cs
@@ -34,14 +34,17 @@ namespace ServiceControl.Audit.Infrastructure
             configuration.GetSettings().Set("NServiceBus.PublishSubscribe.EnablePublishing", false);
 
             var serviceControlLogicalQueue = settings.ServiceControlQueueAddress;
-            if (serviceControlLogicalQueue.IndexOf("@") >= 0)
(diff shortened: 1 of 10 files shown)


WilliamBZA:
Adds checkboxes to select which instance type to install

image

awright18 (approved):

bording:
This seems to assume that at least one checkbox is checked when you click the Add button.

When I unchecked both and then clicked Add, I got a `NullReferenceException`.

Seems like we need some protection around having nothing selected.

WilliamBZA:
I've update this. It validates that at least one instance type is selected.

It also fixes the validation messages [not being displayed on the controls](https://github.com/Particular/ServiceControl/issues/2960).

awright18 (approved):
LGTM

bording (approved):

udidahan:
As a potential user reading the text of those checkboxes, I'm not sure I'd intuitively understand the different between "ServiceControl" and "ServiceControl Audit". Is there something we could do to make this clearer?

WilliamBZA:
@udidahan The wording was there before this PR and wasn't changed here, but I agree that it's something that should be made clearer. Do you know if there's already an issue tracking that?

udidahan:
I don't know.

WilliamBZA:
I created a [follow-up issue](https://github.com/Particular/ServiceControl/issues/3278) to try make the use of names clearer.

PR #3211, opened 2022-11-15
Delete CheckRavenDBIndexErrors custom check

mauroservienti:


WojcikMike (approved):
Looks good, but I think we have approval tests that need to be fixed

PR #3241, opened 2022-11-17
Use an MSBuild project in the solution to build docker images

### LINKED ISSUE #3184: Docker images build process behaves differently when run in CI or locally on developer machines
The docker files define what to include in the image by referencing the build's output. For example, the Amazon SQS image has the following lines: https://github.com/Particular/ServiceControl/blob/fcea37326731282103297002c031ac7fcd99067d/src/docker/servicecontrol.amazonsqs-windows.dockerfile#L5-L7

Those paths are the builds' output paths; they work consistently on developers' machines but don't work in CI. For that reason, CI performs the following when releasing: https://github.com/Particular/ServiceControl/blob/fcea37326731282103297002c031ac7fcd99067d/.github/workflows/release.yml#L135-L163 

The above tries to mimic the same folder layout on a developer machine. However, the content is different. What gets packaged in the zip files is not the build output but a subset. That causes images built locally to differ from the ones built in CI. When smoke testing, we only smoke test the locally produced images as running CI will push them to Docker Hub, making them public.

mauroservienti:
Fix #3184

This PR moves the step of building the docker images inside the ServiceControl solution. By default, the `ServiceControl.DockerImages` is not built when building the solution to reduce the overall build time.

To build it, manually build it from the IDE or the command line using MSBuild.

The project defines an item group with all the supported transport for which a docker image needs to be created. The following is an example of Amazon SQS transport:

```xml
<ItemGroup>
    <SupportedTransport Include="SQS">
      <DockerfileCustomization>amazonsqs</DockerfileCustomization>
      <TransportName>AmazonSQS</TransportName>
      <TransportCustomizationType>"ServiceControl.Transports.SQS.SQSTransportCustomization, ServiceControl.Transports.SQS"</TransportCustomizationType>
    </SupportedTransport>
    <SupportedTransport Include="Another Transport">
        <!-- stuff -->
    </SupportedTransport>
</ItemGroup>
```

The project contains dockerfile "templates" for each ServiceControl instance type and the accompanying sidecar init image. The project during the build process for each of the defined supported transports:

- Cleans the output folder
- Copies to the output folder the dockerfile template renaming it according to the `DockerfileCustomization` property
- Invokes the docker build passing as arguments the `TransportName` and `TransportCustomizationType` defined as ARG in the dockerfile


src/ServiceControl.DockerImages/README.md
```diff
@@ -0,0 +1,8 @@
+﻿  # ServiceControl.DockerImages
+  
+  
+  WARNING
+
+  This project is not automatically built when building the solution
+  to keep the overall build time under control.
+  To build Docker images explicitely build this project.
```
andreasohlund, line 8:
```suggestion
  To build Docker images explicitly build this project.
```

src/ServiceControl.DockerImages/ServiceControl.DockerImages.csproj
```diff
@@ -0,0 +1,46 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <!-- 
+  
+  WARNING
+  This project is not automatically built when building the solution
+  to keep the overall build time under control.
+  To build Docker images explicitely build this project.
```
andreasohlund, line 8:
```suggestion
  To build Docker images explicitly build this project.
```

andreasohlund (approved):

WojcikMike (approved):

andreasohlund (approved):

.gitignore
```diff
@@ -1,5 +1,6 @@
 nugets
 deploy
+dockerfiles
```
andreasohlund, line 3: Pretty sure reposync will revert this :(

.github/workflows/build-release.yml
```diff
@@ -68,95 +68,21 @@ jobs:
...
-              
-              $dockerbuildcmd = "docker build -t particular/" + $dockerImageName +":${{ needs.release.outputs.semver }}" + " -f .\src\docker\" + $_.Name + " ."
-              Invoke-Expression $dockerbuildcmd.ToLower()
-          }
-      - name: Login to Docker Hub
-        uses: docker/login-action@v2.1.0
-        with:
-          username: ${{ secrets.DOCKERHUB_USERNAME }}
-          password: ${{ secrets.DOCKERHUB_TOKEN }}
+        run: msbuild src/ServiceControl.DockerImages/ServiceControl.DockerImages.csproj -p:Configuration=Release -restore
```
danielmarbach, line 83: I think you can use dotnet directly and don't need to pull in msbuild

mauroservienti: Thanks, Daniel. I think it's correct. For now, we decided to keep it consistent with all the other places in GitHub Actions were we build ServiceControl and related stuff.

PR #3244, opened 2022-11-18
Move the push to Octopus in the Docker job

mauroservienti:
This is needed because the list of dockerfile is expanded (from templates) during the Docker images build process and is not available in the solution like it was before.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -39,6 +39,8 @@ jobs:
         env:
           AZURE_KEY_VAULT_CLIENT_SECRET: ${{ secrets.AZURE_KEY_VAULT_CLIENT_SECRET }}
         run: msbuild src -p:Configuration=Release -restore -m -verbosity:minimal
+      - name: Prepare dockerfiles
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -180,6 +107,26 @@ jobs:
...
+      - name: Download zips
+        uses: actions/download-artifact@v3.0.1
+        with:
+          name: zips
+          path: zip/*
+      - name: Deploy
+        uses: Particular/push-octopus-package-action@v1.0.0
+        with:
+          octopus-deploy-api-key: ${{ secrets.OCTOPUS_DEPLOY_API_KEY }}
+          additional-metadata-paths: ServiceControlMetadata.json
```
andreasohlund, line 129: Don't you need to download this one since it was created by the previous job?

mauroservienti: No, because the file is now created in the `docker` job (previously, it was in the `release` job) at line 91, right after building the docker images. Being in `docker` job is already where it's need to be at the right time.

andreasohlund (approved):

mikeminutillo (approved):

PR #3246, opened 2022-11-21
Clarify version mismatch docs

andreasohlund:


docs/packaging.md
```diff
@@ -11,27 +11,29 @@ Each zip has the following folder structure:
...
-  Expected: <empty>
-  But was:  < "Transports/AzureStorageQueue/System.Runtime.CompilerServices.Unsafe.dll", "Transports/AzureStorageQueue/System.Threading.Tasks.Extensions.dll" >
+## Assembly version mismatches
 
-   at Tests.DeploymentPackageTests.DuplicateAssemblyFileSizesShouldMatch() in /_/src/ServiceControlInstaller.Packaging.UnitTests/DeploymentPackageTests.cs:line 19
-```
+There can be an issue when the main instance folder and the selected transport each have a copy of the same assembly. This can happen when the transport/persistence package and the instance each reference the same dependency, but select different versions. At install time, one version or the other will be copied into the instance binary folder and things may break unexpectedly at runtime.
 
-> File sizes should match the ones in the **ServiceControl.Audit** folder. Check versions of dependencies.
+To prevent this, the unit test `DeploymentPackageTests.DuplicateAssemblyShouldHaveMatchingVersions` tests the generated zip files for duplicated assemblies. If their versions match, then the test passes. If not then the test will fail with:
```
mauroservienti, line 26:
```suggestion
To prevent this, the unit test `DeploymentPackageTests.DuplicateAssemblyShouldHaveMatchingVersions` tests if duplicated assemblies might be deployed. If their versions match, then the test passes. If not then the test will fail with:
```

docs/packaging.md
```diff
@@ -11,27 +11,29 @@ Each zip has the following folder structure:
...
+```
 
-> But was:  < "Transports/AzureStorageQueue/System.Runtime.CompilerServices.Unsafe.dll", "Transports/AzureStorageQueue/System.Threading.Tasks.Extensions.dll" >
+> Microsoft.Bcl.AsyncInterfaces.dll has a different version in Instance/ServiceControl.Audit compared to Transports/AzureServiceBus
 
-This is telling you the assemblies that are different from the ones found in the ServiceControl.Audit folder.
+This is telling you that the mismatch is happing in the ServiceControl.Audit and that the instance it self is clashing with the Azure ServiceBus transport and RavenDB 5 persister.
```
mauroservienti, line 36:
```suggestion
This tells that the mismatch is happening in ServiceControl.Audit and that the instance itself is clashing with the Azure ServiceBus transport and RavenDB 5 persister.
```

mauroservienti (approved):
Minor changes

andreasohlund:
damn, had turned on automerge

PR #3254, opened 2022-11-23
Set MinVer properties - 4.27

particularbot6:


ramonsmits:
@bording I'm not really sure if it is required for SC.

bording:
> @bording I'm not really sure if it is required for SC.

It is not. This would only need to be done for repos that released NServiceBus 8 components.

ramonsmits:
> It is not. This would only need to be done for repos that released NServiceBus 8 components.

@bording Not part of the RTM TF but it was more the question if this file needs updating to for MinVer but I guess it is not needed because we tag the release on the master branch?

Issue #3264, opened 2022-11-30
[4.26] Audit instances can't start when using a custom hostname


andreasohlund:
Since RavenDB5 will only allow localhost to be used if running unsecured any attempt to install/run a 4.26 audit instance with hostname set to anything other than `localhost` will fail with 

```
System.AggregateException: Failed to retrieve cluster topology from all known nodes http://servicepulse-dev:33335 -> {"Message":"Running in a potentially unsafe mode. Server certificate information has not been set up and the server address is
not configured within allowed unsecured access address range. Please find the RavenDB settings file *settings.json* in the server directory and fill in your certificate information in either Security.Certificate.Path or Security.Certificate.Load.Exec If you would rather like to keep your server unsecured, please relax the Security.UnsecuredAccessAllowed setting to match the ServerUrl setting value
```

### Workarounds

The following potential workarounds are known at this stage.

#### Use localhost for the audit instance

If the audit instance is running on the same machine as the main instance the audit instance can be reconfigured to use localhost:

1. [Set the hostname to](https://docs.particular.net/servicecontrol/audit-instances/creating-config-file#host-settings-servicecontrol-audithostname) `localhost` for the audit instance
2. [Update the remote instances setting in the main instance](https://docs.particular.net/servicecontrol/servicecontrol-instances/remotes#configuration-version-4-and-later) to use `localhost` as well

#### Configure a custom certificate

Create the instance using localhost as the host name.

Follow the instructions in https://ravendb.net/docs/article-page/5.3/csharp/server/security/authentication/certificate-configuration to secure the instance.

The `settings.json` file mentioned should be in the `{ServiceControl installation directiory}\RavenDBServer` directory.

Change the host name to the desired value.

#### Allow insecure access

⚠️ This allows insecure access to the database from the private network.

Create the instance using localhost as the host name.

Add a `settings.json` file in the `{ServiceControl installation directiory}\RavenDBServer` directory with the following content:

```
{
    "Security.UnsecuredAccessAllowed" : "PrivateNetwork"
}
```

See https://ravendb.net/docs/article-page/5.3/csharp/server/configuration/security-configuration#security.unsecuredaccessallowed for more details

Change the host name to the desired value.

PR #3270, opened 2022-12-01
Only expose RavenDB 5 on localhost

andreasohlund:
Fixes https://github.com/Particular/ServiceControl/issues/3264

### DIFF
--- a/src/ServiceControl.Audit.Persistence.RavenDb5/persistence.manifest
+++ b/src/ServiceControl.Audit.Persistence.RavenDb5/persistence.manifest
@@ -10,11 +10,6 @@
       "Mandatory": true
     },
     {
-      "Name": "ServiceControl.Audit/HostName",
-      "Mandatory": true
-
-    },
-    {
       "Name": "ServiceControl.Audit/DatabaseMaintenancePort",
       "Mandatory": true
     }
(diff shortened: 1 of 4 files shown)


src/ServiceControl.Audit.Persistence.RavenDb5/persistence.manifest
```diff
@@ -9,11 +9,6 @@
       "Name": "ServiceControl.Audit/DBPath",
       "Mandatory": true
     },
-    {
-      "Name": "ServiceControl.Audit/HostName",
-      "Mandatory": true
-
-    },
```
mauroservienti, line 16: Isn't this also used to set up ServiceControl API?

andreasohlund: Yes but this only controls what the persisted needs to know, the instance it self will still have and use the setting

mikeminutillo (approved):
As a consequence of this, the database will not be available outside of the local server, even in maintenance mode. I don't think that's an issue but it should probably be mentioned in the documentation somwhere.

andreasohlund:
@mikeminutillo docs already states that localhost must be used

https://docs.particular.net/servicecontrol/audit-instances/creating-config-file#troubleshooting-ravendb-5

where

3.5 points out that hostname also works

https://docs.particular.net/servicecontrol/audit-instances/creating-config-file#troubleshooting-ravendb-3-5

PR #3279, opened 2022-12-05
📦 Bump Azure.Message.ServiceBus to 7.11.1

ramonsmits:
Bump Azure.Message.ServiceBus to 7.11.1, previously a transitive dependency and using 7.4.0


WilliamBZA (approved):

bording:
@ramonsmits Since you are going ahead with incrementing the transport dependency like we discussed, we should probably have closed this PR to avoid having it merged.

Since it has been, let's make sure to undo this change when we update the transport.

PR #3294, opened 2022-12-08
Fix deprecated set-output calls in release workflows (4.26)

DavidBoike:


bording:
4.27 is about to be released, so this release branch is going soon after. I think we can just close this instead.

DavidBoike:
:+1:

PR #3356, opened 2023-01-12
Exclude the changes and sagatype properties from indexing

andreasohlund:
We are not querying them so this will speed up indexing

See https://ravendb.net/docs/article-page/5.4/java/indexes/using-analyzers#manipulating-field-indexing-behavior for more details

This was also recommended by the RavenDB team as a workaround for https://issues.hibernatingrhinos.com/issue/RavenDB-19762/Failed-to-index-a-large-reduce-result

### DIFF
--- a/src/ServiceControl.Audit.Persistence.RavenDb5/Indexes/SagaDetailsIndex.cs
+++ b/src/ServiceControl.Audit.Persistence.RavenDb5/Indexes/SagaDetailsIndex.cs
@@ -15,17 +15,17 @@ namespace ServiceControl.SagaAudit
                                              doc.SagaType,
                                              Changes = new[]
                                              {
-                        new SagaStateChange
-                        {
-                            Endpoint = doc.Endpoint,
-                            FinishTime = doc.FinishTime,
-                            InitiatingMessage = doc.InitiatingMessage,
-                            OutgoingMessages = doc.OutgoingMessages,
-                            StartTime = doc.StartTime,
-                            StateAfterChange = doc.StateAfterChange,
-                            Status = doc.Status
-                        }
-                    }
+                                                new SagaStateChange
+                                                {
+                                                    Endpoint = doc.Endpoint,
+                                                    FinishTime = doc.FinishTime,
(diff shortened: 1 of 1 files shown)


src/ServiceControl.Audit.Persistence.RavenDb5/Indexes/SagaDetailsIndex.cs
```diff
@@ -52,6 +52,9 @@ into g
                                         .OrderByDescending(x => x.FinishTime)
                                         .ToList()
                                 };
+
+            Index(x => x.SagaType, FieldIndexing.No);
+            Index(x => x.Changes, FieldIndexing.No);
```
SzymonPobiega, line 57: I think we also need to add

```
Stores.Add(x => x.Changes, FieldStorage.Yes);
```

as described in https://ravendb.net/docs/article-page/5.4/Csharp/indexes/using-analyzers#manipulating-field-indexing-behavior

andreasohlund: Tests are all green but I'll add a few new ones to make sure, stat tuned

andreasohlund: @SzymonPobiega we already have tests to check that we return data see https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl.Audit.Persistence.Tests/SagaHistoryTests.cs#L28

I did add one new test to check that we do not index the saga type and changes columns

SzymonPobiega: I think the difference might be that if you don't do `FieldStorage.Yes` then when you query by index, it will have to load the documents to return the results. But now that I think about it, that's actually the correct way as the system is write-heavy, not read-heavy. So 👍

SzymonPobiega (approved):

cquirosj (approved):

PR #3358, opened 2023-01-12
Zip PowerShell module in assets folder

bording:
Follow-up to #3354 

The module needs to be zipped in the `assets` folder to better integrate with the existing Octopus workflows.

This also adds the zip step to the CI workflow so that the module continues to be available in those runs as well.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,12 +73,12 @@ jobs:
       - name: Build
         run: msbuild src -p:Configuration=Release -restore -m -verbosity:minimal
       - name: Zip PowerShell module
-        if: matrix.transport == 'Msmq' && matrix.storage == 'RavenDB35'
+        if: matrix.transport == 'LearningTransport' && matrix.storage == 'SqlDb'
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -72,7 +72,12 @@ jobs:
         uses: microsoft/setup-msbuild@v1.3
       - name: Build
         run: msbuild src -p:Configuration=Release -restore -m -verbosity:minimal
-      - name: Upload packages
+      - name: Zip PowerShell module
+        if: matrix.transport == 'Msmq' && matrix.storage == 'RavenDB35'
```
DavidBoike, line 76: Should we change this to use a condition that is more likely to still exist in ~3 years? (Line 81 too)

Another option is to add a `upload-packages: true` to one of the configs but the addition of the storage aspect of the matrix does complicate that.

:shrug:

DavidBoike (approved):

.github/workflows/ci.yml
bording: Yeah, I could change it to something else in this PR if you want.

Issue #3364, opened 2023-01-19
Audit instances logs to the installation folder


andreasohlund:
As of version 4.26 new audit instances will try to log to the installation folder and fail on startup since its unlikely for the user account to have write permissions there

### Symptoms

ServiceControl audit fails on startup with 

```
System.UnauthorizedAccessException: Access to the path '{installation-path}\RavenDBServer\Packages\NuGet$1\test.file' is denied.
   at System.IO.FileSystem.DeleteFile(String fullPath)
   at System.IO.File.Delete(String path)
   at Raven.Server.Utils.IOExtensions.TryHandlingError(String directory, Int32 i, Exception e)
```

Relates to https://github.com/Particular/ServiceControl/issues/3253

### Workaround 

As a workaround, grant the user account _write_ permissions to the ServiceControl audit instance's installation directory and subdirectories. 

The default installation path for Service Control Audit instances is 
```
C:\Program Files (x86)\Particular Software\<instance name>
```

andreasohlund:
Embedded raven has an explicit `logsPath` option that could be used to solve this, a quick spike proved that it works

https://github.com/Particular/ServiceControl/pull/3367

PR #3367, opened 2023-01-20
Configure embedded RavenDB 5 to use the configured log directory of the audit instance

andreasohlund:
Fixes https://github.com/Particular/ServiceControl/issues/3364

### DIFF
--- a/src/ServiceControl.Audit.Persistence.RavenDb5/EmbeddedDatabase.cs
+++ b/src/ServiceControl.Audit.Persistence.RavenDb5/EmbeddedDatabase.cs
@@ -34,7 +34,7 @@
             {
                 CommandLineArgs = new List<string>
                 {
-                    $"--License.Path=\"{localRavenLicense}\" --Logs.Mode={databaseConfiguration.ServerConfiguration.LogsMode}"
+                    $"--License.Path=\"{localRavenLicense}\" --Logs.Mode={databaseConfiguration.ServerConfiguration.LogsMode} --Indexing.NuGetPackagesPath=\"c:\\bleble\""
                 },
                 AcceptEula = true,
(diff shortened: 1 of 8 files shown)


andreasohlund:
@mauroservienti @DavidBoike it seems like raven uses this path no matter what so that might explain why we couldn't change it using the `settings.json` file. I would assume things like loglevel would work using that approach but haven't tested

andreasohlund:
>Well that’s good, that means we don’t have to worry about rewriting a JSON file if the SC configuration changes, right?

Yea

We do have https://docs.particular.net/servicecontrol/logging#logging-levels-versions-1-9-and-above that now is incorrect that would require a json file but we can handle that as a separate less urgent issue

mikeminutillo:
I can confirm that the Raven Embedded source code [always sets a value](https://github.com/ravendb/ravendb/blob/v5.4/src/Raven.Embedded/ServerOptions.cs#L16) and [passes it on the command line](https://github.com/ravendb/ravendb/blob/v5.4/src/Raven.Embedded/RavenServerRunner.cs#L36). We cannot even [clear out the default](https://github.com/ravendb/ravendb/blob/v5.4/src/Raven.Embedded/RavenServerRunner.cs#L21-L22) so we must always set it.

mikeminutillo:
I made a [slight tweak](https://github.com/Particular/ServiceControl/pull/3367/commits/0466488ea701271f3708fa947bd6dc75fe80b63a) to ensure that the config value is copied into the persister specific settings. Unfortunately the acceptance tests bypass this by using the same server instance for every test, and the persistence tests bypass it because they test below the configuration layer.

mikeminutillo (approved):

tmasternak (approved):

tmasternak (approved):

PR #3385, opened 2023-01-31
Enable tracking successfully processed audit messages

DavidBoike:
This PR adds:

* API to quickly access endpoint throughput data (RavenDB 5 audit instances only)
* `/configuration/remotes` endpoint to show configuration of remote audit instances


src/ServiceControl.Audit.Persistence.RavenDb5/Indexes/AuditCountIndex.cs
```diff
@@ -0,0 +1,40 @@
+﻿namespace ServiceControl.Audit.Persistence.RavenDb.Indexes
+{
+    using System.Linq;
+    using Raven.Client.Documents.Indexes;
+    using ServiceControl.Audit.Auditing;
+    using ServiceControl.Audit.Auditing.MessagesView;
+    using ServiceControl.Audit.Monitoring;
+
+    public class AuditCountIndex : AbstractIndexCreationTask<ProcessedMessage, DailyAuditCount>
```
andreasohlund, line 9: LGTM to me from a code perspective. 

My only concern would be making sure that this doesn't impact performance. 

@mauroservienti given the perf test stuff we built during the RavenDB5 TF how much effort would you say it would be to spin up a VM and do a test to verify?

DavidBoike: Is any of that (how to do those tests) documented?

mauroservienti: Yes-ish 😬

andreasohlund: We have https://github.com/Particular/Platform/blob/main/documentation/service-control-load-testing.md and https://github.com/Particular/RavenDB.StatsCollector

Not much but me or @mauroservienti should be able to help out to get you going.

Did we have anything around how to setup the VM mauro?

mauroservienti: No, but there is nothing special to do. Just install ServiceControl as usual or run it from the command line. Use the load generator with MSMQ to load messages and the stats collector to capture RavenDB stats. Both are command line tools

DavidBoike, line 9: By the way, do y'all think it's OK here to use the DailyAuditCount type, which lives in the base persistence project, as the ReduceResult here? I wrestled with the coupling implications. Alternative would be to use an internal class `AuditCountIndex.Result` but then the structure would mean a LOT of mapping.

Meanwhile I think the complex structure of the result is good, I'm trying to keep the index so that it's completely time-based on the day so that when audit expiry happens and all the audit information for a day evaporates, it's just 1 thing that Raven has to also throw away for this index.

mauroservienti: I don't see any issue directly using the `DailyAuditCount` type. It's already a DTO on its own, no need for more mappings.

DavidBoike:
@andreasohlund Please re-review. I tested everything in my Azure VM (not a load test, since there's no longer anything affecting ingestion) and it's good to go except for the Raven license issue that would prevent us from releasing anything.

andreasohlund (approved):

PR #3439, opened 2023-02-28
Updated NServiceBus.AmazonSQS to 5.7.1

jpalac:


mauroservienti:
We want to wait merging this until we get https://github.com/Particular/NServiceBus.AmazonSQS/pull/1983 in the transport

mauroservienti:
Don't we also need to change the `SQSTransportCustomization` class to support a new query string parameter to enable receiving unwrapped messages? (and we need to update doco for that too)

mikeminutillo:
We can always receive unwrapped messages using this version of the transport. We won't _emit_ unwrapped messages without a configuration flag. Do we need to add that?

mikeminutillo (approved):

PR #3449, opened 2023-03-03
Include NServiceBus version when listing endpoints

andreasohlund:
Just a spike to demo how to extract nsb versions of endpoints from the audit stream.

Other sources: Errors, heartbeats and custom checks

### DIFF
--- a/src/ServiceControl.Audit/Auditing/AuditPersister.cs
+++ b/src/ServiceControl.Audit/Auditing/AuditPersister.cs
@@ -182,6 +182,7 @@
                     HostId = observedEndpoint.HostId,
                     LastSeen = processedMessage.ProcessedAt,
                     Name = observedEndpoint.Name,
+                    NServiceBusVersion = observedEndpoint.NServiceBusVersion,
                     Id = KnownEndpoint.MakeDocumentId(observedEndpoint.Name, observedEndpoint.HostId),
                 };
                 observedEndpoints.Add(uniqueEndpointId, knownEndpoint);
--- a/src/ServiceControl/Monitoring/KnownEndpointIndex.cs
+++ b/src/ServiceControl/Monitoring/KnownEndpointIndex.cs
@@ -9,14 +9,14 @@ namespace ServiceControl.Audit.Monitoring
         public KnownEndpointIndex()
         {
             Map = knownEndpoints => from knownEndpoint in knownEndpoints
-                              select new
-                              {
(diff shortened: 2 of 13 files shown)


src/ServiceControl.Audit/Auditing/AuditPersister.cs
```diff
@@ -182,6 +182,7 @@ static void RecordKnownEndpoints(EndpointDetails observedEndpoint, Dictionary<st
                     HostId = observedEndpoint.HostId,
                     LastSeen = processedMessage.ProcessedAt,
                     Name = observedEndpoint.Name,
+                    NServiceBusVersion = observedEndpoint.NServiceBusVersion,
```
DavidBoike, line 185: What's the context of this code? Is `observedEndpoints` a collection of only the endpoints observed during the batch being ingested, and then updates to known endpoints are done at the same time as the batch is ingested? Because otherwise I don't understand how `LastSeen` would be updated, let alone `NServiceBusVersion`.

What happens if you are doing a phased update of an endpoint's instances, and one endpoint instance is updated to a new NServiceBus version and another is still running the old. Will the observed value in KnownEndpoints oscillate?

src/ServiceControl/Monitoring/KnownEndpointIndex.cs
```diff
@@ -8,15 +8,15 @@ public class KnownEndpointIndex : AbstractIndexCreationTask<KnownEndpoint>
...
-            Map = messages => from message in messages
-                              select new
-                              {
-                                  EndpointDetails_Name = message.EndpointDetails.Name,
-                                  EndpointDetails_Host = message.EndpointDetails.Host,
-                                  message.HostDisplayName,
-                                  message.Monitored,
-                                  message.HasTemporaryId
-                              };
+            Map = knownEndpoints => from knownEndpoint in knownEndpoints
```
DavidBoike, line 11: So to be clear here, you're changing the variable name because this index is on known endpoints (so this index touches a few hundred documents, max) not on processed messages, so it really never should have said `message` in the first place. Correct?

andreasohlund: Yea

src/ServiceControl.Audit/Auditing/AuditPersister.cs
andreasohlund: Yea that code is funky, need to understand this better if we ever go ahead with this

PR #3502, opened 2023-04-24
Hostname fields no longer accept IP addresses

awright18:
Fixes 
- #3503

### DIFF
--- a/src/ServiceControl.Config.Tests/Validation/AddAuditInstanceValidationTests.cs
+++ b/src/ServiceControl.Config.Tests/Validation/AddAuditInstanceValidationTests.cs
@@ -1,11 +1,11 @@
 ﻿namespace ServiceControl.Config.Tests.Validation
 {
     using System.ComponentModel;
-    using NUnit.Framework;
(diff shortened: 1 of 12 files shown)


cquirosj:
Great to have test coverage now for the addition of up addresses!
Do we have tests for the old behavior too? Like hostnames, http://+  and http://* ? I am not sure but I think those are also supported behaviors.

soujay (approved):

mauroservienti:
@awright18 in #3528, we fixed a build-related issue causing the Azure tests to fail. If you rebase, it should be green.

awright18:
@mauroservienti I rebased, and I'm now seeing tests failing for manifest-related issues. Would you mind taking a look at the [CI results](https://github.com/Particular/ServiceControl/actions/runs/5069878011)

mauroservienti:
It's a race condition in the build artifacts we thought we had fixed; it looks like it's not the case. We'll further investigate.

mauroservienti:
#3542 is merged. You can rebase, it should work now 🤞😆

PR #3497, opened 2023-04-18
Fix audit queue not being provisioned by the audit instance

andreasohlund:
Fixes #3487

PR #3507, opened 2023-04-28
Fix audit queue not being provisioned by the audit instance (#3497)

andreasohlund:
Backport to 4.30.1


andreasohlund:
@WilliamBZA rebased, should be green shortly

WilliamBZA (approved):

PR #3541, opened 2023-05-25
Transport and Persistence Manifest Cleanup

jpalac:

PR #3542, opened 2023-05-25
Streamline artifacts dependencies by coping transports and persisters binaries from the artifacts folder and not from the various output folders

mauroservienti:
Replaces https://github.com/Particular/ServiceControl/pull/3541


jpalac:
Should we be able to then remove the project references to the persisters/transports not used directly? So remove all but InMemory and Learning?

mauroservienti:
In theory we can try to go back to the include.props files ne test locally what happens. The only thing we need to make sure is that transports and persisters are built before the test projects.

I suggest we merge this one, also to unblock other PRs going to master. Then we can refine even more.

jpalac (approved):

PR #3546, opened 2023-05-26
Remove stale endpoint instances each hour

WilliamBZA:
Adds a setting to remove stale endpoint instances after they have been stale for the specified time.

Will need a docs PR too


soujay (approved):

WilliamBZA:
Closing for now. Spike showed that it was possible, this can be used as a starting point for this work in the future.

WilliamBZA:
It might also be worthwhile having the frequency at which the cleanup runs also be a setting so that the cleanup can be configured to run daily instead of hourly, or some other setting that makes sense.

PR #3552, opened 2023-05-31
Include learning transport in SCMU

jpalac:
Include the LearningTransport in SCMU if Environment variable ServiceControl_IncludeLearningTransport is true

### DIFF
--- a/src/ServiceControlInstaller.Engine/Instances/TransportNames.cs
+++ b/src/ServiceControlInstaller.Engine/Instances/TransportNames.cs
@@ -37,6 +37,7 @@ namespace ServiceControlInstaller.Engine.Instances
         public const string RabbitMQClassicDirectRoutingTopology = "RabbitMQ - Direct routing topology (classic queues)";
 
         public const string RabbitMQQuorumDirectRoutingTopology = "RabbitMQ - Direct routing topology (quorum queues)";
+        public const string LearningTransport = "Learning Transport";
 
         public static bool IsDeprecated(string transportName)
         {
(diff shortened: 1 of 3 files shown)


src/ServiceControlInstaller.Engine/Instances/TransportNames.cs
```diff
@@ -37,6 +37,7 @@ public static class TransportNames
         public const string RabbitMQClassicDirectRoutingTopology = "RabbitMQ - Direct routing topology (classic queues)";
 
         public const string RabbitMQQuorumDirectRoutingTopology = "RabbitMQ - Direct routing topology (quorum queues)";
+        public const string LearningTransport = "Learning Transport";
```
mikeminutillo, line 40: None of the others include transport in the name

```suggestion
        public const string LearningTransport = "Learning";
```

mikeminutillo (approved):

mauroservienti (approved):

mauroservienti:
Do we want to document this in the README?

jpalac:
> Do we want to document this in the README?

Yes but I'd like to test this first - would actually like to see this in the installer. I have no way of creating the installer locally, so the only way might be to push an alpha on this branch?...

mauroservienti:
I think you can also just F5 the SCMU project, it should work. Not sure it can install instances, thought.

jpalac:
Ok I've verified that it loads the option into the dropdown if the env variable is present. It also created the instance with the "queues" being created in the file system - guess it's ready for smoke testing

PR #3554, opened 2023-05-31
Use attributes to ignore tests instead of filtering by category

mauroservienti:
NOTE: Before merging make sure tests run as expected!

This PR replaces the NUnit test categories with a set of custom attributes that use the `ServiceControl_TESTS_FILTER` environment variable to filter which tests to run. Fixes:

- #3471 

The downside of this approach is that ServiceControl has a set of tests that are transport agnostic (identified in the CI matrix as `Default`). To make sure those tests are not run for every transport permutation the filter attribute must be defined in every test project. Failing to do so results in the tests being executed for every transport, nothing concerning other than the build will take more time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,14 +95,14 @@ jobs:
           # Cleanup of queues starting with `GHA-` handled by https://github.com/Particular/NServiceBus.AmazonSQS/blob/master/.github/workflows/tests-cleanup.yml
           $connectString = "AccessKeyId=${{ secrets.AWS_ACCESS_KEY_ID }};SecretAccessKey=${{ secrets.AWS_SECRET_ACCESS_KEY }};Region=${{ secrets.AWS_REGION }};QueueNamePrefix=GHA-${{ github.run_id }}"
           echo "ServiceControl.TransportTests.SQS.ConnectionString=$connectString" | Out-File -FilePath $Env:GITHUB_ENV -Encoding utf-8 -Append
-      - name: Set test category filter
+      - name: Set test filter
         shell: pwsh
         run: |
           if('${{ matrix.test-category }}' -eq 'Default'){
-            echo "TEST_FILTER=TestCategory!~Transports." | Out-File -FilePath $Env:GITHUB_ENV -Encoding utf-8 -Append
+            echo "TEST_FILTER=${{ matrix.test-category }}" | Out-File -FilePath $Env:GITHUB_ENV -Encoding utf-8 -Append
           }
           else{
-            echo "TEST_FILTER=TestCategory=Transports.${{ matrix.test-category }}" | Out-File -FilePath $Env:GITHUB_ENV -Encoding utf-8 -Append
+            echo "TEST_FILTER=Transports.${{ matrix.test-category }}" | Out-File -FilePath $Env:GITHUB_ENV -Encoding utf-8 -Append
           }
       - name: Run tests
         uses: Particular/run-tests-action@v1.5.0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,7 +98,7 @@ jobs:
       - name: Run tests
         uses: Particular/run-tests-action@v1.5.0
         env:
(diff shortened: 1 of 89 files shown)


mauroservienti:
It's clear that I don't understand how it should work. I might be missing something obvious 😖

.github/workflows/ci.yml
```diff
@@ -95,19 +95,17 @@ jobs:
...
           else{
-            echo "TEST_FILTER=TestCategory=Transports.${{ matrix.test-category }}" | Out-File -FilePath $Env:GITHUB_ENV -Encoding utf-8 -Append
+            echo "TEST_FILTER=Transports.${{ matrix.test-category }}" | Out-File -FilePath $Env:GITHUB_ENV -Encoding utf-8 -Append
           }
       - name: Run tests
         uses: Particular/run-tests-action@v1.5.0
-        with:
-          filter: ${{ env.TEST_FILTER }}
         env:
           ServiceControl/LicenseText: ${{ secrets.LICENSETEXT }}
```
DavidBoike, line 110: Just put your matrix value in here and remove the entire "Set test filter" step above. The whole if is trying to be fancy by having every test that's NOT categorized be a part of the default.

And I would drop the `Transports.` part. Just use what's in the matrix.
```suggestion
          TEST_FILTER: ${{ matrix.test-category }}
          ServiceControl/LicenseText: ${{ secrets.LICENSETEXT }}
```

src/TestHelper/IncludeInTestsAttribute.cs
```diff
@@ -0,0 +1,26 @@
...
+using NUnit.Framework;
+using NUnit.Framework.Interfaces;
+using NUnit.Framework.Internal;
+
+[AttributeUsage(AttributeTargets.Method | AttributeTargets.Class | AttributeTargets.Assembly)]
+public abstract class IncludeInTestsAttribute : Attribute, IApplyToContext
+{
+    public void ApplyToContext(TestExecutionContext context)
+    {
+        var currentFilters = Environment.GetEnvironmentVariable("TEST_FILTER")?.Split(new char[] { ';' }, StringSplitOptions.RemoveEmptyEntries);
```
DavidBoike, line 12: Why even try to split? YAGNI?

mauroservienti: No, because of the need to run tests locally. The difference with SqlP is that SqlP has no "Default" tests, they are all engine bound. In SC while developing locally what if you wanted to run RabbitMQ and Default tests? It's true that one could set the TEST_FILTER to All.

src/ServiceControl.Audit.AcceptanceTests.RavenDB/ServiceControl.Audit.AcceptanceTests.RavenDB.csproj
```diff
@@ -23,6 +24,7 @@
 
   <ItemGroup>
     <Compile Include="..\ServiceControl.Audit.AcceptanceTests\**\*.cs" Exclude="..\ServiceControl.Audit.AcceptanceTests\AcceptanceTestStorageConfiguration.cs;..\ServiceControl.Audit.AcceptanceTests\obj\**\*.*;..\ServiceControl.Audit.AcceptanceTests\bin\**\*.*" />
+    <Compile Remove="..\ServiceControl.Audit.AcceptanceTests\TestsFilter.cs" />
```
DavidBoike, line 27: Seems like some rearranging could make this unnecessary but :shrug:

DavidBoike (approved):

src/ServiceControl.Audit.AcceptanceTests.RavenDB/ServiceControl.Audit.AcceptanceTests.RavenDB.csproj
mauroservienti: Yeah, I could put all those in the Exclude of the previous line, I feel it's more explicit/visible this way.

mikeminutillo (approved):

PR #3572, opened 2023-06-06
Remove az bug workaround

andreasohlund:
MS claims to have released a fix so we should be able to remove


mauroservienti:
The OS base image has probably not been updated yet to include the fix

andreasohlund:
>The OS base image has probably not been updated yet to include the fix

Yea, the windows image was updated last week but perhaps didn't include it
https://github.com/actions/runner-images/releases/tag/win22%2F20230517.1

Made the PR a draft

andreasohlund:
@mauroservienti green!

PR #3577, opened 2023-06-07
If RabbitMQ input queue not found for audit ingestion the startup fails

tmasternak:
This is a first take on making the startup fail-fast when `audit` queue does not exist in the broker. Although these might be valid solutions from a technical perspective I have some doubts if this is a valid approach from the UX perspective.

Currently, the service is going to continue working no matter what kind of errors are happening and we assume that at some point things are "going to get better". The instance is never stopped automatically but only when explicitly requested by the user. With this change, we would introduce one error type that requires a different behavior, only during startup.


jpalac:
Closing in favour of https://github.com/Particular/ServiceControl/pull/3589

Issue #3338, opened 2022-12-21
Release workflow is much slower after changes to how Docker images are built - can now be collapsed to just 3 images


bording:
The changes introduced in #3241 to how docker images are built has had some negative side effects to the time it takes to release ServiceControl, making it much slower.

Previously, the release workflow built the docker images in parallel via a matrix configuration. That has been removed with the move to using a project to build the images instead.

We should try and find a way to keep the benefits of the project-based way of building, but also re-introduce the parallel building that was lost.

WilliamBZA:
This has been done as part of the ServiceControl on Linux work, closing.

PR #3584, opened 2023-06-08
Run non-deploy portions of release workflow when CI workflow runs

bording:
This PR makes the following changes:
- The release workflow now has all the same triggers as the CI workflow, meaning it will run whenever the CI workflow runs.
  - The steps that actually deploy the release artifacts are skipped unless the workflow is triggered from a tag. This means that nothing changes in our release process because of this PR.
  - The release workflow is skipped when Dependabot opens a PR because it would otherwise fail due to the code signing secrets not being available to Dependabot.
- A step to verify the expected number of release artifacts has been added. This means that if something causes the number of artifacts to change, the release workflow will fail. The counts need to be manually adjusted if there is an intentional change to the expected number of artifacts for a release.
- Building the docker images is because [they currently take way too long](https://github.com/Particular/ServiceControl/issues/3338) to include as part of every PR.

The end result of this PR is that you'll see the release workflow running on every PR, making it easy to access release artifacts, like the Windows installer, for testing. This also means that it is easier to prevent the release workflow from being broken accidentally.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,6 +15,11 @@ jobs:
   release:
     runs-on: windows-2022
     steps:
+      - name: Check for secrets
+        env:
+          SECRETS_AVAILABLE: ${{ secrets.SECRETS_AVAILABLE }}
+        shell: pwsh
+        run: exit $(If ($env:SECRETS_AVAILABLE -eq 'true') { 0 } Else { 1 })
       - name: Checkout
         uses: actions/checkout@v3.5.2
         with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,9 +1,14 @@
 name: Release
 on:
   push:
+    branches:
+      - master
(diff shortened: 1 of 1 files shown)


DavidBoike:
Still a TON of time building Docker images…as I'm reviewing, 51m and still counting.

What about:

1. Don't build the docker images or list them out. Reverse that once we coalesce down to 3 images, but at least an installer asset will be available quickly.
2. Only build representative images for one transpot.

bording:
> Still a TON of time building Docker images…as I'm reviewing, 51m and still counting.

> Don't build the docker images or list them out. Reverse that once we coalesce down to 3 images, but at least an installer asset will be available quickly.

Yeah way too painful to have them being built on every PR. I think we'll just have to disable that for now until we are able to reduce the number of images being required.

DavidBoike (approved):

DavidBoike (approved):

Issue #3522, opened 2023-05-16
RavenDB Server process not restarted


ramonsmits:
#### Description

When the RavenDB5 server process is gone. Either because it crashed or was killed SC isn't detecting this state.

#### Expected behavior

SC having a recovery mechanism when the RavenDB process dies.

#### Actual behavior

Nothing happens

#### Versions

4.30.0

### Steps to reproduce

1. Start audit instance with RavenDB 5 engine
2. Kill the sub process that hosts the RavenDB 5 server
3. Observe the logs and notice nothing is logged
4. Observe the RavenDB 5 server process isn't restarted

### Relevant log output

A log file from a customer shows the log flooded with the following error:

```txt
System.Net.WebException: Unable to connect to the remote server ---> System.Net.Sockets.SocketException: No connection could be made because the target machine actively refused it 127.0.0.1:44445
   at System.Net.Sockets.Socket.InternalEndConnect(IAsyncResult asyncResult)
   at System.Net.Sockets.Socket.EndConnect(IAsyncResult asyncResult)
   at System.Net.ServicePoint.ConnectSocketInternal(Boolean connectFailure, Socket s4, Socket s6, Socket& socket, IPAddress& address, ConnectSocketState state, IAsyncResult asyncResult, Exception& exception)
   --- End of inner exception stack trace ---
   at System.Net.HttpWebRequest.EndGetRequestStream(IAsyncResult asyncResult, TransportContext& context)
   at System.Net.Http.HttpClientHandler.GetRequestStreamCallback(IAsyncResult ar)
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at Raven.Client.Http.RequestExecutor.<SendRequestToServer>d__116`1.MoveNext() in C:\Builds\RavenDB-Stable-5.4\54038\src\Raven.Client\Http\RequestExecutor.cs:line 1039
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Raven.Client.Http.RequestExecutor.<ExecuteAsync>d__114`1.MoveNext() in C:\Builds\RavenDB-Stable-5.4\54038\src\Raven.Client\Http\RequestExecutor.cs:line 866
[...]

PR #3586, opened 2023-06-09
Restart RavenDB Server process

jpalac:
Attempts to resolve https://github.com/Particular/ServiceControl/issues/3522

Restarting the RavenDB5 server process if it stops.

To discuss: should this be something that we try a few times in a loop and then exit, or should we not bother restarting at all and exit straight away?


jpalac:
Closed in favour or https://github.com/Particular/ServiceControl/pull/3589

PR #3598, opened 2023-06-14
Improve  warning message when trying to edit-retry  an already edited message

soujay:
- Original Issue: https://github.com/Particular/ServicePulse/issues/1526
- Internal Issue: https://github.com/Particular/PlatformBugs/issues/1224

This PR improves warning message that is seen in the logs when trying to edit-retry  an already edited message

### DIFF
--- a/src/ServiceControl/Recoverability/Editing/EditHandler.cs
+++ b/src/ServiceControl/Recoverability/Editing/EditHandler.cs
@@ -58,7 +58,7 @@
                 }
                 else if (edit.EditId != context.MessageId)
                 {
-                    log.WarnFormat("Discarding edit {0} because the failure ({1}) has already been edited by edit {2}", context.MessageId, message.FailedMessageId, edit.EditId);
+                    log.WarnFormat("Discarding edit & retry request because the failure (FailedMessages/{1}) has already been edited edited in FailedMessageEdit/{2}", message.FailedMessageId, edit.EditId);
                     return;
                 }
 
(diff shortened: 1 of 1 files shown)


src/ServiceControl/Recoverability/Editing/EditHandler.cs
```diff
@@ -58,7 +58,7 @@ public async Task Handle(EditAndSend message, IMessageHandlerContext context)
                 }
                 else if (edit.EditId != context.MessageId)
                 {
-                    log.WarnFormat("Discarding edit {0} because the failure ({1}) has already been edited by edit {2}", context.MessageId, message.FailedMessageId, edit.EditId);
+                    log.WarnFormat("Discarding edit & retry request because the failure (FailedMessages/{1}) has already been edited edited in FailedMessageEdit/{2}", message.FailedMessageId, edit.EditId);
```
mauroservienti, line 61: We can use the builtin mechanism to generate the ID, so if the strategy changes, we're not relying on strings here.

```suggestion
                    log.WarnFormat("Discarding edit & retry request because the failure ({1}) has already been edited by {2}", FailedMessage.MakeDocumentId(message.FailedMessageId), FailedMessageEdit.MakeDocumentId(message.FailedMessageId));
```

mauroservienti: An even better solution would be to use string interpolation

```suggestion
                    log.WarnFormat($"Discarding edit & retry request because the failure ({FailedMessage.MakeDocumentId(message.FailedMessageId)}) has already been edited by {FailedMessageEdit.MakeDocumentId(message.FailedMessageId)}");
```

Not sure ti compiles, I'm adding suggestions in the browser

soujay:
@mauroservienti - Applied your suggestions

mauroservienti (approved):

Issue #2021, opened 2020-05-21
Monitoring instance "Failed to write startup diagnostics"  


ramonsmits:
Was fixed for SC error and audit instances in https://github.com/Particular/ServiceControl/pull/1979 

```txt
-------------------------------------------------------------
ServiceControl Monitoring Version:				4.10.0+0.Branch.release-4.10.Sha.ff59a3f2eca37d4b93ae10a629f43e5c6fd01a78
Selected Transport:					ServiceControl.Transports.SqlServer.SqlServerTransportCustomization, ServiceControl.Transports.SqlServer
-------------------------------------------------------------
2020-05-21 01:45:04.4778|4|Warn|NServiceBus.Transport.SqlServer.SqlConnectionFactory|Maximum connection pooling value (Max Pool Size=N) is not configured on the provided connection string. The default value (100) will be used.
2020-05-21 01:45:04.5948|5|Error|NServiceBus.WriteStartupDiagnostics|Failed to write startup diagnostics
System.UnauthorizedAccessException: Access to the path 'C:\Program Files (x86)\Particular Software\Particular.Sql.Monitoring\.diagnostics\Particular.Sql.Monitoring-configuration.txt' is denied.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options, String msgPath, Boolean bFromProxy)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, Boolean useAsync)
   at NServiceBus.AsyncFile.<WriteBytes>d__0.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at NServiceBus.WriteStartupDiagnostics.<OnStart>d__1.MoveNext()`
```

kbaley:
[...]

PR #3606, opened 2023-06-20
add diagnostics file when new monitoring instance is created

soujay:
This PR is a fix for [Monitoring instance "Failed to write startup diagnostics"   #2021](https://github.com/Particular/ServiceControl/issues/2021)

### DIFF
--- a/src/ServiceControl.Monitoring/Bootstrapper.cs
+++ b/src/ServiceControl.Monitoring/Bootstrapper.cs
@@ -1,6 +1,7 @@
 ﻿namespace ServiceControl.Monitoring
 {
     using System;
+    using System.Configuration;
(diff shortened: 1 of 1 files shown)


src/ServiceControl.Monitoring/Bootstrapper.cs
```diff
@@ -1,6 +1,7 @@
 ﻿namespace ServiceControl.Monitoring
 {
     using System;
+    using System.Configuration;
```
WilliamBZA, line 4: Why is this needed?

soujay: removed test code

WilliamBZA (approved):

Issue #1971, opened 2020-04-21
Bug: Trailing spaces in "Name" can result in hang or otherwise result in trailing period characters in name / window service name for error and audit instances name


ramonsmits:
Trailing spaces in "Name" field can result in a hang of the SCMU or results in trailing period characters in name / window service name for error and audit instances name.

Any data entered in the name field should be trimmed of trailing whitespace or have the trailing whitespace excluded.

mikeminutillo:
This issue has been released in 4.32.1. Details in [the announcement](https://discuss.particular.net/t/servicecontrol-4-32-1-patch-release-available/3640).

PR #3617, opened 2023-06-29
Remove trailing spaces from instance names

soujay:
PR to fix [Trailing spaces in "Name" can result in hang or otherwise result in trailing period characters in name / window service name for error and audit instances name #1971](https://github.com/Particular/ServiceControl/issues/1971)


awright18 (approved):
This is a good change!

Issue #3557, opened 2023-06-01
If the transport label in SCMU is too long it's not visible


mauroservienti:
#### Description

When selecting a transport like in the below screenshot, it might happen that the transport label gets truncated because too long

<img width="539" alt="Screenshot 2023-06-01 at 11 32 48" src="https://github.com/Particular/ServiceControl/assets/1325611/3116556b-330a-4d74-ad32-6dae69814a8d">

#### Expected behavior

View the full selected transport configuration name

### Steps to reproduce

Install an instance and select a transport with a long name/config

mikeminutillo:
This issue has been released in 4.32.1. Details in [the announcement](https://discuss.particular.net/t/servicecontrol-4-32-1-patch-release-available/3640).

PR #3618, opened 2023-06-29
word wrap long transport name

soujay:
This PR is to fix [If the transport label in SCMU is too long it's not visible #3557](https://github.com/Particular/ServiceControl/issues/3557)


awright18 (approved):
This looks good

PR #3648, opened 2023-08-01
SPIKE: Run MSMQ on netstandard2.0 with reference sources

DavidBoike:


DavidBoike:
Spike served its purpose, closing.

PR #3661, opened 2023-08-14
Fix flooding the logs with "Error parsing message type:" for metrics from polymorphic messages 

ramonsmits:
The received value is the value from NServiceBus.EnclosedMessageTypes which can contain multiple types.

This should resolve the issue that `MonitoredEndpointMessageTypeParser.Parse` receives the value with all types and results in the following log entry:

```txt
2023-08-14 10:19:48.5594|8|Warn|ServiceControl.Monitoring.Http.Diagrams.MonitoredEndpointMessageTypeParser|Error parsing message type: XXXX, YYY, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null;ZZZZ, YYY, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null.
System.IO.FileLoadException: The given assembly name or codebase was invalid. (Exception from HRESULT: 0x80131047)
   at System.Reflection.AssemblyName.nInit(RuntimeAssembly& assembly, Boolean forIntrospection, Boolean raiseResolveEvent)
   at System.Reflection.AssemblyName..ctor(String assemblyName)
   at ServiceControl.Monitoring.Http.Diagrams.MonitoredEndpointMessageTypeParser.Parse(String typeName) in /_/src/ServiceControl.Monitoring/Http/Diagrams/MonitoredEndpointMessageTypeParser.cs:line 31
```

Not sure if the behavior should only add the first (top) type or all. This is only taking the first type

### DIFF
--- a/src/ServiceControl.Monitoring.AcceptanceTests/When_querying_queue_length_data.cs
+++ b/src/ServiceControl.Monitoring.AcceptanceTests/When_querying_queue_length_data.cs
@@ -125,7 +125,11 @@
             }
         }
 
-        class SampleMessage : IMessage
+        class SampleMessage : SampleBaseMessage
+        {
+        }
+
+        class SampleBaseMessage : IMessage
         {
         }
 
--- a/src/ServiceControl.Monitoring.AcceptanceTests/When_querying_retries_data.cs
+++ b/src/ServiceControl.Monitoring.AcceptanceTests/When_querying_retries_data.cs
@@ -77,7 +77,11 @@
             public bool ShuttingDown { get; set; }
         }
 
-        class SampleMessage : IMessage
+        class SampleMessage : SampleBaseMessage
+        {
+        }
+
+        class SampleBaseMessage : IMessage
         {
         }
     }
--- a/src/ServiceControl.Monitoring.AcceptanceTests/When_querying_timings_data.cs
+++ b/src/ServiceControl.Monitoring.AcceptanceTests/When_querying_timings_data.cs
@@ -63,7 +63,11 @@
         {
         }
 
-        class SampleMessage : IMessage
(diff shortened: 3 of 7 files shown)


danielmarbach:
I apologize for my inputs coming in with such big delays. I have a few absences this week. These are wild thoughts. I'm fairly new to this code.

The original stack trace showed the problem coming from `MonitoredEndpointMessageTypeParser` and now we have essentially multiple places where we have to incorporate the splitting logic. The parser itself already has to deal with endpoint message types and the controller can already deal with multiple message types

https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl.Monitoring/Http/Diagrams/DiagramApiController.cs#L254-L258

I wonder if it would be better to consolidate things a bit to centralize the parsing and creation of `EndpointMessageType` and make the `MonitoredEndpointMessageTypeParser` rely on that type too. This would essentially make sure polymorphic message types are visualized as multiple message types. Polymorphic types are always separated by `;` see https://github.com/Particular/NServiceBus/blob/f9a03dc930c20dc80b18034f824be51fc2adbacd/src/NServiceBus.Core/Unicast/Messages/MessageMetadata.cs#L62. The parser could take that into account. Core also seems to treat those polymorphic ones as individual ones in multiple places

https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Pipeline/Incoming/DeserializeMessageConnector.cs#L72-L97
https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Pipeline/Incoming/DeserializeMessageConnector.cs#L120-L126
https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Routing/UnicastPublishRouter.cs#L28-L30

danielmarbach (approved):

PR #3663, opened 2023-08-14
Fix ASBS QueueLengthProvider to process duplicate QueueRuntimeProperties

ramonsmits:
Based on a support cases
It seems `GetQueuesRuntimePropertiesAsync` can return the same items multiple times which will cause a `System.ArgumentException: An item with the same key has already been added.` as it uses `.Add` instead of `[]`

### DIFF
--- a/src/ServiceControl.Transports.ASBS/QueueLengthProvider.cs
+++ b/src/ServiceControl.Transports.ASBS/QueueLengthProvider.cs
@@ -85,7 +85,7 @@ namespace ServiceControl.Transports.ASBS
                 while (await enumerator.MoveNextAsync().ConfigureAwait(false))
                 {
                     var queueRuntimeProperties = enumerator.Current;
-                    queuePathToRuntimeInfo.Add(queueRuntimeProperties.Name, queueRuntimeProperties);
+                    queuePathToRuntimeInfo[queueRuntimeProperties.Name] = queueRuntimeProperties;
                 }
             }
             finally
--- a/src/ServiceControl.Transports.ASBS/QueueLengthProvider.cs
+++ b/src/ServiceControl.Transports.ASBS/QueueLengthProvider.cs
@@ -85,6 +85,14 @@ namespace ServiceControl.Transports.ASBS
                 while (await enumerator.MoveNextAsync().ConfigureAwait(false))
                 {
                     var queueRuntimeProperties = enumerator.Current;
+                    if (queuePathToRuntimeInfo.ContainsKey(queueRuntimeProperties.Name))
+                    {
+                        var existingItem = queuePathToRuntimeInfo[queueRuntimeProperties.Name];
+                        if(existingItem.Count!=queueRuntimeProperties.Count)
+                        {
+                            Logger.WarnFormat("Queue <{0}> already processed with different length (length {1} vs {2})", queueRuntimeProperties.Name, queuePathToRuntimeInfo.Count, existingItem.Count);
+                        }
+                    }
                     queuePathToRuntimeInfo[queueRuntimeProperties.Name] = queueRuntimeProperties;
                 }
             }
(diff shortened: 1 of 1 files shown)


src/ServiceControl.Transports.ASBS/QueueLengthProvider.cs
```diff
@@ -85,7 +85,7 @@ public Task Start()
                 while (await enumerator.MoveNextAsync().ConfigureAwait(false))
                 {
                     var queueRuntimeProperties = enumerator.Current;
```
TravisNickels, line 87:
```suggestion
                    var queueRuntimeProperties = enumerator.Current;
                    if (queuePathToRuntimeInfo.ContainsKey(queueRuntimeProperties.Name))
                    {
                        Logger.InfoFormat("There was an unexpected duplicate queue <{0}> that was retrieved from the Azure Service Bus namespace. The existing properties are going to be updated to the new values.", queueRuntimeProperties.Name);
                        Logger.InfoFormat("{0} queues have currently been retrieved.", queuePathToRuntimeInfo.Count);
                    }
```

Is this something like what you were thinking or is this too wordy?

ramonsmits: @TravisNickels Yeah, I like it but thinking about it I think we could combine it with a length check and only report is there is a difference and log it as warning:
```suggestion
                    var queueRuntimeProperties = enumerator.Current;
                    if (queuePathToRuntimeInfo.ContainsKey(queueRuntimeProperties.Name))
                    {
                    	var existingItem = queuePathToRuntimeInfo[queueRuntimeProperties.Name];
                    	if(existingItem.Count!=queueRuntimeProperties.Count)
                    	{
                            Logger.WarnFormat("Queue <{0}> already processed with different length (length {1} vs {2})", queueRuntimeProperties.Name, queuePathToRuntimeInfo.Count, existingItem.Count);
                        }
                    }
```

TravisNickels (approved):

src/ServiceControl.Transports.ASBS/QueueLengthProvider.cs
```diff
@@ -85,7 +85,15 @@ public Task Start()
...
-                    queuePathToRuntimeInfo.Add(queueRuntimeProperties.Name, queueRuntimeProperties);
+                    if (queuePathToRuntimeInfo.ContainsKey(queueRuntimeProperties.Name))
+                    {
+                        var existingItem = queuePathToRuntimeInfo[queueRuntimeProperties.Name];
+                        if(existingItem.Count!=queueRuntimeProperties.Count)
+                        {
+                            Logger.WarnFormat("Queue <{0}> already processed with different length (length {1} vs {2})", queueRuntimeProperties.Name, queuePathToRuntimeInfo.Count, existingItem.Count);
+                        }
+                    }
+                    queuePathToRuntimeInfo[queueRuntimeProperties.Name] = queueRuntimeProperties;
```
ramonsmits, line 96:
```suggestion
                    queuePathToRuntimeInfo[queueRuntimeProperties.Name] = queueRuntimeProperties; // Assuming last write is most up to date
```

TravisNickels:
I accidentally pushed changes to the wrong branch, I did a force push to drop the changes I made in this PR.

src/ServiceControl.Transports.ASBS/QueueLengthProvider.cs
```diff
@@ -85,7 +85,20 @@ public Task Start()
...
+                        {
+                            Logger.WarnFormat(
+                                "Queue <{0}> already processed with different length (length {1} vs {2})",
+                                queueRuntimeProperties.Name,
+                                queuePathToRuntimeInfo.ActiveMessageCount,
+                                existingItem.ActiveMessageCount
+                                );
+                        }
+                    }
+                    queuePathToRuntimeInfo[queueRuntimeProperties.Name] = queueRuntimeProperties; // Assuming last write is most up to date
```
TravisNickels, line 101:
```suggestion
                    if (queuePathToRuntimeInfo.ContainsKey(queueRuntimeProperties.Name))
                    {
                        Logger.WarnFormat("Queue <{0}> already processed.  Updating queue runtime properties.", queueRuntimeProperties.Name);
                    }
                    queuePathToRuntimeInfo[queueRuntimeProperties.Name] = queueRuntimeProperties; // Assuming last write is most up to date
```

So if we want to keep it simple, maybe something like this?

PR #3658, opened 2023-08-11
Finish persistence seam for primary instance

DavidBoike:

PR #3664, opened 2023-08-14
Replace CLR approval test with HTTP API approval test

DavidBoike:
Given the PublicClr test is ostensibly to prevent breaking the interface with ServicePulse and ServiceInsight, it seems an actual readout of only the HTTP routes and the methods they map to is a much better indication of that.

Once merged, I intend to rebase https://github.com/Particular/ServiceControl/pull/3658 on top of it, as the current diff in the PublicClr API is far too large to get anything useful from.


ramonsmits (approved):
@DavidBoike is it actually important to know to what the route is mapped to? I see results are ordered 👍🏼 and its all pretty self explanatory

DavidBoike:
I think it's important to be able to eyeball the controller that maps to that the route that has `{id}` in it maps to a method that has `Guid id` or `string id` in it as a parameter. And if you end up renaming that's fine as that's a simple 1-line diff to deal with.

danielmarbach:
I was off yesterday and didn't have time to look into this before it got merged. I think this is a good improvement to the previous solution. One thing that I still don't entirely understand, why we are not using something like [Pact](https://github.com/pact-foundation/pact-net) to verify the contract.

DavidBoike:
Mostly because something like that would be well outside the scope of our (already overloaded) task force.

ramonsmits:
@danielmarbach interesting, I didn't know that project. Do you know if a platform internals observation/opportunity exists for it? I searched for it but not get any results.

PR #3667, opened 2023-08-18
POC: Smoke-testing tool

DavidBoike:
Not intending to merge. I was convinced this should live in its own repository so that it can be developed independently and be used for smoke-testing future releases of ServiceControl as well.


DavidBoike:
Moving to its own repository

PR #3670, opened 2023-08-23
Testing scenarios

tmasternak:


DavidBoike:
@tmasternak after our chat today were you thinking this would get updated in our release checklist, and not in the repo?

Just skimming this though, I'm not sure if it's prescriptive enough…like it says what to test but not how.

DavidBoike (approved):

PR #3680, opened 2023-08-28
Install RavenDB35 instance by default

DavidBoike:
Validated the changes work both to create a new primary instance, as well as updating a 4.32.0 primary instance to the latest version.


tmasternak (approved):
LGTM

PR #3691, opened 2023-09-04
Resolve docker build issues

ramonsmits:
- Optimized building of docker images by significantly reducing the IO to the docker daemon when running docker build and reduced the number of layers by combining ADD and ENV commands
- Bumped the base image to mcr.microsoft.com/dotnet/framework/runtime:4.8-windowsservercore-ltsc2022 this image is "only" 4.3GB, further reducing build IO and container runtime size
- Added docker sample to test the servicecontrol containers with rabbitmq
- Artifacts to push to docker daemon and the generated docker files are now written in the `ServiceControl.DockerImages` project output folder


DavidBoike (commented):
I'm going to need a walkthrough of these changes but something does have to change with the gitignore path.

.gitignore
```diff
@@ -100,3 +100,5 @@ src/scaffolding.config
 
 # Visual Studio Code
 .vscode
+
+dockerfiles/
```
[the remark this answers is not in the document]
ramonsmits: Resolved via f1537e0

tmasternak (approved):

src/ServiceControl.DockerImages/ServiceControl.DockerImages.csproj
```diff
@@ -94,16 +94,28 @@
...
+    <Exec Command="docker build -f $(DockerfilesFolder)servicecontrol.%(SupportedTransport.DockerfileCustomization).monitoring.init-windows.dockerfile -t particular/servicecontrol.%(SupportedTransport.DockerfileCustomization).monitoring.init-windows:$(MinVerVersion) --build-arg TRANSPORT=%(SupportedTransport.TransportName) --build-arg TRANSPORT_CUSTOMIZATION_TYPE=%(SupportedTransport.TransportCustomizationType) ./../../deploy/" />
+    <Exec Command="docker build -f $(DockerfilesFolder)servicecontrol.%(SupportedTransport.DockerfileCustomization).monitoring-windows.dockerfile -t particular/servicecontrol.%(SupportedTransport.DockerfileCustomization).monitoring-windows:$(MinVerVersion) --build-arg TRANSPORT=%(SupportedTransport.TransportName) --build-arg TRANSPORT_CUSTOMIZATION_TYPE=%(SupportedTransport.TransportCustomizationType) ./../../deploy/" />
+
+    <!-- Tag for local use -->
```
DavidBoike, line 108:
```suggestion
    <!-- Tag as "dev" for easier local testing of containers -->
```

src/ServiceControl.DockerImages/ServiceControl.DockerImages.csproj
```diff
@@ -78,7 +74,13 @@
     <MakeDir Directories="$(DockerfilesFolder)" />
   </Target>
 
+
   <Target Name="ExpandDockerfileTemplates" AfterTargets="CleanGeneratedDockerfiles">
+
+    <PropertyGroup>
+      <DockerfilesFolder>$(OutputPath)/dockerfiles/</DockerfilesFolder>
```
bording, line 81: Why are you moving where the dockerfiles are being created? This isn't necessary for copying the deploy files and causes problems for the release workflow which needs to know where they dockerfiles are.

src/ServiceControl.DockerImages/ServiceControl.DockerImages.csproj
```diff
@@ -90,20 +92,44 @@
     <!-- monitoring instance and sidecar dockerfiles -->
     <Copy SourceFiles="servicecontrol.transport.monitoring.init-windows.dockerfile-template" DestinationFiles="$(DockerfilesFolder)servicecontrol.%(SupportedTransport.DockerfileCustomization).monitoring.init-windows.dockerfile" />
     <Copy SourceFiles="servicecontrol.transport.monitoring-windows.dockerfile-template" DestinationFiles="$(DockerfilesFolder)servicecontrol.%(SupportedTransport.DockerfileCustomization).monitoring-windows.dockerfile" />
+
+    <ItemGroup>
+      <DeployErrorFiles Include="..\..\deploy\Particular.ServiceControl\**\*.*" Exclude="**\Transports\**" />
+      <DeployAuditFiles Include="..\..\deploy\Particular.ServiceControl.Audit\**\*.*" Exclude="**\Transports\**" />
+      <DeployMonitoringFiles Include="..\..\deploy\Particular.ServiceControl.Monitoring\**\*.*" Exclude="**\Transports\**" />
+      <DeployTransportFilesInclude="..\..\deploy\Particular.ServiceControl.Monitoring\Transports\**\*.*" />
```
bording, line 100:
```suggestion
      <DeployTransportFiles Include="..\..\deploy\Particular.ServiceControl.Monitoring\Transports\**\*.*" />
```

src/ServiceControl.DockerImages/ServiceControl.DockerImages.csproj
```diff
@@ -90,20 +92,44 @@
...
+    </ItemGroup>
+
+    <Copy SourceFiles="@(DeployErrorFiles)" DestinationFolder="$(OutputPath)\docker\Particular.ServiceControl\%(RecursiveDir)" SkipUnchangedFiles="true" />
+    <Copy SourceFiles="@(DeployAuditFiles)" DestinationFolder="$(OutputPath)\docker\Particular.ServiceControl.Audit\%(RecursiveDir)" SkipUnchangedFiles="true" />
+    <Copy SourceFiles="@(DeployMonitoringFiles)" DestinationFolder="$(OutputPath)\docker\Particular.ServiceControl.Monitoring\%(RecursiveDir)" SkipUnchangedFiles="true" />
+    <Copy SourceFiles="@(DeployTransportFilesInclude)" DestinationFolder="$(OutputPath)\docker\Transports\%(RecursiveDir)" SkipUnchangedFiles="true" />
```
bording, line 106:
```suggestion
    <Copy SourceFiles="@(DeployTransportFiles)" DestinationFolder="$(OutputPath)\docker\Transports\%(RecursiveDir)" SkipUnchangedFiles="true" />
```

src/ServiceControl.DockerImages/ServiceControl.DockerImages.csproj
```diff
@@ -90,20 +92,44 @@
     <!-- monitoring instance and sidecar dockerfiles -->
     <Copy SourceFiles="servicecontrol.transport.monitoring.init-windows.dockerfile-template" DestinationFiles="$(DockerfilesFolder)servicecontrol.%(SupportedTransport.DockerfileCustomization).monitoring.init-windows.dockerfile" />
     <Copy SourceFiles="servicecontrol.transport.monitoring-windows.dockerfile-template" DestinationFiles="$(DockerfilesFolder)servicecontrol.%(SupportedTransport.DockerfileCustomization).monitoring-windows.dockerfile" />
+
+    <ItemGroup>
```
bording, line 96: Since this target is called `ExpandDockerfileTemplates`, I think it makes more sense to create another template for copying the files out of the deploy folder instead of adding it to this one.

bording (changes requested):

src/ServiceControl.DockerImages/ServiceControl.DockerImages.csproj
ramonsmits: No, this is tested and has correct behavior. I needed to move it here otherwise `$(OutputPath)` does not have a value as its resolved too early.

src/ServiceControl.DockerImages/ServiceControl.DockerImages.csproj
ramonsmits: Done in 67263e063d671bd54ac2500f53e2fb97989021d5

src/ServiceControl.DockerImages/ServiceControl.DockerImages.csproj
ramonsmits: I added a comment in 0ca696be24b9e305f7aca5213683b7e5659422a2 to explain why its there and moved it in 0a59b67a9 to the task that uses it first.

Please share if you know of an alternative. The only alternative I see is using `$(OutputPath)` in each task but that defeats the purpose of `DockerfilesFolder` a bit as each task is than doing its on path concatenation.

bording: @ramonsmits I think you missed the point of my question. We discussed copying the deploy content into the bin folder, but not the dockerfiles. 

The dockerfiles don't need to be in the same place, and still aren't with your changes. Moving where they are breaks the release workflow.

ramonsmits: @bording gotcha, well, I added a .gitignore in there which you didn't like that is why I moved that to the `bin` folder so that the .gitignore isn't needed.

Reverted that in 395cf836fcc070d48b6ad3603c2f248be0bd0c2f with a comment too that that location is used by release CI.

PR #3698, opened 2023-09-11
MessagesViewTransformer

ramonsmits:


DavidBoike:
@ramonsmits We don't need this PR anymore, right?

DavidBoike:
Oops, I meant to just comment, but closing it is probably the right move. Please reopen if not.

PR #3708, opened 2023-09-14
Lifecycle refactoring and fixing of various `IHost` dispose issues

ramonsmits:
Hosted services are dependent on the document store which needs to be initialized before the hosted services are created by the hosting extensions.

During refactoring it was noticed that the created host are not properly being disposed by both commands and tests. All created hosts should now be properly be disposed as the end of its usage.


src/ServiceControl.Persistence.RavenDb5/RavenDbInstaller.cs
```diff
@@ -13,8 +13,7 @@ public RavenDbInstaller(IPersistenceLifecycle lifecycle)
 
         public async Task Install(CancellationToken cancellationToken)
         {
-            await lifecycle.Start(cancellationToken);
-            await lifecycle.Stop(cancellationToken);
+            await lifecycle.Initialize(cancellationToken);
```
DavidBoike, line 16: Should this be disposed then?

src/ServiceControl.Persistence/IPersistenceLifecycle.cs
```diff
@@ -3,9 +3,8 @@
     using System.Threading;
     using System.Threading.Tasks;
 
-    public interface IPersistenceLifecycle // TODO: We don't need this, implementations should just implement IHostedService
+    public interface IPersistenceLifecycle
```
DavidBoike, line 6: Since all the implementations implement IDisposable, shouldn't this as well?

src/ServiceControl.Persistence.RavenDb5/RavenDbInstaller.cs
ramonsmits: No, it should not. The DI container will dispose it once it is disposed.

src/ServiceControl.Persistence/IPersistenceLifecycle.cs
ramonsmits: Implementations currently use IDisposable but could also use IAsyncDisposable. The interface should not demand one over the other.

src/ServiceControl.Persistence.RavenDb5/RavenDbInstaller.cs
ramonsmits: @DavidBoike your comment made me analyze that code again. The installer behavior is completely different. It creates its own service provider but that is not being disposed! Meaning, it is created, returns a lifecycle from that provider but that provider its lifetime isn't disposed.

Its the opposite from how regular persister is used as there a services collection is passed into the persister to register its stuff. Why isn't the same approach take for the installer? Register the installer interface in the services so that the `SetupBootstrapper` can resolve this installer via DI? This way the `CreateInstaller` isn't needed. 

We should either do our own lifetime control or not but why not do it for "runtime" but do it ourselves for "install/setup"?

Made a fix in 6c25641

DavidBoike (approved):

PR #3719, opened 2023-09-29
Store message body attachments on FailedMessage documents

DavidBoike:


src/ServiceControl.Persistence.RavenDb5/RavenDBPersisterSettings.cs
```diff
@@ -27,4 +27,6 @@ class RavenDBPersisterSettings : PersistenceSettings
     public const int DatabaseMaintenancePortDefault = 33334;
     public const int ExpirationProcessTimerInSecondsDefault = 600;
     public const string LogsModeDefault = "Information";
+
+    public override bool MessageBodiesAlwaysStoredInFailedMessage => true;
```
ramonsmits, line 31: Why does this exist?

src/ServiceControl.Persistence.RavenDb5/RavenAttachmentsBodyStorage.cs
```diff
@@ -1,43 +1,61 @@
...
-        public async Task Store(string messageId, string contentType, int bodySize, Stream bodyStream)
+        public Task Store(string uniqueId, string contentType, int bodySize, Stream bodyStream)
+            => throw new NotImplementedException("Only included for interface compatibility with Raven3.5 persister implementation. Raven5 tests should use IIngestionUnitOfWorkFactory to store failed messages/bodies.");
+
+        public async Task<MessageBodyStreamResult> TryFetch(string bodyId)
         {
-            var documentId = MessageBodyIdGenerator.MakeDocumentId(messageId);
+            using var session = documentStore.OpenAsyncSession();
+
+            // BodyId could be a MessageID or a UniqueID, but if a UniqueID then it will be a DeterministicGuid of MessageID and endpoint name and be Guid-parseable
```
ramonsmits, line 29: Why can it be a message ID or unique ID?

src/ServiceControl.Persistence.RavenDb5/ErrorMessagesDataStore.cs
```diff
@@ -730,13 +734,37 @@ await using (var ie = await session.Advanced.StreamAsync(query))
...
+            var result = await bodyStorage.TryFetch(uniqueMessageId);
+
+            if (result == null)
+            {
+                throw new InvalidOperationException("IBodyStorage.TryFetch result cannot be null");
+            }
+
+            if (result.HasResult)
+            {
+                using (result.Stream) // Not strictly required for MemoryStream but might be different behavior in future .NET versions
```
ramonsmits, line 752: Wouldn't it be more efficient to check if Stream is a MemoryStream so that the `CopyToAsync` isn't required?

DavidBoike: This chunk is [copy-pasted from here](https://github.com/Particular/ServiceControl/blob/18bc698d6a67e4721b39ae6ddcf7c7dddb7592b6/src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSender.cs#L114) so I didn't analyze it to that level.

src/ServiceControl.Persistence.RavenDb5/RavenAttachmentsBodyStorage.cs
DavidBoike: Because in my redesign, we're always storing it using the unique ID, and the message metadata will say "look at `messages/{uniqueId}/body` for the body" so in theory, it should always be the unique ID. But previous versions stored the body only by message ID and while in theory, Pulse and Insight should pay attention to the message metadata that points to the uniqueID-based URL, it's impossible for me to say that assumption hasn't leaked into one of those apps. So just following what the API looks like, it's entirely reasonable some client somewhere could try to access by message ID.

src/ServiceControl.Persistence.RavenDb5/RavenAttachmentsBodyStorage.cs
```diff
@@ -1,43 +1,61 @@
...
+            // This is preferred, then we know we're getting the correct message body that is attached to the FailedMessage document
+            if (Guid.TryParse(bodyId, out _))
```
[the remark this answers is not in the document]
DavidBoike: More by the API history than what Core is forcing through our abstraction.

The reason for a Guid.TryParse() is that a uniqueID is always going to be built by DeterministicGuid and will always be parseable as a Guid. This might be true of a MessageID too, but if you had a custom message ID strategy or you had an MSMQ-like `{Guid}\{int}` construction, then we can easily skip this part knowing it's not going to succeed.

src/ServiceControl.Persistence.RavenDb5/RavenDBPersisterSettings.cs
DavidBoike: Because the Raven3.5 persister might store it text-encoded in the document or it might store it in body storage, depending on reasons. When you're doing recoverability you have to figure out where it is, so this just tells it "Raven5 only does it one way".

If we remove the Raven3.5 code then we could simplify everything and remove this property as well.

src/ServiceControl.Persistence.RavenDb5/RavenAttachmentsBodyStorage.cs
ramonsmits: If we test SI + SP and they use the BodyUrl we could simplify this and only use Unique ID?

src/ServiceControl.Persistence.Tests.RavenDb/TestPersistenceImpl.cs
```diff
@@ -70,7 +69,7 @@ public override void BlockToInspectDatabase()
                 return;
             }
 
-            var url = $"http://localhost:{settings.DatabaseMaintenancePort}/studio/index.html#databases/documents?&database=%3Csystem%3E";
+            var url = $"http://localhost:{(Settings as RavenDBPersisterSettings).DatabaseMaintenancePort}/studio/index.html#databases/documents?&database=%3Csystem%3E";
```
ramonsmits, line 72:
```suggestion
            var databaseMaintenanceUrl = ((RavenDBPersisterSettings)Settings).DatabaseMaintenanceUrl
            var url = databaseMaintenanceUrl + $"/studio/index.html#databases/documents?&database=%3Csystem%3E";
```

src/ServiceControl.Persistence.RavenDb5/ErrorMessagesDataStore.cs
DavidBoike: Like we talked about, I'd rather put this in a different PR.

src/ServiceControl.Persistence.Tests.RavenDb/TestPersistenceImpl.cs
DavidBoike: This is an unrelated change too

ramonsmits: Your change was doing an `as` which could resolve to `null`. This change uses a cast which seems to me would be a better fit as the instance should always be that type.

ramonsmits (approved):

tmasternak (approved):

PR #3720, opened 2023-10-02
Move MultiInstance tests to RavenDB 5

DavidBoike:


src/ServiceControl/SagaAudit/SagaUpdatedHandler.cs
```diff
@@ -1,24 +1,102 @@
...
+                var connectionDetails = await connectionBuilder.BuildPlatformConnection();
+
+                if (connectionDetails.ToDictionary().TryGetValue("SagaAudit", out var sagaAuditObj) && sagaAuditObj is JObject sagaAudit)
+                {
+                    auditQueueName = sagaAudit["SagaAuditQueue"].Value<string>();
+                    nextAuditQueueNameRefresh = DateTime.UtcNow.AddMinutes(5);
+                }
+            }
+            catch (Exception)
+            {
```
tmasternak, line 57: This could happen also when we cannot connect to the remote instance that is handling SagaAudit messages. 

In such a case, in order to prevent a message loss scenario I think we should consider:
- Either throwing an exception and making sure SagaAudit messages surface in the FailedMessages view in ServiceControl (when could add a custom recoverability policy to make sure that AuditMessages are moved to the error queue on the first exception)
- or store the messages that cannot be forwarded and tell the users to export/import the `SagaSnapshot` collection by hand.

I prefer the first option.

tmasternak (dismissed):

src/ServiceControl/SagaAudit/SagaUpdatedHandler.cs
```diff
@@ -1,24 +1,102 @@
...
+            try
+            {
+                if (auditQueueName != null && nextAuditQueueNameRefresh > DateTime.UtcNow)
+                {
+                    return;
+                }
+
+                var connectionDetails = await connectionBuilder.BuildPlatformConnection();
+
+                if (connectionDetails.ToDictionary().TryGetValue("SagaAudit", out var sagaAuditObj) && sagaAuditObj is JObject sagaAudit)
```
ramonsmits, line 50: How does this behave when there are multiple audit instances?

src/ServiceControl/SagaAudit/SagaUpdatedHandler.cs
```diff
@@ -1,24 +1,102 @@
...
+                {
+                    auditQueueName = sagaAudit["SagaAuditQueue"].Value<string>();
+                    nextAuditQueueNameRefresh = DateTime.UtcNow.AddMinutes(5);
```
ramonsmits, line 53: Magic number, I would prefer this to be a readonly TimeSpan where this value is also logged when set

src/ServiceControl/SagaAudit/SagaUpdatedHandler.cs
ramonsmits: Agreed, just throw here so that recoverability can deal with it

src/ServiceControl/SagaAudit/SagaUpdatedHandler.cs
```diff
@@ -1,24 +1,102 @@
...
 
-            return store.StoreSnapshot(sagaSnapshot);
+            if (auditQueueName is null || nextAuditQueueNameRefresh < DateTime.UtcNow)
+            {
+                await RefreshAuditQueue();
+            }
+
+            if (auditQueueName is null)
+            {
+                throw new Exception("Could not determine audit queue name to forward saga update message. The ServiceControl remote audit instance ");
```
ramonsmits, line 32: I think the message is incomplete?
```suggestion
                throw new Exception("Could not retrieve audit queue to forward saga update from audit connection info.");
```

ramonsmits (approved):
Ok with this although it feels a bit complex

src/ServiceControl/SagaAudit/SagaUpdatedHandler.cs
DavidBoike: Here is where each audit instance would respond with its own data:

https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl.Audit/Connection/ConnectionController.cs#L16-L31

Here's where the primary would be asking each instance for that:

https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/Connection/RemotePlatformConnectionDetailsProvider.cs#L29-L57

On line 46 you can see it's calling `connection.Add(…)` for each key it gets back which is here:

https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/Connection/PlatformConnectionDetails.cs#L10-L23

So if you have multiple audit instances all providing a `SagaAudit` key, the rest will become `SagaAudit1`, `SagaAudit2`, etc. but we'll take the value from the first non-numbered one.

And ultimately they should all be the same queue anyway. Unless you're on MSMQ and the Audit instance doesn't know about servernames. But if that's the case then the Platform Connection feature has a bug in it. In any case I intend to check that later on.

src/ServiceControl/SagaAudit/SagaUpdatedHandler.cs
DavidBoike: I don't know why you want it to be a separate field, especially since I ended up adding a second timeout for failures so that ingestion does not back up, but I did add the logging so you'll know what's going on from the logs.

src/ServiceControl/SagaAudit/SagaUpdatedHandler.cs
```diff
@@ -1,24 +1,82 @@
...
+            try
+            {
+                if (nextAuditQueueNameRefresh > DateTime.UtcNow)
+                {
+                    return;
+                }
+
+                var connectionDetails = await connectionBuilder.BuildPlatformConnection();
+
+                if (connectionDetails.ToDictionary().TryGetValue("SagaAudit", out var sagaAuditObj) && sagaAuditObj is JObject sagaAudit)
```
ramonsmits, line 56:
```suggestion
				// First instance is named `SagaAudit`, following instance `SagaAudit1`..`SagaAuditN`
                if (connectionDetails.ToDictionary().TryGetValue("SagaAudit", out var sagaAuditObj) && sagaAuditObj is JObject sagaAudit)
```

src/ServiceControl/SagaAudit/SagaUpdatedHandler.cs
```diff
@@ -1,24 +1,82 @@
...
+                {
+                    auditQueueName = sagaAudit["SagaAuditQueue"].Value<string>();
```
ramonsmits, line 58:
```suggestion
					// Pick any audit queue, assume all instance are based on competing consumer
                    auditQueueName = sagaAudit["SagaAuditQueue"].Value<string>();
```

src/ServiceControl.Persistence.RavenDb5/SagaAudit/SagaAuditDestinationCustomCheck.cs
```diff
@@ -0,0 +1,68 @@
...
+        public override Task<CheckResult> PerformCheck()
+        {
+            var failedEndpoints = stateHolder.GetFailedEndpoints();
+
+            if (failedEndpoints.Length == 0)
+            {
+                return passResult;
+            }
+
+            var message = $"In the last 24 hours, the following endpoints have reported saga audit data to the ServiceControl Primary instance. Instead, saga audit data should be sent to the Audit Queue Name configured in the ServiceControl Audit Instance. Affected endpoints: "
```
ramonsmits, line 29: Nice!

ramonsmits (approved):
LGTM, added a few code comment suggestions regarding resolving the audit queue

PR #3724, opened 2023-10-03
Remove confusion about `ServerUrl`

ramonsmits:
Remove confusion about `ServerUrl` by removing `DatabaseMaintenanceUrl` and documenting settings property


DavidBoike:
Ramon and I decided we should check to see what the Audit instance is doing and potentially roll changes to that into this PR as well.

ramonsmits:
Analyzed audit instance. Audit instance only uses `localhost` for the server URL:

https://github.com/Particular/ServiceControl/blob/c4ac6aa6/src/ServiceControl.Audit.Persistence.RavenDb5/RavenDbPersistenceConfiguration.cs#L64

Going to apply the same for primary instance.

DavidBoike (approved):

PR #3725, opened 2023-10-03
Path was still based on RavenDB 3.5 implementation, updated to use RavenDB 5 voron location

ramonsmits:


src/ServiceControl.Persistence.RavenDb5/EmbeddedDatabase.cs
```diff
@@ -174,7 +174,7 @@ static void RecordStartup(RavenDBPersisterSettings settings)
 
         static long DataSize(RavenDBPersisterSettings settings)
         {
-            var datafilePath = Path.Combine(settings.DatabasePath, "data");
+            var datafilePath = Path.Combine(settings.DatabasePath, "Databases", settings.DatabaseName, "Raven.voron");
```
DavidBoike, line 177: The database is more than that file. One database that grows large enough can have multiple data files in it, and the Indexes directory has a directory for each index which each have their own Raven.voron file as well. For production systems those index directories will also get quite large.
```suggestion
            var datafilePath = Path.Combine(settings.DatabasePath, "Databases", settings.DatabaseName);
```

DavidBoike (changes requested):

src/ServiceControl.Persistence.RavenDb5/EmbeddedDatabase.cs
DavidBoike: Nevermind, Voron uses [a single file](https://ravendb.net/articles/disk-space-management-inside-voron) and I missed the bit about this being only for the Database file size, and that the Database Folder Size covered what I was thinking of.

DavidBoike (approved):

PR #3727, opened 2023-10-05
Message retention via document expiration with RavenDB 5

ramonsmits:


DavidBoike (approved):

src/ServiceControl.Persistence.RavenDb5/ErrorMessagesDataStore.cs
```diff
@@ -601,7 +599,7 @@ public async Task<(string[] ids, int count)> UnArchiveMessagesByRange(DateTime f
                            update
                            {{
                                 msg.Status = args.Unresolved
-                                delete msg['@metadata']['@expires']
+                                {ExpirationManager.DeleteExpirationFieldScript}
```
DavidBoike, line 602: From the examples on https://ravendb.net/docs/article-page/6.0/csharp/client-api/operations/patching/set-based it looks like the statements inside the `update` block are going to need semicolons

DavidBoike:
I'm a little concerned that the effort to try to isolate expiration behind an ExpirationManager is wasted and makes things more complex, not less. If you could do it cleanly, that'd be one thing, but there are 3 different ways to enable expiration and 2 different ways to cancel it, some of which work on the session and others that have to alter some javascript command string. It just seems really messy and error-prone.

My vote would be to revert the last 2-ish commits.

tmasternak:
@DavidBoike I think there is still value in keeping all code connected to RavenDB 5 expiry management in a single class. My reasoning is:
- it makes it visible that expiration management is messy - removing `ExpirationManager` would hide this and not solve it
- it is a single place in which all the messiness is contained - a go-to place to figure out what the logic is and were it is used.

PR #3733, opened 2023-10-09
Bump to 5.0 and remove RavenDB 3.5 persister code

DavidBoike:


ramonsmits:
@tmasternak easier to review: https://github.com/Particular/ServiceControl/compare/dec0f8d...rm-raven35

src/ServiceControl.Audit.UnitTests/PersistenceManifestCanLoadFromRoot.cs
```diff
@@ -6,8 +6,13 @@
     [TestFixture]
     public class PersistenceManifestCanLoadFromRoot
     {
-        const string persistenceName = "RavenDB35";
-        const string persistenceType = "ServiceControl.Audit.Persistence.RavenDb.RavenDbPersistenceConfiguration, ServiceControl.Audit.Persistence.RavenDb";
+        const string persistenceName = "InMemory";
+        const string persistenceType = "ServiceControl.Audit.Persistence.InMemory.InMemoryPersistenceConfiguration, ServiceControl.Audit.Persistence.InMemory";
+
+        // TODO: Not really sure why this test was here in the first place. Before removing Raven35 code it was testing that "RavenDB35" could be turned
```
ramonsmits, line 12: @jpalac What is the intent of this test? Git blame shows that you created it.

ramonsmits (approved):

src/ServiceControl.Audit.UnitTests/PersistenceManifestCanLoadFromRoot.cs
jpalac: Hmm good question. When we introduced the `PersistenceManifestLibrary` we added the ability to dynamically load persistence from disk - ie without it being referenced but just the persistence files existing in a specific SC folder. 
I believe that this test was to verify backwards compatibility - ie that it still finds the persistence on disk if it's a referenced assembly (so in the main/root folder as opposed to in the specific persistence folders we introduced)

PR #3740, opened 2023-10-11
Spike RavenDB 6

DavidBoike:
Spike shows the changes necessary to adopt RavenDB 6.0. Indexes can be specified to prefer Corax service-wide, and existing ServiceControl databases will continue to use Lucene unless reset.

Must use at least 6.0.1 as 6.0.0 contains a bug in the Corax index implementation that prevents messages from being retrieved by using an index search on message id.

Given the number of command-line flags being forwarded to RavenDB it might be a good idea to package these up into a configuration file in a program data location, and give the process a pointer to that file when starting the database.

### DIFF
--- a/nuget.config
+++ b/nuget.config
@@ -1,10 +1,14 @@
 <?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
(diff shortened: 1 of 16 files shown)


DavidBoike:
Spike has served its purpose.

PR #3767, opened 2023-10-20
Improve development manifest discovery

bording:
Updates transport and persistence manifest management to improve the way the debug-time F5 experience is implemented.

Previously the projects used artifact copying to move transport and persistence bits around into subdirectories of each app. This is an inversion of normal project dependencies that could cause problems, such as making changes to persistence code and not having the correct code loaded for debugging unless the persistence library was specifically rebuilt before running the project.

In this reconfiguration, artifacts are not copied around in this way, but the transport and persistence loading infrastructure instead understands where those projects will exist on disk during the development process, and can load the assemblies directly from the artifacts of that transport/persistence project, without needing them to exist in a subdirectory of the app's own bin folder.

The loading infrastructure _can_ still load the transport or persistence plugins from a subdirectory as well, which will be the new layout for a production build. This gets us closer to a paradigm that will be necessary for Docker deployments where the binaries for all transport and persistence options are present and config or environment variables are used to choose which plugin to load at runtime.


DavidBoike (approved):

src/ServiceControl.Audit.Persistence/DevelopmentPersistenceLocations.cs
```diff
@@ -0,0 +1,36 @@
...
+    static class DevelopmentPersistenceLocations
+    {
+        public static List<string> ManifestFiles { get; } = new List<string>();
+
+        static DevelopmentPersistenceLocations()
+        {
+            var assembly = typeof(DevelopmentPersistenceLocations).Assembly.Location;
+            var assemblyDirectory = Path.GetDirectoryName(assembly);
+
+            var srcFolder = Path.GetDirectoryName(Path.GetDirectoryName(Path.GetDirectoryName(Path.GetDirectoryName(assemblyDirectory))));
```
DavidBoike, line 15:
```suggestion
            // Becomes null if it navigates past the root of a drive
            var srcFolder = Path.GetDirectoryName(Path.GetDirectoryName(Path.GetDirectoryName(Path.GetDirectoryName(assemblyDirectory))));
```

src/ServiceControl.Persistence/DevelopmentPersistenceLocations.cs
```diff
@@ -0,0 +1,35 @@
...
+    static class DevelopmentPersistenceLocations
+    {
+        public static List<string> ManifestFiles { get; } = new List<string>();
+
+        static DevelopmentPersistenceLocations()
+        {
+            var assembly = typeof(DevelopmentPersistenceLocations).Assembly.Location;
+            var assemblyDirectory = Path.GetDirectoryName(assembly);
+
+            var srcFolder = Path.GetDirectoryName(Path.GetDirectoryName(Path.GetDirectoryName(Path.GetDirectoryName(assemblyDirectory))));
```
DavidBoike, line 15:
```suggestion
            // Becomes null if it navigates past the root of a drive
            var srcFolder = Path.GetDirectoryName(Path.GetDirectoryName(Path.GetDirectoryName(Path.GetDirectoryName(assemblyDirectory))));
```

src/ServiceControl.Transports/DevelopmentTransportLocations.cs
```diff
@@ -0,0 +1,42 @@
...
+    static class DevelopmentTransportLocations
+    {
+        public static List<string> ManifestFiles { get; } = new List<string>();
+
+        static DevelopmentTransportLocations()
+        {
+            var assembly = typeof(DevelopmentTransportLocations).Assembly.Location;
+            var assemblyDirectory = Path.GetDirectoryName(assembly);
+
+            var srcFolder = Path.GetDirectoryName(Path.GetDirectoryName(Path.GetDirectoryName(Path.GetDirectoryName(assemblyDirectory))));
```
DavidBoike, line 15:
```suggestion
            // Becomes null if it navigates past the root of a drive
            var srcFolder = Path.GetDirectoryName(Path.GetDirectoryName(Path.GetDirectoryName(Path.GetDirectoryName(assemblyDirectory))));
```

DavidBoike (approved):
Just adding some suggestions because at first glance that looks like something that could throw. Verified return-null behavior in both .NET and .NET Framework in different versions of LINQPad.

PR #3756, opened 2023-10-17
Run ServiceControl Management as a self-contained app

DavidBoike:

PR #3773, opened 2023-10-24
Optimize binaries

DavidBoike:
_Waiting for https://github.com/Particular/ServiceControl/pull/3756, then will be rebased and retargeted with master as the base branch. There are likely approval tests currently broken as well._

- https://github.com/Particular/ServiceControl/pull/3756

----

* Reduces the size of the self-contained installer from 316 MB to 188 MB (59.4% of original)
* Reduces the size of the PowerShell module from 480 MB to 227 MB (47.3% of original)

It does so by:

1. Storing one set of transport plugins, rather than copying them to all three apps
1. In transport/persistence plugins, not including dependencies of the abstraction project as these binaries will be "merged" as runtime (`<ProjectReference Include="**.csproj" Private="false" ExcludeAssets="runtime" />`)
1. Not including RavenDB server artifacts in each instance
1. Reducing dependencies that were only used to get a Type or String value (Lucene)


src/ServiceControl.Audit.Persistence.RavenDb5/Indexes/MessagesViewIndex.cs
```diff
@@ -34,7 +33,7 @@ public MessagesViewIndex()
 
             Index(x => x.Query, FieldIndexing.Search);
 
-            Analyze(x => x.Query, typeof(StandardAnalyzer).AssemblyQualifiedName);
+            Analyze(x => x.Query, "StandardAnalyzer");
```
DavidBoike, line 36: We were including the entire Lucene package just to get one string. I believe this should work and be "the same" in RavenDB (not cause a re-index on deploy) but need to double-check.

ramonsmits:
```suggestion
            Analyze(x => x.Query, "StandardAnalyzer"); // Not using typeof() to prevent dependency on Lucene
```

src/ServiceControl.Audit.Persistence.RavenDb5/Indexes/MessagesViewIndexWithFullTextSearch.cs
```diff
@@ -35,7 +34,7 @@ public MessagesViewIndexWithFullTextSearch()
 
             Index(x => x.Query, FieldIndexing.Search);
 
-            Analyze(x => x.Query, typeof(StandardAnalyzer).AssemblyQualifiedName);
+            Analyze(x => x.Query, "StandardAnalyzer");
```
ramonsmits, line 37:
```suggestion
            Analyze(x => x.Query, "StandardAnalyzer"); // Not using typeof() to prevent dependency on Lucene
```

src/ServiceControl.Persistence.RavenDb5/ServiceControl.Persistence.RavenDb5.csproj
```diff
@@ -6,12 +6,14 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <ProjectReference Include="..\ServiceControl.Infrastructure\ServiceControl.Infrastructure.csproj" />
-    <ProjectReference Include="..\ServiceControl.Persistence\ServiceControl.Persistence.csproj" />
+    <!-- Private=false & ExcludeAssets=runtime prevent repeatedly including binary dependencies of ServiceControl.Persistence in each persister directory -->
```
ramonsmits, line 9: Makes sense as those assemblies are already there at runtime. I thought copying was only done for packages and executables but likely that is pre SDK behavior.

tmasternak (approved):

bording (approved):

PR #3777, opened 2023-10-26
Added SuppressPropertyChangedWarnings to remove warning

ramonsmits:
The warnings states to add the `SuppressPropertyChangedWarnings` attribute. Not sure if that is the right fix but I'm fed up with the warning :-)


ramonsmits:
Interestingly I don't see it on master anymore so closing

PR #3783, opened 2023-10-30
Load transport information from manifest files

DavidBoike:
Load transport info from manifest files the same way as persistence manifests, although better class name alignment will come in a future PR.

Transports will now be referred to in app.config files by Name and not by TypeName, but loading by old names is still supported. All transport information is now stored in transport.manifest files and not hardcoded.


ramonsmits (approved):

ramonsmits (approved):

src/ServiceControlInstaller.Engine/FileSystem/FileUtils.cs
```diff
@@ -126,6 +117,60 @@ internal static void UnzipToSubdirectory(Stream zipStream, string targetPath, st
...
+
+            static readonly char[] directorySplitChars = new[] { '/', '\\' };
+
+            public ZipFilterEvaluator(string zipFolderNameToExtract, string targetPath)
+            {
+                folderNameSegments = (zipFolderNameToExtract ?? string.Empty).Split(directorySplitChars, StringSplitOptions.RemoveEmptyEntries);
+                this.targetPath = targetPath;
+            }
+
+            public bool Evaluate(ZipEntry zipEntry, out string resultPath)
```
tmasternak, line 134: This method is doing two things: 
- applying the filter on the zip entry segment - which dictates if the entry should be extracted or not
- calculating the target path to which the entry should be extracted

Should we break up the method into two submethods? Should we change the naming from `Evalute`? Maybe to `ShouldExtract`?

tmasternak (dismissed):

src/ServiceControlInstaller.Engine/FileSystem/FileUtils.cs
DavidBoike: Split and renamed

PR #3788, opened 2023-11-01
Require PowerShell 7.2

DavidBoike:
Raises the PowerShell requirement from PowerShell 5.1 to PowerShell 7.2.

PowerShell 5.1 requires multi-targeting for .NET Framework and .NET 6, resulting in bloated file sizes as the resources containing the ServiceControl executables must be copied for both platforms.

PowerShell 7.2 is an LTS release based on .NET 6.


ramonsmits (approved):

src/ServiceControl.Management.PowerShell/Particular.ServiceControl.Management.format.ps1xml
```diff
@@ -15,8 +15,8 @@
                 <PropertyName>Name</PropertyName>
               </ListItem>
               <ListItem>
-                <Label>TypeName</Label>
-                <PropertyName>TypeName</PropertyName>
+                <Label>DisplayName</Label>
```
bording, line 18: I think we might want to make some additional changes here. I'm not sure that there's a point in showing the display name? Also, there's some weirdness around "deprecated" things. For example, the learning transport is showing in the output, but as deprecated.

tmasternak (approved):

PR #3792, opened 2023-11-03
Handle future transport/persistence settings without crashing

DavidBoike:
ServiceControl 5.0 and above will not have a Windows Installer, which means that a user can have multiple versions of Service Control Management that can be used back and forth without having to uninstall/reinstall a different version.

That introduces a bug in ServiceControl 4.x because, by definition, ServiceControl 4 versions will not be able to understand the different transports and persistence options that are introduced in the future.

This bugfix shows some "Unknown" text when an instance's transport or persistence is not known to the current ServiceControl version. This will be shipped as part of ServiceControl 4.33.0, which is intended to be the last minor version of the 4.x timeline, and will be the last to be installed through a Windows Installer.

So if you have a hybrid system and need to be able to install ServiceControl 5.x instances while still managing a few ServiceControl 4.x instances, update Service Control Management to 4.33 so you can open it and view the 5.x instances without crashing.

### DIFF
--- a/src/ServiceControlInstaller.Engine/Instances/ServiceControlAuditPersisters.cs
+++ b/src/ServiceControlInstaller.Engine/Instances/ServiceControlAuditPersisters.cs
@@ -8,7 +8,7 @@
 
     public class ServiceControlAuditPersisters
     {
-        public static IReadOnlyList<PersistenceManifest> LoadAllManifests(string zipFilePath)
+        internal static IReadOnlyList<PersistenceManifest> LoadAllManifests(string zipFilePath)
         {
             using (var zipArchive = ZipFile.OpenRead(zipFilePath))
             {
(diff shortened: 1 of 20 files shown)


src/ServiceControlInstaller.Engine/Instances/ServiceControlAuditPersisters.cs
```diff
@@ -31,5 +31,22 @@ public static IReadOnlyList<PersistenceManifest> LoadAllManifests(string zipFile
...
+        {
+            var manifests = LoadAllManifests(zipFilePath);
+
+            if (string.IsNullOrEmpty(name))
+            {
+                // Must always remain RavenDB35 so that SCMU understands that an instance with no configured value is an old Raven 3.5 instance
+                return manifests.Single(m => m.Name == "RavenDB35");
+            }
+
+            return manifests.FirstOrDefault(m => m.Name == name || m.TypeName == name) ?? new PersistenceManifest
```
ramonsmits, line 45: Be explicit in the string comparisons?

ps. Didn't check the using! Drive by code suggestion!

```suggestion
            return manifests.FirstOrDefault(m => m.Name.Equals(name, StringComparison.OrdinalIgnoreCase) || m.TypeName.Equals(name, StringComparison.OrdinalIgnoreCase)) ?? new PersistenceManifest
```

ramonsmits (approved):

tmasternak (approved):

PR #3805, opened 2023-11-09
Force upgrade for audit instance and added force upgrade to Powershell

ramonsmits:
1. SCMU now also has an *Upgrade Instance* under advanced settings for audit instances.
2. Powershell cmdlets `Invoke-ServiceControlAuditInstanceUpgrade` and `Invoke-ServiceControlInstanceUpgrade` now have `-ForceRecreateDatabase` switch. When set the database path will be moved and the persister manifest will be forced to the latest Raven engine.

- [x]  Ensure https://github.com/Particular/ServiceControl/pull/3806 is merged
- [x]  Rebase on https://github.com/Particular/ServiceControl/pull/3808
- [x]  Refactored to use `Force` switch
- [x]  Ensure https://github.com/Particular/ServiceControl/pull/3808 is merged

PR #3810, opened 2023-11-13
Switch from `!disable` to dedicated settings to disable error/audit queue ingestion

tmasternak:
Merge after
- [x] https://github.com/Particular/ServiceControl/pull/3805

### DIFF
--- a/src/ServiceControl/Infrastructure/Settings/Settings.cs
+++ b/src/ServiceControl/Infrastructure/Settings/Settings.cs
@@ -395,7 +395,7 @@ namespace ServiceBus.Management.Infrastructure.Settings
                 throw new Exception("ServiceBus/ErrorQueue requires a value to start the instance");
             }
 
-            IngestErrorMessages = !SettingsReader.Read("IngestErrorMessages", true);
+            IngestErrorMessages = SettingsReader.Read("IngestErrorMessages", true);
 
             if (!IngestErrorMessages)
             {


src/ServiceControl.Audit/Infrastructure/Settings/Settings.cs
```diff
@@ -44,6 +43,31 @@ public Settings(string serviceName, string transportType = null, string persiste
             EnableFullTextSearchOnBodies = SettingsReader<bool>.Read("EnableFullTextSearchOnBodies", true);
         }
 
+        void LoadAuditQueueInformation()
+        {
+            AuditQueue = SettingsReader<string>.Read("ServiceBus", "AuditQueue", "audit");
+
+            if (string.IsNullOrEmpty(AuditQueue))
+            {
+                throw new Exception("ServiceBus/AuditQueue requires a value to start the instance");
```
ramonsmits, line 52:
```suggestion
                throw new Exception("ServiceBus/AuditQueue value is required to start the instance");
```

src/ServiceControl.Audit/Infrastructure/Settings/Settings.cs
```diff
@@ -44,6 +43,31 @@ public Settings(string serviceName, string transportType = null, string persiste
...
+            }
+
+            IngestAuditMessages = SettingsReader<bool>.Read("ServiceControl", "DisableAuditQueueIngestion", false);
```
ramonsmits, line 55: I think it would be best if the setting key and assigned property share the same name:
```suggestion
            DisableAuditQueueIngestion= SettingsReader<bool>.Read("ServiceControl", "DisableAuditQueueIngestion", false);
```

ramonsmits (approved):
👍🏼 just minor comment to rename a property to align settings keys with code

src/ServiceControl.Audit/Infrastructure/Settings/Settings.cs
DavidBoike: Yet it needs to match https://github.com/Particular/ServiceControl/pull/3798 on the release branch, correct?

Either way if the name is "Ingest…" then the default should not be false, and if it's "Disable…" then the rest of the code needs to invert.

tmasternak: I think proposal makes sense. I'll adjust the naming on both PRs.

PR #3815, opened 2023-11-15
Using SemanticVersion type that supports pre-release suffix

ramonsmits:
Converts using `NuGet.Versioning.SemanticVersion` instead of `System.Version`. This allows using versions that have a pre-release suffix like `5.0.0-alpha.2.24`. This makes it easier for local testing upgrade logic

### DIFF
--- a/src/ServiceControl.Config/UI/Shell/ShellViewModel.cs
+++ b/src/ServiceControl.Config/UI/Shell/ShellViewModel.cs
@@ -2,18 +2,17 @@
 {
     using System;
     using System.Linq;
-    using System.Reflection;
     using System.Threading;
     using System.Threading.Tasks;
     using System.Windows.Input;
     using Caliburn.Micro;
     using Commands;
     using Events;
-    using Extensions;
     using Framework;
     using Framework.Rx;
     using ListInstances;
     using NoInstances;
+    using NuGet.Versioning;
     using ServiceControlInstaller.Engine.Instances;
 
     class ShellViewModel : RxConductor<RxScreen>.OneActive, IHandle<RefreshInstances>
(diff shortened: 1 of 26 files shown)


src/ServiceControl.Config/UI/Shell/ShellViewModel.cs
```diff
@@ -121,34 +118,17 @@ public async Task RefreshInstances()
 
         void LoadAppVersion()
         {
-            var assemblyInfo = typeof(App).Assembly.GetAttribute<AssemblyInformationalVersionAttribute>();
```
DavidBoike, line 124: Good-bye GitVersion!

src/ServiceControl.Config/UI/Shell/VersionCheckerHelper.cs
```diff
@@ -44,57 +45,45 @@ static async Task<List<Release>> GetVersionInformation()
...
-            public Release(Version current)
+            public Release(SemanticVersion current)
             {
-                Tag = current.ToString();
+                Tag = current.ToNormalizedString();
             }
 
-            [JsonProperty("tag")]
-            public string Tag { get; set; }
+            [JsonProperty("tag")] public string Tag { get; set; }
```
DavidBoike, line 68: Why did these change formatting? It's not really germane to this PR and is completely different than attribute usages on most every other project in the company.

DavidBoike (changes requested):

src/ServiceControl.Config/UI/Shell/VersionCheckerHelper.cs
ramonsmits: That was not intentional... maybe some auto formatting quick

DavidBoike (approved):

PR #3832, opened 2023-11-16
ServiceControl Management tweaks

DavidBoike:
1. Allow deleting instances of _older_ major versions
2. Centralize major version edit/delete checks so that PowerShell can share them
3. Fix issue with ServiceControl Management throwing an exception on refresh if PowerShell (or another version) has removed an instance in the background.


.github/workflows/ci.yml
```diff
@@ -33,7 +33,9 @@ jobs:
       - name: Setup .NET SDK
         uses: actions/setup-dotnet@v3.2.0
         with:
-          dotnet-version: 8.0.x
+          dotnet-version: |
+            8.0.x
+            7.0.x
```
ramonsmits, line 38: Why is 7 needed? I don't see a reference to is in the changes.

ramonsmits (approved):

.github/workflows/ci.yml
DavidBoike: To give RavenDB exactly what it wants. Even though the PR to step up to .NET 8 SDK worked in CI, in 90% of CI runs after that, Raven was failing to start in less than 30 seconds on one of the test projects. :shrug:

PR #3840, opened 2023-11-20
Fix incorrect usage of DateTime.Now

ramonsmits:
A lot of code was using `DateTime.Now`. This is the local time and should not be used for capturing time. It is affected by Daylight Saving Time (DST) changes and its slower than `.UtcNow` because it needs to convert the system its UTC time to a local time using the active timezone info.

### DIFF
--- a/src/ServiceControl/LoggingConfigurator.cs
+++ b/src/ServiceControl/LoggingConfigurator.cs
@@ -101,7 +101,7 @@ ServiceControl Version:				{version}
             var logger = LogManager.GetLogger("LoggingConfiguration");
             var logEventInfo = new LogEventInfo
             {
-                TimeStamp = DateTime.Now
+                TimeStamp = DateTime.UtcNow
             };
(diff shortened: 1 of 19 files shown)


src/ServiceControl/LoggingConfigurator.cs
```diff
@@ -101,7 +101,7 @@ public static void ConfigureLogging(LoggingSettings loggingSettings)
             var logger = LogManager.GetLogger("LoggingConfiguration");
             var logEventInfo = new LogEventInfo
             {
-                TimeStamp = DateTime.Now
+                TimeStamp = DateTime.UtcNow
             };
             //logger.InfoFormat("Logging to {0} with LogLevel '{1}'", fileTarget.FileName.Render(logEventInfo), loggingSettings.LoggingLevel.Name);
```
ramonsmits, line 106:
```suggestion
            logger.InfoFormat("Logging to {0} with LogLevel '{1}'", fileTarget.FileName.Render(logEventInfo), loggingSettings.LoggingLevel.Name);
```

DavidBoike (approved):

PR #3842, opened 2023-11-20
Bundle self-contained version of RavenDB Server

DavidBoike:
During a CI build, we will parse the [Directory.Packages.props](https://github.com/Particular/ServiceControl/blob/master/src/Directory.Packages.props) file to discover the current version of RavenDB, and then download the self-contained version (which does not require .NET to be preinstalled on the system) from RavenDB and use that as the resource embedded in the installer.

This means that the .NET and ASP.NET runtimes are no longer needed as prerequisites for installation, leaving the only prerequisite as .NET Framework 4.7.2 for ServiceControl itself.

Once ServiceControl is updated to .NET, and is also deployed as self-contained, there will be no install prerequisites left.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,6 +36,9 @@ jobs:
           dotnet-version: |
             8.0.x
             7.0.x
+      - name: Download RavenDB Server
+        shell: pwsh
+        run: ./tools/download-ravendb-server.ps1
       - name: Build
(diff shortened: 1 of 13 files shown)


bording (approved):

tmasternak (approved):

ramonsmits (approved):
It might be useful to explain the deployment and packaging in the README

PR #3853, opened 2023-11-23
Fix cmdlet connection string validation

ramonsmits:
Powershell cmdlets for creating error, audit and monitoring instance failed for MSMQ because sample connection string value was tested against `null` while it was `""` (empty string)

### DIFF
--- a/src/ServiceControl.Transports.Msmq/transport.manifest
+++ b/src/ServiceControl.Transports.Msmq/transport.manifest
@@ -5,7 +5,7 @@
       "Name": "MSMQ",
       "DisplayName": "MSMQ",
       "TypeName": "ServiceControl.Transports.Msmq.MsmqTransportCustomization, ServiceControl.Transports.Msmq",
-      "SampleConnectionString": "",
+      "SampleConnectionString": null,
       "AvailableInSCMU": true,
       "Default": true,
       "Aliases": [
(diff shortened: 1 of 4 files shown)


src/ServiceControl.Transports.Msmq/transport.manifest
```diff
@@ -5,7 +5,7 @@
       "Name": "MSMQ",
       "DisplayName": "MSMQ",
       "TypeName": "ServiceControl.Transports.Msmq.MsmqTransportCustomization, ServiceControl.Transports.Msmq",
-      "SampleConnectionString": "",
+      "SampleConnectionString": null,
```
DavidBoike, line 8: Not sure if this part was a good idea. Looks like that made a handful of tests fail. Before adjusting the tests I would make sure this doesn’t affect what happens to the validation when setting up an instance in SCMU.

tmasternak: The UI in SCMU [is doing the same check](https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl.Config/UI/InstanceAdd/ServiceControlEditorViewModel.cs#L47) i.e. `IsNullOrEmpty`. That said the tests seem to assume that `string.Empty` is the expected value so I'm 👍 for the change.

tmasternak (approved):

PR #3880, opened 2023-12-05
Package dependency cleanup

RoderickIveans:


bording:
@RoderickIveans Thanks for the PR.

We're in the middle of some major changes in this repo right now, and while we do want to a package dependencies cleanup similar to what you've done here, we have more work we want to do before we make these changes.

Because of that, we're not able to merge this at this time, so I'm going to go ahead and close it.

PR #3923, opened 2024-01-15
Fix dispose not being invoked

ramonsmits:
Fix dispose not being invoked and aligned all command runners to use `using var`


mauroservienti:
Not sure we want to bring this into the `net8` branch. In the primary instance, we drastically changed the way bootstrapping works and got rid of the bootstrapper class.

ramonsmits:
@mauroservienti I know, that is why I applied it on this branch to not have any merge conflicts that need to be resolved later

mauroservienti (approved):

danielmarbach (approved):

Issue #3909, opened 2024-01-04
RavenDB can timeout at start which causes a startup failure that sometimes is unrecoverable


ramonsmits:
### Describe the bug

RavenDB can timeout at start which causes a startup failure that sometimes is unrecoverable.

This can happen then there is some type of corruption in RavenDB that will take it longer than 30 seconds to recover thus start. ServiceControl will than crash at startup. If Windows Service Recovery is correctly setup Windows Service Control Manager will auto restart the instance after 1 minute but likely this fails again.

What is typical is that the "start" of the ServiceControl log file is short but clearly isn't fully initialized. In the Windows Task Manager is can be observed that the process eventually dies. 

### Steps to reproduce

No method known to reproduce.

### Relevant log output

```txt
2024-01-03 14:16:23.942 ERROR Unhandled exception was caught.
Raven.Client.Exceptions.Database.DatabaseLoadTimeoutException: Raven.Client.Exceptions.Database.DatabaseLoadTimeoutException: Database audit after 00:00:30 is still loading, try again later. Database initialization log:
[Load Database] 01/03/2024 14:15:53 :: Database 'audit' : Starting database initialization
[Load Database] 01/03/2024 14:15:53 :: Database 'audit' : Creating db.lock file
[Load Database] 01/03/2024 14:15:58 :: Database 'audit' : Initializing NotificationCenter
[Load Database] 01/03/2024 14:15:58 :: Database 'audit' : Initializing DocumentStorage

   at Raven.Server.Routing.RouteInformation.ThrowDatabaseLoadTimeoutWithLog(StringSegment databaseName, TimeSpan timeout, String log) in C:\Builds\RavenDB-Stable-5.4\54092\src\Raven.Server\Routing\RouteInformation.cs:line 225
   at Raven.Server.Routing.RouteInformation.UnlikelyWaitForDatabaseToLoad(RequestHandlerContext context, Task`1 database, DatabasesLandlord databasesLandlord, StringSegment databaseName) in C:\Builds\RavenDB-Stable-5.4\54092\src\Raven.Server\Routing\RouteInformation.cs:line 211
   at Raven.Server.Routing.RouteInformation.WaitForDb(Task databaseLoading) in C:\Builds\RavenDB-Stable-5.4\54092\src\Raven.Server\Routing\RouteInformation.cs:line 247
[...]

PR #3924, opened 2024-01-15
Defer documentstore creation

ramonsmits:
- Related to: https://github.com/Particular/ServiceControl/issues/3909

Defer creation of DocumentStore so that Initialize returns faster and the process not be killed by Windows Service Control Manager because the service takes over 25 seconds to start.

Its using `.GetAwaiter().GetResult();` which isn't really nice. Could be refactored so that most APIs are converted from async to sync. Will have the Initialize still be async.


ramonsmits:
Closing in favor of the following alternatives:

- https://github.com/Particular/ServiceControl/pull/3925
- https://github.com/Particular/ServiceControl/pull/3926

PR #3925, opened 2024-01-15
Custom lifetime manager to prevent Windows Service start timeout - Generic IHostLifetime wrapper 

ramonsmits:
- Related to: https://github.com/Particular/ServiceControl/issues/3909
- Alternative to: https://github.com/Particular/ServiceControl/pull/3926
- 
A `PersisterInitializingLifetimeWrapper` that is DI configured to use a specific console or windowsservice lifetime.

### DIFF
--- a/src/ServiceControl/Hosting/Commands/CustomLivetimes/HostBuilderLifetimeExtensions.cs
+++ b/src/ServiceControl/Hosting/Commands/CustomLivetimes/HostBuilderLifetimeExtensions.cs
@@ -1,0 +1,33 @@
+﻿namespace Particular.ServiceControl.Commands
+{
+    using global::ServiceControl.Persistence;
+    using Microsoft.Extensions.DependencyInjection;
+    using Microsoft.Extensions.Hosting;
+    using Microsoft.Extensions.Hosting.Internal;
+    using Microsoft.Extensions.Hosting.WindowsServices;
+
+    static class HostBuilderLifetimeExtensions
+    {
+        public static void SetupLifetime(this IHostBuilder hostBuilder, bool runAsWindowsService)
+        {
+            if (runAsWindowsService)
+            {
+                hostBuilder.UseWindowsService();
+                hostBuilder.ConfigureServices(s =>
+                {
+                    s.AddSingleton<WindowsServiceLifetime>();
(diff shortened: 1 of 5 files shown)


src/ServiceControl/Hosting/Commands/CustomLivetimes/HostBuilderLifetimeExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+    static class HostBuilderLifetimeExtensions
+    {
+        public static void SetupLifetime(this IHostBuilder hostBuilder, bool runAsWindowsService)
+        {
+            if (runAsWindowsService)
+            {
+                hostBuilder.UseWindowsService();
+                hostBuilder.ConfigureServices(s =>
+                {
+                    s.AddSingleton<WindowsServiceLifetime>();
```
tmasternak, line 18: do we need this, considering line 15?

src/ServiceControl/Hosting/Commands/CustomLivetimes/PersisterInitializingLifetimeWrapper.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Hosting;
+using ServiceControl.Persistence;
+
+
+sealed class PersisterInitializingLifetimeWrapper : IHostLifetime
```
tmasternak, line 7: could we make this more generic i.e. independent of persisted? Could be called `HostLifetimeWithInitializationPhase`

src/ServiceControl/Hosting/Commands/CustomLivetimes/HostBuilderLifetimeExtensions.cs
ramonsmits: Yes, as that line only registers an interface implementation and not the implementation separately.

src/ServiceControl/Hosting/Commands/CustomLivetimes/PersisterInitializingLifetimeWrapper.cs
ramonsmits: It could but then it should also have a constructor that allows for a more generic initialization but I don't think that is needed.

ramonsmits:
Closing in favor of https://github.com/Particular/ServiceControl/pull/3926

PR #3884, opened 2023-12-05
Update to .NET 8 on Windows

bording:

PR #3927, opened 2024-01-17
Make root controller test pass, but...

mauroservienti:
It requires changing the address the root controller uses, which I believe will have unintended consequences in production. Something is still off.


mauroservienti:
Fixed by https://github.com/Particular/ServiceControl/pull/3884/commits/be7f3d8e67aeb0a030fb24553d1e43dec5c9d088

Issue #3921, opened 2024-01-12
Incorrect ordering of events in Dashboard and Events page


saschanm:
#### Description
When viewing ServiceControl events via ServicePulse either from dashboard (last 10 events) or via Events tab (all events - paged) the returned results are incorrectly showing oldest first rather than most recent first.

#### Expected behavior
1) Dashboard should show most recent 10 events
2) Events tab should show most recent first (page 1)

#### Actual behavior
Dashboard shows the oldest retained events (14 day retention) so need to drill down into events tab to see recent events.
Events tab also in reverse order, so we have to go to last page to see most recent events.

Note also there is another bug that means that last page also includes some old events, adding a further impedence to viewing recent events - see:

- https://github.com/Particular/ServicePulse/issues/1559

####
This appears to have been introduced when upgrading Service Control instance to v5 - currently on v5.0.3 

### Steps to reproduce

Open ServicePulse to dashboard and see most recent events are old.

soujay:
@saschanm - Thank for reporting the bug.

soujay:
This has been released in [ServiceControl 5.0.4](https://github.com/Particular/ServiceControl/releases/tag/5.0.4)

PR #3937, opened 2024-01-19
Fix incorrect ordering of the events

soujay:
- Resolves: https://github.com/Particular/ServiceControl/issues/3921

Removes `OrderingType.Double`, with no value passed [RavenDB 5.x defaults to lexicographic ordering](https://ravendb.net/docs/article-page/5.4/csharp/client-api/session/querying/sort-query-results):

> When using RQL directly, if no ordering type is specified, then the server defaults to lexicographic ordering.

The document property on which is sorted is `RaisedAt`. This value is stored/interpreted as `2020-05-27T00:00:00.0000000Z`. Using lexicographic ordering will work correctly with these values.

### Symptoms

Incorrect order of events in ServicePulse events view.

### Who's affected

ServiceControl 5 instances

### Root cause

Incorrect `OrderingType` was passed to RavenDB query.

### DIFF
--- a/src/ServiceControl.Persistence.RavenDB/EventLogDataStore.cs
+++ b/src/ServiceControl.Persistence.RavenDB/EventLogDataStore.cs
@@ -5,7 +5,6 @@
     using EventLog;
     using Persistence.Infrastructure;
     using Raven.Client.Documents;
-    using Raven.Client.Documents.Session;
 
     class EventLogDataStore : IEventLogDataStore
     {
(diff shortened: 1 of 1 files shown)


kbaley (approved):

ramonsmits:
I see that `OrderingType.Double` is removed. However, the RavenDB docs state:

> When using RQL directly, if no ordering type is specified, then the server defaults to lexicographic ordering.

The document property on which is sorted is `RaisedAt`. If this is indeed lexicographic then it depends on the culture that RavenDB uses on how these timestamps are sorted.

ramonsmits (approved):

PR #3944, opened 2024-01-25
Make primary instance run

mauroservienti:


mauroservienti:
Fix cherry-picked to `net8` branch

PR #3962, opened 2024-02-14
Enable building on non-Windows OS without errors

bording:
These changes let non-Windows operating systems build successfully without any errors.

This PR is about building locally. It doesn't add Linux builds in CI or do anything to make SC run on Linux.

Note: I've also included a bump to `MinVerMinimumMajorMinor` that was needed because 5.0.0 was not tagged on a commit visible to `master`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -37,7 +37,7 @@ jobs:
       - name: Build
         run: dotnet build src --configuration Release -graph --no-restore
       - name: Build ServiceControl Management
-        run: dotnet publish src\ServiceControl.Config\ServiceControl.Config.csproj --configuration Release --runtime win-x64 --no-build --output assets
+        run: dotnet publish src\ServiceControl.Config\ServiceControl.Config.csproj --no-build --output assets
       - name: Sign NuGet packages
         uses: Particular/sign-nuget-packages-action@v1.0.0
         with:
--- a/src/ServiceControlInstaller.Engine/ServiceControlInstaller.Engine.csproj
+++ b/src/ServiceControlInstaller.Engine/ServiceControlInstaller.Engine.csproj
@@ -38,4 +38,10 @@
     </ItemGroup>
   </Target>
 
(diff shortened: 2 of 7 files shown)


src/ServiceControlInstaller.Engine/ServiceControlInstaller.Engine.csproj
```diff
@@ -38,4 +38,10 @@
     </ItemGroup>
   </Target>
 
+  <Target Name="WorkaroundForSqlClientWindowsFormsReference" BeforeTargets="AddTransitiveFrameworkReferences">
```
danielmarbach, line 41: The dependencies of the sql client are a big mess but there seems to be some movement

https://github.com/dotnet/SqlClient/issues/1108

.github/workflows/release.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       - name: Build
         run: dotnet build src --configuration Release -graph --no-restore
       - name: Build ServiceControl Management
-        run: dotnet publish src\ServiceControl.Config\ServiceControl.Config.csproj --configuration Release --runtime win-x64 --no-build --output assets
+        run: dotnet publish src\ServiceControl.Config\ServiceControl.Config.csproj --no-build --output assets
```
danielmarbach, line 40: Release is unnecessary with no-build I guess?

danielmarbach (approved):

danielmarbach:
Changes look good from my end. I can only test the build tomorrow on my linux machine but feel free to move forward if needed. We can also make more changes in dedicated PRs

src/ServiceControlInstaller.Engine/ServiceControlInstaller.Engine.csproj
bording: Yeah, I'm hoping that will be resolved soon and then we can remove these workaround targets.

.github/workflows/release.yml
bording: The .NET 8 SDK changed `dotnet publish` to default to `Release`, so there's no longer a reason to specify it.

PR #3968, opened 2024-02-19
Add initial throughput persistence library

abparticular:


src/Throughput.Persistence/PersistenceManifest.cs
jpalac, line 1: Do we need this? Ie the idea is to use the same persistence as the main instance, so this would be a setting just passed in from the main instance during initialization? The same with dblocation, dblogs and whatever other settings the user has chosen on the main instance.

src/Particular.ThroughputCollector.Persistence/PersistenceManifest.cs
```diff
@@ -0,0 +1,130 @@
+﻿namespace Particular.ThroughputCollector.Persistence;
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text.Json;
+using NServiceBus.Logging;
```
jpalac, line 8: I don't think that we should be referencing nservicebus - I've been using the microsoft.extensions.logging everywhere.

src/Particular.ThroughputCollector.Persistence.RavenDb/Particular.ThroughputCollector.Persistence.RavenDb.csproj
```diff
@@ -0,0 +1,20 @@
...
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="ByteSize" />
+    <!--<PackageReference Include="NServiceBus.CustomChecks" />-->
+    <PackageReference Include="NServiceBus" />
```
jpalac, line 12: I don't think that we should be referencing nservicebus - I've been using the microsoft.extensions.logging everywhere.

src/Particular.ThroughputCollector/Particular.ThroughputCollector.csproj
```diff
@@ -7,11 +7,12 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="System.Configuration.ConfigurationManager" />
+    <ProjectReference Include="..\Particular.ThroughputCollector.Persistence\Particular.ThroughputCollector.Persistence.csproj" />
+    <ProjectReference Include="..\ServiceControl.Configuration\ServiceControl.Configuration.csproj" />
```
jpalac, line 11: We should not be referencing anything related to ServiceControl.

jpalac (commented):
Couple of initial comments in relation to references.

PR #3970, opened 2024-02-19
Self-contained instances

bording:
This PR makes the SC instances fully self-contained for `win-x64`, so there is no dependency on what .NET runtimes are installed on the machine. 

By default, building the solution will still produce framework-dependent output. The `WindowsSelfContained` property needs to be true to get a self-contained build. The release workflows have been updated to set this property.

This change also ensures that when the release workflow builds the PowerShell module, only the `win-x64` versions of the dependencies are included.

To keep the file size of SCMU from growing dramatically by having to bundle 3 additional copies of the .NET runtime, some changes have been made to the `deploy` artifact layout and packaging:

- Each instance folder now contains the `Persisters` and `Transports` folders, and this entire file structure is deployed on disk for an instance. The manifest discovery features are then leveraged to choose the correct component based on the config value.
- [An MSBuild task has been added to Particular.Packaging](https://github.com/Particular/Particular.Packaging/pull/189) that is used to detect duplicate files. Any file that is duplicated in all 3 folders is excluded from the instance zip and is instead put into an `InstanceShared` zip. The actual duplicate files are not removed from the deploy folders, only the zip files are impacted.

ServiceControlInstaller.Engine has been updated to understand the new deployment layout and zip file structure.

Some misc. cleanup and small fixes have also been included as part of this PR.

---
One of these is important enough to call out separately here:

Since we're still using `ConfigurationManager` to read app.config files after our move to .NET, there is a problem that is introduced. On .NET, the config file being looked for is `{assembly}.dll.config`, but on .NET Framework the file is named `{assembly}.exe.config`. All of our existing installer logic is built around creating and updating `exe.config` files, and this is way too embedded into the code to safely change.

Instead, what I've introduced here is an `ReadExeConfiguration` method that the instances call first thing on startup that opens up an existing `exe.config` file and reads those settings into `ConfigurationManager` so that they will be available to the rest of the application. Unfortunately, to update the connection strings in this manner, some private reflection was required to make the `ConnectionStrings` properly not read-only. Given that the `ConfigurationManager` code is a shim to let app.config files be used on .NET, this code is not likely to break/change, but I have added a test to let us know if the reflection starts failing for some reason.

The end result of this change is that ServiceControl instances continue to use the existing `exe.config` file just like all previous versions of SC do.


src/ServiceControl.Audit/Program.cs
```diff
@@ -73,6 +77,33 @@ static Assembly TryLoadAssembly(AssemblyLoadContext loadContext, string folderPa
...
 
+        static void ReadExeConfiguration()
+        {
+            // ConfigurationManager on .NET is looking for {assembly}.dll.config files, but all previous versions of ServiceControl will have {assembly}.exe.config instead.
+            // This code reads in the exe.config files and adds all the values into the ConfigurationManager's collections.
+
+            var assembly = Assembly.GetExecutingAssembly();
+            var location = Path.GetDirectoryName(assembly.Location);
+            var assemblyName = assembly.GetName();
+            var exePath = Path.Combine(location, $"{assemblyName.Name}.exe");
```
danielmarbach, line 88: When executing locally with F5 the only thing I have is the DLL around and there is no exe. This change seems to break the F5 experience. Should we have some conditional logic that falls back to `dll.config`?

src/ServiceControl.Monitoring/Program.cs
```diff
@@ -63,6 +67,33 @@ static Assembly TryLoadAssembly(AssemblyLoadContext loadContext, string folderPa
             return null;
         }
 
+        static void ReadExeConfiguration()
```
danielmarbach, line 70: Same comments as above

src/ServiceControl.MultiInstance.AcceptanceTests/ServiceControl.runsettings
```diff
@@ -1,7 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<RunSettings>
-  <RunConfiguration>
-    <!-- The tests need to be run in x64 otherwise an exception about RavenDB (Voron) not being supported in 32bit mode will be thrown. -->
-    <TargetPlatform>x64</TargetPlatform>
```
danielmarbach, line 5: I have verified and I can still run tests locally so this change seems to be :+1:

src/ServiceControl/Program.cs
```diff
@@ -72,6 +76,33 @@ static Assembly TryLoadAssembly(AssemblyLoadContext loadContext, string folderPa
             return null;
         }
 
+        static void ReadExeConfiguration()
```
danielmarbach, line 79: Same comments as above

src/ServiceControl.Audit/Program.cs
mauroservienti: On macOS, I have the following:

image

The `ServiceControl` Unix executable, the `ServiceControl.dll`, and the corresponding `.dll.config` file.

mauroservienti: Shouldn't this code be guarded by an if checking for the `.exe.config` file existence?

danielmarbach (approved):
@bording discussed the feedback on the PR and will follow through on a bunch of those nitpicks. Giving the approval already

mauroservienti (approved):

src/ServiceControl.Audit/Program.cs
```diff
@@ -73,6 +77,33 @@ static Assembly TryLoadAssembly(AssemblyLoadContext loadContext, string folderPa
             return null;
         }
 
+        static void ReadExeConfiguration()
```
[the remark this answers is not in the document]
bording: I ended up moving it to ServiceControl.Configuration, which is actually shared by all three instances, and makes sense since this is related to configuration anyway.

src/ServiceControl.Audit/Program.cs
```diff
@@ -73,6 +77,33 @@ static Assembly TryLoadAssembly(AssemblyLoadContext loadContext, string folderPa
...
+
+            // This reflection is required because the connection strings collection has had its read only flag set, so it won't let anything be added to it.
+            var type = typeof(ConfigurationElementCollection);
+            var field = type.GetField("_readOnly", BindingFlags.NonPublic | BindingFlags.Instance);
+            field?.SetValue(ConfigurationManager.ConnectionStrings, false);
+
+            foreach (var connectionStringSetting in configuration.ConnectionStrings.ConnectionStrings.Cast<ConnectionStringSettings>())
+            {
+                ConfigurationManager.ConnectionStrings.Add(connectionStringSetting);
+            }
```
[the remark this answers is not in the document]
bording: As part of moving to using `UnsafeAccessor`, I went ahead and added a way to call the `SetReadOnly` method to put the collection back into its original state.

src/ServiceControl.Audit/Program.cs
bording: The issue is here is related to `{assemblyName.Name}.exe` being passed into `OpenExeConfiguration` not existing on non-Windows machines. I've changed the code to use `OpenMappedExeConfiguration`, which lets me pass in the actual config file name, and now this should work correctly on all OSes.

PR #3972, opened 2024-02-20
Licensing component setup

jpalac:
I'm sure this will change - it's a starting point to work from on the other branches


src/Licensing/AuditThroughputCollector.cs
```diff
@@ -0,0 +1,21 @@
+﻿namespace Particular.License
+{
+    using Microsoft.Extensions.Logging;
+
+    class AuditThroughputCollector
+    {
+        public AuditThroughputCollector(string serviceControlAPI, ILogger logger)
```
abparticular, line 7: I think we should use "typed" loggers wherever possible

```suggestion
        public AuditThroughputCollector(string serviceControlAPI, ILogger<AuditThroughputCollector> logger)
```

abparticular (approved):

src/Licensing/ThroughputCollectorHostedService.cs
```diff
@@ -0,0 +1,47 @@
+﻿namespace Particular.License
+{
+    using Microsoft.Extensions.Hosting;
+    using Microsoft.Extensions.Logging;
+    using Particular.License.Contracts;
+
+    class ThroughputCollectorHostedService : IHostedService
+    {
+        public ThroughputCollectorHostedService(ILoggerFactory loggerFactory, PlatformData platformData)
```
johnsimons, line 9: I am unsure if it makes sense to have this class.
I reckon each collector will have its own polling schedule, and hence this class becomes unnecessary.

src/Licensing/ThroughputCollectorHostedService.cs
```diff
@@ -0,0 +1,47 @@
+﻿namespace Particular.License
+{
+    using Microsoft.Extensions.Hosting;
+    using Microsoft.Extensions.Logging;
+    using Particular.License.Contracts;
+
+    class ThroughputCollectorHostedService : IHostedService
```
johnsimons, line 7: It may make more sense to use a [BackgroudService](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/host/hosted-services?view=aspnetcore-8.0&tabs=visual-studio#backgroundservice-base-class), thoughts?

src/ServiceControl/WebApplicationBuilderExtension.cs
```diff
@@ -132,6 +133,14 @@ public static void AddServiceControl(this WebApplicationBuilder hostBuilder, Set
             }
 
             hostBuilder.AddServiceControlComponents(settings, ServiceControlMainInstance.Components);
+
+            hostBuilder.AddThroughputCollector(broker: TransportManifestLibrary.GetName(settings.TransportType),
+               serviceControlAPI: settings.ApiUrl,
+               errorQueue: settings.ErrorQueue,
+               auditQueue: "?",
+               transportConnectionString: settings.TransportConnectionString,
+               persistenceType: PersistenceManifestLibrary.GetName(settings.PersistenceType)
```
johnsimons, line 142: Is this even required? We know we are going to use Raven, correct?

johnsimons (approved):
I've added a few comments for consideration, but still happy to merge this the way it is for now

src/Licensing/ThroughputCollectorHostedService.cs
jpalac: Well we want one entry point from the host, not 3, hence this class.

src/ServiceControl/WebApplicationBuilderExtension.cs
jpalac: It is as when the main instance changes to have options, we also need to have the option. Currently I think we also do an InMemory persister for testing, so that would be passed in.

src/Licensing/ThroughputCollectorHostedService.cs
jpalac: Yeah haven't figured that part out yet - that's step 2 :-)

src/Licensing/ThroughputCollectorHostedService.cs
johnsimons: That can be hidden away in `ThroughputCollectorHostBuilderExtensions.AddThroughputCollector`, thoughts?
Because I reckon we want the collectors themselves registered in DI.

src/Licensing/AuditThroughputCollector.cs
jpalac: If we are passing in the Ilogger from the main service then it will always be the type of that main service. That is typed - only way to get around that is to pass the loggerfactory around instead and i'm not sure we want to do that. Can easily be reviewed later on.

src/ServiceControl/WebApplicationBuilderExtension.cs
johnsimons: >It is as when the main instance changes to have options

There is an assumption here that, as we know, may not be true 😄

src/Licensing/ThroughputCollectorHostedService.cs
jpalac: I see what you mean though - I'll have a look to see what having 3 will be like

src/Licensing/ThroughputCollectorHostBuilderExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+
+    public static class ThroughputCollectorHostBuilderExtensions
+    {
+        public static IHostApplicationBuilder AddThroughputCollector(this IHostApplicationBuilder hostBuilder, string broker, string serviceControlAPI, string errorQueue, string auditQueue, string transportConnectionString, string persistenceType)
+        {
+            var services = hostBuilder.Services;
+            services.AddSingleton(new PlatformData { Broker = broker, ServiceControlAPI = serviceControlAPI, AuditQueue = auditQueue, ErrorQueue = errorQueue, TransportConnectionString = transportConnectionString, PersistenceType = persistenceType });
+            services.AddHostedService<AuditThroughputCollectorHostedService>();
+            services.AddHostedService<BrokerThroughputCollectorHostedService>();
+
```
jpalac, line 17: I'm not injecting the monitoring one here as yet, as we don't know which way that will work - ie the queue or the added persistence.

src/Licensing/AuditThroughputCollector.cs
jpalac: With changing to 3 services, there will be an Ilogger per service

PR #3979, opened 2024-02-22
Address some of the TODOs

danielmarbach:


src/ServiceControl.Audit/Infrastructure/WebApi/HttpResponseExtensions.cs
```diff
@@ -79,9 +79,7 @@ public static void WithPagingLinks(this HttpResponse response, PagingInfo pageIn
                 AddLink(links, pageInfo.Page + 1, "next", path, originalQueryCollection);
             }
 
-            // TODO can this be new StringValues(links.ToArray())) ? we don't know what the separator will be
-            // https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Primitives/src/StringValues.cs#L235
-            response.WithHeader("Link", new StringValues(string.Join(", ", links)));
+            response.WithHeader("Link", new StringValues(links.ToArray()));
```
danielmarbach, line 82: The stringified version of this will also contain a comma but without a space. [This regex should be able to deal with it](https://github.com/Particular/ServiceInsight/blob/master/src/ServiceInsight/ServiceControl/DefaultServiceControl.cs#L724) in ServiceInsight since the space is set to Zero and 1 of the preceding token

mauroservienti (approved):

danielmarbach:
@mauroservienti could you give this another pass? I will then merge it. @bording we can always make further tweaks should your review bring up new insights

mauroservienti (approved):

PR #3986, opened 2024-03-02
Update to NServiceBus 9

bording:
This PR updates ServiceControl to use NServiceBus 9 and uses the MSMQ source packages provided by NSB9 to bring back the MSMQ support.


src/ServiceControl.Monitoring/Hosting/Commands/RunCommand.cs
```diff
@@ -13,8 +13,8 @@ public override async Task Execute(Settings settings)
 
             var hostBuilder = WebApplication.CreateBuilder();
             hostBuilder.AddServiceControlMonitoring((_, __) => Task.CompletedTask, settings, endpointConfiguration);
-            using var app = hostBuilder.Build();
 
+            var app = hostBuilder.Build();
```
danielmarbach, line 17: This is a general comment. Not something we have to address though. I was wondering if it would still be better from a code flow perspective to do using. The app is disposable and from that perspective it would make sense to add a using. Otherwise we are relying on knowledge of implementation details of `RunAsync` which I'm a little bit on the fence but I can see the argument of "yeah but how the code is right now is more aligned with the default templates of MS that also relies on RunAsync doing the dispose"

src/ServiceControl.Monitoring/WebApplicationBuilderExtensions.cs
```diff
@@ -115,7 +111,13 @@ internal static void ConfigureEndpoint(EndpointConfiguration config, Func<ICriti
 
         config.UseSerialization<NewtonsoftJsonSerializer>();
         config.UsePersistence<NonDurablePersistence>();
+
+        var recoverability = config.Recoverability();
+        recoverability.Immediate(c => c.NumberOfRetries(3));
```
danielmarbach, line 116: Do we need to specify the immediate retries and why not simply disable the delayed ones that we don't need? Looking at the change we now effectively went from 5 (default) immediate to 3 and I'm wondering what the implications could be.

src/ServiceControl.Transports.Tests/BasicEndpointSetup.cs
```diff
@@ -19,6 +19,10 @@ public class BasicEndpointSetup : IEndpointSetupTemplate
             endpointConfiguration.UseSerialization<NewtonsoftJsonSerializer>();
             endpointConfiguration.RegisterComponentsAndInheritanceHierarchy(runDescriptor);
 
+            var recoverability = endpointConfiguration.Recoverability();
+            recoverability.Immediate(c => c.NumberOfRetries(3));
```
danielmarbach, line 23: Same comment as above

src/ServiceControl.Monitoring/WebApplicationBuilderExtensions.cs
danielmarbach: Ah you basically aligned it with SC instances I think that is fine.

danielmarbach (approved):

src/ServiceControl/Hosting/Commands/ImportFailedErrorsCommand.cs
```diff
@@ -28,10 +28,10 @@ public override async Task Execute(HostArguments args, Settings settings)
             // TODO: Ideally we would never want to actually bootstrap the web api. Figure out how
             var hostBuilder = WebApplication.CreateBuilder();
             hostBuilder.AddServiceControl(settings, endpointConfiguration, loggingSettings);
-            using var app = hostBuilder.Build();
 
+            using var app = hostBuilder.Build();
```
mauroservienti, line 32: Should we remove the `using` here to keep it aligned with the various other places where we create, start, and stop apps?

danielmarbach: @mauroservienti when you are not using `RunAsync` you have to dispose manually

mauroservienti (approved):

PR #3988, opened 2024-03-06
Adding query by date for audit messages

johnsimons:
This makes it simpler to calculate the throughput based on audit ingestion.
The plan is to call this endpoint from the throughput every hour and read the total count, we don't even need to page the data.


johnsimons:
not required after all

PR #3994, opened 2024-03-07
Acceptance test todos

danielmarbach:


src/ServiceControl.AcceptanceTesting/EndpointConfigurationExtensions.cs
```diff
@@ -1,33 +1,96 @@
...
+
+                foreach (var nestedType in rootType.GetNestedTypes(BindingFlags.Instance | BindingFlags.Static | BindingFlags.Public | BindingFlags.NonPublic).SelectMany(t => GetNestedTypeRecursive(t, builderType)))
+                {
+                    yield return nestedType;
+                }
+            }
         }
 
-        public static void NoDelayedRetries(this EndpointConfiguration configuration)
+        public static async Task DefinePersistence(this EndpointConfiguration config, RunDescriptor runDescriptor, EndpointCustomizationConfiguration endpointCustomizationConfiguration)
```
mauroservienti, line 59: Should this be named `UseInMemoryPersistence`? Because there is no option to configure a different one, right?

mauroservienti (approved):
Only a minor comment. All the rest looks good to go.

src/ServiceControl.AcceptanceTesting/EndpointConfigurationExtensions.cs
danielmarbach: I think currently the method is consistently called `DefinePersistence`. I have only moved those a bit. If you don't mind I would rather not do it now.

mauroservienti (approved):

PR #4004, opened 2024-03-12
Use PeriodicTimer in the async timer

danielmarbach:
This switches the internals of AsyncTimer to using PeriodicTimer which is built-in. To support the due time and to avoid having to change the period of the timer it first delays for the due time and then waits until the period fires. The previous implementation did essentially prolong the execution of the next interval by the time it took the callback to execute. This implementation changes the behavior slightly by checking if the next tick is available, which might already be. If multiple periods are over, those get coalesced into one execution.

> The [PeriodicTimer](https://learn.microsoft.com/en-us/dotnet/api/system.threading.periodictimer?view=net-8.0) behaves like an auto-reset event, in that multiple ticks are coalesced into a single tick if they occur between calls to [WaitForNextTickAsync(CancellationToken)](https://learn.microsoft.com/en-us/dotnet/api/system.threading.periodictimer.waitfornexttickasync?view=net-8.0#system-threading-periodictimer-waitfornexttickasync(system-threading-cancellationtoken)).

Do you see any issues with this potentially firing a bit earlier? It would be nice to be able to rely on the provided period timer.

### DIFF
--- a/src/ServiceControl.Infrastructure/AsyncTimer.cs
+++ b/src/ServiceControl.Infrastructure/AsyncTimer.cs
@@ -1,4 +1,6 @@
-﻿namespace ServiceControl.Infrastructure.BackgroundTasks
+﻿#nullable enable
+
+namespace ServiceControl.Infrastructure.BackgroundTasks
 {
     using System;
     using System.Threading;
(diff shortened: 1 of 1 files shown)


danielmarbach:
Discussed it today during the sync and we have decided that the initial behavior of prolonging the execution by the time it takes to execute the callback is desirable because we are running potentially high intensity IO bound operations.

PR #4007, opened 2024-03-13
Monitoring todo

danielmarbach:


src/ServiceControl/Infrastructure/WebApi/BodyUrlRouteFix.cs
```diff
@@ -1,4 +1,4 @@
-﻿namespace ServiceControl.Infrastructure.OWIN
+namespace ServiceControl.Infrastructure.WebApi
```
danielmarbach, line 1: I think we should consider removing the route URL fixes after some research

mauroservienti (approved):

PR #4022, opened 2024-03-20
Remaining todos

danielmarbach:


src/ServiceControl.Persistence.RavenDB/ErrorMessagesDataStore.cs
```diff
@@ -609,34 +609,26 @@ public async Task RemoveFailedMessageRetryDocument(string uniqueMessageId)
...
-                .AndAlso()
-                .WhereBetween(options => options.LastModified, from.Ticks, to.Ticks)
-                .AndAlso()
-                .WhereEquals(o => o.QueueAddress, queueAddress)
-                .SelectFields<FailedMessage>()
-                .ToQueryable()
-                .TransformToFailedMessageView();
-
-            await using (var ie = await session.Advanced.StreamAsync(query))
+            var query = session
```
danielmarbach, line 615: @mauroservienti you have a bit more knowledge about RavenDB. This whole thing is weird. We basically pulled down the entire FailedMessage to then just use the UniqueMessageId. Then we bring that up to the outer layers which then iterates through those, creates FailedMessageRetry Ids to then delete them 1:1. Horrible. I tried to do a DeleteByQuery but that doesn't work to my knowledge since the collection I query from and have to delete from are different. So I tried to at least materialize less from the server.

As a next small improvement, we could change the `RemoveFailedMessageRetryDocument` accept a number of IDs. I was thinking a bit about streaming IDs back and then the outer layers doing batch deletes per stream, but I wasn't sure of the implications of that whole change, since we would keep the cursor open while we are trying to delete batch deletes. 

This TODO is a bit of a rabbit hole and not even from us :dancing_men:

bording (approved):

PR #4024, opened 2024-03-21
Persistence changes

danielmarbach:
This PR makes the following changes to the persistence seam:

- We can now guard against accidentally restarting the embedded RavenDB server while we are shutting down
- The primary RavenDB persister now uses `IRavenDocumentStoreProvider` model like the audit persister had, which solves the initialization problem that we previously needed the custom `IHostLifetime` implementations to work around.
- Other changes have been made to align persistence concepts between the primary and audit instances, and to better align how the persisters interact with `IHostApplicationBuilder` being used to bootstrap the instances.


mauroservienti (approved):

src/ServiceControl.Audit.Persistence.RavenDB/EmbeddedDatabase.cs
```diff
@@ -28,7 +28,7 @@ public sealed class EmbeddedDatabase : IDisposable
                 isStopping = true;
             });
 
-            // Installer scenarios do not use the host and do not have a lifetime
+            // Test scenarios do not always have a lifetime
```
mauroservienti, line 31: This is the only bit I'm not so happy about. We make this nullable because of tests. We shouldn't have to account for how tests are set up in the production code.

src/ServiceControl.Audit.Persistence.InMemory/InMemoryPersistence.cs
```diff
@@ -17,6 +20,20 @@ public void Configure(IServiceCollection services)
...
+
+    class InMemoryPersistenceInstaller(IHostApplicationLifetime applicationLifetime) : IHostedService
+    {
+        public Task StartAsync(CancellationToken cancellationToken)
+        {
+            applicationLifetime.StopApplication();
+            return Task.CompletedTask;
+        }
+
+        public Task StopAsync(CancellationToken cancellationToken) => Task.CompletedTask;
```
mauroservienti, line 37: It feels off to add an empty installer for the sake of calling `applicationLifetime.StopApplication()` and nothing else

src/ServiceControl.Audit.Persistence.RavenDB/EmbeddedDatabase.cs
danielmarbach: Agreed. I was also thinking it is not ideal but on the other hand we have also other places where we have internal properties and stuff that is only used for testing

src/ServiceControl.Audit.Persistence.InMemory/InMemoryPersistence.cs
danielmarbach: Yeah this is not needed of we switch in the setup bootstrapper to have a using around the host and then start/stop it. I went this approach for now so that we can see the implications of what we discussed today

src/ServiceControl.Audit.Persistence.RavenDB/EmbeddedDatabase.cs
bording: With the latest changes to the PR, all production scenarios have a host available, and I was able to modify the usage in the tests to pass in an `ApplicationLifetime`.

bording (approved):

PR #4025, opened 2024-03-21
Use IHostApplicationLifetime to detect ServiceControl is shutting down and RavenDB doesn't need to be restarted

mauroservienti:
- [ ] Apply the same logic to the Audit instance


danielmarbach:
Pulling it into the other branch

PR #4026, opened 2024-03-21
CheckFreeSpace CustomCheck moved into persister and some aligment

danielmarbach:


src/ServiceControl.Persistence.RavenDB/CustomChecks/CheckMinimumStorageRequiredForIngestion.cs
```diff
@@ -77,11 +69,9 @@ public static void Validate(RavenPersisterSettings settings)
             }
         }
 
-        public const int MinimumStorageLeftRequiredForIngestionDefault = 5;
+        public const int MinimumStorageLeftRequiredForIngestionDefault = 20;
```
danielmarbach, line 72: As far as I understand 5 was never actually the default. Because when you don't set the threshold the default from the settings was applied which was 20

https://github.com/Particular/ServiceControl/pull/4026/files#diff-74de14602613e3840bb68469f19c4929908e0d830e211f05a9b30cb250a72361L426

It doesn't hurt to double check though as a reviewer because I got tired at some point staring at this code.

bording: It's documented as the default being 5: https://docs.particular.net/servicecontrol/creating-config-file

Looking at the two "disk check" custom checks, I think they need to have different default values, otherwise they don't seem to make a lot of sense having both.

It appears that the `DataSpaceRemainingThreshold` value for `CheckFreeDiskSpace` is intended to be the first alarm that disk space is getting low, and then the `MinimumStorageLeftRequiredForIngestion` value for `CheckMinimumStorageRequiredForIngestion` is the critical alarm, where ingestion is actually disabled.

Based on that, it seems like it make sense to keep/fix `MinimumStorageLeftRequiredForIngestion` at 5 and `DataSpaceRemainingThreshold` at 20.

src/ServiceControl.Audit.Persistence.RavenDB/CustomChecks/CheckFreeDiskSpace.cs
```diff
@@ -0,0 +1,84 @@
...
+        public CheckFreeDiskSpace(DatabaseConfiguration databaseConfiguration) : base("ServiceControl.Audit database", "Storage space", TimeSpan.FromMinutes(5))
+        {
+            dataPath = databaseConfiguration.ServerConfiguration.DbPath;
+            percentageThreshold = databaseConfiguration.DataSpaceRemainingThreshold / 100m;
+            Logger.Debug($"Check ServiceControl data drive space remaining custom check starting. Threshold {percentageThreshold:P0}");
+        }
+
+        public override Task<CheckResult> PerformCheck(CancellationToken cancellationToken = default)
+        {
+            var dataPathRoot = Path.GetPathRoot(dataPath);
```
bording, line 23: `dataPath` itself can still be `null` here, when an external RavenDB instance is being used via the `ConnectionString` property on `ServerConfiguration`\`RavenPersisterSettings`.

All 4 of these custom checks should have logic that skips the logic and passes the checks when an external RavenDB server is being used.

It looks like there is a `UseEmbeddedServer` property available on both config types that should directly serve that purpose vs. having to infer it because of a `null` `DbPath`.

Then, if we're embedded, and `DbPath` is `null` that does seem like a throwing scenario.

bording (changes requested):

src/ServiceControl.UnitTests/ApprovalFiles/APIApprovals.PlatformSampleSettings.approved.txt
```diff
@@ -36,6 +36,5 @@
   "MaximumConcurrencyLevel": 10,
   "RetryHistoryDepth": 10,
   "RemoteInstances": [],
-  "DataSpaceRemainingThreshold": 20,
```
bording, line 39: These platform sample tests are supposed to show when configuration changes in order to make sure the the [config files get updated](https://github.com/Particular/Particular.PlatformSample/tree/master/src/Particular.PlatformSample/configs), but this change shows the test isn't all that effective since the setting is still valid but has now disappeared from this approval file...

src/ServiceControl.Audit.Persistence.RavenDB/CustomChecks/CheckFreeDiskSpace.cs
```diff
@@ -0,0 +1,86 @@
...
+            }
+
+            if (!databaseConfiguration.ServerConfiguration.UseEmbeddedServer)
+            {
+                return CheckResult.Pass;
+            }
+
+            if (dataPathRoot == null)
+            {
+                throw new Exception($"Unable to find the root of the data path {dataPathRoot}");
```
bording, line 29: Seems like we need to keep the data path value instead of just storing `dataPathRoot` because this error message is nonsensical now!

Looks like this applies to all 4 checks.

src/ServiceControl/App.config
```diff
@@ -22,7 +22,7 @@ These settings are only here so that we can debug ServiceControl while developin
     <!--<add key="ServiceControl/TransportType" value="SQLServer" />-->
 
     <add key="ServiceControl/PersistenceType" value="RavenDB" />
-    <add key="ServiceControl/DataSpaceRemainingThreshold" value="10" />
+    <add key="ServiceControl/DataSpaceRemainingThreshold" value="20" />
```
bording, line 25: Seems like we could just remove this if we're going to set it to default value?

Looks like don't have this in the other app.config files, so it seems fine to remove.

bording (changes requested):

mauroservienti (approved):
Apart from Brandon's comments, this is good to go.

bording (approved):

PR #4027, opened 2024-03-22
Implement RavenDB persistence for audit throughput

abparticular:


src/Particular.ThroughputCollector.Contracts/ThroughputSource.cs
```diff
@@ -1,6 +1,8 @@
 ﻿namespace Particular.ThroughputCollector.Contracts;
+
 public enum ThroughputSource
 {
+    None,
```
johnsimons, line 5: Is this a valid option, if we create an` EndpointIndentifier`, what scenario would we ever select it?

src/Particular.ThroughputCollector.Persistence.InMemory/InMemoryThroughputDataStore.cs
```diff
@@ -2,57 +2,66 @@
...
-class InMemoryThroughputDataStore : IThroughputDataStore
+class InMemoryThroughputDataStore(PersistenceSettings persistenceSettings) : IThroughputDataStore
 {
-    private readonly List<Endpoint> endpoints;
-    private List<BrokerData> brokerData;
+    private readonly EndpointCollection endpoints = [];
+    private readonly List<BrokerData> brokerData = [];
 
-    public InMemoryThroughputDataStore()
+    public Task<IEnumerable<Endpoint>> GetAllEndpoints(bool includePlatformEndpoints = true, CancellationToken cancellationToken = default)
```
johnsimons, line 16: Should this return `IAsyncEnumerable`?

src/Particular.ThroughputCollector.Persistence.InMemory/InMemoryThroughputDataStore.cs
```diff
@@ -2,57 +2,66 @@
...
-    public async Task<IReadOnlyList<Endpoint>> GetAllEndpoints()
+    public Task<Endpoint?> GetEndpoint(EndpointIdentifier id, CancellationToken cancellationToken = default)
     {
-        return await Task.FromResult(endpoints);
+        endpoints.TryGetValue(id, out var endpoint);
+        return Task.FromResult(endpoint);
     }
 
-    public async Task<Endpoint?> GetEndpointByName(string name, ThroughputSource throughputSource)
+    public Task<IEnumerable<(EndpointIdentifier, Endpoint?)>> GetEndpoints(IEnumerable<EndpointIdentifier> endpointIds, CancellationToken cancellationToken = default)
```
johnsimons, line 31: Should this return IAsyncEnumerable?

src/Particular.ThroughputCollector.Persistence.InMemory/InMemoryThroughputDataStore.cs
```diff
@@ -2,57 +2,66 @@
...
+        var filteredEndpoints = includePlatformEndpoints
+            ? endpoints
+            : endpoints.Where(endpoint => !persistenceSettings.PlatformEndpointNames.Contains(endpoint.Id.Name));
 
-    public Task Setup() => Task.CompletedTask;
+        return Task.FromResult(filteredEndpoints);
+    }
 
-    public async Task<IReadOnlyList<Endpoint>> GetAllEndpoints()
+    public Task<Endpoint?> GetEndpoint(EndpointIdentifier id, CancellationToken cancellationToken = default)
```
johnsimons, line 25: I don't see a reason for the `cancellationToken` to have `default`?

src/Particular.ThroughputCollector.Persistence.InMemory/InMemoryThroughputDataStore.cs
```diff
@@ -114,4 +123,9 @@ public async Task SaveBrokerData(Broker broker, string? scopeType, Dictionary<st
     {
         return await Task.FromResult(brokerData.FirstOrDefault(w => w.Broker == broker));
     }
+
+    class EndpointCollection : KeyedCollection<EndpointIdentifier, Endpoint>
```
johnsimons, line 132: What is this used for?

src/Particular.ThroughputCollector.Persistence.RavenDb/Models/EndpointDocument.cs
```diff
@@ -0,0 +1,16 @@
...
+
+using Particular.ThroughputCollector.Contracts;
+
+class EndpointDocument(EndpointIdentifier id)
+{
+    public EndpointIdentifier EndpointId { get; } = id;
+
+    public string SanitizedName { get; set; } = string.Empty;
+
+    public IList<string> EndpointIndicators { get; } = new List<string>();
```
johnsimons: simplify collection initialisation

src/Particular.ThroughputCollector.Persistence.RavenDb/ThroughputDataStore.cs
```diff
@@ -1,22 +1,131 @@
...
 using Raven.Client.Documents;
 
-class ThroughputDataStore(IDocumentStore store, DatabaseConfiguration databaseConfiguration) : IThroughputDataStore
+class ThroughputDataStore(
+    IDocumentStore store,
+    PersistenceSettings persistenceSettings,
+    DatabaseConfiguration databaseConfiguration) : IThroughputDataStore
 {
-    public async Task<IReadOnlyList<Endpoint>> GetAllEndpoints()
+    const string ThroughputTimeSeriesName = "INC: throughput data";
```
johnsimons, line 12: what does "INC" mean?

src/Particular.ThroughputCollector.Persistence.RavenDb/ThroughputDataStore.cs
```diff
@@ -1,22 +1,131 @@
...
+    }
+
+    public async Task RecordEndpointThroughput(EndpointIdentifier id, IEnumerable<EndpointThroughput> throughput, CancellationToken cancellationToken = default)
+    {
+        if (!throughput.Any())
+        {
+            return;
+        }
+
+        using var session = store.OpenAsyncSession("throughput");
```
johnsimons, line 107: So I assume we can name the sessions, for some kind of telemetry purposes?
Should the session name include more context, eg "Throughput/RecordEndpointThroughput"

src/Particular.ThroughputCollector/AuditThroughput/AuditCommands.cs
```diff
@@ -2,17 +2,21 @@
...
+        public static readonly SemanticVersion MinAuditCountsVersion = new(4, 29, 0);
         // Want 2d audit retention so we get one complete UTC day no matter what time it is.
-        public static Func<RemoteInstanceInformation, bool> ValidRemoteInstances = r => r.SemVer?.Version >= MinAuditCountsVersion && r.Retention >= TimeSpan.FromDays(2);
+        public static Func<RemoteInstanceInformation, bool> ValidRemoteInstances = r =>
+            r.SemanticVersion != null &&
+            r.SemanticVersion >= MinAuditCountsVersion &&
+            r.Retention >= TimeSpan.FromDays(2);
 
-        public static async Task<ServiceControlEndpoint[]> GetKnownEndpoints(IEndpointsApi endpointsApi)
+        public static async Task<IEnumerable<ServiceControlEndpoint>> GetKnownEndpoints(IEndpointsApi endpointsApi)
```
johnsimons, line 19: Use `IAsyncEnumerable`

src/Particular.ThroughputCollector/AuditThroughput/AuditCommands.cs
```diff
@@ -26,24 +30,20 @@ public static async Task<ServiceControlEndpoint[]> GetKnownEndpoints(IEndpointsA
...
-            })
-            .ToArray();
+            });
 
-            return scEndpoints ?? Array.Empty<ServiceControlEndpoint>();
+            return scEndpoints ?? [];
         }
 
-        public static async Task<List<AuditCount>> GetAuditCountForEndpoint(IAuditCountApi auditCountApi, string endpointUrlName)
+        public static async Task<IEnumerable<AuditCount>> GetAuditCountForEndpoint(IAuditCountApi auditCountApi, string endpointUrlName)
```
johnsimons, line 38: Use `IAsyncEnumerable`

src/Particular.ThroughputCollector/AuditThroughput/AuditCommands.cs
```diff
@@ -26,24 +30,20 @@ public static async Task<ServiceControlEndpoint[]> GetKnownEndpoints(IEndpointsA
...
         {
-            return (await auditCountApi.GetEndpointAuditCounts(endpointUrlName)).Select(s =>
-            {
-                return new AuditCount { Count = s.Count, UtcDate = DateOnly.FromDateTime(s.UtcDate) };
-            }).ToList();
+            return (await auditCountApi.GetEndpointAuditCounts(endpointUrlName)).Select(s => new AuditCount { Count = s.Count, UtcDate = DateOnly.FromDateTime(s.UtcDate) });
         }
 
-        public static async Task<List<RemoteInstanceInformation>> GetAuditRemotes(IConfigurationApi configurationApi)
+        public static async Task<List<RemoteInstanceInformation>> GetAuditRemotes(IConfigurationApi configurationApi, CancellationToken cancellationToken = default)
```
johnsimons, line 43: Use `IAsyncEnumerable`

src/Particular.ThroughputCollector/AuditThroughput/AuditThroughputCollectorHostedService.cs
```diff
@@ -1,104 +1,120 @@
...
+    protected override Task ExecuteAsync(CancellationToken stoppingToken)
     {
-        protected override Task ExecuteAsync(CancellationToken stoppingToken)
-        {
-            logger.LogInformation("Starting AuditThroughputCollector Service");
+        logger.LogInformation("Starting AuditThroughputCollector Service");
 
-            backgroundTimer = timeProvider.CreateTimer(async _ => await GatherThroughput(), null, TimeSpan.FromSeconds(30), TimeSpan.FromDays(1));
-            stoppingToken.Register(_ => backgroundTimer?.Dispose(), null);
+        backgroundTimer = timeProvider.CreateTimer(async _ => await GatherThroughput(stoppingToken), null, TimeSpan.FromSeconds(30), TimeSpan.FromDays(1));
```
johnsimons, line 26: I think this needs more work to handle `CancellationTokeException`

src/Particular.ThroughputCollector/BrokerThroughput/BrokerThroughputCollectorHostedService.cs
```diff
@@ -61,20 +61,17 @@ await foreach (var queueName in throughputQuery.GetQueueNames(stoppingToken))
 
         async Task Exec(IQueueName queueName, DateOnly startDate)
         {
-            var endpoint = await dataStore.GetEndpointByName(queueName.QueueName, ThroughputSource.Broker);
+            var endpointId = new EndpointIdentifier();
```
johnsimons, line 64: This looks wrong, don't we need to set the name and source?

johnsimons: Actually, how come we allow to new up a `EndpointIdentifier` without specifying name and source?

src/Particular.ThroughputCollector/IThroughputCollector.cs
```diff
@@ -7,7 +7,7 @@ public interface IThroughputCollector
         Task<List<EndpointThroughputSummary>> GetThroughputSummary();
         Task UpdateUserIndicatorsOnEndpoints(List<EndpointThroughputSummary> endpointsThroughputSummary);
         Task<BrokerSettings> GetBrokerSettingsInformation();
-        Task<ConnectionTestResults> TestConnectionSettings();
+        Task<ConnectionTestResults> TestConnectionSettings(CancellationToken cancellationToken = default);
```
johnsimons, line 10: Well, all these need cancellation tokens, I raised it in https://github.com/orgs/Particular/projects/197/views/1?pane=issue&itemId=57288709

src/Particular.ThroughputCollector/WebApi/ThroughputController.cs
```diff
@@ -50,9 +50,9 @@ public async Task<BrokerSettings> GetThroughputBrokerSettingsInformation()
 
         [Route("throughput/settings/test")]
         [HttpGet]
-        public async Task<ConnectionTestResults> TestThroughputConnectionSettings()
+        public async Task<ConnectionTestResults> TestThroughputConnectionSettings(CancellationToken cancellationToken)
         {
-            return await throughputCollector.TestConnectionSettings();
+            return await throughputCollector.TestConnectionSettings(cancellationToken);
```
johnsimons, line 55: nice

johnsimons (changes requested):
I added a few comments.
In summary, and this is not so much about this specific PR but overall.
- We should not have `CancellationToken cancellationToken = default` in our method signatures. The entry points for out APIs are either a Controller/Action or a BackgroundService, which both have CancellationTokens
- Any method that currently returns `Task<List<...>>` or `Task<IEnumerable<...>>` should change to return `IAsyncEnumerable<....>` this allows more efficient streaming

src/Particular.ThroughputCollector.Persistence.RavenDb/Models/EndpointDocument.cs
abparticular: Yes, fixed it when build failed.  Not sure why I didn't get an error on my dev environment.

src/Particular.ThroughputCollector.Persistence.RavenDb/ThroughputDataStore.cs
```diff
@@ -1,22 +1,131 @@
...
+                endpoint.LastCollectedDate = DateOnly.FromDateTime(timeSeries.LastOrDefault()?.Timestamp ?? DateTime.MinValue);
+            }
+
+            endpoints.Add((id, endpoint));
+        }
+
+        return endpoints;
+    }
+
+    public async Task SaveEndpoint(Endpoint endpoint, CancellationToken cancellationToken = default)
```
jpalac, line 90: If the purpose of this method is to only add an endpoint (as opposed to add or update) then could we rename to AddEndpoint?

src/Particular.ThroughputCollector.Persistence.RavenDb/ThroughputDataStore.cs
```diff
@@ -1,22 +1,131 @@
...
+    {
+        var document = endpoint.ToEndpointDocument();
+
+        using var session = store.OpenAsyncSession("throughput");
+
+        await session.StoreAsync(document, document.GenerateDocumentId(), cancellationToken);
```
jpalac, line 96: So what happens here if the endpoint already exists in the db? I saw that in the InMemory persistence we do a check and remove it if it does.

abparticular:
> I added a few comments. In summary, and this is not so much about this specific PR but overall.
> 
> * We should not have `CancellationToken cancellationToken = default` in our method signatures. The entry points for out APIs are either a Controller/Action or a BackgroundService, which both have CancellationTokens

I'll confess I didn't consider each case where I was adding the tokens.  I will review them.

However, I don't agree the Controllers and BackgroundService are the only entry points.  I know that there are some test contexts/cases that call methods without an available CancellationToken.

I would argue providing a default token for publicly available methods is better than forcing callers to provide one.

src/Particular.ThroughputCollector/AuditThroughput/AuditThroughputCollectorHostedService.cs
```diff
@@ -1,104 +1,120 @@
...
+                if (endpoint == null)
                 {
-                    if (!await ThroughputRecordedForYesterday(endpoint.Name, utcYesterday))
-                    {
-                        var auditCounts = await AuditCommands.GetAuditCountForEndpoint(auditCountApi, endpoint.UrlName);
-                        //for each endpoint record the audit count for the day we are currently doing as well as any others that are available
-                        await dataStore.RecordEndpointThroughput(SCEndpointToEndpoint(endpoint, auditCounts));
-                    }
+                    endpoint = ConvertToEndpoint(knownEndpointsLookup[endpointId], auditCounts);
+                    await dataStore.SaveEndpoint(endpoint, cancellationToken);
```
jpalac, line 65: Why do we need to do this Save here, if the RecordEndpointThroughput method which is called below will add the endpoint if it doesn't exist? I think we should only have one of the methods on the datastore - it's unnecessarily confusing having the two.

src/Particular.ThroughputCollector.Persistence.RavenDb/ThroughputDataStore.cs
johnsimons: Sorry, so what is "throughput" argument passed?

src/Particular.ThroughputCollector/AuditThroughput/AuditCommands.cs
johnsimons: Yes, we need to refactor `ScatterGater` to pass cancelationtokens and IAsyncEnumeration

johnsimons:
> > I added a few comments. In summary, and this is not so much about this specific PR but overall.
> > 
> > * We should not have `CancellationToken cancellationToken = default` in our method signatures. The entry points for out APIs are either a Controller/Action or a BackgroundService, which both have CancellationTokens
> 
> I'll confess I didn't consider each case where I was adding the tokens. I will review them.
> 
> However, I don't agree the Controllers and BackgroundService are the only entry points. I know that there are some test contexts/cases that call methods without an available CancellationToken.
> 
> I would argue providing a default token for publicly available methods is better than forcing callers to provide one. Let the caller decide whether or not it needs to be able to cancel the operation.

I guess I want to ensure callers in the context of ServiceControl are not forgetting to do it.
Also, Tests, also have cancellation tokens

src/Particular.ThroughputCollector.Persistence.InMemory/InMemoryThroughputDataStore.cs
johnsimons: https://ravendb.net/docs/article-page/5.4/csharp/client-api/session/querying/how-to-stream-query-results

abparticular:
Oh, I forgot to add we have code analyzers that raise errors when an available CancellationToken is not passed into an async method that accepts one.

src/Particular.ThroughputCollector.Persistence.InMemory/InMemoryThroughputDataStore.cs
abparticular: Lol thanks.  I am aware of the interface, but it kept throwing exceptions when I tried to use it.  I will be creating a minimal repro and raising an issue with RavenDB.

src/Particular.ThroughputCollector/ThroughputCollectorHostBuilderExtensions.cs
```diff
@@ -1,86 +1,97 @@
...
+    }
+
+    public static IHostApplicationBuilder AddThroughputCollectorPersistence(this IHostApplicationBuilder hostBuilder, string persistenceType, string? errorQueue = null, string? serviceControlQueue = null)
+    {
+        hostBuilder.Services.AddPersistence(persistenceType, settings =>
+        {
+            if (!string.IsNullOrEmpty(errorQueue))
             {
-                serviceControlAPI = serviceControlAPI.Replace("http://0.0.0.0", "http://localhost");
+                settings.PlatformEndpointNames.Add(errorQueue);
```
jpalac, line 86: Why are we storing the platform queue names twice? on here as well as on the throughputSettings class.

abparticular:
> Also, Tests, also have cancellation tokens

E.g. [EndpointTests.cs](https://github.com/Particular/ServiceControl/blob/TF-3601/src/Particular.ThroughputCollector.Persistence.Tests/EndpointsTests.cs) are all calling datastore methods, but don't have an ambient CancellationToken.

src/Particular.ThroughputCollector.Persistence.RavenDb/ThroughputDataStore.cs
```diff
@@ -1,22 +1,131 @@
...
+
+        var endpoint = document?.ToEndpoint();
+        if (endpoint != null)
+        {
+            var timeSeries = await session
+                .IncrementalTimeSeriesFor(documentId, ThroughputTimeSeriesName)
+                .GetAsync(token: cancellationToken);
+
+            endpoint.LastCollectedDate = DateOnly.FromDateTime(timeSeries.LastOrDefault()?.Timestamp ?? DateTime.MinValue);
+        }
```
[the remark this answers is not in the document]
abparticular: Yes the parameter on line 37 tells RavenDb to also load the associated time series for the document.  When it queries it "again" on line 43, it is pulling it out of the session and not actually round tripping to the DB.

src/Particular.ThroughputCollector.Persistence.RavenDb/ThroughputDataStore.cs
abparticular: "Add or overwrite" would be a better description than "add or update".

src/Particular.ThroughputCollector.Persistence.RavenDb/ThroughputDataStore.cs
abparticular: StoreAsync will overwrite it.

src/Particular.ThroughputCollector.Persistence.RavenDb/ThroughputDataStore.cs
abparticular: It specifies which database to use.  The default database in the document store is the SC one.

src/Particular.ThroughputCollector/AuditThroughput/AuditThroughputCollectorHostedService.cs
abparticular: My plan is to ultimately remove the side effect of creating the endpoint in RecordEndpointThroughput.  The endpoint has other parameters (e.g. sanitized name, indicators, etc) which do not make sense to pass into RecordEndpointThroughput.

src/Particular.ThroughputCollector/ThroughputCollectorHostBuilderExtensions.cs
abparticular: I need the queue names in the persistence assemblies to do DB side filtering of the queues, and `ThroughputSettings` is not available.

I left them on ThroughputSettings because `BrokerThroughputCollectorHostedService` needs them as well.

src/Particular.ThroughputCollector.Persistence.RavenDb/ThroughputDataStore.cs
jpalac: It's only doing an add, so AddEndpoint would make sense.

src/Particular.ThroughputCollector/AuditThroughput/AuditThroughputCollectorHostedService.cs
jpalac: AppendEndpointThroughput also does the adding of an endpoint if it doesn't exist. I would like to clean all that up so that it's very clear as to what is happening in every method, without making too many db calls.
I'd be happy with just the 2 methods:

- CreateEndpointThroughput (which only creates it ensuring that it doesn't overwrite any existing ones)
- AppendEndpointThroughput (which appends or creates it if it doesn't exist)

I don't have an issue with both methods also adding the endpoint itself if it doesn't exist - it is part of the one record in the db afterall.

johnsimons (approved):

src/Particular.ThroughputCollector/ThroughputCollectorHostBuilderExtensions.cs
jpalac: I'm thinking perhaps we don't save those queues at ? I don't think that @johnsimons is doing that on the broker side. That would mean we wouldn't need these on the persistence side.

abparticular: I'm actually thinking the opposite in that they should be included in the summary as well, just marked as a platform queue (for lack of a better term).  Although how we treat platform queues is probably better discussed as a separate topic outside this PR.

johnsimons: It seems we need to move `ThroughputSettings` to `Particular.ThroughputCollector.Contracts`

jpalac: @abparticular that has been moved. The existing tool removes those error etc queues and does not include them in the report (for most transports - it's not consistent).

PR #4029, opened 2024-03-23
Move persistence config into a dedicated AddServiceControlAuditInstallers extension

mauroservienti:
The same could be done for the primary and monitoring instances.

In the same `SetupCommand`, we create queues, if requested, without starting the Host. It might be worth investigating if it makes sense to align queue creation with the storage setup and make it so that's controlled by the host.

### DIFF
--- a/src/ServiceControl.Audit/Infrastructure/Hosting/Commands/SetupCommand.cs
+++ b/src/ServiceControl.Audit/Infrastructure/Hosting/Commands/SetupCommand.cs
@@ -8,6 +8,7 @@
     using Persistence;
     using Settings;
     using Transports;
+    using WebApi;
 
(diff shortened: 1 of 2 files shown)


danielmarbach:
One of the things I was thinking about lately is that actually there is no persistence installer really. All it does is initialize the persistence lifecycle. This happens during "installation" time as well as during "runtime". Technically with only RavenDB and "InMemory" be supported the entire concept of installation for the persistence could go away. 

FYI I'm sharing my thoughts here but that doesn't mean I'm opposed to this change. I think further encapsulating the installer (if we keep it) is a good thing.

danielmarbach:
Pulling this in for now into the other branch

PR #4031, opened 2024-03-26
moving things to endpoint

johnsimons:


src/ServiceControl.Transports/IThroughputQuery.cs
```diff
@@ -35,11 +37,11 @@ public class QueueThroughput
 {
     public DateOnly DateUTC { get; set; }
     public long TotalThroughput { get; set; }
-    public string? Scope { get; set; }
-    public string[] EndpointIndicators { get; set; } = [];
 }
 
 public interface IQueueName
```
abparticular, line 42: I think this should be renamed to IQueueDetails or something - it's not just the name any more

abparticular (approved):

src/ServiceControl.Transports/IThroughputQuery.cs
johnsimons: good point

johnsimons: Done

PR #4032, opened 2024-03-26
Use AssemblyLoadContext to load plugin assemblies

bording:
This PR introduces a `PluginAssemblyLoadContext` which is used to load the transport and persister plugin assemblies. This approach ensures that any native libraries that are part of the dependencies of those projects can be properly resolved.

With this in place, the assembly resolving code called wired up to the `AssemblyLoadContext.Default.Resolving` event is no longer needed, so it has been removed.

The `Main` methods have been further simplified to use top-level statements.


src/ServiceControl.Audit/Infrastructure/Settings/Settings.cs
```diff
@@ -42,6 +47,20 @@ public Settings(string serviceName = null, string transportType = null, string p
             ServiceControlQueueAddress = SettingsReader.Read<string>(SettingsRootNamespace, "ServiceControlQueueAddress");
             TimeToRestartAuditIngestionAfterFailure = GetTimeToRestartAuditIngestionAfterFailure();
             EnableFullTextSearchOnBodies = SettingsReader.Read(SettingsRootNamespace, "EnableFullTextSearchOnBodies", true);
+
+            TransportResolver = manifest =>
```
danielmarbach, line 51: Nit: can we make this a static lambda?

src/ServiceControl.Audit/Infrastructure/Settings/Settings.cs
```diff
@@ -42,6 +47,20 @@ public Settings(string serviceName = null, string transportType = null, string p
...
+            {
+                var assemblyPath = Path.Combine(manifest.Location, $"{manifest.AssemblyName}.dll");
+                var loadContext = new PluginAssemblyLoadContext(assemblyPath);
+                return loadContext;
+            };
+
+            PersistenceResolver = manifest =>
```
danielmarbach, line 58: Nit: can we make this a static lambda?

src/ServiceControl.Infrastructure/PluginAssemblyLoadContext.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.Reflection;
+using System.Runtime.Loader;
+
+public class PluginAssemblyLoadContext(string assemblyPath) : AssemblyLoadContext(assemblyPath)
+{
+    readonly AssemblyDependencyResolver resolver = new(assemblyPath);
+
+    protected override Assembly Load(AssemblyName assemblyName)
+    {
+        if (Default.Assemblies.Any(a => a.FullName == assemblyName.FullName))
```
danielmarbach, line 13: I wasn't sure if this is really something we should focus on. Still mentioning for completness. Do you have an idea how expensive that any is vs have a more efficient data structure like a hashset?

src/ServiceControl.Monitoring/Settings.cs
```diff
@@ -54,9 +56,10 @@ public ITransportCustomization LoadTransportCustomization()
         {
             try
             {
-                TransportType = TransportManifestLibrary.Find(TransportType);
-
-                var customizationType = Type.GetType(TransportType, true);
+                var transportManifest = TransportManifestLibrary.Find(TransportType);
+                var assemblyPath = Path.Combine(transportManifest.Location, $"{transportManifest.AssemblyName}.dll");
+                var loadContext = new PluginAssemblyLoadContext(assemblyPath);
```
danielmarbach, line 61: Here we don't need the functions? Would it be worthwhile to still introduce those for similarity reasons?

src/ServiceControl.Transports.ASQ/transport.manifest
```diff
@@ -1,9 +1,9 @@
 ﻿{
-  "Version": "1.0.0",
   "Definitions": [
     {
       "Name": "AzureStorageQueue",
       "DisplayName": "Azure Storage Queues",
+      "AssemblyName": "ServiceControl.Transports.ASQ",
       "TypeName": "ServiceControl.Transports.ASQ.ASQTransportCustomization, ServiceControl.Transports.ASQ",
```
danielmarbach, line 7: I know we can't change this now because of backward compatibility and in general I also like having the assembly name separate. What looks awkward now is that we have the assembly name and the type name contains it too.

src/ServiceControl/Infrastructure/Settings/Settings.cs
```diff
@@ -58,8 +62,18 @@ public class Settings
             DataSpaceRemainingThreshold = GetDataSpaceRemainingThreshold();
             TimeToRestartErrorIngestionAfterFailure = GetTimeToRestartErrorIngestionAfterFailure();
             DisableExternalIntegrationsPublishing = SettingsReader.Read(SettingsRootNamespace, "DisableExternalIntegrationsPublishing", false);
+
+            PersistenceResolver = manifest =>
```
danielmarbach, line 66: Nit: Can we make this a static lambda?

Would it be worthwhile to align the transport loading with a function too?

danielmarbach (commented):
Few preliminary thoughts (I know this is draft but I wanted to share it before I'm off off)

src/ServiceControl/Infrastructure/Settings/Settings.cs
bording: > Would it be worthwhile to align the transport loading with a function too?

Yes, that was the plan! First, I wanted to see what tests were failing and get those working, and then go back and clean things up.

src/ServiceControl.Infrastructure/PluginAssemblyLoadContext.cs
bording: I'm not sure how I could have another data structure there since `Default.Assemblies` could contain more items each time `Load` is called. It doesn't seem to make much sense to me to populate a new HashSet each call.

I could swap the `Any` with a foreach though.

src/ServiceControl.Monitoring/Settings.cs
bording: Yep, that was the next step in the changes I wanted to make!

src/ServiceControl.Transports.ASQ/transport.manifest
bording: I had looked into this a bit, and I believe we can change this if we want without migration / back compat concerns. However, I have currently left it in because the assembly portion is actually required to be there for the `Type.GetType` calls to be able to properly load the types.

If we wanted to remove it from the manifest, we'd then have to reassemble at runtime to pass into the `Type.GetType` calls. Not sure that's worth doing?

src/ServiceControl.AcceptanceTesting/InfrastructureConfig/ConfigureEndpointLearningTransport.cs
```diff
@@ -46,6 +45,7 @@ public Task Cleanup()
         }
 
         public string Name => "Learning";
```
PhilBastian, line 47: since these names are also used in tests, is there any reason to not make them static constants or an enum in order to prevent typos?

src/ServiceControl.Audit.Persistence/PersistenceManifest.cs
```diff
@@ -78,28 +78,16 @@ static string GetAssemblyDirectory()
...
-        public static string GetPersistenceFolder(string persistenceType)
+        public static PersistenceManifest Find(string persistenceType)
         {
             if (persistenceType == null)
             {
                 throw new Exception("No persistenceType has been configured. Either provide a Type or Name in the PersistenceType setting.");
             }
 
-            var persistenceManifestDefinition = PersistenceManifests.FirstOrDefault(w => w.IsMatch(persistenceType));
+            var persistenceManifest = PersistenceManifests.FirstOrDefault(w => w.IsMatch(persistenceType));
```
PhilBastian, line 88:
```suggestion
            return PersistenceManifests.FirstOrDefault(w => w.IsMatch(persistenceType));
```

src/ServiceControl.Infrastructure/PluginAssemblyLoadContext.cs
```diff
@@ -0,0 +1,41 @@
...
+
+    protected override Assembly Load(AssemblyName assemblyName)
+    {
+        foreach (var assembly in Default.Assemblies)
+        {
+            if (assembly.FullName == assemblyName.FullName)
+            {
+                return null;
+            }
+        }
```
PhilBastian, line 18:
```suggestion
        if (Default.Assemblies.Any(assembly => assembly.FullName == assemblyName.FullName)) return null;
```

PhilBastian: my assumption from reading this code is that if the assembly full name is in Defauly.Assemblies then it's already loaded and therefore doesn't need to be loaded here? Perhaps provide a comment to this effect

PhilBastian (approved):

mauroservienti (approved):

src/ServiceControl.AcceptanceTesting/InfrastructureConfig/ConfigureEndpointLearningTransport.cs
bording: Mostly just that I didn't see a lot of benefit in doing that vs just making the changes. If there was a typo, it would be pretty immediately obvious since it would no longer match the value in a manifest file, and the the persister or transport would not be found, and the tests would fail in a rather dramatic fashion.

src/ServiceControl.Transports.ASQ/transport.manifest
bording: I did also move where the check occurs, so that it only runs if the plugin folder actually has a copy of the assembly also. That makes it have to check the default context much less often.

src/ServiceControl.Audit.Persistence/PersistenceManifest.cs
bording: Took a look, and would prefer not to make this change since there are two `PersistenceManifest` classes and a `TransportManifest` class that all have the same basic code shape. I wouldn't want to change `TransportManifest` in a similar way, so I'd prefer to keep them all the same.

Also, having the local there does make it easier to see the value when debugging, so in general I don't mind having it there.

PR #4081, opened 2024-04-15
Remove Newtonsoft.Json as much as possible

danielmarbach:
This PR attempts to remove Newtonsoft.Json as much as possible and switches over to System.Text.Json. I have analyzed a large portion of the message types and they are using mostly very simple types that should be 1:1 compatible with System.Text.Json. In the cases when ServiceControl doesn't send messages to itself the types are locally defined and we assume a standard configured newtonsoft json serializer. So we should be able to assume we are not doing advanced tricks that [would not be compatible with System.Text.Json](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/migrate-from-newtonsoft?pivots=dotnet-8-0#differences-in-default-behavior).

RavenDB heavily relies on Newtonsoft.Json so I left it in the persistence seam.


src/ServiceControlInstaller.Engine/Configuration/ServiceControl/RemoteInstanceConverter.cs
```diff
@@ -1,35 +1,28 @@
...
 
     public class RemoteInstanceSetting
     {
+        [JsonPropertyName("api_uri")]
         public string ApiUri { get; set; }
+
+        [JsonPropertyName("queue_address")]
         public string QueueAddress { get; set; }
     }
 }
```
mauroservienti, line 28: What's the reason to prefer attributes over setting snake casing in the serializer options?

danielmarbach: The engine is NET6. The System.Text.Json version in .NET6 doesn't have the snake casing naming policy. I could have taken a dependency to system.text.json v8 directly but I figured this would be overkill

mauroservienti: Got it, thanks!

bording (approved):

PR #4082, opened 2024-04-16
Load Throughput Collector's persistence assembly into same AssemblyLoadContext as primary instance persistence seam

abparticular:


src/ServiceControl/Licensing/ThroughputComponent.cs
```diff
@@ -35,10 +37,47 @@ public override void Setup(Settings settings, IComponentInstallationContext cont
...
+        AddPersistence(settings, hostBuilder.Services);
+        context.RegisterInstallationTask(serviceProvider => serviceProvider.GetRequiredService<ThroughputPersistence.IPersistenceInstaller>().Install());
+    }
+
+    public void AddPersistence(Settings settings, IServiceCollection services)
+    {
+        if (!addedDefaultContextAssemblyResolver)
+        {
+            // Only add the assembly resolver if this is the first time the method has been called
+            AssemblyLoadContext.Default.Resolving += ResolvePersistenceAssemblyInDefaultContext;
```
johnsimons, line 49: This is re-introducing what Brandon had to remove because this does not deal with unmanaged DLLs.
So it works now because our datastorage is managed, but once we have unmanaged DLLs, it won't.

abparticular: No, this is only necessary because some tests specify different settings and load assemblies into the Default AssemblyLoadContext.  For those tests, this resolver is required to locate the persistence assemblies.

PR #4091, opened 2024-04-17
Moved assembly loading code into PersistenceFactory

jpalac:
Moved the loading code out of the component to make it cleaner, and this way other components can easier load their persistence into the service.

### DIFF
--- a/src/ServiceControl/Persistence/PersistenceFactory.cs
+++ b/src/ServiceControl/Persistence/PersistenceFactory.cs
@@ -55,7 +55,7 @@ namespace ServiceControl.Persistence
                 : loadContext;
         }
 
-        public static object LoadComponentPersistence(Settings settings, string componentPersistenceAssemblyPath, string componentPersistenceType)
+        public static T LoadComponentPersistence<T>(Settings settings, string componentPersistenceAssemblyPath, string componentPersistenceType)
         {
             if (!loadedComponentPersisters.Contains(componentPersistenceType))
             {
@@ -83,7 +83,12 @@ namespace ServiceControl.Persistence
             ?? throw new InvalidOperationException($"Could not load type '{componentPersistenceType}' for requested persistence type '{hostPersistenceManifest.Name}' from '{loadContext.Name}' load context");
 
             loadedComponentPersisters.Add(componentPersistenceType);
-            return Activator.CreateInstance(type);
+            if (Activator.CreateInstance(type) is not T typedComponentPersistence)
+            {
+                throw new InvalidOperationException($"{componentPersistenceType} is not of type {typeof(T).Name}");
+            }
+
+            return typedComponentPersistence;
         }
     }
 }
--- a/src/ServiceControl/Persistence/PersistenceFactory.cs
+++ b/src/ServiceControl/Persistence/PersistenceFactory.cs
@@ -1,13 +1,18 @@
 namespace ServiceControl.Persistence
 {
     using System;
+    using System.Collections.Generic;
+    using System.IO;
     using System.Linq;
+    using System.Reflection;
     using System.Runtime.Loader;
     using ServiceBus.Management.Infrastructure.Settings;
     using ServiceControl.Infrastructure;
 
     static class PersistenceFactory
     {
+        static List<string> loadedComponentPersisters = [];
+
         public static IPersistence Create(Settings settings, bool maintenanceMode = false)
         {
             var persistenceConfiguration = CreatePersistenceConfiguration(settings);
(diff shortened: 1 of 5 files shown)


src/ServiceControl/Persistence/PersistenceFactory.cs
```diff
@@ -49,5 +54,36 @@ public static AssemblyLoadContext DetermineLoadContext(Settings settings, string
                 ? new PluginAssemblyLoadContext(assemblyPath)
                 : loadContext;
         }
+
+        public static object LoadComponentPersistence(Settings settings, string componentPersistenceAssemblyPath, string componentPersistenceType)
```
abparticular, line 58:
```suggestion
        public static T LoadComponentPersistence<T>(Settings settings, string componentPersistenceAssemblyPath, string componentPersistenceType)
```

src/ServiceControl/Persistence/PersistenceFactory.cs
```diff
@@ -49,5 +54,36 @@ public static AssemblyLoadContext DetermineLoadContext(Settings settings, string
...
+                {
+                    pluginLoadContext.AddResolver(componentPersistenceAssemblyPath);
+                }
+            }
+
+            var type = Type.GetType(componentPersistenceType, loadContext.LoadFromAssemblyName, null, true)
+            ?? throw new InvalidOperationException($"Could not load type '{componentPersistenceType}' for requested persistence type '{hostPersistenceManifest.Name}' from '{loadContext.Name}' load context");
+
+            loadedComponentPersisters.Add(componentPersistenceType);
+            return Activator.CreateInstance(type);
```
abparticular, line 86:
```suggestion
            if (Activator.CreateInstance(type) is not T typedComponentPersistence)
            {
                throw new InvalidOperationException($"{componentPersistenceType} is not of type {typeof(T).Name}");
            }

            return typedComponentPersistence;
```

src/ServiceControl/Persistence/PersistenceFactory.cs
```diff
@@ -1,13 +1,18 @@
...
+    using System.IO;
     using System.Linq;
+    using System.Reflection;
     using System.Runtime.Loader;
     using ServiceBus.Management.Infrastructure.Settings;
     using ServiceControl.Infrastructure;
 
     static class PersistenceFactory
     {
+        static List<string> loadedComponentPersisters = [];
```
abparticular, line 14: I'm not sure why this needs to be a list? Isn't a single bool that tracks whether the resolver got added enough? I don't think we need to add it for each type.

src/ServiceControl/Persistence/PersistenceFactory.cs
```diff
@@ -49,5 +54,36 @@ public static AssemblyLoadContext DetermineLoadContext(Settings settings, string
...
         }
+
+        public static object LoadComponentPersistence(Settings settings, string componentPersistenceAssemblyPath, string componentPersistenceType)
+        {
+            if (!loadedComponentPersisters.Contains(componentPersistenceType))
+            {
+                AssemblyLoadContext.Default.Resolving += ResolvePersistenceAssemblyInDefaultContext;
+            }
+
+            Assembly ResolvePersistenceAssemblyInDefaultContext(AssemblyLoadContext context, AssemblyName name) =>
```
abparticular, line 65: Hmm I just realized looking at this that I actually have a bug here.  It actually should be something like

```suggestion
            static Assembly ResolvePersistenceAssemblyInDefaultContext(AssemblyLoadContext context, AssemblyName name)
            {
                var path = Path.Combine(<componentPersistenceLocation>, $"{name.Name}.dll");

                return File.Exists(path)
                ? context.LoadFromAssemblyPath(path)
                : null;
            }
```

abparticular (commented):
Yes I think it's better.  Just have a few tweaks.

src/ServiceControl/Persistence/PersistenceFactory.cs
jpalac: Wouldn't each component type have its own resolver with different params?

src/ServiceControl/Persistence/PersistenceFactory.cs
abparticular: Because it is returning the component's persistence assembly regardless of which assembly it was actually trying to resolve (given by the AssemblyName parameter).  

This will fail if it is trying to locate a dependency of our persistence assembly, since we would actually return the persistence assembly itself.  Also while debugging I have seen instances where it is trying to resolve assemblies that are unrelated to our persistence seam, and in those cases we should be returning null, instead of loading our persistence assembly.

src/ServiceControl/Persistence/PersistenceFactory.cs
abparticular: Yes I think you are right, although not the path, but the location.  In which case I would use a Hashset instead of a List.  I would also confirm the closure is capturing the correct value.

jpalac:
@abparticular happy for you to merge this into your branch and make the adjustments you mentioned.

abparticular (approved):

PR #4090, opened 2024-04-17
Reproduction of issue, not being able to find type

johnsimons:
@bording I am having an issue as part of [#tf-cs3601-enable-the-platform-report-on-endpoint](https://particularsoftware.slack.com/archives/C06CUKQ3668) that I believe is related to the latest changes to do with `AssemblyLoadContext`.

I have prepared this reproduction with hopefully the minimum changes required to replicate the problem.

I am adding a "new feature" to ServiceControl in this reproduction.
This feature includes a Controller/Action and a datastore that is implemented in RavenDB (though it is just hard coded).

The interface for this new datastore is declared in a different assembly. This is important because this is what makes it fail.

When you run this repro, go to http://localhost:33333/api/new, and you should see the following error in the console of SC:
```
2024-04-17 11:13:28.8356|22|Info|Microsoft.AspNetCore.HttpLogging.HttpLoggingMiddleware|Duration: 7.6784ms
2024-04-17 11:13:28.8365|22|Error|Microsoft.AspNetCore.Server.Kestrel|Connection id "0HN2UFEBGQOD5", Request id "0HN2UFEBGQOD5:00000001": An unhandled exception was thrown by the application.|System.InvalidOperationException: Unable to resolve service for type 'ServiceControl.Persistence.NewFeature.INewFeatureDataStore' while attempting to activate 'ServiceControl.NewFeature.Web.NewFeatureController'.
   at Microsoft.Extensions.DependencyInjection.ActivatorUtilities.ThrowHelperUnableToResolveService(Type type, Type requiredBy)
   at lambda_method354(Closure, IServiceProvider, Object[])
   at Microsoft.AspNetCore.Mvc.Controllers.ControllerFactoryProvider.<>c__DisplayClass6_0.<CreateControllerFactory>g__CreateController|0(ControllerContext controllerContext)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.Next(State& next, Scope& scope, Object& state, Boolean& isCompleted)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.InvokeInnerFilterAsync()
--- End of stack trace from previous location ---
   at Microsoft.AspNetCore.Mvc.Infrastructure.ResourceInvoker.<InvokeFilterPipelineAsync>g__Awaited|20_0(ResourceInvoker invoker, Task lastTask, State next, Scope scope, Object state, Boolean isCompleted)
[...]

PR #4092, opened 2024-04-17
Moving entry point to be explicit

johnsimons:
This is what the code would look like if we were explicit about our dependencies.

In this PR, we have moved our data storage registrations to `ServiceControl..Persistence.RavenDB`, and reference our implementations directly, no dynamic loading.

This is more in line with the existing code patterns. I think introducing yet another code pattern is not ideal.

I also noticed that the ServiceControl primary instance data layer has no "in-memory" implementation, which makes me wonder if we should at all, given it is just used for testing.

At the moment, this code does not work because https://github.com/Particular/ServiceControl/pull/4090


src/ServiceControl/Infrastructure/WebApi/HostApplicationBuilderExtensions.cs
```diff
@@ -33,6 +34,7 @@ public static void AddServiceControlApi(this IHostApplicationBuilder builder)
                 options.ModelBinderProviders.Insert(0, new SortInfoModelBindingProvider());
             });
             controllers.AddApplicationPart(Assembly.GetExecutingAssembly());
+            controllers.AddApplicationPart(typeof(ThroughputController).Assembly);
```
johnsimons, line 37: being explicit about our Controller

src/ServiceControl.Persistence.RavenDB/RavenPersistence.cs
```diff
@@ -72,7 +76,11 @@ public void AddPersistence(IServiceCollection services)
         services.AddSingleton<SagaAuditDestinationCustomCheck.State>();
     }
 
-    public void AddInstaller(IServiceCollection services) => ConfigureLifecycle(services);
+    public void AddInstaller(IServiceCollection services)
+    {
+        ConfigureLifecycle(services);
+        services.AddThroughputRavenPersistence();
```
johnsimons, line 82: entry point for installer

src/ServiceControl.Persistence.RavenDB/RavenPersistence.cs
```diff
@@ -52,6 +54,8 @@ public void AddPersistence(IServiceCollection services)
 
         services.AddSingleton<OperationsManager>();
 
+        services.AddThroughputRavenPersistence();
```
johnsimons, line 57: entry point for our persistence registrations

PR #4098, opened 2024-04-18
Refactor Raven connection string and embedded server startup

johnsimons:
This intent here is to separate the concern of bringing a RavenDB server up (example when using the embedded server) from getting a `IDocumentStore`.

I also exposed the store connection string as a "dependency", this way a component (example the throughput component) can depend on the connectionstring, and create its own store, example in our TF we have a separate database, so it is nice to just get a connection string and we can create our own store.

This change also starts treating both embedded or external RavenDB servers the same, so there is no distinctions how they both operate.

This changes are relying on the order of registration of `IHostedServices` to ensure there are no deadlocks.


src/ServiceControl.Persistence.RavenDB/EmbeddedDatabase.cs
```diff
@@ -19,12 +18,10 @@ namespace ServiceControl.Persistence.RavenDB
 
     public sealed class EmbeddedDatabase : IDisposable
     {
-        EmbeddedDatabase(RavenPersisterSettings configuration, IHostApplicationLifetime lifetime)
+        EmbeddedDatabase(RavenPersisterSettings configuration)
```
johnsimons, line 21: Given we are using an `IHostedService` I don't see a need for having any knowledge of `IHostApplicationLifetime`

src/ServiceControl.Persistence.RavenDB/RavenEmbeddedHostedService.cs
```diff
@@ -0,0 +1,34 @@
+namespace ServiceControl.Persistence.RavenDB;
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Hosting;
+
+class RavenEmbeddedHostedService(RavenPersisterSettings configuration)
```
johnsimons, line 8: This is now exclusively just used to start and stop an embedded server

src/ServiceControl.Persistence.RavenDB/RavenEmbeddedHostedService.cs
```diff
@@ -0,0 +1,34 @@
...
+
+    public Task StopAsync(CancellationToken cancellationToken)
+    {
+        embeddedDatabase.Dispose();
+        return Task.CompletedTask;
+    }
+
+    public async Task<string> GetConnectionString()
+    {
+        while (embeddedDatabase == null)
```
johnsimons, line 27: Not sure if this is the best way, but the idea is to wait till the server is ready

src/ServiceControl.Persistence.RavenDB/RavenEmbeddedPersistenceLifecycle.cs
```diff
@@ -1,58 +0,0 @@
...
-            while (true)
-            {
-                cancellationToken.ThrowIfCancellationRequested();
-
-                try
-                {
-                    documentStore = await database.Connect(cancellationToken);
-                    return;
-                }
-                catch (DatabaseLoadTimeoutException e)
```
johnsimons, line 39: I don't know if we still need some of this logic, thoughts?

PR #4119, opened 2024-04-29
Make FailureGroupsUnarchiveController public

mauroservienti:


danielmarbach:
Given that we have a route test that scans all controllers should we augment that one to throw when controllers are not public (can be done as a dedicated PR)

mauroservienti:
> Given that we have a route test that scans all controllers should we augment that one to throw when controllers are not public (can be done as a dedicated PR)

That would be good, yes.

PR #4120, opened 2024-04-29
Improve HttpApiRoutes approval tests

mauroservienti:


bording (approved):

danielmarbach (approved):
Generally I would prefer us tracking the controllers that violate the visibility rule in the foreach and then Assert the collection is empty at the end to see one failure for all controllers instead of having a failure per controller that violates it but given this assertion solves the problem that we will never have this problem anymore I don't think it is worth the effort changing

PR #4121, opened 2024-04-29
Add the Particular-ServicePulse-Version to the audit instance accepted headers

mauroservienti:


src/ServiceControl.Audit/Infrastructure/WebApi/Cors.cs
```diff
@@ -10,7 +10,7 @@ public static CorsPolicy GetDefaultPolicy()
 
             builder.AllowAnyOrigin();
             builder.WithExposedHeaders(["ETag", "Last-Modified", "Link", "Total-Count", "X-Particular-Version"]);
-            builder.WithHeaders(["Origin", "X-Requested-With", "Content-Type", "Accept"]);
+            builder.WithHeaders(["Origin", "X-Requested-With", "Content-Type", "Accept", "Particular-ServicePulse-Version"]);
```
danielmarbach, line 13: We already have special code in the scatter gather API that treats some headers as special. At the moment we never directly talk to an audit instance. We are always going over the "gateway". HttpForwarder in YARP also has the possibility to remove certain types of headers.

I know we raised a general discussion about this header but I wonder **if we stick to it** was should only allow this at the gateway boundaries and not at the aggregated instances.

PR #4140, opened 2024-05-03
Move throughput persistence code into Service Control primary instance

abparticular:


danielmarbach:
LGTM from a persistence integration standpoint. I have pushed a small tweak

PR #4086, opened 2024-04-16
Spike serialization context

danielmarbach:

PR #4142, opened 2024-05-03
CustomTimeSpanConverter to stay compatible with previous version of the saga audit plugin that used SimpleJson and a custom converter

danielmarbach:
The SagaAudit plugin has previously used SimpleJson and [had a custom implementation for TimeSpan that converted TimeSpan values into the short format ('g')](https://github.com/Particular/NServiceBus.SagaAudit/blob/release-4.0/src/NServiceBus.SagaAudit/MessageSerializationStrategy.cs#L36-L40). The behavior of this is that instead of constantly writing the 'c' format like `hh:mm:ss` (8 chars) it writes in certain circumstances `h:mm:ss` (7 chars). 

Newtonsoft.Json can send and receive the 7 or the 8 chars format. 
System.Text.Json only handles the 8 chars format. More details in https://github.com/dotnet/runtime/issues/100538

The changes in the Audit plugin make sense and given only new instances of ServiceControl that use System.Text.Json are affected, we have decided to implement a custom timespan converter for this very specific case to make deserializing `SagaUpdatedMessage` more relaxed.

I have also included parts of the spike of https://github.com/Particular/ServiceControl/pull/4086 into this PR to source gen the type information.

### DIFF
--- a/src/ServiceControl/SagaAudit/SagaUpdatedHandler.cs
+++ b/src/ServiceControl/SagaAudit/SagaUpdatedHandler.cs
@@ -37,6 +37,11 @@
 
         async Task RefreshAuditQueue()
         {
+            if (nextAuditQueueNameRefresh > DateTime.UtcNow)
+            {
+                return;
+            }
+
             await semaphore.WaitAsync();
             try
             {
--- a/src/ServiceControl/SagaAudit/SagaUpdatedHandler.cs
+++ b/src/ServiceControl/SagaAudit/SagaUpdatedHandler.cs
@@ -11,18 +11,12 @@
     using ServiceControl.Infrastructure;
     using ServiceControl.Persistence;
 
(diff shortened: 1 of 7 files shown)


src/ServiceControl/SagaAudit/SagaUpdatedHandler.cs
```diff
@@ -43,6 +37,11 @@ public async Task Handle(SagaUpdatedMessage message, IMessageHandlerContext cont
 
         async Task RefreshAuditQueue()
         {
+            if (nextAuditQueueNameRefresh > DateTime.UtcNow)
```
bording, line 40: What caused this check to be added? It was added as part of the "Cleanup of the SagaUpdatedHandler" commit, but this doesn't seem like cleanup.

danielmarbach: I did split the commit

mauroservienti (approved):

PR #4147, opened 2024-05-06
Simplify the primary instance SagaAudit data store

mauroservienti:
Primary instances stored SagaAudit messages received in the primary input queue. Since introducing Audit instances, SagaAudit messages should go to the audit queue, not the primary instance input queue. Until ServiceControl 5.1.x, the primary instance upon receiving a SagaAudit message was:

- Enumerating all audit instances, picking one (and caching the result for 5 minutes)
- Storing the message into a modified data store that, instead of storing, was adding to an in-memory dictionary the endpoint name that was still sending SagaAudit messages to the primary instance
- Forwarding the SagaAudit message to the selected audit instance

At the same time, a primary instance custom check was reported to ServicePulse, listing any endpoint still sending messages to the primary instance.

This PR removes all that and keeps:

- Enumerating all audit instances, picking one (and caching the result for 5 minutes)
- Forwarding the SagaAudit message to the selected audit instance

The new code also logs a warning explaining that the forwarding functionality will be removed in version 6.


src/ServiceControl.Persistence.RavenDB/SagaAudit/NoOpSagaAuditDataStore.cs
```diff
@@ -0,0 +1,14 @@
...
+    using System;
+    using System.Threading.Tasks;
+    using ServiceControl.Persistence.Infrastructure;
+    using ServiceControl.SagaAudit;
+
+    class NoOpSagaAuditDataStore : ISagaAuditDataStore
+    {
+        public Task<bool> StoreSnapshot(SagaSnapshot sagaSnapshot) => Task.FromResult(false);
+
+        public Task<QueryResult<SagaHistory>> GetSagaById(Guid sagaId) => Task.FromResult(QueryResult<SagaHistory>.Empty());
```
danielmarbach, line 12: Technically, this task could be cached but given this is not really on the critical path I'd say leave it as is

src/ServiceControl/SagaAudit/GetSagaByIdApi.cs
```diff
@@ -11,18 +11,11 @@ namespace ServiceControl.SagaAudit
...
-            base(dataStore, settings, httpClientFactory)
-        {
-        }
-
-        protected override async Task<QueryResult<SagaHistory>> LocalQuery(SagaByIdContext input)
-        {
-            var result = await DataStore.GetSagaById(input.SagaId);
-            return result;
-        }
+        protected override Task<QueryResult<SagaHistory>> LocalQuery(SagaByIdContext input)
```
danielmarbach, line 17: Technically, this task could be cached but given this is not really on the critical path I'd say leave it as is

danielmarbach (approved):
LGTM.

Do you know when the custom check was introduced? Has sufficient time passed so that we can remove it?

mauroservienti:
> Do you know when the custom check was introduced? Has sufficient time passed so that we can remove it?

I don't. I was already marked for removal in v5.

src/ServiceControl.Persistence.RavenDB/SagaAudit/NoOpSagaAuditDataStore.cs
danielmarbach: No longer relevant with #4175

PR #4149, opened 2024-05-07
Introduce a scatter/gather with no local call

mauroservienti:
https://github.com/Particular/ServiceControl/pull/4147 surfaced that the scatter/gather API was designed assuming a local query always happens. That's not the case anymore for SagaAudits.

This PR attempts to introduce a scatter/gather API that doesn't require a local data store.


src/ServiceControl/CompositeViews/Messages/ScatterGatherApi.cs
```diff
@@ -21,32 +21,22 @@ public abstract class ScatterGatherApiBase
...
         }
+        protected Settings Settings { get; }
+        protected IHttpClientFactory HttpClientFactory { get; }
 
-        protected TDataStore DataStore { get; }
-        Settings Settings { get; }
-        IHttpClientFactory HttpClientFactory { get; }
-
-        public async Task<QueryResult<TOut>> Execute(TIn input, string pathAndQuery)
+        protected void ExecuteRemotes(List<Task<QueryResult<TOut>>> tasks, string pathAndQuery)
```
mauroservienti, line 37: @bording, The only thing I dislike about the proposed changes is this `ExecuteRemotes` method. Looking at the API from the user perspective, it is not clear what the List should be.

Another option would be to design this like:

```suggestion
        protected List<Task<QueryResult<TOut>>> ExecuteRemotes(string pathAndQuery)
```

And then create the list instance inside the method. That was my original design, which comes with the downside of not being able to initialize the list with the right size.

danielmarbach:
I wonder if it is worth complicating the already quite complex inheritance chain for one special case given that you could still kill the NoOp store and simply implement the local query by return an empty query result (which would also get rid of the NoOp implementation and additional interface. We would achieve the same with less complexity

mauroservienti:
🤷 Good question, @danielmarbach. https://github.com/Particular/ServiceControl/pull/4147 already solves the compelling problem by simplifying stuff. This further improves the design by allowing a cleaner approach for a specific use case that we might face more in the future. That said, I prefer this approach to the NoOp one, but it's just a preference. That's why this is up for discussion.

danielmarbach:
@mauroservienti this is what I meant https://github.com/Particular/ServiceControl/pull/4175

It avoids the oddities you talked about in https://github.com/Particular/ServiceControl/pull/4149#discussion_r1592846598

mauroservienti:
Withdrawn in favor of #4175

PR #4174, opened 2024-05-16
Look for posix compliant variables

johnsimons:
Our code that looks for env vars is not entirely [POSIX compliant](https://stackoverflow.com/a/2821183).
This PR adds also looking for all UPPER case env vars.

### DIFF
--- a/src/ServiceControl.Configuration/EnvironmentVariableSettingsReader.cs
+++ b/src/ServiceControl.Configuration/EnvironmentVariableSettingsReader.cs
@@ -27,6 +27,13 @@ static class EnvironmentVariableSettingsReader
             return true;
         }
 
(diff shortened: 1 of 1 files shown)


src/ServiceControl.Configuration/EnvironmentVariableSettingsReader.cs
```diff
@@ -27,6 +27,13 @@ public static bool TryRead<T>(SettingsRootNamespace settingsNamespace, string na
             return true;
         }
 
+        // POSIX comliant https://stackoverflow.com/a/2821183
```
danielmarbach, line 30:
```suggestion
        // POSIX compliant https://stackoverflow.com/a/2821183
```

danielmarbach (approved):

PR #3965, opened 2024-02-16
Incorporate Throughput Tool into ServiceControl

jpalac:
Result of TF https://github.com/Particular/CustomerSuccess/issues/3601

This PR brings the [Throughput Tool](https://github.com/Particular/Particular.EndpointThroughputCounter) functionality into ServiceControl.

PR #4193, opened 2024-05-23
Integrated the transport seam better into the host to be able to leverage DI (similar to the persistence seam)

danielmarbach:
This PR is the outcome of a discussion related to changes done as part of https://github.com/Particular/ServiceControl/pull/3965 which had to introduce a `ThroughputQueryProvider` type on the transport customization. By aligning the transport customization integration well enough with how the persistence seam integrates into the DI container it becomes possible to register types directly that then the transport infrastructure can resolve. This is demonstrated by migrating the queue length query providers to actual hosted services.


src/ServiceControl.Transports/TransportCustomization.cs
```diff
@@ -33,6 +35,10 @@ public abstract class TransportCustomization<TTransport> : ITransportCustomizati
 
         protected abstract void CustomizeTransportForMonitoringEndpoint(EndpointConfiguration endpointConfiguration, TTransport transportDefinition, TransportSettings transportSettings);
 
+        public virtual void AddTransport(IServiceCollection services)
```
johnsimons, line 38: @danielmarbach, if we are making this seam aware of asp.net/di, it makes me wonder if the `IProvideQueueLength` could be turned into a `BackgroundService` and registered accordingly. Thoughts?
It would be one less abstraction that we need to take care.

danielmarbach: Yes that would be my idea as well

src/ServiceControl.Transports/IProvideQueueLength.cs
```diff
@@ -1,16 +1,9 @@
 ﻿namespace ServiceControl.Transports
 {
-    using System;
-    using System.Threading.Tasks;
+    using Microsoft.Extensions.Hosting;
 
-    public interface IProvideQueueLength
+    public interface IProvideQueueLength : IHostedService
```
johnsimons, line 5: I feel that the implementers need to decide whether using `IHostedService` is the right thing to do.
in other words, should this interface **not** implement `IHostedService `?

src/ServiceControl.Transports/AbstractQueueLengthProvider.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        protected Action<QueueLengthEntry[], EndpointToQueueMapping> Store { get; private set; } = store;
+
+        protected string ConnectionString { get; } = settings.ConnectionString;
+
+        public override async Task StartAsync(CancellationToken cancellationToken)
+        {
+            // BackgroundService in .NET awaits the returned task to be completed. Queue providers might be doing
+            // some heavy lifting before yielding so we are forcing things to be offloaded here by default.
+            await Task.Yield();
```
johnsimons, line 19: This feels strange to me, I feel that should be up to the implementations

src/ServiceControl.Transports.SqlServer/QueueLengthProvider.cs
```diff
@@ -38,36 +35,27 @@ public void TrackEndpointInputQueue(EndpointToQueueMapping queueToTrack)
             tableSizes.TryAdd(sqlTable, 0);
         }
 
-        public Task Start()
+        protected override async Task ExecuteAsync(CancellationToken stoppingToken)
         {
-            stop = new CancellationTokenSource();
-
-            poller = Task.Run(async () =>
+            while (!stoppingToken.IsCancellationRequested)
```
johnsimons, line 40: Would the `PeriodicTimer` be a better fit for this background task?
I used it in https://github.com/Particular/ServiceControl/blob/1b522c5bd43c7ea55a0025720ccdad778affcc7e/src/Particular.LicensingComponent/BrokerThroughput/BrokerThroughputCollectorHostedService.cs#L34 and feels nicer.
I am just wondering.

johnsimons (approved):
LGTM

src/ServiceControl.Transports.SqlServer/QueueLengthProvider.cs
johnsimons: And yes, I noticed this is existing code and they are all the same pattern!

src/ServiceControl.Transports/AbstractQueueLengthProvider.cs
danielmarbach: Yeah it is weird but this is actually an ongoing issue in the .NET runtime repository that MS will eventually fix. Given that all our queue length providers follow the same pattern and this only affects the startup and not really the hot path I figured it is better to put it into a cross-cutting place

src/ServiceControl.Transports.SqlServer/QueueLengthProvider.cs
danielmarbach: Yeah the reason I use the while with the delay pattern is because the existing code was already doing that. At least from that perspective I wanted to stay compatible. I was aware the period timer pattern exists. It is also documented here https://learn.microsoft.com/en-us/aspnet/core/fundamentals/host/hosted-services?view=aspnetcore-8.0&tabs=visual-studio#asynchronous-timed-background-task

Can you elaborate on the "feels nicer" part?

src/ServiceControl.Transports/IProvideQueueLength.cs
danielmarbach: That's an excellent question. I was going down that path because the existing code and test was starting and stopping things to make it work. By not adding the hosted service interface the test code would become slightly more complex because it would have to retrieve the `AbstractQueueLengthProvider` and then start that. This would also mean we then technically need three registrations of which one would only ever be used by the transport tests. We can't just retrieve the `IHostedService` instance because while there is currently only one, there might be more later. So I'm kind of torn. I'll merge the PR and think about this a bit more.

johnsimons: Yeah, I can see the complexity.
TBH, given this is not a public API, I am fine with it 😄

PR #4198, opened 2024-05-28
Add test to verify that default azure credentials can be used with ASB

andreasohlund:


PhilBastian (approved):

andreasohlund:
@danielmarbach made the base class split we talked about, please take a look

danielmarbach (approved):

PR #4205, opened 2024-05-29
Fix problems with PowerShell dependency resolver

bording:
This PR makes the following changes to the PowerShell dependency resolver:

- When having to fallback to the default `runtime` section the deps.json file, assembly file path will now be properly extracted even when the path does not start with `lib`
- Resource assemblies will now be properly loaded if they exist.

Without these changes, Microsoft.Data.SqlClient may not be loaded properly by the PowerShell module.

### DIFF
--- a/src/ServiceControl.Management.PowerShell/DependencyResolver.cs
+++ b/src/ServiceControl.Management.PowerShell/DependencyResolver.cs
@@ -45,16 +45,26 @@ class DependencyResolver
     {
         ArgumentNullException.ThrowIfNull(assemblyName);
 
-        var library = dependencyContext.RuntimeLibraries.SingleOrDefault(r => r.Name.Equals(assemblyName.Name, StringComparison.Ordinal));
+        var name = assemblyName.Name;
+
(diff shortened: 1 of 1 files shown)


lailabougria (approved):
LGTM

lailabougria:
Would it be worth going the extra mile here and adding some tests?

NChaganlal:
Talking to @bording offline (he is on leave) and the reason why we didn't add tests was:

> "because there is no infrastructure in place for adding those tests (no test project for the PowerShell module project) and TBH adding tests around that code could be tricky because of it being related to the low-level nature of it (assembly loading)"

NChaganlal:
@adamralph and I evaluated the effort of writing tests and it feels like there is a bit of effort in making sure the infrastructure is in place to support acceptance tests for Powershell. Additionally, Powershell code is targeting .net6-windows which also requires that the tests are written in the same version of .net for reliability reasons. We realised that the ROI at this point in time is not worth it so we can ensure we unblock the customer. Tests should be considered as part of a normal release.

PR #4212, opened 2024-06-03
Check for Raven server version, license, and cluster configuration in exernally-managed mode

DavidBoike:


src/ServiceControl.RavenDB/StartupChecks.cs
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        public static async Task EnsureServiceControlLicense(IDocumentStore store, CancellationToken cancellationToken = default)
+        {
+            var licenseResponse = await store.Maintenance.Server.SendAsync(new GetLicenseOperation(), cancellationToken);
+            if (licenseResponse.LicensedTo != "ParticularNservicebus (Israel)")
+            {
+                throw new Exception("Wrong license");
+            }
+        }
```
boblangley, line 30: Do we need this anymore since we are rolling our own image?

The important check is the version. If they want the license they use our image. I think we can remove this check.

src/ServiceControl.RavenDB/StartupChecks.cs
```diff
@@ -0,0 +1,66 @@
...
+
+                if (response.TryGet<string>("Id", out var id) && response.TryGet<string>("LicensedTo", out var licensedTo))
+                {
+                    Result = new LicenseResult(id, licensedTo);
+                }
+            }
+        }
+
+        record class LicenseResult(string Id, string LicensedTo);
+    }
```
boblangley, line 65:
```suggestion
```

boblangley (approved):

PR #4228, opened 2024-06-14
Resolves IDE0270 issues

ramonsmits:
Resolves:

> [IDE0270] Null check can be simplified (https://learn.microsoft.com/dotnet/fundamentals/code-analysis/style-rules/ide0270)


ramonsmits:
Would be overwritten by reposync

DavidBoike (approved):

PR #4236, opened 2024-06-18
Make setup Di compatible

johnsimons:


src/ServiceControl/ComponentInstallationContext.cs
```diff
@@ -2,21 +2,16 @@ namespace Particular.ServiceControl
...
-        public List<Assembly> IndexAssemblies { get; } = [];
-
-        public List<Func<Task>> InstallationTasks { get; } = [];
+        public List<Func<IServiceProvider, Task>> InstallationTasks { get; } = [];
 
         public void CreateQueue(string queueName) => Queues.Add(queueName);
 
-        public void AddIndexAssembly(Assembly assembly) => IndexAssemblies.Add(assembly);
-
-        public void RegisterInstallationTask(Func<Task> setupTask) => InstallationTasks.Add(setupTask);
```
danielmarbach, line 20: I always found this super funky honestly. We have a method but the list is mutable anyway so you could easily just call InstallationTasks.Add. And then why do they need to be of type `Func<Task> or `Func<IServiceProvider, Task>` when we could define a dedicated installer interface that a component can register as part of AddServiceControlInstallers. That could also had an installation component that runs all that stuff automatically on start.

johnsimons: This was introduced in https://github.com/Particular/ServiceControl/pull/2906, which we have abandoned, so I ended up removing it, given it was not used anyway.

abparticular (approved):

PR #4247, opened 2024-06-19
Load RavenDB client certificate if located next to the executable

DavidBoike:


src/ServiceControl.Audit.Persistence.RavenDB/RavenExternalPersistenceLifecycle.cs
```diff
@@ -35,10 +37,19 @@ public async Task Initialize(CancellationToken cancellationToken = default)
             {
                 await initializeSemaphore.WaitAsync(cancellationToken);
 
+                // Look for raven-client-certificate.pfx in same directory as application code
+                var potentialCertPath = Path.GetFullPath(Path.Combine(GetType().Assembly.Location, "..", "raven-client-certificate.pfx"));
```
bording, line 41: That path doesn't look right. Should be up two levels from the persistence assembly when everything is in a production layout.

Seeing that now though, this would never work correctly for development scenarios, so I'm wondering if we need to rethink the location.

bording (changes requested):

src/ServiceControl.Audit.Persistence.RavenDB/RavenExternalPersistenceLifecycle.cs
boblangley: We had discussed this before. It is easier to bind shorter paths. Could we just look somewhere like `/cert` or similar?

bording: > We had discussed this before. It is easier to bind shorter paths. Could we just look somewhere like `/cert` or similar?

The goal here is to make it a path that would be valid regardless if it's inside a container or not. That way this is something that could work whenever a certificate exists.

Having it live next to the application means, in the container scenario, you'd be mounting the cert to `/app/raven-client-certificate.pfx`

DavidBoike: Brandon: you're right, the path would not be right, and would be different during dev and production. I'll find something else that does the entry point or something.

Bob: Brandon and I discussed earlier today, this is a get out of jail free card that isn't container-implementation-specific. It's just a feature/get-out-of-jail-free card that could theoretically even be used in Windows without assuming anything about platform. We also tested you could bind a single file directly with `-v path/to/pfx:/app/raven-client-certificate.pfx` this way which doesn't seem terribly long?

boblangley: Yeah, that works.

bording (approved):

src/ServiceControl.Audit.Persistence.RavenDB/RavenExternalPersistenceLifecycle.cs
```diff
@@ -35,10 +38,21 @@ public async Task Initialize(CancellationToken cancellationToken = default)
             {
                 await initializeSemaphore.WaitAsync(cancellationToken);
 
+                // Look for raven-client-certificate.pfx in same directory as application code
+                var applicationDirectory = Path.GetDirectoryName(Assembly.GetEntryAssembly()!.Location) ?? string.Empty;
```
DavidBoike, line 42: @bording with the change to null coalesce to string, should the `!.Location` be changed to `?.Location`? Docs say it can return null when called from unmanaged code…which I don't see ever happening? But maybe `?.` would be safer?

PR #4263, opened 2024-06-26
Fix for when sending to queues on a different machine

jpalac:


src/ServiceControl.Monitoring/Infrastructure/ReportThroughputHostedService.cs
```diff
@@ -27,7 +28,14 @@ protected override async Task ExecuteAsync(CancellationToken cancellationToken)
...
                     {
-                        logger.LogError(ex, $"Error obtaining throughput from Monitoring for {ReportSendingIntervalInMinutes} minutes interval.");
+                        if (ex.InnerException is not null and QueueNotFoundException)
+                        {
+                            logger.LogError($"Error obtaining throughput from Monitoring for {ReportSendingIntervalInMinutes} minutes interval: {ex?.InnerException?.Message}");
+                        }
+                        else
+                        {
+                            logger.LogError(ex, $"Error obtaining throughput from Monitoring for {ReportSendingIntervalInMinutes} minutes interval.");
+                        }
```
abparticular, line 38:
```suggestion
                        var exception = ex.InnerException is QueueNotFoundException qnfe ? qnfe : ex;

                        logger.LogError(exception, $"Error obtaining throughput from Monitoring for {ReportSendingIntervalInMinutes} minutes interval.");
```

abparticular (approved):

src/ServiceControl.Monitoring/Infrastructure/ReportThroughputHostedService.cs
jpalac: I want to display a different error and not log the whole exception if it's just a queue not found thing.

PR #4264, opened 2024-06-26
Fix sql query name

jpalac:


src/ServiceControl.Transports.SqlServer/SqlServerQuery.cs
```diff
@@ -14,7 +14,7 @@ namespace ServiceControl.Transports.SqlServer;
 public class SqlServerQuery(
     ILogger<SqlServerQuery> logger,
     TimeProvider timeProvider,
-    TransportSettings transportSettings) : BrokerThroughputQuery(logger, "SqlTransport")
+    TransportSettings transportSettings) : BrokerThroughputQuery(logger, "SQLServer")
```
johnsimons, line 17: This is what is in the tool - https://github.com/Particular/Particular.EndpointThroughputCounter/blob/51db173151ee0337c74db013d35d701ec881a4a5/src/AppCommon/Commands/SqlServerCommand.cs#L158C37-L158C49

jpalac: hmm - ok well in that case i'll update SP to expect that value

jpalac:
not doing

PR #4269, opened 2024-06-27
Add sanitized name cleanser to brokerquery and use it in summary and report grouping

jpalac:


src/ServiceControl.Transports.SQS/AmazonSQSQuery.cs
```diff
@@ -292,4 +292,6 @@ public override string SanitizeEndpointName(string endpointName)
             return endpointName;
         }
     }
+
+    public override string SanitizedEndpointNameCleanser(string endpointName) => endpointName.ToLower();
```
johnsimons, line 296: I checked SQS and the queue name is case sensitive.
```suggestion

```

jpalac: removed it

src/ServiceControl.Transports/BrokerThroughput/IBrokerThroughputQuery.cs
```diff
@@ -20,4 +20,5 @@ public interface IBrokerThroughputQuery
     KeyDescriptionPair[] Settings { get; }
     Task<(bool Success, List<string> Errors, string Diagnostics)> TestConnection(CancellationToken cancellationToken);
     string SanitizeEndpointName(string endpointName);
+    string SanitizedEndpointNameCleanser(string endpointName);
```
abparticular, line 23: Do we need another method? Is it not possible for `SanitizeEndpointName` to include any necessary "cleansing"?

src/Particular.LicensingComponent/ThroughputCollector.cs
```diff
@@ -214,6 +214,13 @@ string Mask(string stringToMask)
         }
     }
 
+    string CleanseSanitizedName(string endpointName)
+    {
+        return throughputQuery == null ? defaultSanitizedEndpointNameCleanser(endpointName) : throughputQuery.SanitizedEndpointNameCleanser(endpointName);
+    }
+
+    Func<string, string> defaultSanitizedEndpointNameCleanser = endpointName => endpointName;
+
```
abparticular, line 223:
```suggestion
    string CleanseSanitizedName(string endpointName)
    {
        return throughputQuery == null ? endpointName : throughputQuery.SanitizedEndpointNameCleanser(endpointName);
    }
```

src/ServiceControl.Transports/BrokerThroughput/IBrokerThroughputQuery.cs
jpalac: no as that's used for something else - to store the sanitized name. We are then using the cleansing to cleanse the audit and monitoring sanitized name to that it matches what the broker has.

johnsimons: yeah, i was thinking the same but i can see the reason

johnsimons (approved):

PR #4270, opened 2024-06-27
Turn off licensing

jpalac:


abparticular (approved):

bording:
We've created the `release-5.3` by branching from an earlier point and cherry-picking changes, so this won't be needed.

PR #4274, opened 2024-06-28
Improve docs for sql env vars

johnsimons:


src/ServiceControl.Transports.SqlServer/SqlServerQuery.cs
```diff
@@ -147,8 +148,12 @@ protected override void InitialiseCore(FrozenDictionary<string, string> settings
     public static class SqlServerSettings
     {
         public static readonly string ConnectionString = "SqlServer/ConnectionString";
-        public static readonly string ConnectionStringDescription = "A single database connection string that will provide at least read access to all queue tables.";
+
+        public static readonly string ConnectionStringDescription =
+            "Database connection string that will provide at least read access to all queue tables.";
```
jpalac, line 153: If there's no database specified (ie it's not a "single" db connection) then how would SC know which database/catalog to use?

johnsimons: it would error, and not start

johnsimons: The connection string requires the initial catalog to be set

jpalac: So why remove the word "single" from the description then?

johnsimons: because there is no "multiple"

jpalac (approved):
Can you please make the same updates on the params in docs PR?

johnsimons:
> Can you please make the same updates on the params in docs PR?

already done that

PR #4275, opened 2024-06-28
Switch to readonly dictionary instead

danielmarbach:
By looking at the usage of the dictionary it feels to me it is not worth the overhead. Frozen collections in general can take multiple miliseconds to initialize and that overhead is only worth taking when you actually read from those many many times. By having had a quick look at the usage of the settings in the queries it looks like we usually extract concrete information from the settings in the initialization path and then use that stored state to do the runtime operations. So effectively the frozen dictionary is not read on the hot path over and over again.

### DIFF
--- a/src/Particular.LicensingComponent.UnitTests/AuditThroughputCollectorHostedService_Tests.cs
+++ b/src/Particular.LicensingComponent.UnitTests/AuditThroughputCollectorHostedService_Tests.cs
@@ -1,8 +1,8 @@
 ﻿namespace Particular.LicensingComponent.UnitTests;
 
 using System;
-using System.Collections.Frozen;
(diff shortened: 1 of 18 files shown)


johnsimons:
> Or was it meant to be "just readonly"?

yes that was the intention

abparticular (approved):

jpalac (approved):

danielmarbach:
Collabed with David on fixing the rabbitmq setup action
https://github.com/Particular/ServiceControl/pull/4286 solves this. Needs a response base and https://github.com/Particular/ServiceControl/pull/4275/commits/f923909d6e363e2fe67196eaffacd52646d8bb89 should be dropped

PR #4280, opened 2024-07-01
Cache http client in sqs

johnsimons:
The http client was not being cached

### DIFF
--- a/src/ServiceControl.Transports.SQS/AmazonSQSQuery.cs
+++ b/src/ServiceControl.Transports.SQS/AmazonSQSQuery.cs
@@ -16,8 +16,8 @@ using Amazon.Runtime;
 using Amazon.Runtime.CredentialManagement;
 using Amazon.SQS;
 using Amazon.SQS.Model;
+using BrokerThroughput;
 using Microsoft.Extensions.Logging;
-using ServiceControl.Transports.BrokerThroughput;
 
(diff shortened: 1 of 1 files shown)


src/ServiceControl.Transports.SQS/AmazonSQSQuery.cs
```diff
@@ -175,7 +175,10 @@ public static class AmazonSQSSettings
 
     class AwsHttpClientFactory : HttpClientFactory
     {
-        public override HttpClient CreateHttpClient(IClientConfig clientConfig) => new(new SocketsHttpHandler { PooledConnectionLifetime = TimeSpan.FromMinutes(2) });
+        readonly HttpClient httpClient =
```
johnsimons, line 178: We actually want to reuse the client all the time

src/ServiceControl.Transports.SQS/AmazonSQSQuery.cs
```diff
@@ -175,7 +175,10 @@ public static class AmazonSQSSettings
...
+            new(new SocketsHttpHandler { PooledConnectionLifetime = TimeSpan.FromMinutes(2) });
+
+        public override HttpClient CreateHttpClient(IClientConfig clientConfig) => httpClient;
         public override bool DisposeHttpClientsAfterUse(IClientConfig clientConfig) => false;
```
johnsimons, line 182: this is needed so the client is never disposed

jpalac (approved):

PR #4281, opened 2024-07-01
Convert to use transport directly

johnsimons:


src/Particular.LicensingComponent/MonitoringThroughput/MonitoringThroughputHostedService.cs
```diff
@@ -0,0 +1,34 @@
...
+using Microsoft.Extensions.Logging;
+using NServiceBus.Transport;
+using ServiceControl.Transports;
+using Shared;
+
+internal class MonitoringThroughputHostedService(ITransportCustomization transportCustomization, TransportSettings transportSettings, ILogger<MonitoringThroughputHostedService> logger, MonitoringService monitoringService) : IHostedService
+{
+    private TransportInfrastructure? transportInfrastructure;
+
+    private async Task Handle(MessageContext message, CancellationToken cancellationToken)
```
jpalac, line 13: It doesn't work - this never gets hit - messages are not being processed from the queue.

src/Particular.LicensingComponent/MonitoringThroughput/MonitoringThroughputHostedService.cs
```diff
@@ -0,0 +1,34 @@
+namespace Particular.LicensingComponent.MonitoringThroughput;
+
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using NServiceBus.Transport;
+using ServiceControl.Transports;
+using Shared;
+
+internal class MonitoringThroughputHostedService(ITransportCustomization transportCustomization, TransportSettings transportSettings, ILogger<MonitoringThroughputHostedService> logger, MonitoringService monitoringService) : IHostedService
```
jpalac, line 9: We use BackgroundService for all the other ones - should we keep it in sync?

jpalac (changes requested):
Doesn't seem to work - the messages sent from monitoring are not being received.

src/Particular.LicensingComponent/MonitoringThroughput/MonitoringThroughputHostedService.cs
johnsimons: I fixed that, we need to explicitly start the receiver

src/Particular.LicensingComponent/MonitoringThroughput/MonitoringThroughputHostedService.cs
johnsimons: I looked at it, but unfortunately the signature methods of our transport don't really line up well with a single `run` method, hence the reason to do it that way.
This is the way the other are also done.

jpalac (approved):

PR #4314, opened 2024-07-18
Disable report download if using a broker but there is no broker data for last 30 days

jpalac:
Currently the report download button on the Usage page is enabled as long as there is 1 day of data in the last 30 days.
This could produce incomplete reports if a broker transport is used but not configured correctly, and the data is only coming from a subset of endpoints that have Audit and/or Monitoring configured.

This change ensures that the report download button is not enabled if a broker transport is being used, but there's no data from the broker.

### DIFF
--- a/src/Particular.LicensingComponent.UnitTests/Infrastructure/ThroughputCollectorTestFixture.cs
+++ b/src/Particular.LicensingComponent.UnitTests/Infrastructure/ThroughputCollectorTestFixture.cs
@@ -1,6 +1,7 @@
 ﻿namespace Particular.LicensingComponent.UnitTests.Infrastructure
 {
     using System;
+    using System.Linq;
     using System.Threading.Tasks;
     using Contracts;
     using Microsoft.Extensions.DependencyInjection;
@@ -16,21 +17,24 @@
         [SetUp]
         public virtual Task Setup()
         {
-            configuration = new ThroughputTestsConfiguration();
+            var testMethod = GetType().GetMethod(TestContext.CurrentContext.Test.MethodName);
+            var attribute = testMethod?.GetCustomAttributes(typeof(UseNonBrokerTransportAttribute), false).FirstOrDefault();
 
-            return configuration.Configure(SetThroughputSettings, SetExtraDependencies);
+            return configuration.Configure(SetThroughputSettings, SetExtraDependencies, attribute is not UseNonBrokerTransportAttribute);
         }
 
         [TearDown]
-        public virtual Task Cleanup()
-        {
-            return configuration?.Cleanup();
-        }
+        public virtual Task Cleanup() => configuration?.Cleanup();
 
         protected ILicensingDataStore DataStore => configuration.LicensingDataStore;
 
         protected IThroughputCollector ThroughputCollector => configuration.ThroughputCollector;
 
-        protected ThroughputTestsConfiguration configuration;
+        protected ThroughputTestsConfiguration configuration = new();
+
+        [AttributeUsage(AttributeTargets.Method)]
+        public class UseNonBrokerTransportAttribute : Attribute
+        {
+        }
     }
 }
--- a/src/Particular.LicensingComponent/ThroughputCollector.cs
+++ b/src/Particular.LicensingComponent/ThroughputCollector.cs
@@ -97,7 +97,7 @@ public class ThroughputCollector(ILicensingDataStore dataStore, ThroughputSettin
     async Task<ReportGenerationState> GetReportGenerationStateForBroker(CancellationToken cancellationToken)
     {
         var reportCanBeGenerated = await dataStore.IsThereThroughputForLastXDaysForSource(30, ThroughputSource.Broker, false, cancellationToken);
-        var reason = reportCanBeGenerated ? string.Empty : $"Need at least one day of usage data in the last 30 days from {transport} broker.";
+        var reason = reportCanBeGenerated ? string.Empty : "Need at least one day of usage data in the last 30 days from the configured broker.";
 
         return new ReportGenerationState(transport)
         {
--- a/src/Particular.LicensingComponent/ThroughputCollector.cs
+++ b/src/Particular.LicensingComponent/ThroughputCollector.cs
@@ -96,6 +96,15 @@ public class ThroughputCollector(ILicensingDataStore dataStore, ThroughputSettin
             ReportCanBeGenerated = await dataStore.IsThereThroughputForLastXDays(30, cancellationToken)
         };
 
+        //ensure that if there is a broker that we only allow generation if there is broker data in the last 30 days
+        if (reportGenerationState.ReportCanBeGenerated && throughputQuery != null)
+        {
+            if (!await dataStore.IsThereThroughputForLastXDaysForSource(30, ThroughputSource.Broker, false, cancellationToken))
+            {
+                reportGenerationState.ReportCanBeGenerated = false;
+            }
+        }
+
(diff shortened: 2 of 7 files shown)


johnsimons (approved):
LGTM.

src/Particular.LicensingComponent/ThroughputCollector.cs
```diff
@@ -96,6 +96,15 @@ public async Task<ReportGenerationState> GetReportGenerationState(CancellationTo
...
+        {
+            if (!await dataStore.IsThereThroughputForLastXDaysForSource(30, ThroughputSource.Broker, false, cancellationToken))
+            {
+                reportGenerationState.ReportCanBeGenerated = false;
+            }
+        }
+
         if (!reportGenerationState.ReportCanBeGenerated)
         {
             reportGenerationState.Reason = "24 hours worth of data needs to exist in the last 30 days.";
```
abparticular, line 110: I think we need to modify the reason based on what the issue actually is.  

Otherwise situations can arise where the customer sees that there is recent throughput data, but they're still not able to generate a report because it isn't broker data. And the reason they're given is that there is no recent data.

jpalac: @abparticular  There's limited space on the screen - we can't really add any more text than that, If you have a suggestion that doesn't change the length please give one. 
They would also be seeing a big red message on the screen saying that there are connection issues to the broker pointing them to fix it. So it should be pretty obvious in general what the issue is.

src/Particular.LicensingComponent/ThroughputCollector.cs
```diff
@@ -89,19 +89,33 @@ public async Task<List<EndpointThroughputSummary>> GetThroughputSummary(Cancella
...
+    public async Task<ReportGenerationState> GetReportGenerationState(CancellationToken cancellationToken) =>
+        throughputQuery == null ?
+            await GetReportGenerationStateForNonBroker(cancellationToken) :
+            await GetReportGenerationStateForBroker(cancellationToken);
+
+    async Task<ReportGenerationState> GetReportGenerationStateForBroker(CancellationToken cancellationToken)
     {
-        var reportGenerationState = new ReportGenerationState(transport)
+        var reportCanBeGenerated = await dataStore.IsThereThroughputForLastXDaysForSource(30, ThroughputSource.Broker, false, cancellationToken);
+        var reason = reportCanBeGenerated ? string.Empty : $"Need at least one day of usage data in the last 30 days from ${transport} broker.";
```
johnsimons, line 100: `transport` prop here, will display an unformatted string, eg `RabbitMQ.ClassicConventionalRouting`
In SP we format these strings, see https://github.com/Particular/ServicePulse/blob/ca4c4c8700a53051052151a13a0e86ef21189d83/src/Frontend/src/stores/ThroughputStore.ts#L54

src/Particular.LicensingComponent.UnitTests/Infrastructure/ThroughputCollectorTestFixture.cs
```diff
@@ -16,21 +17,24 @@ abstract class ThroughputCollectorTestFixture
...
 
         protected ILicensingDataStore DataStore => configuration.LicensingDataStore;
 
         protected IThroughputCollector ThroughputCollector => configuration.ThroughputCollector;
 
-        protected ThroughputTestsConfiguration configuration;
+        protected ThroughputTestsConfiguration configuration = new();
+
+        [AttributeUsage(AttributeTargets.Method)]
+        public class UseNonBrokerTransportAttribute : Attribute
```
johnsimons, line 36: Feels a bit off to introduce this attribute to customise the setup of the test, I think refactoring the test may be simpler

src/Particular.LicensingComponent/ThroughputCollector.cs
johnsimons: has anyone tested the length of this string in SP? What does it look like with a long transport name?

johnsimons (changes requested):

src/Particular.LicensingComponent/ThroughputCollector.cs
abparticular: Thank you, nice spot.  I actually meant to get the transport from the broker query which I believe is short form.

src/Particular.LicensingComponent.UnitTests/Infrastructure/ThroughputCollectorTestFixture.cs
abparticular: I felt the attribute was preferable rather than splitting related tests across different test fixtures.

src/Particular.LicensingComponent/ThroughputCollector.cs
```diff
@@ -89,19 +89,33 @@ public async Task<List<EndpointThroughputSummary>> GetThroughputSummary(Cancella
...
+    public async Task<ReportGenerationState> GetReportGenerationState(CancellationToken cancellationToken) =>
+        throughputQuery == null ?
+            await GetReportGenerationStateForNonBroker(cancellationToken) :
+            await GetReportGenerationStateForBroker(cancellationToken);
+
+    async Task<ReportGenerationState> GetReportGenerationStateForBroker(CancellationToken cancellationToken)
     {
-        var reportGenerationState = new ReportGenerationState(transport)
+        var reportCanBeGenerated = await dataStore.IsThereThroughputForLastXDaysForSource(30, ThroughputSource.Broker, false, cancellationToken);
+        var reason = reportCanBeGenerated ? string.Empty : $"Need at least one day of usage data in the last 30 days from ${throughputQuery!.MessageTransport} broker.";
```
johnsimons, line 100: Still not exactly like what SP does https://github.com/Particular/ServicePulse/blob/ca4c4c8700a53051052151a13a0e86ef21189d83/src/Frontend/src/stores/ThroughputStore.ts#L54, but I guess it is better

johnsimons: for Sql Server actually displays "SqlTransport" 😞

johnsimons: I think I would make this generic and just say "Need at least one day of usage data in the last 30 days from the configured broker."

src/Particular.LicensingComponent/ThroughputCollector.cs
```diff
@@ -89,19 +89,33 @@ public async Task<List<EndpointThroughputSummary>> GetThroughputSummary(Cancella
...
+    }
 
-        if (!reportGenerationState.ReportCanBeGenerated)
-        {
-            reportGenerationState.Reason = "24 hours worth of data needs to exist in the last 30 days.";
-        }
+    async Task<ReportGenerationState> GetReportGenerationStateForNonBroker(CancellationToken cancellationToken)
+    {
+        var reportCanBeGenerated = await dataStore.IsThereThroughputForLastXDays(30, cancellationToken);
+        var reason = reportCanBeGenerated ? string.Empty : $"Need at least one day of usage data in the last 30 days.";
```
johnsimons, line 112: So should this say something like 
```suggestion
        var reason = reportCanBeGenerated ? string.Empty : $"Need at least one day of usage data in the last 30 days from audits and/or monitoring instances.";
```

jpalac: I think that would be too long

johnsimons: yeah ok

johnsimons (approved):

PR #4317, opened 2024-07-22
dont query star for SQL table check

SimonCropp:
also use raw string to make the query string alloc smaller


SimonCropp:
looks like the failure is unrelated to my change?

PR #4328, opened 2024-07-22
cache sqltable query

SimonCropp:
* cache sql QueueLengthQuery on table to avoid calculating it every time in the fast path
 * use _fullTableName in calculation of query and re-use it in quals, hashcode, and tostring
 * remove SqlNameHelper.Unquote since not used
 * enabled nullable checks for SqlNameHelper and SqlTable


DavidBoike:
This looks like a good change but we decided we didn't have time to fully kick the tires and still release 5.5 today. Should be considered for the next release.

SimonCropp:
@DavidBoike can u re open and  disable that stale PR bot. clearly the Particular process moves slower than that bot closes PRs

DavidBoike:
Your comment alone would have resulted in the `stale` label being removed within 24 hours, but thanks for the reminder, I looked at this and I'm going to get it shipped in 5.7.0 next week.

PR #4354, opened 2024-08-08
Use chiseled base image

DavidBoike:
Changing the base image for the three application images from `mcr.microsoft.com/dotnet/aspnet:8.0` to `mcr.microsoft.com/dotnet/aspnet:8.0-noble-chiseled-composite` decreases the overall image size (as reported by `docker images`) drastically. Sizes are from the `linux/amd64` architecture image.

| Image | Previous Image Size | Updated Size |
|-|:-:|:-:|
| particular/servicecontrol | 391.57 MB | 213.54 MB |
| particular/servicecontrol-audit | 385.85 MB | 207.81 MB |
| particular/servicecontrol-monitoring | 375.04 MB | 197.01 MB |

The difference of 178.03 MB in each case is the difference between in size between base images.

### DIFF
--- a/src/HealthCheckApp/HealthCheckApp.csproj
+++ b/src/HealthCheckApp/HealthCheckApp.csproj
@@ -1,0 +1,16 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <AssemblyName>healthcheck</AssemblyName>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <InvariantGlobalization>true</InvariantGlobalization>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Artifact Include="$(OutputPath)" FileExclude="ServiceControl.dll.config;nsb*.txt" DirExclude=".db;.logs" DestinationFolder="$(ArtifactsPath)healthcheck" />
+  </ItemGroup>
+
+</Project>
--- a/src/HealthCheckApp/HealthCheckApp.csproj
+++ b/src/HealthCheckApp/HealthCheckApp.csproj
@@ -1,16 +1,11 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <AssemblyName>healthcheck</AssemblyName>
-    <OutputType>Exe</OutputType>
     <TargetFramework>net8.0</TargetFramework>
+    <OutputType>Exe</OutputType>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
-    <InvariantGlobalization>true</InvariantGlobalization>
+    <AssemblyName>healthcheck</AssemblyName>
   </PropertyGroup>
-
-  <ItemGroup>
-    <Artifact Include="$(OutputPath)" FileExclude="ServiceControl.dll.config;nsb*.txt" DirExclude=".db;.logs" DestinationFolder="$(ArtifactsPath)healthcheck" />
-  </ItemGroup>
 
 </Project>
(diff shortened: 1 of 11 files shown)


src/ServiceControl.Monitoring/Dockerfile
```diff
@@ -26,9 +25,9 @@ LABEL org.opencontainers.image.source=https://github.com/Particular/ServiceContr
 EXPOSE 33633
 
 COPY --from=build /deploy/Particular.ServiceControl.Monitoring ./app
+COPY --from=build /deploy/healthcheck ./healthcheck
```
bording, line 28:
```suggestion
COPY --from=build /healthcheck /healthcheck
```

src/ServiceControl/Dockerfile
```diff
@@ -25,12 +24,13 @@ LABEL org.opencontainers.image.source=https://github.com/Particular/ServiceContr
 
 EXPOSE 33333
 COPY --from=build /deploy/Particular.ServiceControl ./app
+COPY --from=build /deploy/healthcheck ./healthcheck
```
bording, line 27:
```suggestion
COPY --from=build /healthcheck /healthcheck
```

src/HealthCheckApp/HealthCheckApp.csproj
```diff
@@ -0,0 +1,16 @@
...
+    <AssemblyName>healthcheck</AssemblyName>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <InvariantGlobalization>true</InvariantGlobalization>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Artifact Include="$(OutputPath)" FileExclude="ServiceControl.dll.config;nsb*.txt" DirExclude=".db;.logs" DestinationFolder="$(ArtifactsPath)healthcheck" />
```
bording, line 13: We can remove this entirely if we decide to take my suggested changes to the Dockerfiles.

src/HealthCheckApp/HealthCheckApp.csproj
```diff
@@ -0,0 +1,16 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <AssemblyName>healthcheck</AssemblyName>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
```
bording, line 6: Nit: Let's move this up to the top, and move `AssemblyName` to the bottom.

bording (changes requested):

src/HealthCheckApp/Program.cs
```diff
@@ -0,0 +1,19 @@
...
+
+var response = await http.GetAsync(url).ConfigureAwait(false);
+response.EnsureSuccessStatusCode();
+
+if (response.Content.Headers.ContentType?.MediaType != "application/json" || response.Content.Headers.ContentLength == 0)
+{
+    throw new Exception("Invalid ContentType or Length");
+}
+
+Console.WriteLine("OK");
```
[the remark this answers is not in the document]
DavidBoike: You do get this:

```
"Log": [
{
	"Start": "2024-08-08T22:00:18.440376136Z",
	"End": "2024-08-08T22:00:18.61019214Z",
	"ExitCode": 0,
	"Output": "OK\n"
},
{
	"Start": "2024-08-08T22:00:48.612150175Z",
	"End": "2024-08-08T22:00:48.786716911Z",
	"ExitCode": 0,
	"Output": "OK\n"
},
{ …
```

You can't get away from having the Output property in there though, such as `"Output": ""` as we did previously when we were sending the wget output to /dev/null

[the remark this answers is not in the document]
boblangley: > To help debug failing probes, any output text (UTF-8 encoded) that the command writes on stdout or stderr will be stored in the health status and can be queried with docker inspect. Such output should be kept short (only the first 4096 bytes are stored currently).

It doesn't look like this goes into the normal logs and should be OK.

boblangley (approved):

bording (approved):

boblangley (approved):

PR #4360, opened 2024-08-12
Bump version to 5.7

WilliamBZA:


src/Custom.Build.props
```diff
@@ -15,7 +15,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <MinVerMinimumMajorMinor>5.4</MinVerMinimumMajorMinor>
+    <MinVerMinimumMajorMinor>5.6</MinVerMinimumMajorMinor>
```
ramonsmits, line 18: Shouldn't master be bumped to 5.7?

ramonsmits: Maybe we should ask David/Brandon for a review?

ramonsmits:
Currently builds on master have the wrong tag:

- https://github.com/Particular/ServiceControl/pkgs/container/servicecontrol

Last is `5.6.0-alpha.0.19` but that likely should have been something like `5.7.0-alpha.0.1`.

src/Custom.Build.props
ramonsmits, line 18: @DavidBoike @bording Shouldn't master be bumped to 5.7?

```suggestion
    <MinVerMinimumMajorMinor>5.7</MinVerMinimumMajorMinor>
```

bording (approved):

PR #4361, opened 2024-08-12
Updates to packaging/deployment markdown files

DavidBoike:


docs/deployment.md
```diff
@@ -1,24 +1,39 @@
 # Deployment
 
-All ServiceControl deployment options listed below rely on the files produced by [packaging](packaging.md).
+All ServiceControl deployment options listed below rely on the files produced by [packaging](packaging.md). During packaging, deployment binaries are created in the `deploy` directory, some of which are packaged into zip files in the `zips` folder, and other deployment artifacts are built from these pieces.
 
 ## ServiceControl installer
 
-A [windows installer is provided for download](https://docs.particular.net/servicecontrol/installation) containing both ServiceControl Management Utility (SCMU) and [Powershell commands](https://docs.particular.net/servicecontrol/powershell).
+The zips are packaged as embedded resources in the ServiceControl Management Utility (the "installer").
```
bording, line 7: As mentioned in the packaging doc, they are actually embedded in the installer assembly, not directly in the SCMU exe. Does this matter for this documentation? Not sure,

docs/deployment.md
```diff
@@ -1,24 +1,39 @@
...
-All ServiceControl deployment options listed below rely on the files produced by [packaging](packaging.md).
+All ServiceControl deployment options listed below rely on the files produced by [packaging](packaging.md). During packaging, deployment binaries are created in the `deploy` directory, some of which are packaged into zip files in the `zips` folder, and other deployment artifacts are built from these pieces.
 
 ## ServiceControl installer
 
-A [windows installer is provided for download](https://docs.particular.net/servicecontrol/installation) containing both ServiceControl Management Utility (SCMU) and [Powershell commands](https://docs.particular.net/servicecontrol/powershell).
+The zips are packaged as embedded resources in the ServiceControl Management Utility (the "installer").
 
-When installing an instance using one of the above method instance files, persister (if applicable) and transport files are unzipped into the same folder.
+When installing an instance using one of the above method instance files, the following binaries are unzipped together to form a complete instance:
```
bording, line 9: > When installing an instance using one of the above method instance files

Huh?

docs/deployment.md
```diff
@@ -1,24 +1,39 @@
...
+
+- Application & persistence files
+- Selected transport
+- RavenDB server files (ServiceControl and Audit only)
```
bording, line 13: This list of files doesn't seem to line up with what actually gets unzipped.

For example, all transports are on disk, not just the one selected.

docs/deployment.md
```diff
@@ -1,24 +1,39 @@
...
 
 A configuration file generated based on:
 
 - User input (and defaults)
-- Contents of the `persister.manifest` file of the selected persister
-
-## NuGet package
+- Contents of the `transport.manifest` and `persister.manifest` file for the selected transport and persister.
```
bording, line 18: I know a version of this line already existed, but in what way are the contents of the manifest files used to create a config file?

docs/packaging.md
```diff
@@ -1,15 +1,14 @@
...
-- `/Transports` contains a folder for each supported transport
-- `/Persisters` contains a folder for each supported persister (Only the Audit instance use this currently)
-- `/<product>` contains the instance files specific to the type of instance
-  - `/ServiceControl`
-  - `/ServiceControl.Audit`
-  - `/ServiceControl.Monitoring`
+- `ServiceControl.zip`
+- `ServiceControl.Audit.zip`
+- `ServiceControl.Monitoring.zip`
+- `InstanceShared.zip` - Contains the transport assemblies that are shared by each instance
```
bording, line 10: This will contain any files that are shared across all 3 instance zips, not just transports.

bording (changes requested):

docs/deployment.md
DavidBoike: Yeah just meant to mean "In there somewhere" and I can't think of a better way to say it without repeating everything in packaging.md.

docs/deployment.md
DavidBoike: Added a couple lines, should be closer to the full truth now.

docs/deployment.md
bording: Maybe this could just say that SCMU is self-contained single exe that has everything inside it?

DavidBoike: This good?
```suggestion
The zips are packaged as embedded resources in the ServiceControl Management Utility. Although technicall embedded in a specific assembly, ServiceControl Management is shipped as a self-contained, single-file executable that has everything inside it.
```

bording (approved):

Issue #3980, opened 2024-02-22
DiagramApiController.GetSingleEndpointMetrics throws an error for certain messages


kbaley:
#### Description

For certain messages where NServiceBus.ContentType is `TaggedLongValueOccurrence`, [this line in DiagramApiController](https://github.com/Particular/ServiceControl/blob/c03f2ae3acb39a91aa5a162c3e84f61662ccd750/src/ServiceControl.Monitoring/Http/Diagrams/DiagramApiController.cs#L149), reports null and fails. This causes ServicePulse to show an error that it can't connect to ServiceControl on the monitoring page to view metrics for a single endpoint.

The cause appears to be a malformed body for these messages. When the body is [deserialized](https://github.com/Particular/ServiceControl/blob/c03f2ae3acb39a91aa5a162c3e84f61662ccd750/src/ServiceControl.Monitoring/Messaging/TaggedLongValueWriterOccurrenceSerializerDefinition.cs#L50) for these malformed messages, it looks like one of the types it is trying to parse is empty. So the current theory is that when the TaggedLongValueOccurrence message is initially created, it's not able to get all the types it needs.

#### Expected behavior

The [GetSingleEndpointMetrics](https://github.com/Particular/ServiceControl/blob/c03f2ae3acb39a91aa5a162c3e84f61662ccd750/src/ServiceControl.Monitoring/Http/Diagrams/DiagramApiController.cs#L56) controller method shouldn't report an HTTP exception.

#### Actual behavior

The method throws a NullReferenceException

#### Versions

5.0.3-5.0.5 for sure. Probably others close to this version.       

### Steps to reproduce

For endpoints that exhibit this behaviour, it starts right after it starts processing messages. If the SC monitoring instance is restarted, the monitoring page for the endpoint works until messages start getting reported.

### Relevant log output

```shell
Visiting /monitored-endpoints/<EndpointName>?history=5 for a problem endpoint locally (i.e. using localhost on the SC server itself) gives the following error (for version 5.0.3):


{
"message": "An error has occurred.",
"exceptionMessage": "Object reference not set to an instance of an object.",
"exceptionType": "System.NullReferenceException",
[...]

PR #4365, opened 2024-08-13
handle messagetype being null

PhilBastian:
fixes the symptoms of #3980 

handle messageType being null when processing an endpoint. Note that any null endpoints get filtered out of the displayed results set, in [messageTypes.ts](https://github.com/Particular/ServicePulse/blob/5af9e3a68b66b9a7ba750cd937c0af42353b97e4/src/Frontend/src/components/monitoring/messageTypes.ts#L54):

```js
      // filter out system message types
      .filter((mt) => mt.id && mt.typeName)
```

There is another discussion to be held as to whether unknown and uninferrable messagetypes should be passed through as null, since this prevents metrics being shown for these messages, but this is an imperfect first step to prevent an exception when handling this case

### DIFF
--- a/src/ServiceControl.Monitoring/Infrastructure/EndpointMessageType.cs
+++ b/src/ServiceControl.Monitoring/Infrastructure/EndpointMessageType.cs
@@ -4,7 +4,7 @@
     {
         public EndpointMessageType(string endpointName, string enclosedMessageTypes)
         {
-            var index = enclosedMessageTypes.IndexOf(';');
+            var index = (enclosedMessageTypes ?? "").IndexOf(';');
 
             var firstType = index != -1
                 ? enclosedMessageTypes.Substring(0, index)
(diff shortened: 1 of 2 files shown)


src/ServiceControl.Monitoring/Infrastructure/EndpointMessageType.cs
```diff
@@ -4,7 +4,7 @@
     {
         public EndpointMessageType(string endpointName, string enclosedMessageTypes)
         {
-            var index = enclosedMessageTypes.IndexOf(';');
+            var index = (enclosedMessageTypes ?? "").IndexOf(';');
 
             var firstType = index != -1
```
johnsimons, line 9: I assume it is ok to set `MessageType` to `null`?

johnsimons (approved):
A small comment regarding what are the implications of having `MessageType` set top `null`, but otherwise it makes sense

src/ServiceControl.Monitoring/Infrastructure/EndpointMessageType.cs
PhilBastian: it is handled fine by ServicePulse, which filters out such message types

PhilBastian: added filter servicecontrol side (any metrics gathered against a null messagetype will be discarded anyway)

johnsimons:
@PhilBastian i thought we were going to see if we could filter out these messages at deserialisation time?

PR #4383, opened 2024-08-19
Auto-expire disconnected instance in Monitoring screens

johnsimons:
Auto-prunes endpoint instances that have not reported any metrics for 1 hour

This is part of fixing https://github.com/Particular/ServicePulse/issues/1257

### DIFF
--- a/src/ServiceControl.Monitoring/Infrastructure/HistoryPeriod.cs
+++ b/src/ServiceControl.Monitoring/Infrastructure/HistoryPeriod.cs
@@ -63,7 +63,7 @@
             new(TimeSpan.FromMinutes(10), numberOfIntervals: 60, delayedIntervals: 1),
             new(TimeSpan.FromMinutes(15), numberOfIntervals: 60, delayedIntervals: 1),
             new(TimeSpan.FromMinutes(30), numberOfIntervals: 60, delayedIntervals: 1),
-            new HistoryPeriod(TimeSpan.FromMinutes(LargestHistoryPeriod), 60, 1)
+            new(TimeSpan.FromMinutes(LargestHistoryPeriod), 60, 1)
         ];
 
         public const int LargestHistoryPeriod = 60;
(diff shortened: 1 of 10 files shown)


src/ServiceControl.Monitoring/Infrastructure/RemoveExpiredEndpointInstances.cs
```diff
@@ -0,0 +1,56 @@
...
+                try
+                {
+                    foreach (MonitoredEndpoint monitoredEndpoint in metricsApi.GetAllEndpointsMetrics())
+                    {
+                        foreach (string endpointInstanceId in monitoredEndpoint.EndpointInstanceIds)
+                        {
+                            if (activityTracker.IsExpired(new EndpointInstanceId(monitoredEndpoint.Name,
+                                    endpointInstanceId)))
+                            {
+                                metricsApi.DeleteEndpointInstance(monitoredEndpoint.Name, endpointInstanceId);
```
awright18, line 39: If the Endpoint Instance is removed, will this affect the 30 days' worth of historical data gathered by the licensing component?

johnsimons: Good question @awright18 
It is not affected, we collect that data every 5 minutes, so whatever we get that is what we send to the error instance, the error instance saved the data in storage.
So when an endpoint instance is removed, it just means we no longer send data for that endpoint instance which is pretty much what we are currently doing except we currently send zero.

src/ServiceControl.Monitoring/Infrastructure/HistoryPeriod.cs
```diff
@@ -63,7 +63,9 @@ public override bool Equals(object obj)
             new(TimeSpan.FromMinutes(10), numberOfIntervals: 60, delayedIntervals: 1),
             new(TimeSpan.FromMinutes(15), numberOfIntervals: 60, delayedIntervals: 1),
             new(TimeSpan.FromMinutes(30), numberOfIntervals: 60, delayedIntervals: 1),
-            new(TimeSpan.FromMinutes(60), numberOfIntervals: 60, delayedIntervals: 1)
+            new HistoryPeriod(TimeSpan.FromMinutes(LargestHistoryPeriod), 60, 1)
```
PhilBastian, line 66: either explicitly specify the type of every item, or remove from this one

PhilBastian (approved):

src/ServiceControl.Monitoring/Infrastructure/HistoryPeriod.cs
johnsimons:
```suggestion
            new(TimeSpan.FromMinutes(LargestHistoryPeriod), 60, 1)
```

PR #4403, opened 2024-08-27
Exit setup-and-run mode if setup fails

DavidBoike:
### Symptoms

When using the `--setup-and-run`, if something causes the setup phase to fail, ServiceControl will still proceed to the run phase, which could then result in additional errors related to missing queues or other environmental mismatches.

### Who's affected

Anyone using `--setup-and-run` option, most likely deploying ServiceControl using containers.

### Root cause

Failure to check the return code of the process executing the setup phase.

### DIFF
--- a/src/ServiceControl.Infrastructure/IntegratedSetup.cs
+++ b/src/ServiceControl.Infrastructure/IntegratedSetup.cs
@@ -11,14 +11,14 @@
         const string SetupAndRunCmd = "--setup-and-run";
         const string SetupCmd = "--setup";
 
-        public static async Task Run()
+        public static async Task<int> Run()
         {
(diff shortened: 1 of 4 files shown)


src/ServiceControl.Infrastructure/IntegratedSetup.cs
```diff
@@ -52,7 +52,12 @@ public static Task Run()
             process.BeginOutputReadLine();
             process.BeginErrorReadLine();
 
-            return process.WaitForExitAsync();
+            await process.WaitForExitAsync().ConfigureAwait(false);
+
+            if (process.ExitCode != 0)
+            {
+                Environment.Exit(process.ExitCode);
```
bording, line 59: I'm wondering if it would be better to just return the exit code from this method instead, and let then let the caller `return exitCode` when non-zero.

bording (approved):

PR #4411, opened 2024-09-01
Ensuring Azure ManagementUrl setting accepts urls without the last slash

johnsimons:
It seems the logic did not cater for url without the last slash so urls like `https://management.azure.com/` and `https://management.azure.com` did not match.

This PR ensures those match


jpalac (commented):
The URLs in the description are the same - should they be?
Also - on [this line](https://github.com/Particular/ServiceControl/pull/4411/files#diff-62388c8a9373994d3a79fc7e4753c6dedbb030368d79ee4b76721cedb805b62dR110) - should it be checking and/or printing the parsed url?

johnsimons:
> The URLs in the description are the same - should they be?

Fixed the description, I think github was trying to be smart

johnsimons:
> Also - on [this line](https://github.com/Particular/ServiceControl/pull/4411/files#diff-62388c8a9373994d3a79fc7e4753c6dedbb030368d79ee4b76721cedb805b62dR110) - should it be checking and/or printing the parsed url?

Yeah that is fine, we display what the user entered

jpalac (approved):

PR #4435, opened 2024-09-10
The frontend needs to be aware when no plugin is sending heartbeats

johnsimons:
So that we can display "no plugin installed" in the frontend

### DIFF
--- a/src/ServiceControl/Monitoring/EndpointInstanceMonitor.cs
+++ b/src/ServiceControl/Monitoring/EndpointInstanceMonitor.cs
@@ -117,11 +117,11 @@ namespace ServiceControl.Monitoring
                 Monitored = Monitored,
                 MonitorHeartbeat = Monitored,
 
-                HeartbeatInformation = new HeartbeatInformation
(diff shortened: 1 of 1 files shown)


src/ServiceControl/Monitoring/EndpointInstanceMonitor.cs
```diff
@@ -117,11 +117,11 @@ public EndpointsView GetView()
                 Monitored = Monitored,
                 MonitorHeartbeat = Monitored,
 
-                HeartbeatInformation = new HeartbeatInformation
+                HeartbeatInformation = lastSeen.HasValue ? new HeartbeatInformation
                 {
                     ReportedStatus = status == HeartbeatStatus.Alive ? HeartbeatMonitoringStatus.Beating : HeartbeatMonitoringStatus.Dead,
                     LastReportAt = lastSeen ?? DateTime.MinValue
```
PhilBastian, line 123:
```suggestion
                    LastReportAt = lastSeen.Value
```

PhilBastian (approved):

PR #4446, opened 2024-09-12
Load trialstartdate from ravenDB

WilliamBZA:
A spike to workshop an idea of how to validate the trial start date from RavenDB.

Overrides the active license if the license is a trial license, and if the start dates don't match.

We could also add hashing and date range checking to make this more thorough.

### DIFF
--- a/src/ServiceControl.Persistence.Tests.RavenDB/LicenseMetadata/LicenseMetadataTests.cs
+++ b/src/ServiceControl.Persistence.Tests.RavenDB/LicenseMetadata/LicenseMetadataTests.cs
@@ -1,0 +1,36 @@
+﻿namespace ServiceControl.Persistence.Tests.RavenDB.Operations
+{
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+    using Microsoft.Extensions.DependencyInjection;
+    using NUnit.Framework;
+    using ServiceControl.Persistence.RavenDB;
+
+    [TestFixture]
+    class LicenseMetadataTests : RavenPersistenceTestBase
+    {
+        public LicenseMetadataTests() =>
+            RegisterServices = services =>
+            {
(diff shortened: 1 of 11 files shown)


src/ServiceControl.Persistence/LicenseMetadata.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace ServiceControl.Persistence
+{
+    using System;
+
+    public class LicenseMetadata
```
SzymonPobiega, line 5: Nitpicking: why don't we call it based on what it is -- the trial information?

src/ServiceControl.Persistence.Tests.RavenDB/LicenseMetadata/LicenseMetadataTests.cs
```diff
@@ -0,0 +1,36 @@
+﻿namespace ServiceControl.Persistence.Tests.RavenDB.Operations
```
WilliamBZA, line 1:
```suggestion
﻿namespace ServiceControl.Persistence.Tests.RavenDB.LicenseMetadata
```

irinascurtu (approved):

DavidBoike:
Is this really the best idea? There's nothing container-specific about this, so this will affect Windows-hosted ServiceControl instances as well. With the introduction of this PR, ServiceControl will have a completely separate concept of licensing than anything else running on the system. I could see this becoming pretty confusing.

Plus if we ever want to create new storage options for ServiceControl, this is now an additional use case that must be covered in every single one.

I thought the solution to containers "forgetting" about a trial license was https://github.com/Particular/ServicePulse/pull/2027 ? Why was that not sufficient? 

I understand that with the possibility that someone may not be running endpoints, there has to be some way to replicate the "F5 and browser tab opens" license reminder experience. But ServicePulse still seems like a good _single_ place to implement that. Was using the browser's `localStorage` explored?

PR #4455, opened 2024-09-19
Ingest control messages to update list of connected applications

WilliamBZA:
This would allow connected applications to send control messages to the error queue which will then be ingested by ServiceControl


irinascurtu (approved):

src/ServiceControl/Licensing/LicenseController.cs
```diff
@@ -52,6 +53,8 @@ public class LicenseInfo
             public string InstanceName { get; set; }
 
             public string LicenseStatus { get; set; }
+
+            public string[] ConnectedApplications { get; set; }
```
tmasternak, line 57: Should this be part of the license information or maybe part of the root controller instead? This is a runtime information and for now does not influence the type/content of the license.

SzymonPobiega: I would say it should be a persisted collection because who knows what we are going to do with it in the future ;-)

SzymonPobiega:
I generally prefer the enricher approach https://github.com/Particular/ServiceControl/pull/4456

PR #4456, opened 2024-09-19
Use message enricher to extract connected applications

WilliamBZA:
An alternative to the following where the enricher looks for specific headers to determine which applications are connected:

- #4455
- [ ] Change connected applications to be mass transit


irinascurtu (approved):

src/ServiceControl/Licensing/LicenseController.cs
```diff
@@ -27,7 +27,8 @@ public ActionResult<LicenseInfo> License(bool refresh)
                 Status = activeLicense.IsValid ? "valid" : "invalid",
                 LicenseType = activeLicense.Details.LicenseType ?? string.Empty,
                 InstanceName = settings.InstanceName ?? string.Empty,
-                LicenseStatus = activeLicense.Details.Status
+                LicenseStatus = activeLicense.Details.Status,
+                ConnectedApplications = settings.ConnectedApplications.ToArray()
```
SzymonPobiega, line 31: Should we make it a persistent document with an in-memory cache to avoid writes if we already have a specific value there?

WilliamBZA: Connected applications are now persisted, and they've been moved off of the license object to the root controller

irinascurtu (approved):

src/ServiceControl/Operations/ConnectedApplicationsEnricher.cs
```diff
@@ -0,0 +1,27 @@
...
+        static List<string> MassTransitHeaders =
+        [
+            "MassTransitHeader1",
+            "MassTransitHeader2",
+            "MassTransitHeader3"
+        ];
+
+        public void Enrich(ErrorEnricherContext context)
+        {
+            if (!alreadyAddedMassTransit && context.Headers.Any(incomingHeader => MassTransitHeaders.Contains(incomingHeader.Key)))
```
ramonsmits, line 18: This is confusing. It tries to be both generic (`ConnectedApplicationsEnricher`) and specific (`alreadyAddedMassTransit`)

I think we should go: 

1. Be specific and have an enricher per "application/component/feature" and check for the existence of any `MT-%` or probe a subset of headers like the current implementation and call this `MatchMassTransitEnricher`
2. A bit more generic look for a specific header like `Feature` and capture all values but that requires the sender to add a header that contains the product identifier
3. Make it even more generic and iterate over a bunch of registered `FeatureProbe` and for each match store each corresponding "identifier" so that the task of the enricher is to take care of storing and caching any matches and have very simple "probe" implementations

WilliamBZA: I envisaged this extending to other applications. e.g.

```cs
        static string MassTransitHeaderPrefix = "MT-";
        static string MicrosoftEventingFrameworkHeaderPrefix = "MS-";

        public void Enrich(ErrorEnricherContext context)
        {
            if (!alreadyAddedMassTransit && context.Headers.Any(incomingHeader => incomingHeader.Key.StartsWith(MassTransitHeaderPrefix, System.StringComparison.InvariantCultureIgnoreCase)))
            {
                alreadyAddedMassTransit = true;
                _ = connectedApplicationsDataStore.Add("MassTransitConnector");
            } else if (!alreadyAddedMicrosoftEventingFramework && context.Headers.Any(incomingHeader => incomingHeader.Key.StartsWith(MicrosoftEventingFrameworkHeaderPrefix , System.StringComparison.InvariantCultureIgnoreCase)))
            {
                alreadyAddedMicrosoftEventingFramework = true;
                _ = connectedApplicationsDataStore.Add("MicrosoftEventingFramework");
            }
        }

        static bool alreadyAddedMassTransit = false;
        static bool alreadyAddedMicrosoftEventingFramework = false;

```

tmasternak (approved):

src/ServiceControl/Operations/MassTransitDetector.cs
```diff
@@ -0,0 +1,27 @@
...
+    class MassTransitDetector(IConnectedApplicationsDataStore connectedApplicationsDataStore) : IEnrichImportedErrorMessages
+    {
+        public void Enrich(ErrorEnricherContext context)
+        {
+            if (alreadyDetected)
+            {
+                return;
+            }
+
+            if (context.Headers.Any(h => IsMassTransitHeader(h.Key)))
```
johnsimons, line 15: Could there be contention here in multi-threads?

PR #4484, opened 2024-09-27
Update to RavenDB 6.2

bording:
This PR makes the following changes related to updating ServiceControl to use RavenDB 6.2:

- When running in external RavenDB mode, the check to prevent startup when a cluster is detected has been removed.
- When running in external RavenDB mode, the version startup check no longer requires an exact version match between server and client. It now verifies that the server product version is >= client product version. This aligns with the [RavenDB Client/Server Compatibility](https://ravendb.net/docs/article-page/6.2/csharp/client-api/faq/backward-compatibility#compatibility---ravendb-4.2-and-higher) statment.
- All indexes will now use the Corax search engine by default. Existing indexes currently using Lucene will be updated after they have been reset.
- The `servicecontrol-ravendb` container has been updated to ensure that RavenDB data has been correctly modified to be migrated from RavenDB 5 to RavenDB 6.


src/ServiceControl.Persistence.RavenDB/DatabaseSetup.cs
```diff
@@ -29,16 +29,17 @@ public async Task Execute(CancellationToken cancellationToken)
...
 
             if (dbRecord == null)
             {
                 try
                 {
-                    await documentStore.Maintenance.Server.SendAsync(
-                        new CreateDatabaseOperation(new DatabaseRecord(databaseName)), cancellationToken);
+                    var databaseRecord = new DatabaseRecord(databaseName);
+                    databaseRecord.Settings.Add("Indexing.Auto.SearchEngineType", "Corax");
+                    databaseRecord.Settings.Add("Indexing.Static.SearchEngineType", "Corax");
```
DavidBoike, line 40: Would this affect only brand new databases or would it amend existing databases as well?

bording: Only new ones. We need to add something more if we want to update existing databases also.

DavidBoike: Well, an existing database is going to have all the indexes anyway, unless we were to add one which we're not currently planning, so I guess it's a difference without distinction really.

.gitattributes
```diff
@@ -1,2 +1,3 @@
 # Auto detect text files and perform LF normalization
 * text=auto
+*.sh text eol=lf
```
DavidBoike, line 3: We need to add this to RepoStandards

src/ServiceControl.Audit.Persistence.RavenDB/DatabaseSetup.cs
```diff
@@ -1,69 +1,74 @@
...
-            try
-            {
-                await documentStore.Maintenance.ForDatabase(configuration.Name).SendAsync(new GetStatisticsOperation(), cancellationToken);
-            }
-            catch (DatabaseDoesNotExistException)
+            await CreateDatabase(documentStore, configuration.Name, cancellationToken);
+            await UpdateDatabaseSettings(documentStore, configuration.Name, cancellationToken);
+
+            await DeleteLegacySagaDetailsIndex(documentStore, cancellationToken);
+            await CreateIndexes(documentStore, cancellationToken);
```
DavidBoike, line 26: Should we be concerned at all that these 2 things are order-dependent? Because the legacy saga details index has the same name as the one that CreateIndexes will create. It's just whether it includes a specific argument in the Reduce definition that's different.

Should this be part of CreateIndexes() with a comment about why it has to go first?

src/ServiceControl.Persistence.RavenDB/DatabaseSetup.cs
```diff
@@ -16,35 +18,66 @@ public async Task Execute(CancellationToken cancellationToken)
...
 
-            var expirationConfig = new ExpirationConfiguration
-            {
-                Disabled = false,
-                DeleteFrequencyInSec = settings.ExpirationProcessTimerInSeconds
-            };
+            await IndexCreation.CreateIndexesAsync(typeof(DatabaseSetup).Assembly, documentStore, null, null, cancellationToken);
 
-            await documentStore.Maintenance.SendAsync(new ConfigureExpirationOperation(expirationConfig), cancellationToken);
+            await ConfigureExpiration(settings, cancellationToken);
```
DavidBoike, line 26: Does this have immediate effect, and not require a database reset that UpdateDatabaseSettings() is doing as well? (Same for Audit above.)

src/ServiceControl.Audit.Persistence.RavenDB/DatabaseSetup.cs
```diff
@@ -1,69 +1,74 @@
...
+
+            await ConfigureExpiration(documentStore, cancellationToken);
```
DavidBoike, line 28: Does this have immediate effect, and not require a database reset that UpdateDatabaseSettings() is doing as well? (Same for Primary below)

src/ServiceControl.RavenDB/EmbeddedDatabase.cs
```diff
@@ -146,7 +147,8 @@ public async Task<IDocumentStore> Connect(CancellationToken cancellationToken)
                 Conventions = new DocumentConventions
                 {
                     SaveEnumsAsIntegers = true
-                }
+                },
+                SkipCreatingDatabase = true
```
DavidBoike, line 151: I assume this is so we can create the DB(s) with Corax settings in the DatabaseSetup, correct?

src/ServiceControlInstaller.Engine.UnitTests/Validation/SelfContainedRavenDBTest.cs
```diff
@@ -15,16 +15,16 @@ public void CheckForSelfContainedRavenDB()
...
+            var ravenServerExe = Path.Combine(ravenServerPath, "Raven.Server.exe");
+            var offlineOperationsUtility = Path.Combine(ravenServerPath, "rvn.exe");
 
             Assert.Multiple(() =>
             {
                 Assert.That(ravenStudio, Does.Exist); // No matter what
-                Assert.That(Directory.Exists(runtimes), Is.EqualTo(isLocal));  // Only in local development
-                Assert.That(File.Exists(systemCollections), Is.EqualTo(isCI)); // Only on CI
-                Assert.That(File.Exists(aspNetCoreHttp), Is.EqualTo(isCI));    // Only on CI
+                Assert.That(Directory.Exists(ravenServerDll), Is.EqualTo(isLocal));  // Only in local development
```
DavidBoike, line 25: _**Directory**_.Exists("some-path\Raven.Server.dll") ? Directory.Exists to a DLL file?

src/ServiceControl.RavenDB/sc-container-startup.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+set -euo pipefail
+
+echo "Checking RavenDB data for proper migration to RavenDB 6"
+
+LEGACY_PATH="/opt/RavenDB/Server/RavenData"
+
+if [[ -d "$LEGACY_PATH" ]]; then
+  echo "RavenDB data is being mounted to the wrong location. See the upgrade guide for details on how to fix this."
```
DavidBoike, line 9:
```suggestion
  echo "RavenDB data is being mounted to the wrong location. Refer to the article 'Upgrade ServiceControl from Version 5 to Version 6' in the documentation for steps that must be taken to update ServiceControl to this version.
```

src/ServiceControl.RavenDB/sc-container-startup.sh
```diff
@@ -0,0 +1,29 @@
...
+  exit 1
+fi
+
+FILECOUNT="$(find "$RAVEN_DataDir" \! -writable -printf '.' | wc -c)"
+
+if [[ $FILECOUNT -ne 0 ]]; then
+  echo "RavenDB data is not writable by the ravendb user. See the upgrade guide for details on how to fix this."
```
DavidBoike, line 16:
```suggestion
  echo "RavenDB data in the mounted data volume is not writable by the ravendb user. Refer to the article 'Upgrade ServiceControl from Version 5 to Version 6' in the documentation for steps that must be taken to update ServiceControl to this version."
```

src/ServiceControl.RavenDB/sc-container-startup.sh
```diff
@@ -0,0 +1,29 @@
...
+  exit 1
+fi
+
+FILECOUNT="$(find "$RAVEN_DataDir" \! -readable -printf '.' | wc -c)"
+
+if [[ $FILECOUNT -ne 0 ]]; then
+  echo "RavenDB data is not readable by the ravendb user. See the upgrade guide for details on how to fix this."
```
DavidBoike, line 23:
```suggestion
  echo "RavenDB data in the mounted data volume is not readable by the ravendb user. Refer to the article 'Upgrade ServiceControl from Version 5 to Version 6' in the documentation for steps that must be taken to update ServiceControl to this version."
```

DavidBoike (commented):
Here's my suggestions for (hopefully) better preflight error messages.

src/ServiceControl.Audit.Persistence.RavenDB/DatabaseSetup.cs
bording: I think the comment in the method itself is enough to explain why it needs to go first, so but I've gone ahead and moved the call to be inside `CreateIndexes`.

src/ServiceControl.Persistence.RavenDB/DatabaseSetup.cs
bording: It was in this order before, so I don't know if it takes effect immediately or not. I would hope so, but I'll check.

src/ServiceControl.RavenDB/EmbeddedDatabase.cs
bording: Yes, with this not being set before, we never actually hitting the path in `CreateDatabase` that would... create the database.

src/ServiceControlInstaller.Engine.UnitTests/Validation/SelfContainedRavenDBTest.cs
bording: Yeah, should have updated that as part of the changes I made to the test. Fixed.

DavidBoike (approved):

boblangley (approved):

PR #4486, opened 2024-10-01
move instance concurrency config from manifest to transport customisation

PhilBastian:


src/ServiceControl.Transports/TransportCustomization.cs
```diff
@@ -11,6 +11,13 @@ namespace ServiceControl.Transports
     using NServiceBus.Features;
     using NServiceBus.Transport;
 
+    public enum EndpointType
```
jpalac, line 14: Should we take this out into its own file?

jpalac (commented):
I guess the only thing to verify is that the `transportSettings.MaxConcurrency` setting is never null before it is used - ie those "if" statements are never true.
Is there a test for that we can somehow write?

jpalac (approved):

src/ServiceControl/Infrastructure/Settings/Settings.cs
```diff
@@ -57,7 +57,8 @@ public Settings(
             }
 
             ProcessRetryBatchesFrequency = TimeSpan.FromSeconds(30);
-            MaximumConcurrencyLevel = SettingsReader.Read(SettingsRootNamespace, "MaximumConcurrencyLevel", TransportManifestLibrary.Find(TransportType)?.DefaultPrimaryMaximumConcurrencyLevel ?? 10);
+            //TODO: is this overridden by SCMU??
```
danielmarbach, line 60: Is this TODO still needed?

PR #4490, opened 2024-10-02
RabbitMQQueueDetails parsing doesn't account for duplicated elements

danielmarbach:
Fixes part of https://github.com/Particular/PlatformBugs/issues/1371

In some cases, RabbitMQ management API might return json elements with duplicated keys. 

For example, with one of the customers it returned a duplicated `memory` element

```json
{
    "filtered_count": 44,
    "item_count": 44,
    "items": [
        {
            "arguments": {
                "x-queue-type": "quorum"
            },
            "auto_delete": false,
            "consumer_capacity": 1,
            "consumer_utilisation": 1,
            "consumers": 1,
            "durable": true,
            "effective_policy_definition": {},
            "exclusive": false,
            "leader": "rabbit@7509bf5e7b2a",
            "members": [
                "rabbit@7509bf5e7b2a"
            ],
            "memory": 112564,
            "memory": 112564,
```

By accessing the json object dictionary, the dictionary is initialized and cannot contain multiple keys. [JsonElement is more flexible in that regard](https://github.com/dotnet/runtime/issues/71784).

Both this tool and ServiceControl lack proper testing infrastructure in place to reproduce these problems so I ended up manually loading a fake Json payload from a filestream into the parser.


danielmarbach:
In https://github.com/Particular/Particular.EndpointThroughputCounter/pull/848 I went down the rabbit hole of adding more coverage for the paging etc. which didn't exist but ran out of time to catch up here with things too.

DavidBoike (approved):

jpalac (approved):

PR #4495, opened 2024-10-03
tests for instance max concurrency setting

PhilBastian:


src/ServiceControl.Transports.Tests/ServiceControlEndpointTests.cs
```diff
@@ -32,6 +32,32 @@ public async Task Should_configure_endpoint()
             Assert.That(ctx.EndpointsStarted, Is.True);
         }
 
+        [TestCase(15)]
```
danielmarbach, line 35: Could those tests all be moved into the shared folder except the postgres one?

danielmarbach:
Thanks for going the extra mile ❤️😍

danielmarbach:
A few more thoughts

We already have

ServiceControlEndpointTests with Should_set_max_concurrency_for_ProvisionQueues which seems to be off because the max concurrency setting is not relevant for queue provisioning. The name of the class also suggests those tests were intended to test setting up the primary. 
ServiceControlMonitoringEndpointTests seems to indicate these tests were meant to setup the monitoring instance
But we don't have anything for Customizing the Audit Instance (I guess because the main endpoint is send-only).

So something seems to be slightly off with our test layering

src/ServiceControl.Transports.Tests/ServiceControlEndpointTests.cs
PhilBastian: I considered it, but since the PostgreSQL and SQL Server transports have different defaults, it would require exposing the default value as a public property on the transport customisation. This doesn't seem like the right path to follow in order to unify some tests...?

danielmarbach: See #4499

SzymonPobiega (approved):

andreasohlund (approved):

PR #4530, opened 2024-10-14
Simplified trial license validation logic

tmasternak:
if trail:

- and no `endDate` in db => store value in db
- and `endDate` in db > 14 days from now => invalidate the license,
- else use the db-based `endDate`


src/ServiceControl/Licensing/ActiveLicense.cs
```diff
@@ -9,56 +9,44 @@
...
     public class ActiveLicense
     {
-        public ActiveLicense() => Refresh();
-
         public bool IsValid { get; set; }
 
         public LicenseDetails Details { get; set; }
 
-        public void Refresh()
+        public async Task Refresh(ITrialLicenseMetadataProvider trialLicenseMetadataProvider, CancellationToken cancellationToken)
```
ramonsmits, line 16: Injection `ITrailLicenseMetadataProvider` here feels off.

src/ServiceControl.Persistence/ITrialLicenseMetadataProvider.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace ServiceControl.Persistence
+{
+    using System;
+    using System.Threading;
+    using System.Threading.Tasks;
+
+    public interface ITrialLicenseMetadataProvider
```
SzymonPobiega, line 7: Nitpicking: metadata is the data about data, right? So shouldn't this be just `ITrialLicenseDataProvider`? In other words, I don't see the _meta_ aspect here.

SzymonPobiega (approved):

irinascurtu (approved):

PR #4545, opened 2024-10-21
Fix endpoint name display and grouping for Usage

jpalac:
For some brokers, the name in KnownEndpoints is different to that on the broker (ie Azure is all lower case)

In those scenarios, for [Usage](https://docs.particular.net/servicepulse/usage), we want to display the name that the user selected in their NSB setup as opposed to the broker modified name, and ensure we are grouping like endpoints (ie same endpoint but data coming from broker, Audit and/or Monitoring) correctly for max throughput calculations.

This PR fixes:
- the endpoint name matching when updating the EndpointType.
- the endpoint name matching for calculating maximum throughput for the endpoint

### DIFF
--- a/src/Particular.LicensingComponent/ThroughputCollector.cs
+++ b/src/Particular.LicensingComponent/ThroughputCollector.cs
@@ -77,7 +77,7 @@ public class ThroughputCollector(ILicensingDataStore dataStore, ThroughputSettin
             var endpointSummary = new EndpointThroughputSummary
             {
                 //want to display the endpoint name to the user if it's different to the sanitized endpoint name
-                Name = endpointGroupPerQueue.FirstOrDefault(endpoint => string.Compare(endpoint.Id.Name, endpointGroupPerQueue.Key, false) != 0)?.Id.Name ?? endpointGroupPerQueue.Key,
+                Name = endpointGroupPerQueue.FirstOrDefault(endpoint => string.Equals(endpoint.Id.Name, endpointGroupPerQueue.Key, StringComparison.OrdinalIgnoreCase))?.Id.Name ?? endpointGroupPerQueue.Key,
                 UserIndicator = UserIndicator(endpointGroupPerQueue) ?? (isKnownEndpoint ? Contracts.UserIndicator.NServiceBusEndpoint.ToString() : string.Empty),
                 IsKnownEndpoint = isKnownEndpoint,
                 MaxDailyThroughput = data.Max()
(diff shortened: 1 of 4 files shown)


src/Particular.LicensingComponent/ThroughputCollector.cs
```diff
@@ -77,7 +77,7 @@ public async Task<List<EndpointThroughputSummary>> GetThroughputSummary(Cancella
             var endpointSummary = new EndpointThroughputSummary
             {
                 //want to display the endpoint name to the user if it's different to the sanitized endpoint name
-                Name = endpointGroupPerQueue.FirstOrDefault(endpoint => endpoint.Id.Name != endpoint.SanitizedName)?.Id.Name ?? endpointGroupPerQueue.Key,
+                Name = endpointGroupPerQueue.FirstOrDefault(endpoint => string.Compare(endpoint.Id.Name, endpointGroupPerQueue.Key, false) != 0)?.Id.Name ?? endpointGroupPerQueue.Key,
```
WilliamBZA, line 80:
```suggestion
                Name = endpointGroupPerQueue.FirstOrDefault(endpoint => string.Equals(endpoint.Id.Name, endpointGroupPerQueue.Key, StringComparison.OrdinalIgnoreCase))?.Id.Name ?? endpointGroupPerQueue.Key,
```

src/Particular.LicensingComponent/ThroughputCollector.cs
```diff
@@ -77,7 +77,7 @@ public async Task<List<EndpointThroughputSummary>> GetThroughputSummary(Cancella
             var endpointSummary = new EndpointThroughputSummary
             {
                 //want to display the endpoint name to the user if it's different to the sanitized endpoint name
-                Name = endpointGroupPerQueue.FirstOrDefault(endpoint => endpoint.Id.Name != endpoint.SanitizedName)?.Id.Name ?? endpointGroupPerQueue.Key,
+                Name = endpointGroupPerQueue.FirstOrDefault(endpoint => !string.Equals(endpoint.Id.Name, endpointGroupPerQueue.Key, StringComparison.Ordinal))?.Id.Name ?? endpointGroupPerQueue.Key,
```
johnsimons, line 80: So this fixes it, but it seems `endpointGroupPerQueue.Key` could return the sanitised value?

jpalac: yes - it's meant to - that's why we are comparing the sanitized value to the endpoint name, and if the endpoint name is different, display that

johnsimons (approved):

johnsimons (approved):

WilliamBZA (approved):

WilliamBZA (approved):

PR #4557, opened 2024-10-23
SQL Server and PostgreSQL can indicate report 0 as the queue length value

### LINKED ISSUE #4556: SQL Server and PostgreSQL can indicate report `0` as the queue length value
### Symptoms

When using PostgreSQL or SQL Server transport, the 1-minute queue length diagram (shown in ServicePulse) can report `0` as the queue length value for some of the points on the graph. This happens even if the queue is non-empty and results in a characteristic saw-tooth shape of the graph. 

image

### Who's affected

All PostgreSQL and SQL Server transport users.

### Root cause

The problem is caused by the frequency (every `1` second both for PostgreSQL and SQL Server) at which the queue length measurements are made. The points on the Queue Length Diagram represent an average value for a given time interval that starts at a given time. The length of this interval depends on the visualization period chosen in the UI and by default is set to `1` minute. Every such period is divided into `60` intervals which means that for the `1` minute history period each point represents a `1` second bucket.

https://github.com/Particular/ServiceControl/blob/038bc172cf89a9fd71a7ca84d6c20c60f2d57d7e/src/ServiceControl.Monitoring/Infrastructure/HistoryPeriod.cs#L59-L66

With the query executing every `1` second, it is very likely that, due to jitter in scheduling and query execution times in some intervals, there won't be any measurements made. This in turn causes points on the diagram to indicate the queue length value of `0`.

### Patched version

- https://github.com/Particular/ServiceControl/releases/tag/6.0.1
- https://github.com/Particular/ServiceControl/releases/tag/5.11.1

tmasternak:
Fixes #4556

### DIFF
--- a/src/ServiceControl.Transports.PostgreSql/QueueLengthProvider.cs
+++ b/src/ServiceControl.Transports.PostgreSql/QueueLengthProvider.cs
@@ -132,7 +132,7 @@ class QueueLengthProvider : AbstractQueueLengthProvider
 
     static readonly ILog Logger = LogManager.GetLogger<QueueLengthProvider>();
 
-    static readonly TimeSpan QueryDelayInterval = TimeSpan.FromSeconds(1);
+    static readonly TimeSpan QueryDelayInterval = TimeSpan.FromMilliseconds(200);
 
     const int QueryChunkSize = 10;
 }
--- a/src/ServiceControl.Transports.SqlServer/QueueLengthProvider.cs
+++ b/src/ServiceControl.Transports.SqlServer/QueueLengthProvider.cs
@@ -130,7 +130,7 @@
 
         static readonly ILog Logger = LogManager.GetLogger<QueueLengthProvider>();
 
-        static readonly TimeSpan QueryDelayInterval = TimeSpan.FromSeconds(1);
+        static readonly TimeSpan QueryDelayInterval = TimeSpan.FromMilliseconds(200);
 
         const int QueryChunkSize = 10;
     }


PhilBastian:
the query interval was increased originally because of how much it is hammering the database, which was theorised to potentially increase the likelihood of connection starvation. I'm thinking this needs to be looked at with greater depth, potentially changing how the interval logic works e.g. change it so that if there is no reading in an interval it will take the previous value

@andreasohlund @SzymonPobiega

jpalac:
> the query interval was increased originally because of how much it is hammering the database, which was theorised to potentially increase the likelihood of connection starvation. I'm thinking this needs to be looked at with greater depth, potentially changing how the interval logic works e.g. change it so that if there is no reading in an interval it will take the previous value

@tmasternak do the changes made to the transport address the original reason as to why this interval was decreased?

tmasternak:
@jpalac @PhilBastian great questions. Let me give the best answers I have.

TL;DR: I propose to go with `200` ms for SQL Server and PostgreSQL. I think this should not affect connection starvation. If we get feedback from users that the load on db is too much we can provide an option to change the value (as we do for ASB). Let's capture re-working the query mechanism in a dedicated improvement issue.  

Let me start with SQL Server, which is potentially easier to answer. The value for queue length measurement interval was changed relatively recently (~2 months ago) and my understanding is that it was done to align the value with PostgreSQL. For SQL Server we never observed connection starvation and/or received feedback from users that the value is problematic. As a result, I propose to revert the change and make the interval for SQL Server `200` ms as it was previously.

When it comes to PostgreSQL I assumed that the queue length measurement mechanism would never use more than one connection at any given point. Looking at the code, it seems that we open a single connection and then update queue sizes, batching multiple select statements in a single client request.

https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl.Transports.PostgreSql/QueueLengthProvider.cs#L93-L99

As a result, I think that decreasing the query interval should not affect the connection pooling problem (to be verified for sure during pre-release testing). 

We could still be hammering the database at too much of an interval but as mentioned previously `200` ms value seems to be fine for most/all SQL Server users so I would assume the same should be the case for PostgreSQL.

Finally, when it comes to:

> I'm thinking this needs to be looked at with greater depth, potentially changing how the interval logic works e.g. change it so that if there is no reading in an interval it will take the previous value

I fully agree, I don't think this can be done as part of a bug fix though. The work needed is too big. I propose opening a dedicated improvement type of issue to capture that future work.

jpalac:
From memory the issue with Postgreql was that even though the connection was released, by the time a new one was requested (200ms), the released one was not yet available, hence a new one would be used. As a result a number (up to 8?) comnections were being used, which is significant when the limit in pgres is 100.

Again, perhaps the work you've done on the transport has reduced that impact.

Is there an issue with changing the SQLServer one back, but having the pgres one run less often?

tmasternak:
> From memory the issue with Postgreql was that even though the connection was released, by the time a new one was requested (200ms), the released one was not yet available, hence a new one would be used. As a result a number (up to 8?) comnections were being used, which is significant when the limit in pgres is 100.

We could try testing that, but I'm not sure we can reliably do that with the time we have.

> Is there an issue with changing the SQLServer one back, but having the pgres one run less often?

@jpalac yes, this might be the best approach.

jpalac:
> > Is there an issue with changing the SQLServer one back, but having the pgres one run less often?
> 
> @jpalac yes, this might be the best approach.

Ok - perhaps lets move ahead with that then?

SzymonPobiega:
I've seen the same result in the Learning transport during the NDC Porto conference while I was running the demo continuously. This makes me think that the solution might be deeper. Is far as I understand, the current assumption is that we need to query frequently enough to fill each bucket with data. 

Maybe another solution would be to distinguish buckets for which we have no data and those that have value of 0. Then we could assume that when we have no data, the value is the same as the previous value of, if we have values on both sides, set the bucket to the average of the two neighbors?

tmasternak:
> Maybe another solution would be to distinguish buckets for which we have no data and those that have value of 0. Then we could assume that when we have no data, the value is the same as the previous value of, if we have values on both sides, set the bucket to the average of the two neighbors?

@SzymonPobiega I've created dedicated issue and captured your solution idea there https://github.com/Particular/ServicePulse/issues/2156

tmasternak:
> Again, perhaps the work you've done on the transport has reduced that impact.

@jpalac I've done some more testing and I think we should be good with 200 ms for PostgreSQL. I've tweaked the connection string of the endpoints in the demo setup and all Service Control instances with the "ApplicationName" part in the connection string. This allowed me to measure the number of active connections coming from various parts of this simple system. 

The query used was:

```t-sql
SELECT application_name, count(*) 
FROM pg_stat_activity
where datname = 'nservicebus'
group by application_name;
```

Here are the results:

image

This indicates that connection string pooling for queue length monitoring works as expected i.e. there is only one active connection in the pool at any given point. Secondly, there are 53 active connections in total with endpoints taking 8 connections each, which means that we could spin up to ~5 endpoints more and not reach the default 100 active connections limit in PostgreSQL.

PR #4614, opened 2024-11-15
Adding audience to the metrics api client

johnsimons:


andreasohlund (approved):

johnsimons:
@andreasohlund This PR was failing, so I changed the code, see the latest commit.
How did you go with this fix for the customer in the other code base?

andreasohlund:
The customer for the other use case confirmed that it worked as expected

PR #4620, opened 2024-11-18
What is going on with the self-contained build?

DavidBoike:


DavidBoike:
Looks like a bug in the 9.0 SDK related to self-contained builds, perhaps surfacing only when the runtime version is different than the SDK.

PR #4657, opened 2024-12-06
Add extra logging for support case

TravisNickels:
Related to ASB package for support case logging


src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSenderDequeuer.cs
```diff
@@ -142,9 +142,25 @@ public virtual async Task Run(string forwardingBatchId, Predicate<MessageContext
...
+
+                    Log.Debug("Stopping message receiver.");
+                    await messageReceiver.StopReceive(cancellationToken);
+                    Log.Debug("Message receiver stopped.");
+                }
+                catch (Exception ex)
+                {
+                    Log.Error("An error occurred while stopping the transport receiver.", ex);
+                    throw;
+                }
```
boblangley, line 163: I would remove the try/catch.

I also added some suggestions for the debug messages.

```suggestion
               Log.Debug("Awaiting syncEvent.Task");
               _ = await syncEvent.Task;
               Log.Debug("syncEvent.Task completed");

               Log.Debug("Disposing registration.");
               registration?.Dispose();
               Log.Debug("Registration disposed");

               Log.Debug("Stopping message receiver");
               await messageReceiver.StopReceive(cancellationToken);
               Log.Debug("Message receiver stopped");     
```

boblangley (changes requested):
We can also remove the command line nupkg as it isn't needed.

PR #4755, opened 2025-01-29
Update to NServiceBus.RabbitMQ 10.0.0 and use management client

TravisNickels:
This pull request upgrades the RabbitMQ transport to version 10.

The new transport version brings a new requirement for ServiceControl to operate: The [RabbitMQ management plugin](https://www.rabbitmq.com/docs/management) must be enabled, and the plugin's [statistics and metrics collection must not be disabled](https://www.rabbitmq.com/docs/management#disable-stats).

By default, the settings to access the API will be inferred from the connection string, but the following new connection string settings have been added if custom values are needed:
- `ManagementApiUrl`
- `ManagementApiUserName`
- `ManagementApiPassword`

The transport also does [delivery limit validation](https://docs.particular.net/transports/rabbitmq/connection-settings#delivery-limit-validation) by default, but that can be controlled with the new `ValidateDeliveryLimits` connection string setting.

Other changes in this PR:
- ServiceControl.Transports.RabbitMQ now has access to the transport internals, which means all the code previously duplicated from the transport has been removed.
- The `RabbitMQQuery` and `QueueLengthProvider` classes now access the management API through the transport's management client instead of having their own custom logic.
  - The transport customization classes have been updated to allow access to the `ManagementClient` instance that the transport creates during initialization.
- The separate licensing component settings to access the management API are no longer used. They have been replaced by the above mentioned connection string settings.
  - If any existing settings are found during the upgrade process, they are migrated to the connection string instead.
 - The `DisableRemoteCertificateValidation` and `UseExternalAuthMechanism` are once again working properly instead of being ignored.


src/ServiceControl.Transports.RabbitMQ/ConnectionFactory.cs
```diff
@@ -4,13 +4,15 @@
...
+    using System.Threading;
+    using System.Threading.Tasks;
     using global::RabbitMQ.Client;
 
     class ConnectionFactory
     {
         readonly string endpointName;
         readonly global::RabbitMQ.Client.ConnectionFactory connectionFactory;
-        readonly object lockObject = new object();
+        readonly SemaphoreSlim semaphoreSlim = new(1, 1);
```
ramonsmits, line 15: Don't forget to properly dispose the semaphore

TravisNickels: Thanks for the heads up.  I'll take another look at this

ramonsmits: > I thought that the `semaphoreSlim.Release()` did the dispose,

No, it does not. The semaphore is not single use but re-usable.

bording (approved):

PR #4762, opened 2025-01-31
Emit metrics via otel instead of custom format

andreasohlund:
This replaces the homegrown metrics in the audit instance with OTEL to allow for easier consumption in visualization tools like Prometheus, Azure Monitor etc


src/ServiceControl.Audit/Auditing/AuditPersister.cs
```diff
@@ -146,10 +132,10 @@ public async Task<IReadOnlyList<MessageContext>> Persist(IReadOnlyList<MessageCo
...
                     {
                         // this can throw even though dispose is never supposed to throw
-                        using (bulkInsertCommitDurationMeter.Measure())
-                        {
-                            await unitOfWork.DisposeAsync();
-                        }
+                        var commitSw = Stopwatch.StartNew();
+                        await unitOfWork.DisposeAsync();
+                        commitSw.Stop();
+                        bulkInsertCommitDurationMeter.Record(commitSw.ElapsedMilliseconds);
```
ramonsmits, line 138: Stop is not needed

```suggestion
bulkInsertCommitDurationMeter.Record(commitSw.ElapsedMilliseconds);
```

src/ServiceControl.Audit/Auditing/AuditPersister.cs
```diff
@@ -103,12 +88,13 @@ public async Task<IReadOnlyList<MessageContext>> Persist(IReadOnlyList<MessageCo
...
                         }
 
-                        using (sagaAuditBulkInsertDurationMeter.Measure())
-                        {
-                            await unitOfWork.RecordSagaSnapshot(sagaSnapshot);
-                        }
+                        var sagaSw = Stopwatch.StartNew();
+                        await unitOfWork.RecordSagaSnapshot(sagaSnapshot);
+                        sagaSw.Stop();
+
```
ramonsmits, line 94: Stop is not needed:
```suggestion
```

src/ServiceControl.Audit/Auditing/AuditPersister.cs
```diff
@@ -89,12 +73,13 @@ public async Task<IReadOnlyList<MessageContext>> Persist(IReadOnlyList<MessageCo
...
                         }
 
-                        using (auditBulkInsertDurationMeter.Measure())
-                        {
-                            await unitOfWork.RecordProcessedMessage(processedMessage, context.Body);
-                        }
+                        var auditSw = Stopwatch.StartNew();
+                        await unitOfWork.RecordProcessedMessage(processedMessage, context.Body);
+                        auditSw.Stop();
+
```
ramonsmits, line 79: Stop is not needed:
```suggestion
```

andreasohlund:
@lailabougria can you spot anything off in terms of how we emit the telemetry? naming etc

src/ServiceControl.Audit/Auditing/AuditIngestor.cs
```diff
@@ -32,26 +29,16 @@ ITransportCustomization transportCustomization
...
-            var enrichers = new IEnrichImportedAuditMessages[]
-            {
-                new MessageTypeEnricher(),
-                new EnrichWithTrackingIds(),
-                new ProcessingStatisticsEnricher(),
-                new DetectNewEndpointsFromAuditImportsEnricher(endpointInstanceMonitoring),
-                new DetectSuccessfulRetriesEnricher(),
-                new SagaRelationshipsEnricher()
-            }.Concat(auditEnrichers).ToArray();
+            var enrichers = new IEnrichImportedAuditMessages[] { new MessageTypeEnricher(), new EnrichWithTrackingIds(), new ProcessingStatisticsEnricher(), new DetectNewEndpointsFromAuditImportsEnricher(endpointInstanceMonitoring), new DetectSuccessfulRetriesEnricher(), new SagaRelationshipsEnricher() }.Concat(auditEnrichers).ToArray();
```
ramonsmits, line 32: @andreasohlund update your Rider formatting so that this will not be put on a single line

src/ServiceControl.Audit/HostApplicationBuilderExtensions.cs
```diff
@@ -61,13 +63,35 @@ public static void AddServiceControlAudit(this IHostApplicationBuilder builder,
...
         services.AddPersistence(persistenceSettings, persistenceConfiguration);
 
         NServiceBusFactory.Configure(settings, transportCustomization, transportSettings, onCriticalError, configuration);
         builder.UseNServiceBus(configuration);
 
+        if (!string.IsNullOrEmpty(settings.OtelMetricsUrl))
+        {
+            if (!Uri.TryCreate(settings.OtelMetricsUrl, UriKind.Absolute, out var otelMetricsUri))
+            {
+                throw new UriFormatException($"Invalid OtelMetricsUrl: {settings.OtelMetricsUrl}");
```
ramonsmits, line 75: This was returned in a search:

>The environment variable OTEL_EXPORTER_OTLP_ENDPOINT exists as a standardized way to specify the OpenTelemetry OTLP endpoint. By default, it points to http://localhost:4317 for gRPC and http://localhost:4318 for HTTP/protobuf.
There are also some related environment variables:
>
>OTEL_EXPORTER_OTLP_PROTOCOL for specifying the protocol (grpc, http/protobuf)
>OTEL_EXPORTER_OTLP_HEADERS for additional headers
>OTEL_EXPORTER_OTLP_TIMEOUT for request timeouts


Maybe instead we use a more verbose name?

    OTEL_EXPORTER_OTLP_ENDPOINT

If not maybe immediately we should ensure that environment variable expansion is active so you could do:

```xml
<add key="ServiceControl.Audit/OTEL_EXPORTER_OTLP_ENDPOINT" value="%OTEL_EXPORTER_OTLP_ENDPOINT%" />
```
            
or for that matter:

```env
SERVICECONTROL_AUDIT_OTEL_EXPORTER_OTLP_ENDPOINT=%OTEL_EXPORTER_OTLP_ENDPOINT%
```

src/ServiceControl.Audit/HostApplicationBuilderExtensions.cs
```diff
@@ -61,13 +63,35 @@ public static void AddServiceControlAudit(this IHostApplicationBuilder builder,
...
+            }
+            builder.Services.AddOpenTelemetry()
+                .ConfigureResource(b => b.AddService(
+                    serviceName: "Particular.ServiceControl.Audit",
```
ramonsmits, line 79: I think this needs to be added to `AuditMetrics`

src/ServiceControl.Audit/Auditing/AuditPersister.cs
```diff
@@ -301,16 +283,12 @@ await messageDispatcher.Value.Dispatch(new TransportOperations(messagesToEmit.To
...
-        readonly Meter auditBulkInsertDurationMeter;
-        readonly Meter sagaAuditBulkInsertDurationMeter;
-        readonly Meter bulkInsertCommitDurationMeter;
-        readonly IMessageSession messageSession;
-        readonly Lazy<IMessageDispatcher> messageDispatcher;
+        readonly Counter<long> storedAudits = AuditMetrics.Meter.CreateCounter<long>($"{AuditMetrics.Prefix}.stored_audit_messages");
+        readonly Counter<long> storedSagas = AuditMetrics.Meter.CreateCounter<long>($"{AuditMetrics.Prefix}.stored_saga_audits");
+        readonly Histogram<double> auditBulkInsertDuration = AuditMetrics.Meter.CreateHistogram<double>($"{AuditMetrics.Prefix}.bulk_insert_duration_audits", unit: "ms");
+        readonly Histogram<double> sagaAuditBulkInsertDuration = AuditMetrics.Meter.CreateHistogram<double>($"{AuditMetrics.Prefix}.bulk_insert_duration_sagas", unit: "ms");
+        readonly Histogram<double> auditCommitDuration = AuditMetrics.Meter.CreateHistogram<double>($"{AuditMetrics.Prefix}.commit_duration_audits", unit: "ms");
```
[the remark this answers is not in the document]
ramonsmits: Especially because here its also not clear what the name will be. I think its better if all names are listed in a single location including there prefix/scope.

src/ServiceControl.Audit/Infrastructure/AuditMetrics.cs
```diff
@@ -0,0 +1,10 @@
+namespace ServiceControl.Audit;
+
+using System.Diagnostics.Metrics;
+
+static class AuditMetrics
+{
+    public const string MeterName = "Particular.ServiceControl";
+    public static readonly Meter Meter = new(MeterName, "0.1.0");
+    public static readonly string Prefix = "particular.servicecontrol.audit";
+}
```
ramonsmits, line 10: I think this should better define its intend:

- No need for Audit in the name as the namespace already is clear
- Its for OpenTelemetry
- These are constants

```suggestion
static class TelemetryGlobals
{
    public const string Source = "Particular.ServiceControl";
    public const string Scope = "particular.servicecontrol.audit";
    public static readonly Meter Meter = new(MeterName, "0.1.0");
}
```

src/ServiceControl.Audit/Infrastructure/Settings/Settings.cs
```diff
@@ -109,6 +109,7 @@ public string RootUrl
         public int Port { get; set; }
 
         public bool PrintMetrics => SettingsReader.Read<bool>(SettingsRootNamespace, "PrintMetrics");
+        public string OtelMetricsUrl { get; set; } = SettingsReader.Read<string>(SettingsRootNamespace, nameof(OtelMetricsUrl));
```
ramonsmits, line 112: See other comment on `OTEL_EXPORTER_OTLP_ENDPOINT`

```suggestion
        const string OTEL_EXPORTER_OTLP_ENDPOINT = nameof(OTEL_EXPORTER_OTLP_ENDPOINT);
        []
        public string OtelExporterOltpEndpointUrl { get; init; } = SettingsReader.Read<string>(SettingsRootNamespace, OTEL_EXPORTER_OTLP_ENDPOINT);
```

src/ServiceControl.Audit/Auditing/AuditPersister.cs
```diff
@@ -89,12 +73,11 @@ public async Task<IReadOnlyList<MessageContext>> Persist(IReadOnlyList<MessageCo
                             Logger.Debug("Adding audit message for bulk storage");
                         }
 
-                        using (auditBulkInsertDurationMeter.Measure())
-                        {
-                            await unitOfWork.RecordProcessedMessage(processedMessage, context.Body);
-                        }
+                        var auditSw = Stopwatch.StartNew();
+                        await unitOfWork.RecordProcessedMessage(processedMessage, context.Body);
+                        auditBulkInsertDuration.Record(auditSw.ElapsedMilliseconds);
```
ramonsmits, line 78: There are a lot of things done in this method. Even multiple measurements.
I think we should consider doing the measures in the `unitOfWork` methods itself. The measurement is then simply that whole method and isn't making this method even worse then it already is :-)

ramonsmits (dismissed):

src/ServiceControl.Audit/Auditing/AuditIngestion.cs
```diff
@@ -261,9 +257,10 @@ async Task Loop()
         readonly IAuditIngestionUnitOfWorkFactory unitOfWorkFactory;
         readonly Settings settings;
         readonly Channel<MessageContext> channel;
-        readonly Meter batchSizeMeter;
-        readonly Meter batchDurationMeter;
-        readonly Counter receivedMeter;
+        readonly Histogram<long> auditBatchSize = AuditMetrics.Meter.CreateHistogram<long>($"{AuditMetrics.Prefix}.batch_size_audits");
+        readonly Histogram<double> auditBatchDuration = AuditMetrics.Meter.CreateHistogram<double>($"{AuditMetrics.Prefix}.batch_duration_audits", unit: "ms");
+        readonly Histogram<double> auditMessageSize = AuditMetrics.Meter.CreateHistogram<double>($"{AuditMetrics.Prefix}.audit_message_size", unit: "kilobytes");
+        readonly Counter<long> receivedAudits = AuditMetrics.Meter.CreateCounter<long>($"{AuditMetrics.Prefix}.received_audits");
```
ramonsmits, line 263: There is something wrong with the naming here. THese shouldn't have a prefix like this:

```suggestion
        readonly Counter<long> receivedAudits = AuditMetrics.Meter.CreateCounter<long>(
            "received_total",
            description: "Total number of audit messages received"
            );
```

src/ServiceControl.Audit/Auditing/AuditPersister.cs
```diff
@@ -301,16 +283,12 @@ await messageDispatcher.Value.Dispatch(new TransportOperations(messagesToEmit.To
...
         }
 
-        readonly Counter ingestedAuditMeter;
-        readonly Counter ingestedSagaAuditMeter;
-        readonly Meter auditBulkInsertDurationMeter;
-        readonly Meter sagaAuditBulkInsertDurationMeter;
-        readonly Meter bulkInsertCommitDurationMeter;
-        readonly IMessageSession messageSession;
-        readonly Lazy<IMessageDispatcher> messageDispatcher;
+        readonly Counter<long> storedAudits = AuditMetrics.Meter.CreateCounter<long>($"{AuditMetrics.Prefix}.stored_audit_messages");
```
ramonsmits, line 286: Same here, these prefixes shouldn't be used and isn't according to spec. The `Meter` should define the prefix/scope AFAIK:

- https://learn.microsoft.com/en-us/dotnet/core/diagnostics/metrics-instrumentation

```c#
var meter = meterFactory.Create("Particular.ServiceControl.Audit");
_hatsSold = meter.CreateCounter<int>("stored_messages_total");
```

or

```c#
var meter = meterFactory.Create("Particular.ServiceControl");
_hatsSold = meter.CreateCounter<int>("audit.stored_messages_total");
```

but as we have seperate instance my preference would be the 1st.

src/ServiceControl.Audit/HostApplicationBuilderExtensions.cs
ramonsmits: ok, while testing envvar expansion behavior I stumbled upon the settings reader:

https://github.com/Particular/ServiceControl/blob/ea0cb3269b1d7f7aa4ed6645e309d67f1a01742b/src/ServiceControl.Configuration/EnvironmentVariableSettingsReader.cs#L37-L52

This means we can't use it with a prefix as that logic would still load the value directly from `OTEL_EXPORTER_OTLP_ENDPOINT`.

src/ServiceControl.Audit/Auditing/AuditPersister.cs
andreasohlund: Fixed

src/ServiceControl.Audit/Auditing/AuditIngestion.cs
andreasohlund: Names shortened

src/ServiceControl.Audit/Auditing/AuditIngestion.cs
```diff
@@ -261,9 +257,10 @@ async Task Loop()
         readonly IAuditIngestionUnitOfWorkFactory unitOfWorkFactory;
         readonly Settings settings;
         readonly Channel<MessageContext> channel;
-        readonly Meter batchSizeMeter;
-        readonly Meter batchDurationMeter;
-        readonly Counter receivedMeter;
+        readonly Histogram<long> auditBatchSize = AuditMetrics.Meter.CreateHistogram<long>($"{AuditMetrics.Prefix}.batch_size_audits");
```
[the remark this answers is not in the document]
andreasohlund: Fixed

src/ServiceControl.Audit/Infrastructure/DurationRecorder.cs
```diff
@@ -0,0 +1,12 @@
...
+
+using System;
+using System.Diagnostics;
+using System.Diagnostics.Metrics;
+
+record DurationRecorder(Histogram<double> Histogram) : IDisposable
+{
+    readonly Stopwatch sw = Stopwatch.StartNew();
+
+    public void Dispose() => Histogram.Record(sw.ElapsedMilliseconds);
```
ramonsmits, line 11: This is what @danielmarbach means with the ValueStopwatch, the issue isn't using Stopwatch but not use properties that result in too many unneeded conversions and operations. Stopwatch is super slim and was why I mentioned its not needed to use `.Stop`.

Calling ElapsedMilliseconds is enough:

https://github.com/dotnet/runtime/blob/v9.0.1/src/libraries/System.Private.CoreLib/src/System/Diagnostics/Stopwatch.cs#L107-L110

Furhter optimizing really isn't needed for our use case. That would only be needed when measuring microseconds type of durations.

ramonsmits (approved):

PR #4764, opened 2025-01-31
Enable message editing by default

tmasternak:
### Description

This makes the message editing (before retry) enabled on by default.

### Plan of Action

- [ ] Update documentation https://docs.particular.net/servicepulse/intro-editing-messages#enabling-the-feature


danielmarbach:
When it is enabled by default, we probably should also flip the settings from "Allow" to "Disallow" right?

udidahan:
Was there some other follow-up on the topic that could potentially be linked as the reason for why this was closed?

Issue #4759, opened 2025-01-30
When using a RavenDB PaaS it's impossible to use a RavenDB client certificates stored in Azure KeyVault


mauroservienti:
#### Is your improvement related to a problem? Please describe.

When hosting ServiceControl in Azure Container Apps, using a RavenDB PaaS offering, and storing the RavenDB client certificate (required to connect to the RavenDB hosted service) in Azure KeyVault, the certificate cannot be used securely from within the ServiceControl container instance.
     
#### Describe alternatives you've considered

One option is to create a custom ServiceControl container from the `Particular/ServiceControl` one, copying in the certificate. Although this is a possible workaround, it has a high-security implication.

Another option is to copy the certificate pfx file from the secret mount, add it to the container, and then, using the “command” property in bicep, copy the pfx to the correct folder before starting ServiceControl. This is impractical because the ServiceControl image is _distroless_ and has no shell.

Both options, for different reasons, sound impractical.

DavidBoike:
https://github.com/Particular/ServiceControl/pull/4767 will provide the ability to load a certificate and (optionally) a password either from environment variables, or by pointing to a mounted path.

DavidBoike:
This has been released in [ServiceControl 6.3.0](https://github.com/Particular/ServiceControl/releases/tag/6.3.0).

Check out the updated documentation for container deployments of the [Error instance](https://docs.particular.net/servicecontrol/servicecontrol-instances/deployment/containers#required-settings-ravendb-connection-string-ravendb-client-certificate) and [Audit instance](https://docs.particular.net/servicecontrol/audit-instances/deployment/containers#required-settings-ravendb-connection-string-ravendb-client-certificate) (both sections are the same) describing how to use the new environment variables to specify the RavenDB client certificate. Using the path option should allow you to use the certificate that is mounted into the container from Azure KeyVault.

PR #4767, opened 2025-01-31
Multiple options for configuring RavenDB client certificate

DavidBoike:
Adds the following configuration options as environment variables for connecting to external RavenDB instances, generally for use when deployed as a container:

* `RAVENDB_CLIENTCERTIFICATEPATH`: Provide a path to the client certificate, for example, to a path loaded in a mounted secrets volume.
* `RAVENDB_CLIENTCERTIFICATEBASE64`: Provide the client certificate as a Base64-encoded string.
* `RAVENDB_CLIENTCERTIFICATEPASSWORD`: If using a password-protected *.pfx certificate, supply the password to read it. If using a non-protected certificate, do not include this option.

If the certificate path or Base64 options are not used, the app will still look for a certificate in a static path in the application root. When deployed as a container, this path is `/app/raven-client-certificate.pfx`.

* Fixes https://github.com/Particular/ServiceControl/issues/4759

### DIFF
--- a/src/ServiceControl.Audit.Persistence.RavenDB/ServerConfiguration.cs
+++ b/src/ServiceControl.Audit.Persistence.RavenDB/ServerConfiguration.cs
@@ -20,9 +20,9 @@
         }
 
         public string ConnectionString { get; }
-        public string ClientCertificatePath { get; internal set; }
-        public string ClientCertificateBase64 { get; internal set; }
-        public string ClientCertificatePassword { get; internal set; }
+        public string ClientCertificatePath { get; set; }
+        public string ClientCertificateBase64 { get; set; }
+        public string ClientCertificatePassword { get; set; }
         public bool UseEmbeddedServer { get; }
         public string DbPath { get; internal set; } //Setter for ATT only
         public string ServerUrl { get; }
--- a/src/ServiceControl.Audit.Persistence.RavenDB/ServerConfiguration.cs
+++ b/src/ServiceControl.Audit.Persistence.RavenDB/ServerConfiguration.cs
@@ -18,6 +18,7 @@
         }
 
         public string ConnectionString { get; }
+        public string ClientCertificateBase64 { get; internal set; }
         public bool UseEmbeddedServer { get; }
         public string DbPath { get; internal set; } //Setter for ATT only
         public string ServerUrl { get; }
--- a/src/ServiceControl.Audit.Persistence.RavenDB/ServerConfiguration.cs
+++ b/src/ServiceControl.Audit.Persistence.RavenDB/ServerConfiguration.cs
@@ -22,6 +22,7 @@
         public string ConnectionString { get; }
         public string ClientCertificatePath { get; internal set; }
         public string ClientCertificateBase64 { get; internal set; }
+        public string ClientCertificatePassword { get; internal set; }
         public bool UseEmbeddedServer { get; }
         public string DbPath { get; internal set; } //Setter for ATT only
         public string ServerUrl { get; }
--- a/src/ServiceControl.Audit.Persistence.RavenDB/ServerConfiguration.cs
+++ b/src/ServiceControl.Audit.Persistence.RavenDB/ServerConfiguration.cs
@@ -1,6 +1,8 @@
 ﻿namespace ServiceControl.Audit.Persistence.RavenDB
 {
-    public class ServerConfiguration
+    using ServiceControl.RavenDB;
+
+    public class ServerConfiguration : IRavenClientCertificateInfo
     {
         public ServerConfiguration(string connectionString)
         {
@@ -18,6 +20,7 @@
         }
 
         public string ConnectionString { get; }
+        public string ClientCertificatePath { get; internal set; }
         public string ClientCertificateBase64 { get; internal set; }
         public bool UseEmbeddedServer { get; }
         public string DbPath { get; internal set; } //Setter for ATT only
--- a/src/ServiceControl.RavenDB/RavenClientCertificate.cs
+++ b/src/ServiceControl.RavenDB/RavenClientCertificate.cs
@@ -3,12 +3,26 @@
 namespace ServiceControl.RavenDB;
 
 using System.Reflection;
+using System.Security.Cryptography;
 using System.Security.Cryptography.X509Certificates;
 
 public static class RavenClientCertificate
 {
-    public static X509Certificate2? FindClientCertificate()
+    public static X509Certificate2? FindClientCertificate(string? base64String)
     {
+        if (base64String is not null)
+        {
+            try
+            {
+                var bytes = Convert.FromBase64String(base64String);
+                return new X509Certificate2(bytes);
+            }
+            catch (Exception x) when (x is FormatException or CryptographicException)
+            {
+                throw new Exception("Could not read the RavenDB client certificate from the configured Base64 value.", x);
+            }
+        }
+
         var applicationDirectory = Path.GetDirectoryName(Assembly.GetEntryAssembly()?.Location) ?? string.Empty;
         var certificatePath = Path.Combine(applicationDirectory, "raven-client-certificate.pfx");
 
--- a/src/ServiceControl.RavenDB/RavenClientCertificate.cs
+++ b/src/ServiceControl.RavenDB/RavenClientCertificate.cs
@@ -8,19 +8,24 @@ using System.Security.Cryptography.X509Certificates;
 
 public static class RavenClientCertificate
 {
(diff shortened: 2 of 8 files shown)


src/ServiceControl.Audit.Persistence.RavenDB/ServerConfiguration.cs
```diff
@@ -18,6 +20,9 @@ public ServerConfiguration(string dbPath, string serverUrl, string logPath, stri
         }
 
         public string ConnectionString { get; }
+        public string ClientCertificatePath { get; internal set; }
```
danielmarbach, line 23: Not that it really matters. Just pointing out. Here the setters are internal and in other files they are public.

src/ServiceControl.Audit.Persistence.RavenDB/ServerConfiguration.cs
```diff
@@ -1,6 +1,8 @@
 ﻿namespace ServiceControl.Audit.Persistence.RavenDB
 {
-    public class ServerConfiguration
+    using ServiceControl.RavenDB;
+
+    public class ServerConfiguration : IRavenClientCertificateInfo
```
danielmarbach, line 5: Please don't hesitate to ignore. I do not find the interface is particularly helpful. We could also just have a record/class encapsulating these settings that is then passed to the reader code

src/ServiceControl.RavenDB/RavenClientCertificate.cs
```diff
@@ -0,0 +1,47 @@
...
+
+public static class RavenClientCertificate
+{
+    public static X509Certificate2? FindClientCertificate(IRavenClientCertificateInfo certInfo)
+    {
+        if (certInfo.ClientCertificateBase64 is not null)
+        {
+            try
+            {
+                var bytes = Convert.FromBase64String(certInfo.ClientCertificateBase64);
```
danielmarbach, line 17: I agree, it is probably better to throw when the values are not valid instead of falling back to reading the other variables

src/ServiceControl.RavenDB/RavenClientCertificate.cs
```diff
@@ -0,0 +1,47 @@
...
+                return new X509Certificate2(bytes, certInfo.ClientCertificatePassword);
+            }
+            catch (Exception x) when (x is FormatException or CryptographicException)
+            {
+                throw new Exception("Could not read the RavenDB client certificate from the configured Base64 value.", x);
+            }
+        }
+
+        if (certInfo.ClientCertificatePath is not null)
```
danielmarbach, line 26: Would it be good to do a file check here and throw if it doesn't? Just checking when we have paths we should always assume we have at least read access because the client needs to read from this path too, right?

danielmarbach (approved):
The things mentioned are just suggestions

src/ServiceControl.RavenDB/RavenClientCertificate.cs
```diff
@@ -0,0 +1,51 @@
...
+public static class RavenClientCertificate
+{
+    public static X509Certificate2? FindClientCertificate(IRavenClientCertificateInfo certInfo)
+    {
+        if (certInfo.ClientCertificateBase64 is not null)
+        {
+            try
+            {
+                var bytes = Convert.FromBase64String(certInfo.ClientCertificateBase64);
+                return new X509Certificate2(bytes, certInfo.ClientCertificatePassword);
```
danielmarbach, line 18: I have done some more research and found it the constructor here is obsolete. We should be switching over to the certificate loader instead. Example https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509certificateloader.loadpkcs12?view=net-9.0#system-security-cryptography-x509certificates-x509certificateloader-loadpkcs12(system-byte()-system-string-system-security-cryptography-x509certificates-x509keystorageflags-system-security-cryptography-x509certificates-pkcs12loaderlimits)

src/ServiceControl.RavenDB/RavenClientCertificate.cs
```diff
@@ -0,0 +1,51 @@
...
+
+using System.Reflection;
+using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
+
+public static class RavenClientCertificate
+{
+    public static X509Certificate2? FindClientCertificate(IRavenClientCertificateInfo certInfo)
+    {
+        if (certInfo.ClientCertificateBase64 is not null)
```
danielmarbach, line 13: After thinking about this more I was wondering if we could do better. Given we have already Azure.Identity in place why not provide direct key vault integration? That seems to be the more future-proof and secure version compared to base64 encoding the cert (or we might want to think about providing both for non-key vault integration or other clouds)

All we would need is DefaultCredentials and the key vault URI

johnsimons:
Given that
>  If the certificate path or Base64 options are not used, the app will still look for a certificate in a static path in the application root. When deployed as a container, this path is /app/raven-client-certificate.pfx.

Do we actually need `RAVENDB_CLIENTCERTIFICATEPATH`?
So, a user would have to mount a path and also set that env var versus just mounting the path to the right path.
For example, with mount a path and also set that env var:
```
docker run -e RAVENDB_CLIENTCERTIFICATEPATH=/app/mycert -v /mycertpath:/app/mycert particular/servicecontrol:latest
```

Example with mount a path only:
```
docker run -v /mycertpath/raven.pfx:/app/raven-client-certificate.pfx:ro particular/servicecontrol:latest
```


Regarding `RAVENDB_CLIENTCERTIFICATEBASE64`.
I find mapping paths much easier, and I wonder if this is needed at all. I am raising it because it is actually quite challenging to set a cert using this method, e.g., doing it in base64 and setting the env var. This is just my experience when using this method to [set the license](https://docs.particular.net/nservicebus/licensing/#license-management-environment-variable).

src/ServiceControl.RavenDB/RavenClientCertificate.cs
DavidBoike: The replacement APIs aren't available until .NET 9 and we're currently on .NET 8, so we'll have to do that later.

src/ServiceControl.RavenDB/RavenClientCertificate.cs
DavidBoike: I'm not sure that is appropriate to do outside of a more general "provide tighter integration with cloud native things" and understanding what the corresponding AWS integrations would be. Honestly feels a bit off to talk about a cloud-native way of using a not-cloud-native storage mechanism.

DavidBoike:
@johnsimons If we were only talking about Docker Desktop, you'd be right. That's why the previous ServiceControl on containers task force did not anything like what is in this PR…we assumed that being able to map the file directly to the needed path would be enough.

But it wasn't. In the Azure Container Apps environment (and maybe others) you can point a container at a KeyVault URI and it will mount as a volume with one or more secrets exposed as files within that directory. But you can't mount a single file to a single path like you can with a Docker Desktop volume mounting.

As for the Base64 option, mapping paths is indeed much easier in Docker Desktop. But to map paths in Container Apps, for example, you first need files somewhere in a way that can even be mounted to the container in the first place. This is what our internal installation is currently doing, and it's a big pain requiring a lot of bicep code to have the storage account, and then create the file share, and then map the file share to the container environment, and then make the mapped share available to the container, and then map the container file share to a container volume. Then we currently "get to" pay for a minimum 100GB file share in order to deliver one 4Kb file.

Compared to that, throwing `Convert.ToBase64String(File.ReadAllBytes(path)).Dump();` into LINQPad once and then dumping the result into a secret which can be mapped in one line is…nice.

src/ServiceControl.RavenDB/RavenClientCertificate.cs
danielmarbach: I think it is a fair assessment. I was more thinking along the lines of given we are doing extensions to the configuration API based on specific assumptions of an environment, it would be OK to look at this specific environment to figure out what a more native way of doing it could be. With `CertificateClient` and `DefaultCredentials` and an URI to the key vault you could download the certificate and store it "in memory". I was more worried about doing something that could be seen as glaringly "oh please don't do that" and then having to support it over time seems at least worth a brief stop to evaluate our options.

But I guess the argument here is the ability to provide the certificate "binary" via an environment variable is by definition "cross cloud" even when we haven't investigated the other use cases.

danielmarbach:
> Compared to that, throwing `Convert.ToBase64String(File.ReadAllBytes(path)).Dump();` into LINQPad once and then dumping the result into a secret which can be mapped in one line is…nice.

or with any decent shell ;)

base64 -i path

PR #4758, opened 2025-01-30
Investigation hanging audit instance

ramonsmits:

PR #4782, opened 2025-02-06
Support cancellation of web api and audit storage operations

ramonsmits:
When ingestion hangs and the service looks unresponse a user might want to stop the instance. Especially for the batch ingestion is makes sense to support cancellation as this is a long running operations.

This is a port and continuation of my investigation PR:

- https://github.com/Particular/ServiceControl/pull/4758


ramonsmits:
@danielmarbach @andreasohlund There are multiple commits, one for audit, one for error, and one for domain events. Issue it that seems like a very deep rabbit hole as many handlers invoke storage and all of these do not support cancellation.

As audit is the most affected I think it makes sense to support cancellation there and for the remaning parts to create cruft issues as it is a lot of effort to add cancellation to all those operations.

What is your opinion?

src/ServiceControl/Operations/ErrorIngestion.cs
```diff
@@ -92,7 +92,7 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
                         batchSizeMeter.Mark(contexts.Count);
                         using (batchDurationMeter.Measure())
                         {
-                            await ingestor.Ingest(contexts);
+                            await ingestor.Ingest(contexts, stoppingToken);
                         }
                     }
                     catch (OperationCanceledException) when (stoppingToken.IsCancellationRequested)
```
andreasohlund, line 98: Throwing here will clear the messsage contexts without setting them to completed or failed and leave the OnMessage calls stuck. Is that a problem?

src/ServiceControl.Audit/Auditing/AuditPersister.cs
```diff
@@ -53,7 +54,7 @@ public async Task<IReadOnlyList<MessageContext>> Persist(IReadOnlyList<MessageCo
             {
 
                 // deliberately not using the using statement because we dispose async explicitly
-                unitOfWork = await unitOfWorkFactory.StartNew(contexts.Count);
+                unitOfWork = await unitOfWorkFactory.StartNew(contexts.Count, cancellationToken);
```
andreasohlund, line 57: Should we pass the token to Task.When all as well?

andreasohlund: Should we also pass the token to RecordProcessedMessage, RecordSagaSnapshot and RecordKnownEndpoint?

danielmarbach:
> for the remaning parts to create cruft issues as it is a lot of effort to add cancellation to all those operations.

FYI During the modernization efforts we deliberately had to stop at some point. It was just too much effort and also not in all cases glaringly obvious what the effects of cancellation might have on the stability of the code paths calling it so we did cut off somewhere.

src/ServiceControl.Audit/Auditing/AuditPersister.cs
ramonsmits: > Task.When

Task.WhenAll does not have a CT arg.

ramonsmits: > Should we also pass the token to RecordProcessedMessage, RecordSagaSnapshot and RecordKnownEndpoint?

Added in latest commit

src/ServiceControl.Audit/Auditing/AuditIngestion.cs
```diff
@@ -180,7 +180,7 @@ async Task EnsureStopped(CancellationToken cancellationToken)
             try
             {
                 await startStopSemaphore.WaitAsync(cancellationToken);
-                await StopAndTeardownInfrastructure(cancellationToken);
+                await StopAndTeardownInfrastructure(new CancellationToken(canceled: true));
```
ramonsmits, line 183: @danielmarbach what do you think about this? 

This should result in terminating ASAP but still release any resources.

I'm not sure if transports would properly teardown on stop but if not that should be considered a but right?

danielmarbach: Is the idea here to make sure no transports waits for pending receive operations?

mauroservienti: Yes

ramonsmits: > Is the idea here to make sure no transports waits for pending receive operations?

@danielmarbach yes, but ideally still correctly teardown any resources (because we mix both in Stop..)

mauroservienti (approved):

src/ServiceControl.Audit.Persistence.RavenDB/RavenAttachmentsBodyStorage.cs
```diff
@@ -11,15 +12,15 @@ class RavenAttachmentsBodyStorage(
...
             {
                 return Task.CompletedTask;
             }
 
             return bulkInsert.AttachmentsFor(bodyId)
-                .StoreAsync("body", bodyStream, contentType);
+                .StoreAsync("body", bodyStream, contentType, cancellationToken);
         }
 
         public async Task<StreamResult> TryFetch(string bodyId)
```
andreasohlund, line 26: Should this have a token as well? (I know this is on a different code path)

ramonsmits: @andreasohlund there are many more places that can support cancellation but this PR tries to focus on the majority of the audit/error ingestion IO.

andreasohlund (approved):

src/ServiceControl/Operations/ErrorIngestion.cs
ramonsmits: Changes have been heavily refactored since comments and code has been modified to ensure task are completed as cancelled

src/ServiceControl.Audit/Auditing/AuditIngestion.cs
ramonsmits: Added code comments to clarify why the cancellationtoken is passed in the cancelled state

src/ServiceControl.Audit.Persistence.RavenDB/RavenAttachmentsBodyStorage.cs
ramonsmits: Added additional comments to ensure all audit storage operations support cancellation

PR #4789, opened 2025-02-10
fix batch message deduplication logic to not update batch on already batched message

PhilBastian:
reverts logic changed in https://github.com/Particular/ServiceControl/pull/3709

### DIFF
--- a/src/ServiceControl.Persistence.RavenDB/RetryDocumentDataStore.cs
+++ b/src/ServiceControl.Persistence.RavenDB/RetryDocumentDataStore.cs
@@ -113,7 +113,7 @@
                 }
             };
 
-            return new PatchCommandData(FailedMessageRetry.MakeDocumentId(messageId), null, patch: patchRequest, patchIfMissing: patchRequest);
+            return new PatchCommandData(FailedMessageRetry.MakeDocumentId(messageId), null, patch: new PatchRequest { Script = "" }, patchIfMissing: patchRequest);
         }
 
         public async Task GetBatchesForAll(DateTime cutoff, Func<string, DateTime, Task> callback)


src/ServiceControl.Persistence.RavenDB/RetryDocumentDataStore.cs
```diff
@@ -113,7 +113,7 @@ static ICommandData CreateFailedMessageRetryDocument(string batchDocumentId, str
                 }
             };
 
-            return new PatchCommandData(FailedMessageRetry.MakeDocumentId(messageId), null, patch: patchRequest, patchIfMissing: patchRequest);
+            return new PatchCommandData(FailedMessageRetry.MakeDocumentId(messageId), null, patch: new PatchRequest { Script = "" }, patchIfMissing: patchRequest);
```
jpalac, line 116: I'm guessing there was no option to change this to a create command instead of a patch?

PhilBastian: not that I can find. Hoping @danielmarbach or @DavidBoike can advise if there's a more efficient method to only create a record if it doesn't already exist

danielmarbach: Would something like

```csharp
static ICommandData CreateFailedMessageRetryDocument(string batchDocumentId, string messageId)
{
    var patchRequest = new PatchRequest
    {
        Script = @"
            if (this.RetryBatchId === null) {
                this.FailedMessageId = args.MessageId;
                this.RetryBatchId = args.BatchDocumentId;
            }",
        Values =
        {
            { "MessageId", FailedMessageIdGenerator.MakeDocumentId(messageId) },
            { "BatchDocumentId", batchDocumentId }
        }
    };
    return new PatchCommandData(FailedMessageRetry.MakeDocumentId(messageId), null, patch: patchRequest, patchIfMissing: patchRequest);
}
```

work? It has been I while since I used the patching API

PhilBastian: I don't think there's a case where this document will have a null `RetryBatchId`, since it's `FailedMessageRetry` not `FailedMessage`. The goal here is to only create the document if it doesn't exist, never update an existing document, which the code does with the blank `PatchRequest`, but if there's an alternative, more idiomatic, Raven way of "create only if not exists" then I'd prefer to use that

jpalac (approved):

src/ServiceControl.Persistence.RavenDB/RetryDocumentDataStore.cs
danielmarbach: I'm not sure I understand the comment. The idea was to have a consistent and idempotent patch request for both cases. This would achieve that. I don't think you can use a create command here because create when the document exists would fail and that's not what you want here

PhilBastian: I believe having a single patch request for both cases obfuscates the intended logic that we don't want an update on anything that already exists. Previous releases have had `patch` set to `null`, but that's no longer allowed

DavidBoike: > but that's no longer allowed

Aha, that probably explains why I made the change in the first place, likely not understanding the implications at the time. Makes this PR make a lot more sense to me now, and sounds like you've probably got the right approach, as far as I'm concerned.

PR #4803, opened 2025-02-12
Better names for exposed audit metrics

andreasohlund:


src/ServiceControl.Audit/Auditing/AuditIngestion.cs
```diff
@@ -187,7 +187,7 @@ async Task OnMessage(MessageContext messageContext, CancellationToken cancellati
                 await channel.Writer.WriteAsync(messageContext, cancellationToken);
                 await taskCompletionSource.Task;
 
-                ingestedMessagesCounter.Add(1);
+                successfulMessagesCounter.Add(1, Telemetry.GetIngestedMessageTags(messageContext.Headers));
                 messageSize.Record(messageContext.Body.Length / 1024.0);
```
ramonsmits, line 191: Shouldn't this become a `messaging.message.body.size` tag?

https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#messaging-attributes

> [messaging.message.body.size](https://opentelemetry.io/docs/specs/semconv/attributes-registry/messaging/)	int	The size of the message body in bytes. [17]	1439	Opt-In

andreasohlund: I think that makes sense, will do

src/ServiceControl.Audit/Auditing/AuditIngestion.cs
```diff
@@ -187,8 +188,7 @@ async Task OnMessage(MessageContext messageContext, CancellationToken cancellati
                 await channel.Writer.WriteAsync(messageContext, cancellationToken);
                 await taskCompletionSource.Task;
 
-                ingestedMessagesCounter.Add(1);
-                messageSize.Record(messageContext.Body.Length / 1024.0);
+                successfulMessagesCounter.Add(1, tags);
             }
```
ramonsmits, line 192: I noticed in core that it also sets a tag for when the returned task is cancelled. Only issue here is that I think core does so incorrectly.

It should only add that tag when the cancellationtoken was set AND the task result is cancelled.

Not sure about how usable it but we could do:

```
finally
{
    if(cancellationToken.IsCancellationRequestion)
    tags.Add("cancelled", true);
}
```

Just before we do the `.Add`

ramonsmits (approved):

PR #4818, opened 2025-02-14
Fix watchdog re-creation of TransportInfrastructure and cancellation improvements

ramonsmits:
- Watchdog didn't invoke TransportInfrastructure.Shutdown
- Added maximum start timeout
- Watchdog stop now supports cancellation
- Watchdog stop no longer hiding failing stop and throws

### DIFF
--- a/src/ServiceControl.Audit/Auditing/AuditIngestion.cs
+++ b/src/ServiceControl.Audit/Auditing/AuditIngestion.cs
@@ -76,13 +76,18 @@
                 await startStopSemaphore.WaitAsync(cancellationToken);
                 logger.Debug("Ensure started. Start/stop semaphore acquired");
 
-                if (!unitOfWorkFactory.CanIngestMore())
+                var canIngest = unitOfWorkFactory.CanIngestMore();
+
+                logger.DebugFormat("Ensure started {0}", canIngest);
+
+                if (canIngest)
+                {
+                    await SetUpAndStartInfrastructure(cancellationToken);
+                }
+                else
                 {
                     await StopAndTeardownInfrastructure(cancellationToken);
-                    return;
                 }
-
-                await SetUpAndStartInfrastructure(cancellationToken);
             }
             catch (Exception e)
             {
--- a/src/ServiceControl.Audit/Auditing/AuditIngestion.cs
+++ b/src/ServiceControl.Audit/Auditing/AuditIngestion.cs
@@ -130,7 +130,7 @@
                 await auditIngestor.VerifyCanReachForwardingAddress();
                 await queueIngestor.StartReceive(cancellationToken);
 
-                logger.Info("Started infrastructure");
+                logger.Info(LogMessages.StartedInfrastructure);
             }
             catch (Exception e)
             {
@@ -163,7 +163,7 @@
                 }
 
                 queueIngestor = null;
-                logger.Info("Stopped infrastructure");
+                logger.Info(LogMessages.StoppedInfrastructure);
             }
             catch (Exception e)
             {
@@ -335,5 +335,11 @@
         readonly IHostApplicationLifetime applicationLifetime;
 
         static readonly ILog logger = LogManager.GetLogger<AuditIngestion>();
+
+        internal static class LogMessages
+        {
+            internal const string StartedInfrastructure = "Started infrastructure";
+            internal const string StoppedInfrastructure = "Stopped infrastructure";
+        }
     }
 }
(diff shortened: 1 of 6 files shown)


src/ServiceControl.Infrastructure/Watchdog.cs
```diff
@@ -36,47 +41,43 @@ public Task OnFailure(string failure)
...
 
-                bool? failedOnStartup = null;
+                bool startup = true;
 
                 while (!shutdownTokenSource.IsCancellationRequested)
                 {
                     try
                     {
+                        using var cancellationTokenSource = CancellationTokenSource.CreateLinkedTokenSource(shutdownTokenSource.Token);
+                        cancellationTokenSource.CancelAfter(15000);
```
andreasohlund, line 57: Lets add a comment as to why we choose 15s?

src/ServiceControl.Infrastructure/Watchdog.cs
```diff
@@ -87,25 +88,28 @@ public Task Start(Action onFailedOnStartup)
...
 
-        public Task Stop()
+        public async Task Stop(CancellationToken cancellationToken)
         {
-            shutdownTokenSource.Cancel();
-            return watchdog;
+            try
+            {
+                log.Debug($"Stopping watching process {taskName}");
+                await shutdownTokenSource.CancelAsync().ConfigureAwait(false);
```
andreasohlund, line 101: Nitpick: We should need configureawait in this codebase

src/ServiceControl.Audit/Auditing/AuditIngestion.cs
```diff
@@ -76,18 +76,18 @@ async Task EnsureStarted(CancellationToken cancellationToken)
...
-                logger.DebugFormat("Ensure started {0}", canIngest);
-
-                if (canIngest)
-                {
-                    await SetUpAndStartInfrastructure(cancellationToken);
-                }
-                else
+                if (!unitOfWorkFactory.CanIngestMore())
                 {
+                    logger.Warn("Ensure started. Storage unable to ingest more. Ingestion will be paused");
```
ramonsmits, line 81: @andreasohlund already is done in these methods, now we're double logging

andreasohlund (approved):

src/ServiceControl.Audit/Auditing/AuditIngestion.cs
andreasohlund: we are only logging INFO in there, I think we should log INFO if we stop due to a requested shutdown and WARN when we stop due to the persister being unhealthy?

ramonsmits: This code doesn't know if the persister is unhealthy. You mean when `unitOfWorkFactory.CanIngestMore()` returns `false`?

I'm ok with such a warning but as a state change. This method is invoked I think every minute so those logs would be written every cycle.

andreasohlund: Yea we can tailor the message to say something like "Persister unable to ingest more data, pausing" or something.

Thinking about this more, why are we not using the circuit breaker that we use elsewhere here?

src/ServiceControl.Infrastructure/Watchdog.cs
mauroservienti: Should or should not?

src/ServiceControl.Infrastructure.Tests/WatchdogTests.cs
```diff
@@ -27,16 +28,16 @@ public async Task It_shuts_down_gracefully()
...
             await started.Task;
 
-            await dog.Stop();
+            await dog.Stop(TestContext.CurrentContext.CancellationToken);
 
             await stopped.Task;
         }
 
-        [Test]
+        [Test, Ignore("When ensurestopped throws an exception, stop should also throw an exception as that is an ungraceful stop")]
```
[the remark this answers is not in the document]
andreasohlund: I agree that we can remove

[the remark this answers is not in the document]
ramonsmits: @mauroservienti @andreasohlund I've taken another look at the test and refactored it.

src/ServiceControl.Infrastructure/Watchdog.cs
ramonsmits: @andreasohlund I've searched for `.ConfigureAwait(` and there are 82 matches

andreasohlund (approved):

mauroservienti (approved):

PR #4821, opened 2025-02-18
Update to Azure Service Bus 5

SzymonPobiega:
Azure Service Bus transport 5 introduces new event delivery topology using topic-per-type approach. This change affects how ServiceControl publishes [integration events](https://docs.particular.net/servicecontrol/contracts). 

Behavior: 
- By default ServiceControl uses the new [topic-per-event topology](https://docs.particular.net/transports/azure-service-bus/topology)
- If a topic to publish is provided via connection string `TopicName` property, ServiceControl uses the migration topology that allows publishing events in the way they can be processed by older versions of the transport 
- If custom settings are needed they can be provided via `ServiceControl.Transport.ASBS/Topology` setting. This option allows mixing topic-per-event and single-topic approaches if there are subscribers that use both versions


SzymonPobiega:
@danielmarbach I pushed the latest approach that takes advantage of `AddTransportForPrimaryCore` to only add mappings when needed.

src/ServiceControl.Transports.ASBS/ASBSTransportCustomization.cs
```diff
@@ -40,6 +50,44 @@ protected override void AddTransportForPrimaryCore(IServiceCollection services,
...
+                    EventsToMigrateMap = [
+                        "ServiceControl.Contracts.CustomCheckFailed",
+                        "ServiceControl.Contracts.CustomCheckSucceeded",
+                        "ServiceControl.Contracts.HeartbeatRestored",
+                        "ServiceControl.Contracts.HeartbeatStopped",
+                        "ServiceControl.Contracts.FailedMessagesArchived",
+                        "ServiceControl.Contracts.FailedMessagesUnArchived",
+                        "ServiceControl.Contracts.MessageFailed",
+                        "ServiceControl.Contracts.MessageFailureResolvedByRetry",
+                        "ServiceControl.Contracts.MessageFailureResolvedManually"
```
danielmarbach, line 75: It seems those two are above the character counts of 50 chars and would have been shortened to the MD5 hash. The rest is fine.

Furthermore, there might be cases when the ServiceControl endpoint name is longer than 50 chars we would have shortened it automatically into an MD5 hash. We might need to keep this consistent at least for the migration topology and probably discuss what we do for the new topology

src/ServiceControl.Transports.ASBS/ServiceControl.Transports.ASBS.csproj
```diff
@@ -15,7 +16,7 @@
     <PackageReference Include="Azure.Monitor.Query" />
     <PackageReference Include="Azure.ResourceManager.ServiceBus" />
     <PackageReference Include="NServiceBus.CustomChecks" />
-    <PackageReference Include="NServiceBus.Transport.AzureServiceBus" />
+    <PackageReference Include="NServiceBus.Transport.AzureServiceBus"/>
```
danielmarbach, line 19: I think the default is with space

src/ServiceControl.Transports.ASBS/ServiceControl.Transports.ASBS.csproj
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <TargetFramework>net8.0</TargetFramework>
     <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
+    <NoWarn>CS0618</NoWarn>
```
danielmarbach, line 6: Could we remove the global suppression and have the in code suppression?

src/ServiceControl.Transports.ASBS/ASBSTransportCustomization.cs
SzymonPobiega: > It seems those two are above the character counts of 50 chars and would have been shortened to the MD5 hash

Why would that matter on the publisher side? We are not creating rules here.

SzymonPobiega: > Furthermore, there might be cases when the ServiceControl endpoint name is longer than 50 chars we would have shortened it automatically into an MD5 hash

Again, doesn't it matter only for the subscribers? The publisher just ensures the _bundle_ topic is present.

danielmarbach: It is a good point. As of now ServiceControl did never subscribe and therefore it doesn't matter as long as we don't subscribe to any of our integration events

danielmarbach:
Approving but we should probably wait with merging until we have the final version right?

danielmarbach (approved):

ramonsmits:
> * uses topic-per-event topology

@SzymonPobiega @danielmarbach 

That is a HUGE breaking change. Please don't make that the default! It would break any subscribers to SC events if these would not be using the new topology. As SC needs to be wirecompatible with prior Core versions you can't make this the default without explicit consent.

Another option is explicit opt-in during upgrade OR make hybrid the default.

Please don't do a hidden breaking transport change.

PR #4822, opened 2025-02-18
Allow setting shutdown timeout for all instances

mauroservienti:
Tis PR exposes a new setting (`ServiceControl*/ShutdownTimeout`) to define the process shutdown timeout for all instances. The default value of 5 seconds has been selected based on the most restrictive hosting platform, containers on Linux. Linux onctainers allow for a maximum of ten seconds, and five were chosen to be as conservative as possible. That is less of an issue because when running ServiceControl using containers, we always run with an external RavenDB server, and the ServiceControl shutdown process is lightweight.

On the other hand, when installing Windows as a service, either via SCMU or PowerShell, the installation process sets the shutdown timeout to 2 minutes. By default, Windows allows services to shut down for a maximum of 125 seconds. However, every service is given only 30 seconds unless it requests additional time. For that reason, this PR also includes a custom `WindowsServiceWithRequestTimeout` component that requests additional time as defined by the new `ShutdownTimeout` setting when the service control manager invokes the stop. Unfortunately, there is no way to request extra time when the operating system is shutting down; in that case, there is a high chance that the ServiceControl embedded database doesn't shut down cleanly.

The new `WindowsServiceWithRequestTimeout` is registered when installing ServiceControl on Windows as a service and either run regularly or in maintenance mode.

- [ ] Document the new setting


src/ServiceControl/Infrastructure/Settings/Settings.cs
```diff
@@ -180,6 +181,11 @@ public TimeSpan HeartbeatGracePeriod
 
         public bool DisableHealthChecks { get; set; }
 
+        // Windows services allow a maximum of 125 seconds when stopping a service.
+        // When shutting down or restarting the OS we have no control over the
+        // shutdown timeout
+        public TimeSpan ShutdownTimeout { get; set; } = TimeSpan.FromMinutes(2);
```
andreasohlund, line 187: When running in docker the timeout is 10s on linux and 30s on windows

https://docs.docker.com/reference/cli/docker/container/stop/#:~:text=The%20default%20timeout%20can%20be,30%20seconds%20for%20Windows%20containers.

We should make sure that the SCMU sets this value since it knows the host will be a windows service and similar for our docker compose examples we should set the value as well.

That said I wonder if selecting the "maximum allowed timeout for any of the known hosting environments" is the best thing to do? 

This means that where we miss to set this the instance will always be killed by the host

mauroservienti: You say we should not have a default value but always set it according to the running platform. Is this correct?

Another option (which I'm not sure I like) would be to detect somehow in code the hosting platform and set a default value accordingly.

andreasohlund: >You say we should not have a default value but always set it according to the running platform. Is this correct?

Not sure, I think a default is good but not sure what it should be set to.

Regarding detecting inside: I have done some research about figuring out a good "instance id" and concluded that it's easier to set it from the outside since there you know what it would be.

In this case: SCMU => knows that it will be a Windows service we can set it to the maximum value that can be requested

https://stackoverflow.com/questions/13454054/what-is-the-maximum-time-windows-service-wait-to-process-stop-request-and-how-to

Are we requesting additional time when running as a win service?


Side note: I just realized that in docker we run with a separate database so there we can be more aggressive when shutting down since we don't have to wait for the database to stop3

mauroservienti: > Are we requesting additional time when running as a win service?

Our understanding is the host does that automatically by setting https://github.com/Particular/ServiceControl/blob/599b508e14145013374f6792563626ee553df188/src/ServiceControl.Audit/HostApplicationBuilderExtensions.cs#L48

src/ServiceControl.Audit/Infrastructure/Settings/Settings.cs
```diff
@@ -152,6 +153,11 @@ public int MaxBodySizeToStore
 
         public bool EnableFullTextSearchOnBodies { get; set; }
 
+        // Windows services allow a maximum of 125 seconds when stopping a service.
+        // When shutting down or restarting the OS we have no control over the
+        // shutdown timeout
+        public TimeSpan ShutdownTimeout { get; set; } = TimeSpan.FromMinutes(2);
```
ramonsmits, line 159: This only gives 5 seconds of "cancellation" which I think is sufficient because it should be a sort of let everything go and log about it but do not delay termination

src/ServiceControl/Infrastructure/Settings/Settings.cs
ramonsmits: The default should be 2 minutes, here is why:

- With external mode, 10 seconds should be plenty and the container is resilient to termination. If we would be running for longer then 10 seconds, we likely were not able to adhere to any cancellation anyway.

- With internal mode, the platform is Windows and we can keep running with 2 minutes. We could even run longer but likely the Windows host isn't configured to allow for it. We WANT to be able to run "cancellation" because we want to ensure the child process gets killed so not cause any issues when the service instance gets restarted.

- With internal mode but a reboot, in that case we will be killed anyway. We extend stop to allow for the database to gracefully exit. In case of a reboot we'll be killed when taking longer than 30 seconds.


i.e. Setting ShutdownTimeout to 2 minutes is good. In external we would be gracefully existing in a few seconds. If not, we got other problems.

ramonsmits (approved):

src/ServiceControl/Infrastructure/Settings/Settings.cs
mauroservienti: Take a look at the last two commits:

- https://github.com/Particular/ServiceControl/pull/4822/commits/9c0ab387be8ab3e998d07afe96b29b7dc5bfacf4
- https://github.com/Particular/ServiceControl/pull/4822/commits/2ce99db721763ce06e8ef4c25fe10d64343f4b3f

Aren't those more explicit?

src/ServiceControlInstaller.Engine/Configuration/ServiceControl/ServiceControlAppConfig.cs
```diff
@@ -32,6 +32,12 @@ protected override void UpdateSettings()
             settings.Set(ServiceControlSettings.EnableFullTextSearchOnBodies, details.EnableFullTextSearchOnBodies.ToString(), version);
             settings.Set(ServiceControlSettings.RemoteInstances, RemoteInstanceConverter.ToJson(details.RemoteInstances), version);
 
+            // Windows services allow a maximum of 125 seconds when stopping a service.
+            // When shutting down or restarting the OS we have no control over the
+            // shutdown timeout. This is by the installer engine that is run _only_ on
+            // Windows via SCMU or PowerShell
+            settings.Set(ServiceControlSettings.ShutdownTimeout, "00:02:00");
```
ramonsmits, line 39: Not that familiar with this. This is added during an upgrade to the config?

src/ServiceControl/Infrastructure/Settings/Settings.cs
```diff
@@ -180,6 +181,10 @@ public TimeSpan HeartbeatGracePeriod
 
         public bool DisableHealthChecks { get; set; }
 
+        // The default value is set to the maximum allowed time by the most restrictive
+        // hosting platform, which Docker Linux containers. 
+        public TimeSpan ShutdownTimeout { get; set; } = TimeSpan.FromSeconds(10);
```
ramonsmits, line 186: 10 seconds is when we get killed, you set the timeout to allow for cancellation.

As in:

- if we're interested in cancellation, set it to less than the host its kill duration
- if we're not interested in cancellation, set this to a high value that gives us time to run a graceful termination

With external, I think 5 seconds is already plenty
With internal, I want to wait as long as possible

So I like your last commits, but if we want to see cancellation to take place this number should be even lower:

```suggestion
        // The default value is set to the maximum allowed time by the most restrictive
        // hosting platform, which Docker Linux containers minus a few seconds
        // to actually allow for cancellation and logging to take place
        public TimeSpan ShutdownTimeout { get; set; } = TimeSpan.FromSeconds(5);
```

andreasohlund:
Should this PR also update SCMU to set it to a higher value on upgrade? (since there we know its running as a win service)

mauroservienti:
Sure thing, @andreasohlund. I'll continue working on it.

src/ServiceControlInstaller.Engine/Configuration/ServiceControl/ServiceControlAppConfig.cs
mauroservienti: Either way, upgrade or new instances

andreasohlund (approved):

src/ServiceControl.Hosting/WindowsServiceWithRequestTimeout.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        RequestAdditionalTime(hostOptions.ShutdownTimeout + CancellationDuration);
+        base.OnStop();
+    }
+
+    protected override void OnShutdown()
+    {
+        RequestAdditionalTime(hostOptions.ShutdownTimeout + CancellationDuration);
+        base.OnShutdown();
+    }
```
mauroservienti, line 38: @ramonsmits, according to the [documentation](https://learn.microsoft.com/en-us/dotnet/api/system.serviceprocess.servicebase.requestadditionaltime?view=net-9.0-pp#system-serviceprocess-servicebase-requestadditionaltime(system-timespan)), OnShutdown is ignored.

```suggestion
```

mauroservienti:
Rebased from `master` because the solution file changed

PR #4823, opened 2025-02-18
Logging events to diagnose stop sequence

ramonsmits:
Log when application is done to determine if process was killed and disposing of RavenDB embedded when used.

### DIFF
--- a/src/ServiceControl.Audit/Program.cs
+++ b/src/ServiceControl.Audit/Program.cs
@@ -7,31 +7,40 @@ using ServiceControl.Audit.Infrastructure.Settings;
 using ServiceControl.Configuration;
 using ServiceControl.Infrastructure;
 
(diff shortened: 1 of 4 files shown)


src/ServiceControl.Audit/Program.cs
```diff
@@ -7,31 +7,40 @@
...
 
-await new CommandRunner(arguments.Command).Execute(arguments, settings);
+    await new CommandRunner(arguments.Command).Execute(arguments, settings);
 
-return 0;
+    return 0;
+}
+finally
+{
+    // Leave a trail in the logs to determine if the process was killed
```
mauroservienti, line 43: Should this be a TODO?

```suggestion
    // TODO Leave a trail in the logs to determine if the process was killed
```

PR #4824, opened 2025-02-18
Allow JIT killing of Raven.Embedded.EmbeddedServer by tearing it down as part of stop sequence

ramonsmits:
Teardown Raven.Embedded.EmbeddedServer as part of stop sequence so that we kill the instance if needed as adhering to cancellation is not possible in host builder root container dispose.

When the Embedded RavenDB client supports a Stop with cancellation this code can be simplified.

### DIFF
--- a/src/ServiceControl.Audit.Persistence.RavenDB/IRavenPersistenceLifecycle.cs
+++ b/src/ServiceControl.Audit.Persistence.RavenDB/IRavenPersistenceLifecycle.cs
@@ -6,6 +6,6 @@
     interface IRavenPersistenceLifecycle
     {
         Task Initialize(CancellationToken cancellationToken = default);
-        Task Stop(CancellationToken cancellationToken);
+        Task Stop(CancellationToken cancellationToken = default);
     }
 }
--- a/src/ServiceControl.Audit.Persistence.RavenDB/IRavenPersistenceLifecycle.cs
+++ b/src/ServiceControl.Audit.Persistence.RavenDB/IRavenPersistenceLifecycle.cs
@@ -6,5 +6,6 @@
     interface IRavenPersistenceLifecycle
     {
         Task Initialize(CancellationToken cancellationToken = default);
+        Task Stop(CancellationToken cancellationToken);
     }
 }
--- a/src/ServiceControl.Audit.Persistence.RavenDB/RavenEmbeddedPersistenceLifecycle.cs
+++ b/src/ServiceControl.Audit.Persistence.RavenDB/RavenEmbeddedPersistenceLifecycle.cs
@@ -72,6 +72,8 @@ namespace ServiceControl.Audit.Persistence.RavenDB
             }
         }
 
+        public Task Stop(CancellationToken cancellationToken) => database!.Stop(cancellationToken);
+
         public void Dispose()
         {
             documentStore?.Dispose();
--- a/src/ServiceControl.Audit.Persistence.RavenDB/RavenEmbeddedPersistenceLifecycle.cs
+++ b/src/ServiceControl.Audit.Persistence.RavenDB/RavenEmbeddedPersistenceLifecycle.cs
@@ -72,7 +72,7 @@ namespace ServiceControl.Audit.Persistence.RavenDB
             }
         }
 
-        public Task Stop(CancellationToken cancellationToken) => database!.Stop(cancellationToken);
+        public Task Stop(CancellationToken cancellationToken = default) => database!.Stop(cancellationToken);
 
         public void Dispose()
         {
--- a/src/ServiceControl.Audit.Persistence.RavenDB/RavenExternalPersistenceLifecycle.cs
+++ b/src/ServiceControl.Audit.Persistence.RavenDB/RavenExternalPersistenceLifecycle.cs
@@ -64,6 +64,8 @@ namespace ServiceControl.Audit.Persistence.RavenDB
             }
         }
 
+        public Task Stop(CancellationToken cancellationToken) => Task.CompletedTask; // We are not stopping an external instance
+
         public void Dispose() => documentStore?.Dispose();
 
         IDocumentStore? documentStore;
--- a/src/ServiceControl.Audit.Persistence.RavenDB/RavenExternalPersistenceLifecycle.cs
+++ b/src/ServiceControl.Audit.Persistence.RavenDB/RavenExternalPersistenceLifecycle.cs
@@ -64,7 +64,7 @@ namespace ServiceControl.Audit.Persistence.RavenDB
             }
         }
 
-        public Task Stop(CancellationToken cancellationToken) => Task.CompletedTask; // We are not stopping an external instance
+        public Task Stop(CancellationToken cancellationToken = default) => Task.CompletedTask; // We are not stopping an external instance
 
         public void Dispose() => documentStore?.Dispose();
 
--- a/src/ServiceControl.Persistence.RavenDB/IRavenPersistenceLifecycle.cs
+++ b/src/ServiceControl.Persistence.RavenDB/IRavenPersistenceLifecycle.cs
@@ -8,6 +8,6 @@ namespace ServiceControl.Persistence.RavenDB
     interface IRavenPersistenceLifecycle
     {
         Task Initialize(CancellationToken cancellationToken = default);
-        Task Stop(CancellationToken cancellationToken);
+        Task Stop(CancellationToken cancellationToken = default);
     }
 }
--- a/src/ServiceControl.Persistence.RavenDB/IRavenPersistenceLifecycle.cs
+++ b/src/ServiceControl.Persistence.RavenDB/IRavenPersistenceLifecycle.cs
@@ -8,5 +8,6 @@ namespace ServiceControl.Persistence.RavenDB
     interface IRavenPersistenceLifecycle
     {
         Task Initialize(CancellationToken cancellationToken = default);
+        Task Stop(CancellationToken cancellationToken);
     }
 }
--- a/src/ServiceControl.RavenDB/EmbeddedDatabase.cs
+++ b/src/ServiceControl.RavenDB/EmbeddedDatabase.cs
@@ -66,7 +66,6 @@ namespace ServiceControl.RavenDB
             Logger.InfoFormat("Loading RavenDB license from {0}", licenseFileNameAndServerDirectory.LicenseFileName);
             var serverOptions = new ServerOptions
             {
-                GracefulShutdownTimeout = TimeSpan.FromHours(1), // During Stop/Dispose we manually control this
                 CommandLineArgs =
                 [
                     $"--Logs.Mode={databaseConfiguration.LogsMode}",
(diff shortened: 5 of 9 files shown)


mauroservienti (approved):

src/ServiceControl.RavenDB/EmbeddedDatabase.cs
```diff
@@ -162,23 +162,70 @@ public async Task<IDocumentStore> Connect(CancellationToken cancellationToken)
...
+            }
 
             shutdownTokenSource.Cancel();
-            Logger.Debug("Disposing RavenDB server");
-            EmbeddedServer.Instance.Dispose();
-            Logger.Debug("Dispose RavenDB server");
+
+            if (EmbeddedServer.Instance != null)
+            {
+                serverOptions.GracefulShutdownTimeout = TimeSpan.Zero;
```
andreasohlund, line 223: Should we add a comment here to explain how we are using the "backdoor" to be able to adjust the shutdown timeout?

src/ServiceControl.RavenDB/EmbeddedDatabase.cs
```diff
@@ -262,6 +309,7 @@ static long DirSize(DirectoryInfo d)
         readonly EmbeddedDatabaseConfiguration configuration;
         readonly CancellationToken shutdownCancellationToken;
         readonly CancellationTokenRegistration applicationStoppingRegistration;
+        static ServerOptions serverOptions;
```
andreasohlund, line 312: Does it need to be a static? or would a normal field do?

src/ServiceControl.Audit.Persistence.RavenDB/IRavenPersistenceLifecycle.cs
```diff
@@ -6,5 +6,6 @@
     interface IRavenPersistenceLifecycle
     {
         Task Initialize(CancellationToken cancellationToken = default);
+        Task Stop(CancellationToken cancellationToken);
```
danielmarbach, line 9:
```suggestion
        Task Stop(CancellationToken cancellationToken = default);
```

src/ServiceControl.Audit.Persistence.RavenDB/RavenEmbeddedPersistenceLifecycle.cs
```diff
@@ -72,6 +72,8 @@ public async Task Initialize(CancellationToken cancellationToken = default)
             }
         }
 
+        public Task Stop(CancellationToken cancellationToken) => database!.Stop(cancellationToken);
```
danielmarbach, line 75:
```suggestion
        public Task Stop(CancellationToken cancellationToken = default) => database!.Stop(cancellationToken);
```

src/ServiceControl.Audit.Persistence.RavenDB/RavenExternalPersistenceLifecycle.cs
```diff
@@ -64,6 +64,8 @@ public async Task Initialize(CancellationToken cancellationToken = default)
             }
         }
 
+        public Task Stop(CancellationToken cancellationToken) => Task.CompletedTask; // We are not stopping an external instance
```
danielmarbach, line 67:
```suggestion
        public Task Stop(CancellationToken cancellationToken = default) => Task.CompletedTask; // We are not stopping an external instance
```

src/ServiceControl.Audit.Persistence.RavenDB/IRavenPersistenceLifecycle.cs
johnsimons: I would like to understand this suggestion better, @danielmarbach.
I can understand suggesting such a change for a public nuget lib like our NServiceBus Nugets.
But for an app like ServiceControl, I can't see a good reason to suggest not passing the cancellation token, and that is what I have been doing in apps code.
Can you expand on why you think this is a good change for apps?

src/ServiceControl.RavenDB/EmbeddedDatabase.cs
```diff
@@ -162,23 +175,75 @@ public async Task<IDocumentStore> Connect(CancellationToken cancellationToken)
...
+            await shutdownTokenSource.CancelAsync();
+
+            // TODO: await EmbeddedServer.Instance.StopAsync(cancellationToken);
+
+            var processId = await EmbeddedServer.Instance.GetServerProcessIdAsync(cancellationToken);
+
+            var waitForCancellationTask = Task.Delay(Timeout.Infinite, cancellationToken);
+            var firstTask = await Task.WhenAny(
+                Task.Run(() => EmbeddedServer.Instance.Dispose(), cancellationToken),
+                waitForCancellationTask
```
johnsimons, line 199: @ramonsmits I am trying to understand the reason for this code.
I thought we wanted to give a chance for RavenDB server to end on its own, and then if it takes a long time, then we resort to killing it, or am I misunderstanding?

johnsimons:
I was just looking at what `EmbededServer` is, and it seems to be just a wrapper for the RavenDB server process.
Which makes me wonder if we could do a better job by just implementing that code ourselves.
Has this been considered at all?

src/ServiceControl.RavenDB/EmbeddedDatabase.cs
ramonsmits: > I thought we wanted to give a chance for RavenDB server to end on its own, and then if it takes a long time, then we resort to killing it, or am I misunderstanding?

@johnsimons that is exactly what this code should do:

- Launch the dispose as a task, and the timer
- Wait for either to complete
- If its the timer task it ran out of time and we kill the child process

johnsimons:
> We could clone that code into our code base but it would just be better if they would have a client.

Do we even need to do that, like I said it is just a wrapper around Process.Start .... Process.End, TBH, not very useful

src/ServiceControl.RavenDB/EmbeddedDatabase.cs
```diff
@@ -162,23 +175,75 @@ public async Task<IDocumentStore> Connect(CancellationToken cancellationToken)
...
+                    await ravenChildProcess.WaitForExitAsync(CancellationToken.None);
+                }
+                catch (Exception e)
+                {
+                    Logger.ErrorFormat("Failed to kill RavenDB server PID {0}\n{1}", processId, e);
+                }
+            }
+
+            serverOptions = null!;
+            Logger.Debug("Stopped RavenDB server");
```
johnsimons, line 220:
```suggestion
            var processId = await EmbeddedServer.Instance.GetServerProcessIdAsync(cancellationToken);

            try
            {
                await Task.Run(() => EmbeddedServer.Instance.Dispose(), cancellationToken);
            }
            catch (OperationCanceledException) when (shutdownCancellationToken.IsCancellationRequested)
            {
                try
                {
                    Logger.WarnFormat("Killing RavenDB server PID {0} child process because host cancelled", processId);
                    var ravenChildProcess = Process.GetProcessById(processId);
                    ravenChildProcess.Kill(entireProcessTree: true);
                    // Kill only signals
                    Logger.WarnFormat("Waiting for RavenDB server PID {0} to exit... ", processId);
                    await ravenChildProcess.WaitForExitAsync(CancellationToken.None);
                }
                catch (Exception e)
                {
                    Logger.ErrorFormat("Failed to kill RavenDB server PID {0}\n{1}", processId, e);
                }
            }

            serverOptions = null!;
            Logger.Debug("Stopped RavenDB server");
```

src/ServiceControl.RavenDB/EmbeddedDatabase.cs
ramonsmits: @johnsimons it can take many minutes for the Raven process to finish but the dispose of their wrapper uses 30 seconds, which at the moment is longer that windows shutdown duration thus our delay is cancelled usually before the shutdown. 

If we kill it, and its already done... that is fine.

src/ServiceControl.RavenDB/EmbeddedDatabase.cs
ramonsmits: Your `await Task.Run(() => EmbeddedServer.Instance.Dispose(), cancellationToken);` will block until the Dispose completes. The cancellation token that you're passing there AFAIK only affects the task queue and  will not abandon any task passed. As in, the the CT is already set when Task.Run is invoked, it throws on OCE but once the task is executing it blocks until the task completes.

src/ServiceControl.RavenDB/EmbeddedDatabase.cs
```diff
@@ -162,23 +178,87 @@ public async Task<IDocumentStore> Connect(CancellationToken cancellationToken)
...
+            }
 
             shutdownTokenSource.Cancel();
-            Logger.Debug("Disposing RavenDB server");
-            EmbeddedServer.Instance.Dispose();
-            Logger.Debug("Dispose RavenDB server");
+
+            if (serverOptions != null)
+            {
+                // Set GracefulShutdownTimeout to Zero and exist ASAP, under normal operation instance would already
```
ramonsmits, line 254:
```suggestion
                // Set GracefulShutdownTimeout to Zero and exit ASAP, under normal operation instance would already
```

src/ServiceControl.Persistence.RavenDB/IRavenPersistenceLifecycle.cs
```diff
@@ -8,5 +8,6 @@ namespace ServiceControl.Persistence.RavenDB
     interface IRavenPersistenceLifecycle
     {
         Task Initialize(CancellationToken cancellationToken = default);
+        Task Stop(CancellationToken cancellationToken);
```
mauroservienti, line 11:
```suggestion
        Task Stop(CancellationToken cancellationToken = default);
```

mauroservienti (approved):

src/ServiceControl.RavenDB/EmbeddedDatabase.cs
```diff
@@ -162,23 +178,87 @@ public async Task<IDocumentStore> Connect(CancellationToken cancellationToken)
...
+        {
+            Logger.Debug("Stopping RavenDB server");
+            EmbeddedServer.Instance.ServerProcessExited -= OnServerProcessExited;
+
+            await shutdownTokenSource.CancelAsync();
+
+            // EmbeddedServer does not have have an async Stop method, the Dispose operation blocks and waits
+            // until its GracefulShutdownTimeout is reached and then does a Process.Kill
+            //
+            // This logic gets the child process ID uses a Task.Delay with infinite.
```
andreasohlund, line 198: This line makes no sense to me?

ramonsmits: @andreasohlund please review latest commit. I think this is an improvement.

src/ServiceControl.RavenDB/EmbeddedDatabase.cs
ramonsmits: Slightly changed behavior and added code comments to clarify cancellation behavior

andreasohlund (approved):

PR #4831, opened 2025-02-22
MessagesViewIndex change

ramonsmits:


src/ServiceControl.Audit.Persistence.RavenDB/Indexes/MessagesViewIndex.cs
```diff
@@ -12,23 +12,24 @@ public MessagesViewIndex()
         {
             Map = messages =>
                 from message in messages
+                let metadata = message.MessageMetadata
```
ramonsmits, line 15: @mauroservienti I was banging my head against the wall... but I think let somehow is not supported. Could you quickly review the CI build? The error is about object but it does not make sense to me.

Issue it to that I for some reason cannot run these test locally so I cannot easily run them

ramonsmits:
Closing, not really much to gain except a slightly improved array allocation which isn't going to make a big difference performance wise

PR #4834, opened 2025-02-24
Improving provenance of image

johnsimons:
Using native action to login to ghcr.io.
Replaced multiple tagging commands with a single one Using build-push-action instead of script.
This adds the provenance by default, see https://docs.docker.com/build/ci/github-actions/attestations/#default-provenance.

This gives the image a better score in DockerHub scout health score.

This is the same changes as https://github.com/Particular/ServiceControl.Connector.MassTransit/pull/182

### DIFF
--- a/.github/workflows/build-containers.yml
+++ b/.github/workflows/build-containers.yml
@@ -12,10 +12,16 @@ jobs:
         include:
           - name: servicecontrol
             project: ServiceControl
+            title: ServiceControl
+            description: Gather status, performance and monitoring data for multiple endpoints from a single location.
           - name: servicecontrol-audit
             project: ServiceControl.Audit
+            title: ServiceControl Audit Instance
+            description: Provide valuable information about the message flow through a system.
           - name: servicecontrol-monitoring
             project: ServiceControl.Monitoring
+            title: ServiceControl Monitoring Instance
+            description: Track the health of a distributed system.
       fail-fast: false
     steps:
       - name: Check for secrets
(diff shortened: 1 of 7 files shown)


irinascurtu (approved):

.github/workflows/build-containers.yml
```diff
@@ -38,26 +35,26 @@ jobs:
...
-              --annotation "index:org.opencontainers.image.description=${{ matrix.description }}" \
-              --annotation "index:org.opencontainers.image.created=$(date '+%FT%TZ')" \
-              --annotation "index:org.opencontainers.image.revision=${{ github.sha }}" \
-              --annotation "index:org.opencontainers.image.authors=Particular Software" \
-              --annotation "index:org.opencontainers.image.vendor=Particular Software" \
-              --annotation "index:org.opencontainers.image.version=${{ env.MinVerVersion }}" \
-              --annotation "index:org.opencontainers.image.source=https://github.com/${{ github.repository }}/tree/${{ github.sha }}" \
-              --annotation "index:org.opencontainers.image.url=https://hub.docker.com/r/particular/${{ matrix.name }}" \
-              --annotation "index:org.opencontainers.image.documentation=https://docs.particular.net/servicecontrol/" \
-              --annotation "index:org.opencontainers.image.base.name=mcr.microsoft.com/dotnet/aspnet:8.0-jammy-chiseled-composite-extra" \
```
DavidBoike, line 61: By removing all of this stuff, that means we're not setting any annotations at the index level, only the image level. Is there a way to maintain this data at the index level as well using the build actions?

DavidBoike (changes requested):

.github/workflows/build-containers.yml
johnsimons: @DavidBoike do we need it? What does it do for us?
Also, how do I check what is set and what is not?
I am asking because if you look at the cli call it produces, it creates some metadata, see https://github.com/Particular/ServiceControl/actions/runs/13491741985/job/37690969138?pr=4834#step:9:175, so maybe it does set it.

.github/workflows/build-containers.yml
```diff
@@ -38,26 +35,26 @@ jobs:
         with:
           version: ${{ env.MinVerVersion }}
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v3.9.0
+        uses: docker/setup-buildx-action@v3
```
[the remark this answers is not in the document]
johnsimons: @DavidBoike I read the link:
```
 When using workflow actions, we use an explicit patch version (i.e. @v2.3.4) like this:

- name: Checkout
  uses: actions/checkout@v2.3.4
  with:
    fetch-depth: 0
Updates to workflow action versions can be [automatically applied via dependabot](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-actions-up-to-date-with-dependabot).
```

It doesn't really explain why it is preferred.
I don't really care, but in my opinion, it is pointless to have Dependbot doing these updates constantly.

.github/workflows/build-containers.yml
DavidBoike: If you log in to GHCR with your PAT, and then look at an existing manifest, let's say for `6.4.0-alpha.0.22` using the following: (`--raw` gives it to you as raw JSON that doesn't strip stuff out)

```
docker buildx imagetools inspect ghcr.io/particular/servicecontrol:6.4.0-alpha.0.22 --raw
```

You'll see all the annotations there, this is the "index" level, the thing that points to the amd64/arm64 digests as well as the "unknown" architecture digests that are attestation metadata for each of the 2 architecures.

You can also pull the amd64 metadata by referencing the SHA:

```
docker buildx imagetools inspect ghcr.io/particular/servicecontrol:6.4.0-alpha.0.22@sha256:8997a5697b96a0d0a363a9f6b922d337847c3fa85a20a38b90e2a58739f9b91c --raw
```

No annotations/attestations, just pointing to the layers.

If you do an attestation manifest:

```
docker buildx imagetools inspect ghcr.io/particular/servicecontrol:6.4.0-alpha.0.22@sha256:c97784a1668a0aeb58ac319881f29c1111a2f83773e8459b9e32a769f1c3cdd1 --raw
```

Not much to see there.

So all to say the hope was that we were annotating everything at the "index" level which would apply to the whole version including all architectures, which this PR is currently losing.

If I repeat for the tag `pr-4834` built by this PR:

```
docker buildx imagetools inspect ghcr.io/particular/servicecontrol:pr-4834 --raw

docker buildx imagetools inspect ghcr.io/particular/servicecontrol:pr-4834@sha256:9f4649f3780631024b3c4d24baf220fdcc53b2e23e25b3b872ef96ab3c0cf0a0 --raw

docker buildx imagetools inspect ghcr.io/particular/servicecontrol:6.4.0-alpha.0.22@sha256:e616aee45bbb8f7bfb29dcce810c269a3004dd88cc7085834c046860a3c6c942 --raw
```

First one loses the annotations and doesn't really "replace" it anywhere.

So Docker Hub doesn't currently show this anywhere (other registries might? GHCR only shows the description on pages [like this one](https://github.com/Particular/ServiceControl/pkgs/container/servicecontrol/362181683?tag=6.4.0-alpha.0.22) but that's just a staging area) but given the annotations are a standard it seems reasonable to assume that someday it might?? So while not critical it does concern me to just throw it away.

.github/workflows/build-containers.yml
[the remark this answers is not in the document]
DavidBoike: For workflows that participate in CI (which this does) we want as close to repeatable builds as we can (within reason) and we want the feedback via a failed Dependabot PR that there's something that breaks so there are no surprises when we go to release. That is more "noise" but we have processed in place to auto-merge green builds of trusted actions.

.github/workflows/build-containers.yml
johnsimons: @DavidBoike I think this may work [`14c63fd` (#4834)](https://github.com/Particular/ServiceControl/pull/4834/commits/14c63fd1ce33fd71536c3645fa407421391241d5)
but there is still a bit of repetition, but at least it is all in the same file.
What do you think?

DavidBoike: From `docker buildx imagetools inspect ghcr.io/particular/servicecontrol:pr-4834 --raw ` looks like it does the trick for the annotations, but I have to admit I don't even remember (if I ever knew) how to validate that the labels are applied.

johnsimons: I'll check the labels

DavidBoike (approved):
Removing my "request changes"

PhilBastian (approved):

PR #4835, opened 2025-02-24
Use correct name for the instance setup phase

andreasohlund:


src/ServiceControlInstaller.Engine/Instances/MonitoringInstance.cs
```diff
@@ -188,13 +188,9 @@ public void ApplyConfigChange()
             {
                 try
                 {
-                    QueueCreation.RunQueueCreation(this);
+                    InstanceSetup.Run(this);
                 }
-                catch (QueueCreationFailedException ex)
```
andreasohlund, line 193: We had no logic that needed explicit exception classes to be used so removed them. This now also catches plain `Exception` which previously would have crashed the installer

src/ServiceControlInstaller.Engine/Setup/InstanceSetup.cs
```diff
@@ -0,0 +1,70 @@
+﻿namespace ServiceControlInstaller.Engine.Setup;
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using Instances;
+
+static class InstanceSetup
```
andreasohlund, line 8: This is the renamed QueueCreation class

src/ServiceControlInstaller.Engine/Setup/InstanceSetup.cs
```diff
@@ -0,0 +1,70 @@
...
+        processStartupInfo.EnvironmentVariables.Add("INSTANCE_NAME", instanceName);
+
+        var p = Process.Start(processStartupInfo);
+        if (p != null)
+        {
+            var error = p.StandardError.ReadToEnd();
+            p.WaitForExit((int)TimeSpan.FromMinutes(1).TotalMilliseconds);
+            if (!p.HasExited)
+            {
+                p.Kill();
```
andreasohlund, line 56: @ramonsmits I wonder if we should do a WaitForExit here as well?

src/ServiceControlInstaller.Engine/Setup/InstanceSetup.cs
```diff
@@ -0,0 +1,70 @@
...
+            RedirectStandardError = true
+        };
+
+        processStartupInfo.EnvironmentVariables.Add("INSTANCE_NAME", instanceName);
+
+        var p = Process.Start(processStartupInfo);
+        if (p != null)
+        {
+            var error = p.StandardError.ReadToEnd();
+            p.WaitForExit((int)TimeSpan.FromMinutes(1).TotalMilliseconds);
```
andreasohlund, line 53: Use 10 min

src/ServiceControlInstaller.Engine/Setup/InstanceSetup.cs
```diff
@@ -0,0 +1,70 @@
...
+            if (!p.HasExited)
+            {
+                p.Kill();
```
andreasohlund, line 56: Don't kill since it can break things, we should report back instead

ramonsmits (approved):
As discussed, timeout duration and killing behavior will be addressed in a secondary PR.

src/ServiceControl.Config/Framework/Modules/InstallerModule.cs
```diff
@@ -67,7 +67,7 @@ internal async Task<ReportCard> Add(ServiceControlInstallableBase details, IProg
             {
                 progress.Report(5, 9, "Registering URL ACLs...");
                 instanceInstaller.RegisterUrlAcl();
-                progress.Report(6, 9, "Creating queues...");
+                progress.Report(6, 9, "Running instance setup (can take a very long time)...");
```
mauroservienti, line 70:
```suggestion
                progress.Report(6, 9, "Instance setup in progress, this could take several minutes...");
```

src/ServiceControl.Config/Framework/Modules/InstallerModule.cs
```diff
@@ -141,7 +141,7 @@ internal ReportCard Upgrade(ServiceControlBaseService instance, ServiceControlUp
 
             UpgradeOptions(upgradeOptions, instance);
 
-            progress.Report(++currentStep, totalSteps, "Running Queue Creation...");
+            progress.Report(++currentStep, totalSteps, "Running instance setup (can take a very long time)...");
```
mauroservienti, line 144:
```suggestion
            progress.Report(++currentStep, totalSteps, "Instance setup in progress, this could take several minutes...");
```

PR #4842, opened 2025-02-26
Allow RavenDB search engine type to be configured

andreasohlund:


andreasohlund:
After talking to @bording we decided to instead make sure that we honor any manual overrides made by the user. That way indexes can be modified in maintenance mode and kept that way.

The idea is to use https://ravendb.net/docs/article-page/5.0/csharp/client-api/operations/maintenance/indexes/set-index-lock as the way to signal this.

Ie: 

1. Make the change
2. Lock the index to tell SC to not reset it on restart


https://github.com/Particular/ServiceControl/pull/4849 will make sure that we honour this

PR #4849, opened 2025-02-28
Add tests to make sure that indexes being locked  prevents index resets on instance restarts

andreasohlund:
This means that user can opt-into Lucene by doing the following in Raven Studio:

1. Overriding search engine type to Lucene for a specific index
2. Lock the index [with LockMode set to Ignore](https://ravendb.net/docs/article-page/7.0/csharp/client-api/operations/maintenance/indexes/set-index-lock#lock-modes)

### DIFF
--- a/src/ServiceControl.Audit.Persistence.Tests.RavenDB/IndexSetupTests.cs
+++ b/src/ServiceControl.Audit.Persistence.Tests.RavenDB/IndexSetupTests.cs
@@ -58,7 +58,7 @@ class IndexSetupTests : PersistenceTestFixture
 
         await DatabaseSetup.CreateIndexes(configuration.DocumentStore, true, CancellationToken.None);
 
-        WaitForIndexDefinitionUpdate(indexWithCustomConfigStats);
+        await WaitForIndexDefinitionUpdate(indexWithCustomConfigStats);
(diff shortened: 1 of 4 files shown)


src/ServiceControl.Audit.Persistence.Tests.RavenDB/IndexSetupTests.cs
```diff
@@ -0,0 +1,58 @@
...
+
+    [Test]
+    public async Task Indexes_should_be_reset_on_setup()
+    {
+        var index = new MessagesViewIndexWithFullTextSearch { Configuration = { ["Indexing.Static.SearchEngineType"] = SearchEngineType.Lucene.ToString() } };
+
+        await IndexCreation.CreateIndexesAsync([index], configuration.DocumentStore);
+
+        //TODO: find a better way
+        await Task.Delay(1000);
```
andreasohlund, line 44: @mauroservienti do you know of a way to wait for the Raven to complete the operations to avoid this hack?

andreasohlund:
@mauroservienti I called YAGNI on try catch and log the exception if index is locked with error

mauroservienti (approved):

PR #4852, opened 2025-03-02
Do not kill instance setup to minimize the risk for a corrupted database

andreasohlund:


mauroservienti (commented):
I don't think that waiting forever is the best approach here. We should be waiting longer than 1 minute for sure, but then show users a warning message that explains what's going on and what they could do to mitigate the problem. E.g., open the RavenDB studio, pause the message view index, go back to SCMU and try again.

ramonsmits:
Indeed, reporting progress would be a nice to have but the primary reason for needing to wait very long is because the database could be recovering.

I still think we should wait "till infinity" for setup and never kill it. Abandoning it to me is incorrect as now a process is running on the system for which the user doesn't know WHAT process it is and what to do with it.

The only thing is that we would improvement the reporting. Stating, "this can take a very long time to complete. If this takes longer then 10 minutes review the logs and contact Particular Support if assistance is required"


Only if the installer itself supports "cancellation" then it could have a CancellationTokenSource that triggers after for example 5 minutes but what if the setup could run and wait until the database is in a "healthy" state?

src/ServiceControlInstaller.Engine/Setup/InstanceSetup.cs
```diff
@@ -46,25 +46,16 @@ static void Run(string installPath, string exeName, string instanceName, bool sk
...
-                throw new Exception($"{exeName} threw an error when performing setup. This typically indicates a configuration error. The error output from {exeName} was:\r\n {error}");
-            }
-        }
-        else
+        var error = p.StandardError.ReadToEnd();
+
+        // we will wait "forever" since killing the setup is dangerous and can lead to the database being in an invalid state
+        p.WaitForExit();
+
+        if (p.ExitCode != 0)
```
ramonsmits, line 56: I'm not sure, but I think when an exception is thrown the exit code is negative. If we add any validation we should return a positive exit code. If its possitive the message below could be slight different

Like:

    var message = $"{exeName} failed, reason:\r\n {error}");

andreasohlund:
@mauroservienti I've discussed with Ramon and we think infinite wait is the way to go for now, comment added to explain

ramonsmits (approved):

PR #4863, opened 2025-03-07
Limit throughput data returned to 14 months

johnsimons:
The throughput calculation is supposed to only look at the last 12 months of data.
This PR adds a filter to the data, so it only considers the last 14 months.
We use 14 months to ensure we have whole months and not partial months.


andreasohlund:
@johnsimons the title read like an improvement so we relabeled it, feel free to make it a bug again and change the title if needed

PR #4879, opened 2025-03-13
Prevent correct connection string being overwritten during upgrade of primary instance

abparticular:


bording:
We ended up handling this is a different way: https://github.com/Particular/ServiceControl/pull/4755/commits/6ca6c1d88f95e0d915e73a55a23e67834cc44795

PR #4891, opened 2025-03-19
Add exponential backoff when the async timer fails

andreasohlund:
We've seen logs being flooded by the coding using the async timer during failure scenarios.

To prevent this an exponential backoff strategy capped at 60s will be used for consecutive failures executing the async timer.

### DIFF
--- a/src/ServiceControl.Infrastructure/AsyncTimer.cs
+++ b/src/ServiceControl.Infrastructure/AsyncTimer.cs
@@ -57,7 +57,7 @@ public class TimerJob
         }, CancellationToken.None);
     }
 
-    public async Task Stop()
+    public async Task Stop(CancellationToken cancellationToken)
     {
         if (tokenSource == null)
         {
@@ -67,16 +67,18 @@ public class TimerJob
         await tokenSource.CancelAsync().ConfigureAwait(false);
         tokenSource.Dispose();
 
-        if (task != null)
+        if (task == null)
         {
-            try
-            {
-                await task.ConfigureAwait(false);
-            }
-            catch (OperationCanceledException) when (tokenSource.IsCancellationRequested)
-            {
-                //NOOP
-            }
+            return;
+        }
+
+        try
+        {
+            Task.WaitAll([task], cancellationToken);
+        }
+        catch (OperationCanceledException) when (tokenSource.IsCancellationRequested)
+        {
+            //NOOP
         }
     }
 
--- a/src/ServiceControl.Infrastructure/AsyncTimer.cs
+++ b/src/ServiceControl.Infrastructure/AsyncTimer.cs
@@ -24,11 +24,15 @@ public class TimerJob
             {
                 await Task.Delay(due, token).ConfigureAwait(false);
 
+                var consecutiveFailures = 0;
+
                 while (!token.IsCancellationRequested)
                 {
                     try
                     {
                         var result = await callback(token).ConfigureAwait(false);
+
+                        consecutiveFailures = 0;
                         if (result == TimerJobExecutionResult.DoNotContinueExecuting)
                         {
                             tokenSource.Cancel();
@@ -46,6 +50,11 @@ public class TimerJob
                     }
                     catch (Exception ex)
                     {
+                        consecutiveFailures++;
+                        var exponentialBackoffDelay = TimeSpan.FromSeconds(int.Max(60, consecutiveFailures * consecutiveFailures));
+
+                        await Task.Delay(exponentialBackoffDelay, token).ConfigureAwait(false);
+
                         errorCallback(ex);
                     }
                 }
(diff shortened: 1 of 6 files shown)


src/ServiceControl.Infrastructure/AsyncTimer.cs
```diff
@@ -1,96 +1,106 @@
...
+        if (task == null)
+        {
+            return;
         }
 
-        Task task;
-        CancellationTokenSource tokenSource;
+        try
+        {
+            Task.WaitAll([task], cancellationToken);
```
andreasohlund, line 86: Is there a better way to do this?

src/ServiceControl.Infrastructure/AsyncTimer.cs
```diff
@@ -1,96 +1,106 @@
...
-            }, CancellationToken.None);
-        }
+                    catch (OperationCanceledException) when (token.IsCancellationRequested)
+                    {
+                        break;
+                    }
+                    catch (Exception ex)
+                    {
+                        consecutiveFailures++;
+                        var exponentialBackoffDelay = TimeSpan.FromSeconds(int.Min(60, consecutiveFailures * consecutiveFailures));
```
andreasohlund, line 54: Could the lead to a math error if the situation persists for a long time and consecutiveFailures becomes high?

src/ServiceControl/Monitoring/HeartbeatMonitoringHostedService.cs
```diff
@@ -24,17 +24,7 @@ public async Task StartAsync(CancellationToken cancellationToken)
...
-            try
-            {
-                await timer.Stop();
-            }
-            catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
-            {
-                //NOOP, invoked Stop does not
-            }
-        }
+        public Task StopAsync(CancellationToken cancellationToken) => timer.Stop(cancellationToken);
```
mauroservienti, line 27: Isn't this changing the previous behavior? If this throws an OperationCanceledException and cancellationToken.IsCancellationRequested is true, the exception will bubble up instead of being suppressed like before. Was that intentional?

src/ServiceControl.Infrastructure/AsyncTimer.cs
mauroservienti: Will `consecutiveFailures * consecutiveFailures` be downcasted to an `int`?

andreasohlund: Changed the approach to be safe

src/ServiceControl/Monitoring/HeartbeatMonitoringHostedService.cs
andreasohlund: You are right but that now happens in the async timer so end result should be the same? (see https://github.com/Particular/ServiceControl/pull/4891/files#diff-96d7e0170b214f88063570076afc55fe8451f84549113c204650a38ea9cec8c9R88)

src/ServiceControl.Infrastructure/AsyncTimer.cs
mauroservienti: thinking out laud, isn't there a `task.Wait()` thing?

andreasohlund: Duh, I looked for it in the wrong place 😬 

Fixed now

src/ServiceControl.Infrastructure/AsyncTimer.cs
```diff
@@ -1,96 +1,106 @@
...
-            }, CancellationToken.None);
-        }
+                    catch (OperationCanceledException) when (token.IsCancellationRequested)
+                    {
+                        break;
+                    }
+                    catch (Exception ex)
+                    {
+                        consecutiveFailures++;
+                        var backoffDelay = TimeSpan.FromSeconds(int.Min(6, consecutiveFailures) * 10);
```
ramonsmits, line 54: The flooding happens because this Task.Delay isn't happening. This PR does improve by adding a Task.Delay in the catch. I do find this section a bit hard to read.

How about:
```suggestion
                        const int MaxDelayDurationInSeconds = 60;
                        var delayInSeconds = consecutiveFailures * 10;
                        var backoffDelay = TimeSpan.FromSeconds(int.Min(MaxDelayDurationInSeconds, delayInSeconds));
```

src/ServiceControl.Infrastructure/AsyncTimer.cs
```diff
@@ -1,96 +1,106 @@
...
-                {
-                    await task.ConfigureAwait(false);
-                }
-                catch (OperationCanceledException) when (tokenSource.IsCancellationRequested)
-                {
-                    //NOOP
-                }
-            }
+        await tokenSource.CancelAsync().ConfigureAwait(false);
+        tokenSource.Dispose();
```
ramonsmits, line 77:
```suggestion
        tokenSource.Dispose();
        tokenSource = null;
```

ramonsmits (approved):

PR #4896, opened 2025-03-20
use char override for string.Join

SimonCropp:
avoids putting a string on the heap


ramonsmits (approved):

bording:
Merging into `master` via #4902

PR #4898, opened 2025-03-20
Remove analyser for configure await

andreasohlund:
Since I don't see a use case where SC would need to set it

https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2007#when-to-suppress-warnings

### DIFF
--- a/src/.editorconfig
+++ b/src/.editorconfig
@@ -27,9 +27,8 @@ insert_final_newline = false
 dotnet_diagnostic.CA1068.severity = error
 dotnet_diagnostic.CA1710.severity = error
 dotnet_diagnostic.CA1711.severity = error
-dotnet_diagnostic.CA2007.severity = error
+dotnet_diagnostic.CA2007.severity = none
 dotnet_diagnostic.CA2016.severity = error
-
 
 #### .NET Coding Conventions ####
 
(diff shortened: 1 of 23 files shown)


src/.editorconfig
```diff
@@ -27,10 +27,9 @@ insert_final_newline = false
 dotnet_diagnostic.CA1068.severity = error
 dotnet_diagnostic.CA1710.severity = error
 dotnet_diagnostic.CA1711.severity = error
-dotnet_diagnostic.CA2007.severity = error
+dotnet_diagnostic.CA2007.severity = none
```
mauroservienti, line 30: Won't this be reset by repo sync?

mauroservienti (approved):

src/.editorconfig
andreasohlund: ugh, didn't think about that, is there a workaround for that?

danielmarbach:
Agreed the analyzer is not required here

src/.editorconfig
bording: No there isn't. If want to make this change, you need to add an .editorconfig file to every project to disable it.

andreasohlund: Got it so the only (future) option would be to have different editor configs for "library" repos vs "non-library" repos where that setting could be different?

bording: There's no way to have that sort of variation given the way RepoSync is implemented.

mauroservienti:
Repo sync will override this change if we merge it. Closing.

PR #4918, opened 2025-03-31
New endpoint to support migration of SI to SP

johnsimons:
As part of the ongoing work to migrate SI to SP, we identified that paging in SI does not return the correct results once the user reaches page 2 or higher. The only correct page is page 1.

The TF working on this has evaluated whether paging is needed at all for the screens in SP and decided that, to provide accurate results, we are not migrating paging, given the issues.

This PR introduces a new endpoint that:
- returns only the top x messages
- supports filtering the messages by time sent range
- fixed a bug where the results could return more than the page size limit

### DIFF
--- a/src/ServiceControl.Audit.Persistence.RavenDB/Extensions/RavenQueryExtensions.cs
+++ b/src/ServiceControl.Audit.Persistence.RavenDB/Extensions/RavenQueryExtensions.cs
@@ -1,6 +1,7 @@
 ﻿namespace ServiceControl.Audit.Persistence.RavenDB.Extensions
 {
     using System;
+    using System.Globalization;
     using System.Linq;
     using System.Linq.Expressions;
     using Indexes;
@@ -18,6 +19,45 @@
 
             return source;
         }
+
+        public static IQueryable<MessagesViewIndex.SortAndFilterOptions> FilterBySentTimeRange(this IQueryable<MessagesViewIndex.SortAndFilterOptions> source, string range)
+        {
+            if (string.IsNullOrWhiteSpace(range))
+            {
+                return source;
+            }
+
+            var filters = range.Split(SplitChars, StringSplitOptions.None);
+            DateTime from, to;
+            try
+            {
+
+                if (filters.Length == 2)
+                {
+                    from = DateTime.Parse(filters[0], CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind);
+                    to = DateTime.Parse(filters[1], CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind);
+                    source.Where(m => m.TimeSent >= from && m.TimeSent <= to);
+
+                }
+                else
+                {
+                    from = DateTime.Parse(filters[0], CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind);
+                    source.Where(m => m.TimeSent >= from);
+                }
+            }
+            catch (Exception)
+            {
+                throw new Exception(
+                    "Invalid sent time date range, dates need to be in ISO8601 format and it needs to be a range eg. 2016-03-11T00:27:15.474Z...2016-03-16T03:27:15.474Z");
+            }
+
+            return source;
+        }
+
+        static string[] SplitChars =
+        {
+            "..."
+        };
 
         public static IQueryable<TSource> Paging<TSource>(this IQueryable<TSource> source, PagingInfo pagingInfo)
             => source.Skip(pagingInfo.Offset).Take(pagingInfo.PageSize);
--- a/src/ServiceControl.Audit.Persistence.RavenDB/Extensions/RavenQueryExtensions.cs
+++ b/src/ServiceControl.Audit.Persistence.RavenDB/Extensions/RavenQueryExtensions.cs
@@ -1,7 +1,6 @@
 ﻿namespace ServiceControl.Audit.Persistence.RavenDB.Extensions
 {
     using System;
-    using System.Globalization;
     using System.Linq;
     using System.Linq.Expressions;
     using Indexes;
@@ -20,44 +19,25 @@
             return source;
         }
 
-        public static IQueryable<MessagesViewIndex.SortAndFilterOptions> FilterBySentTimeRange(this IQueryable<MessagesViewIndex.SortAndFilterOptions> source, string range)
+        public static IQueryable<MessagesViewIndex.SortAndFilterOptions> FilterBySentTimeRange(this IQueryable<MessagesViewIndex.SortAndFilterOptions> source, DateTimeRange range)
         {
-            if (string.IsNullOrWhiteSpace(range))
+            if (range == null)
             {
                 return source;
             }
 
-            var filters = range.Split(SplitChars, StringSplitOptions.None);
-            DateTime from, to;
-            try
+            if (range.From.HasValue)
             {
-
-                if (filters.Length == 2)
-                {
-                    from = DateTime.Parse(filters[0], CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind);
-                    to = DateTime.Parse(filters[1], CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind);
-                    source.Where(m => m.TimeSent >= from && m.TimeSent <= to);
-
-                }
-                else
-                {
-                    from = DateTime.Parse(filters[0], CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind);
-                    source.Where(m => m.TimeSent >= from);
-                }
+                source = source.Where(m => m.TimeSent >= range.From);
             }
-            catch (Exception)
+
+            if (range.To.HasValue)
             {
-                throw new Exception(
-                    "Invalid sent time date range, dates need to be in ISO8601 format and it needs to be a range eg. 2016-03-11T00:27:15.474Z...2016-03-16T03:27:15.474Z");
+                source = source.Where(m => m.TimeSent <= range.To);
             }
 
             return source;
         }
-
-        static string[] SplitChars =
-        {
-            "..."
-        };
 
         public static IQueryable<TSource> Paging<TSource>(this IQueryable<TSource> source, PagingInfo pagingInfo)
             => source.Skip(pagingInfo.Offset).Take(pagingInfo.PageSize);
--- a/src/ServiceControl.Audit.Persistence/IAuditDataStore.cs
+++ b/src/ServiceControl.Audit.Persistence/IAuditDataStore.cs
@@ -14,10 +14,10 @@
     {
         Task<QueryResult<IList<KnownEndpointsView>>> QueryKnownEndpoints(CancellationToken cancellationToken);
         Task<QueryResult<SagaHistory>> QuerySagaHistoryById(Guid input, CancellationToken cancellationToken);
-        Task<QueryResult<IList<MessagesView>>> GetMessages(bool includeSystemMessages, PagingInfo pagingInfo, SortInfo sortInfo, CancellationToken cancellationToken);
-        Task<QueryResult<IList<MessagesView>>> QueryMessages(string searchParam, PagingInfo pagingInfo, SortInfo sortInfo, CancellationToken cancellationToken);
-        Task<QueryResult<IList<MessagesView>>> QueryMessagesByReceivingEndpointAndKeyword(string endpoint, string keyword, PagingInfo pagingInfo, SortInfo sortInfo, CancellationToken cancellationToken);
-        Task<QueryResult<IList<MessagesView>>> QueryMessagesByReceivingEndpoint(bool includeSystemMessages, string endpointName, PagingInfo pagingInfo, SortInfo sortInfo, CancellationToken cancellationToken);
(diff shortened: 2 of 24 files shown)


src/ServiceControl.Audit/Auditing/MessagesView/GetMessages2Controller.cs
```diff
@@ -0,0 +1,56 @@
...
+using Infrastructure;
+using Infrastructure.WebApi;
+using Microsoft.AspNetCore.Mvc;
+using Persistence;
+
+[ApiController]
+[Route("api")]
+public class GetMessages2Controller(IAuditDataStore dataStore) : ControllerBase
+{
+    [Route("messages2")]
```
ramonsmits, line 15: Maybe better as `messages-v2` or `messages/v2`?

src/ServiceControl.Audit.Persistence/IAuditDataStore.cs
```diff
@@ -14,10 +14,10 @@ public interface IAuditDataStore
     {
         Task<QueryResult<IList<KnownEndpointsView>>> QueryKnownEndpoints(CancellationToken cancellationToken);
         Task<QueryResult<SagaHistory>> QuerySagaHistoryById(Guid input, CancellationToken cancellationToken);
-        Task<QueryResult<IList<MessagesView>>> GetMessages(bool includeSystemMessages, PagingInfo pagingInfo, SortInfo sortInfo, CancellationToken cancellationToken);
-        Task<QueryResult<IList<MessagesView>>> QueryMessages(string searchParam, PagingInfo pagingInfo, SortInfo sortInfo, CancellationToken cancellationToken);
-        Task<QueryResult<IList<MessagesView>>> QueryMessagesByReceivingEndpointAndKeyword(string endpoint, string keyword, PagingInfo pagingInfo, SortInfo sortInfo, CancellationToken cancellationToken);
-        Task<QueryResult<IList<MessagesView>>> QueryMessagesByReceivingEndpoint(bool includeSystemMessages, string endpointName, PagingInfo pagingInfo, SortInfo sortInfo, CancellationToken cancellationToken);
+        Task<QueryResult<IList<MessagesView>>> GetMessages(bool includeSystemMessages, PagingInfo pagingInfo, SortInfo sortInfo, string timeSentRange, CancellationToken cancellationToken);
```
ramonsmits, line 17: It seems this can be null, an optional argument seems better. Also, at this level I think we should avoid strings and be already as  type-safe as we can. I would prefer a tuple `(DateTime From, DateTime To)` or a range record/class type just like the other typed arguments.

src/ServiceControl.Audit.Persistence.RavenDB/Extensions/RavenQueryExtensions.cs
```diff
@@ -19,6 +20,45 @@ static class RavenQueryExtensions
...
+            catch (Exception)
+            {
+                throw new Exception(
+                    "Invalid sent time date range, dates need to be in ISO8601 format and it needs to be a range eg. 2016-03-11T00:27:15.474Z...2016-03-16T03:27:15.474Z");
+            }
+
+            return source;
+        }
+
+        static string[] SplitChars =
```
ramonsmits, line 57: Why isn't this just 
```suggestion
        const string SplitChars = "...";
```

src/ServiceControl.Audit.Persistence.RavenDB/Extensions/RavenQueryExtensions.cs
```diff
@@ -19,6 +20,45 @@ static class RavenQueryExtensions
...
+                }
+                else
+                {
+                    from = DateTime.Parse(filters[0], CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind);
+                    source.Where(m => m.TimeSent >= from);
+                }
+            }
+            catch (Exception)
+            {
+                throw new Exception(
```
ramonsmits, line 50: Looking at the logic it supports not only a range, but also a single value.

src/ServiceControl.Audit.Persistence.RavenDB/Extensions/RavenQueryExtensions.cs
```diff
@@ -19,6 +20,45 @@ static class RavenQueryExtensions
...
+            var filters = range.Split(SplitChars, StringSplitOptions.None);
+            DateTime from, to;
+            try
+            {
+
+                if (filters.Length == 2)
+                {
+                    from = DateTime.Parse(filters[0], CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind);
+                    to = DateTime.Parse(filters[1], CultureInfo.InvariantCulture, DateTimeStyles.RoundtripKind);
+                    source.Where(m => m.TimeSent >= from && m.TimeSent <= to);
```
ramonsmits, line 39: Does the TO not need to be exclusive?
```suggestion
                    source.Where(m => m.TimeSent >= from && m.TimeSent < to);
```

src/ServiceControl/CompositeViews/Messages/ScatterGatherApiMessageView.cs
```diff
@@ -51,7 +52,7 @@ protected override IList<MessagesView> ProcessResults(TInput input, QueryResult<
                 combined.Sort(comparer);
             }
 
-            return combined;
+            return combined.Take(input.PagingInfo.PageSize).ToList();
```
ramonsmits, line 55: This is a fix right? Better to isolated that in a separate PR.

src/ServiceControl/CompositeViews/Messages/GetMessages2Controller.cs
```diff
@@ -0,0 +1,66 @@
...
+[ApiController]
+[Route("api")]
+public class GetMessages2Controller(
+    GetAllMessagesApi allMessagesApi,
+    GetAllMessagesForEndpointApi allMessagesEndpointApi,
+    SearchApi searchApi,
+    SearchEndpointApi searchEndpointApi)
+    : ControllerBase
+{
+    [Route("messages2")]
```
ramonsmits, line 19: Same comment

ramonsmits (changes requested):
Can you explain why this new endpoint it needed? Second, why is the filtering needed on "time sent"? This to be feels more sorting/seeking requirement but that could conflicting with SortInfo.

As each index uses a different type of ordering we need to "seek" using the ordered index attributes.

This is a new API and the sorting/seeking seems incorrect.

src/ServiceControl.Audit.Persistence.RavenDB/Extensions/RavenQueryExtensions.cs
johnsimons: no, the range is inclusive, it uses the between syntax of Raven

src/ServiceControl.Audit/Auditing/MessagesView/GetMessagesController.cs
```diff
@@ -70,7 +70,7 @@ public async Task<IActionResult> Get(string id, CancellationToken cancellationTo
         [HttpGet]
         public async Task<IList<MessagesView>> Search([FromQuery] PagingInfo pagingInfo, [FromQuery] SortInfo sortInfo, string q, CancellationToken cancellationToken)
         {
-            var result = await dataStore.QueryMessages(q, pagingInfo, sortInfo, cancellationToken);
+            var result = await dataStore.QueryMessages(q, pagingInfo, sortInfo, null, cancellationToken);
```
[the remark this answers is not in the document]
johnsimons: Fixed

src/ServiceControl.Audit/Auditing/MessagesView/GetMessages2Controller.cs
johnsimons: Yeah, hard to name things

src/ServiceControl.Audit.Persistence/IAuditDataStore.cs
johnsimons: Done

src/ServiceControl.Audit.Persistence/Infrastructure/DateTimeRange.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Globalization;
+
+public class DateTimeRange
+{
+    public DateTime? From { get; }
+    public DateTime? To { get; }
+
+    public DateTimeRange(string from = null, string to = null)
+    {
+        if (from != null)
```
PhilBastian, line 13: exception handling for invalid date string?

src/ServiceControl.Persistence/Infrastructure/DateTimeRange.cs
```diff
@@ -0,0 +1,28 @@
...
+using System.Globalization;
+
+public class DateTimeRange
+{
+    public DateTime? From { get; }
+    public DateTime? To { get; }
+
+    public DateTimeRange(string from = null, string to = null)
+    {
+        if (from != null)
```
PhilBastian, line 13: exception handling for invalid date string?

src/ServiceControl/CompositeViews/Messages/ScatterGatherApiMessageView.cs
PhilBastian: I don't think this is a fix to the existing behaviour - this is what's required by the new behaviour (getting first x results only)

That being said, for old versions of SP that still use the old API, this is going to result in even more broken behaviour where they can't even see large chunks of messages. It would make sense to have this behaviour conditional on the new API only

PhilBastian (approved):

src/ServiceControl.Audit.Persistence/Infrastructure/DateTimeRange.cs
johnsimons: Exceptions in REST apis are hard, they don't really bubble to the UI,
So not sure what can we do better?

src/ServiceControl/CompositeViews/Messages/ScatterGatherApiMessageView.cs
johnsimons: Happy to toggel it only conditionally though

src/ServiceControl.Audit.Persistence/Infrastructure/DateTimeRange.cs
PhilBastian: The point is that this is currently throwing an exception. At least ignore the value if it's not valid so that the client still receives data

src/ServiceControl/CompositeViews/Messages/ScatterGatherApiMessageView.cs
PhilBastian: current behaviour on a 50 page size
- first page: SGa 1-50 + SGb 1-50
- second page: SGa 50-100 + SGb 50-100
this solution
- first page: SGa 1-25 + SGb 1-25 (approx, depending on sort)
- second page: SGa 50-75 + SGb 50-75.
The user will never see 25-50 from either of the ScatterGather endpoints

src/ServiceControl.Audit.Persistence/Infrastructure/DateTimeRange.cs
johnsimons: But then they don't know they made a mistake, and they would get unexpected results.
This is a tricky topic that, unfortunately, our REST API does not handle well.
So, I am cautious about introducing a new pattern that is not well thought out.

src/ServiceControl/CompositeViews/Messages/ScatterGatherApiMessageView.cs
johnsimons: That is not the way it works.
We still return 50 results from all the instances, we then merge them all together and sort them and return the top 50.

johnsimons:
@ramonsmits can you unblock this PR, so we can merge it?

src/ServiceControl.Audit/Auditing/MessagesView/GetMessages2Controller.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Infrastructure;
+using Infrastructure.WebApi;
+using Microsoft.AspNetCore.Mvc;
+using Persistence;
+
+[ApiController]
+[Route("api")]
+public class GetMessages2Controller(IAuditDataStore dataStore) : ControllerBase
```
ramonsmits, line 13: Wouldn't it be possible to use `ApiVersionAttribute` by using the Microsoft.AspNetCore.Mvc.Versioning package?

This is just pseudo typing but it would then become something like:

```suggestion
[ApiController]
[Route("api")]
[ApiVersion("2.0")]
public class GetMessages2Controller(IAuditDataStore dataStore) : ControllerBase
```

This also allows for a more clean usage of the deprecated API later on.

johnsimons: Not in this PR

johnsimons: @ramonsmits I am not too concerned with api versioning given that we control the clients and this is not a publicly documented api so we can evolve it anyway we want as long our clients do not break. 
So if you feel strongly about api versioning we can have that discussion separately from this work. 
But otherwise can you unblock the PR so we can proceed with the port of SI to SP.

PR #4930, opened 2025-04-07
Handle the RegexMatchTimeoutException

soujay:
This PR handles the RegexMatchTimeoutException by returning the GetNonStandardClassification

### DIFF
--- a/src/ServiceControl/Recoverability/Grouping/Groupers/ExceptionTypeAndStackTraceFailureClassifier.cs
+++ b/src/ServiceControl/Recoverability/Grouping/Groupers/ExceptionTypeAndStackTraceFailureClassifier.cs
@@ -1,6 +1,7 @@
 namespace ServiceControl.Recoverability
 {
     using System.Linq;
+    using System.Text.RegularExpressions;
 
     public class ExceptionTypeAndStackTraceFailureClassifier : IFailureClassifier
     {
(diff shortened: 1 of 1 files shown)


src/ServiceControl/Recoverability/Grouping/Groupers/ExceptionTypeAndStackTraceFailureClassifier.cs
```diff
@@ -28,21 +29,28 @@ public string ClassifyFailure(ClassifiableMessageDetails failure)
...
-            if (firstStackTraceFrame != null)
+                if (firstStackTraceFrame != null)
+                {
+                    return exception.ExceptionType + ": " + firstStackTraceFrame.ToMethodIdentifier();
+                }
+            }
+            catch (RegexMatchTimeoutException)
             {
-                return exception.ExceptionType + ": " + firstStackTraceFrame.ToMethodIdentifier();
+                return GetNonStandardClassification(exception.ExceptionType);
```
ngallegos, line 50: Do we need this if allowing it to fall through with a comment would execute the same line?

```suggestion
                // Intentionally empty - if the parsing fails, treat it as if exception.Message is null or empty, not throw an exception
```

DavidBoike (approved):

bording (approved):

PR #4938, opened 2025-04-09
Support Azure Service Bus Geo-Recovery alias for querying queue names

DavidBoike:
Azure Service Bus namespaces using the Geo-Recovery feature define an alias that points to a premium namespace, but can perform a one-time disaster failover to a secondary premium namespace. Most data access use cases used by ServiceControl handle this naturally via DNS, except for enumerating queues as part of throughput monitoring.

This PR introduces a DNS query on the provided Service Bus domain name so that if a Geo-Recovery alias is used, it can be mapped to the true Service Bus namespace handling the requests.

### DIFF
--- a/src/ServiceControl.Transports.ASBS/AzureQuery.cs
+++ b/src/ServiceControl.Transports.ASBS/AzureQuery.cs
@@ -30,7 +30,6 @@ public class AzureQuery(ILogger<AzureQuery> logger, TimeProvider timeProvider, T
     ArmClient? armClient;
     string? resourceId;
     ArmEnvironment armEnvironment;
-    MetricsQueryAudience metricsQueryAudience;
 
     protected override void InitializeCore(ReadOnlyDictionary<string, string> settings)
     {
@@ -103,7 +102,7 @@ public class AzureQuery(ILogger<AzureQuery> logger, TimeProvider timeProvider, T
             Diagnostics.AppendLine("Client secret set");
         }
 
-        (armEnvironment, metricsQueryAudience) = GetEnvironment();
+        (armEnvironment, var metricsQueryAudience) = GetEnvironment();
 
         if (managementUrl == null)
         {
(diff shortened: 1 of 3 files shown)


ngallegos (approved):

src/ServiceControl.Transports.ASBS/AzureQuery.cs
```diff
@@ -286,6 +291,37 @@ public override async IAsyncEnumerable<IBrokerQueue> GetQueueNames(
         throw new Exception($"Could not find a ServiceBus named \"{serviceBusName}\"");
     }
 
+    async Task<HashSet<string>> GetValidNamespaceNames(CancellationToken cancellationToken = default)
+    {
+        var validNamespaces = new HashSet<string>(StringComparer.OrdinalIgnoreCase) { serviceBusName };
+
+        // ArmEnvironment Audience Values: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/resourcemanager/Azure.ResourceManager/src/ArmEnvironment.cs
+        // Service Bus Domains: https://learn.microsoft.com/en-us/rest/api/servicebus/
+        var serviceBusCloudDomain = armEnvironment.Audience switch
```
bording, line 300: Instead of a switch, we could use something like a dictionary to use the `ArmEnvironment` as the lookup key. That would avoid needing to hard code the management URLs as well as the domain sufffixes.

src/ServiceControl.Transports.ASBS/AzureQuery.cs
```diff
@@ -286,6 +291,37 @@ public override async IAsyncEnumerable<IBrokerQueue> GetQueueNames(
         throw new Exception($"Could not find a ServiceBus named \"{serviceBusName}\"");
     }
 
+    async Task<HashSet<string>> GetValidNamespaceNames(CancellationToken cancellationToken = default)
```
bording, line 294: We do we need the `HashSet` at all if we're only ever going to return a single value?

bording (approved):

PR #4943, opened 2025-04-22
Refine test to show minimal metadata needed for errors to be imported successfully

andreasohlund:
I did some testing to figure out which minimal set of metadata is needed for error imports to work.

Notes:

- Requiring the `ProcessingMachine` feels off, I think we should default it to null or unknown to be more flexible when importing
- MessageId header is needed for some reason, even though we have the native message id from the transport

I will raise follow-up PRs to address the above


src/ServiceControl.AcceptanceTests/Recoverability/MessageFailures/When_processing_message_with_missing_metadata_failed.cs
```diff
@@ -89,30 +98,23 @@ protected override TransportOperations CreateMessage(MyContext context)
...
 
-                    return new TransportOperations(
-                        new TransportOperation(outgoingMessage, new UnicastAddressTag("error"))
-                    );
+                    return new TransportOperations(new TransportOperation(outgoingMessage, new UnicastAddressTag("error")));
                 }
             }
         }
 
-        public class MyContext : ScenarioContext
```
ramonsmits, line 115: I see several changes in this diff which are unrelated to the actual change. IMHO it woul be better to do these in a separate PR.

src/ServiceControl.AcceptanceTests/Recoverability/MessageFailures/When_processing_message_with_missing_metadata_failed.cs
```diff
@@ -89,30 +98,23 @@ protected override TransportOperations CreateMessage(MyContext context)
                     var headers = new Dictionary<string, string>
                     {
                         [Headers.MessageId] = context.MessageId,
-                        [Headers.ProcessingEndpoint] = context.EndpointNameOfReceivingEndpoint,
```
ramonsmits, line 92: Not having any error info should IMHO not be allowed. It was DLQ'd for a reason. It could be an processing error (exception) or another reason in the infrastructure..

THe SP UI is pretty useless for failures without errors as the grouping options would not work.

But for this test to pass it is indeed not required.

src/ServiceControl/Operations/ErrorIngestor.cs
```diff
@@ -37,13 +37,7 @@ public ErrorIngestor(Metrics metrics,
...
             var ingestedMeter = metrics.GetCounter("Error ingestion - ingested");
 
-            var enrichers = new IEnrichImportedErrorMessages[]
-            {
-                new MessageTypeEnricher(),
-                new EnrichWithTrackingIds(),
-                new ProcessingStatisticsEnricher()
-
-            }.Concat(errorEnrichers).ToArray();
+            var enrichers = new IEnrichImportedErrorMessages[] { new MessageTypeEnricher(), new EnrichWithTrackingIds(), new ProcessingStatisticsEnricher() }.Concat(errorEnrichers).ToArray();
```
ramonsmits, line 40: Is this really better? :-)

ramonsmits:
> * Requiring the `ProcessingMachine` feels off, I think we should default it to null or unknown to be more flexible when importing

Indeed, that should be optional but requires SP/SC changes.

> * MessageId header is needed for some reason, even though we have the native message id from the transport

The header might be auto generated based on the transport message ID in some transports. 

Where are you referring to? It isn't obvious in the diff from this PR.

src/ServiceControl.AcceptanceTests/Recoverability/MessageFailures/When_processing_message_with_missing_metadata_failed.cs
andreasohlund: Agreed, at least a message should be provided. Since the UI becomes useless without it, perhaps we should add it here?

src/ServiceControl.AcceptanceTests/Recoverability/MessageFailures/When_processing_message_with_missing_metadata_failed.cs
andreasohlund: I've rolled back the other cleanup, fixes to this file feels close enough to be included

src/ServiceControl.AcceptanceTests/Recoverability/MessageFailures/When_processing_message_with_missing_metadata_failed.cs
andreasohlund: When ingesting the DLQ messages from ASB the only 2 things available is DLQ Reason and DLQ Description, I chose to map those like this:

NServiceBus.ExceptionInfo.ExceptionType: Dead letter reason
NServiceBus.ExceptionInfo.Message: Dead letter description


This leads to a bit of a funky UI experience since SP assumes a stack trace to be available and prints a `0` instead when displaying the grouping 
image

When showing the actual message it looks better

image

andreasohlund:
@ramonsmits I've created separate tests for what is needed for ingestion to work vs what is needed for the UI to render reasonably. See if you like it

src/ServiceControl.AcceptanceTests/Recoverability/MessageFailures/When_ingesting_failed_message_with_missing_headers.cs
```diff
@@ -0,0 +1,131 @@
...
+    {
+        var context = await Define<TestContext>(c =>
+            {
+                c.AddMinimalRequiredHeaders();
+
+                // This is needed for ServiceControl to be able to detect both endpoint (via failed q header) and host via the processing machine header
+                // Missing endpoint or host will cause a null ref in ServicePulse
+                c.Headers[Headers.ProcessingMachine] = "MyMachine";
+
+                c.Headers["NServiceBus.ExceptionInfo.ExceptionType"] = "SomeExceptionType";
```
ramonsmits, line 51: Use `https://github.com/Particular/NServiceBus/blob/9.2.7/src/NServiceBus.Core/Recoverability/Faults/FaultsHeaderKeys.cs`

src/ServiceControl.AcceptanceTests/Recoverability/MessageFailures/When_ingesting_failed_message_with_missing_headers.cs
```diff
@@ -0,0 +1,131 @@
...
+
+    [Test]
+    public async Task TimeSent_should_not_be_casted()
+    {
+        var sentTime = DateTime.Parse("2014-11-11T02:26:58.000462Z");
+
+        var context = await Define<TestContext>(c =>
+            {
+                c.AddMinimalRequiredHeaders();
+                c.Headers.Add("NServiceBus.TimeSent", DateTimeOffsetHelper.ToWireFormattedString(sentTime));
```
ramonsmits, line 80: Use `Headers.TimeSent`

src/ServiceControl.AcceptanceTests/Recoverability/MessageFailures/When_ingesting_failed_message_with_missing_headers.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        public string UniqueMessageId => DeterministicGuid.MakeId(MessageId, EndpointNameOfReceivingEndpoint).ToString();
+
+        public Dictionary<string, string> Headers { get; } = [];
+
+        public FailedMessageView Failure { get; set; }
+
+        public void AddMinimalRequiredHeaders()
+        {
+            Headers["NServiceBus.FailedQ"] = EndpointNameOfReceivingEndpoint;
```
ramonsmits, line 112: Use `FaultsHeaderKeys.FailedQ`

ramonsmits (approved):
LGTM except for the string values

andreasohlund:
Good catch, replaced with constants now

PR #4945, opened 2025-04-24
Only require endpoint name to be available when detecting receiving endpoint details

andreasohlund:
This lowers the bar further when it comes to importing failures into service control.


andreasohlund:
Talking to @johnsimons I realized that even though the change is trivial more testing should be done and changes to the API considered. 

Raised https://github.com/Particular/ServiceControl/issues/4963 to track this

PR #4992, opened 2025-05-23
Document the reasons why servicecontrol.throughput queue exists

johnsimons:


docs/throughput-collection.md
```diff
@@ -0,0 +1,27 @@
...
+Broker collection is also collected by querying the borker directly every 24 hours.
+The monitoring metrics collection is collected by the metrics intance sending a direct message to a well known queue that by default is named "servicecontrol.throughput".
+
+
+### Why is the "servicecontrol.throughput" queue hardcoded?
+The TF that initially created this new feature decided that by hardcoding the queue name, it would be simpler for the customers to get started because they don't need to configure anything up front.
+This was decided based on simplicity of upgrade vs maintanability.
+
+### Why is the "servicecontrol.throughput" not a sub queue of the error instance?
+Because this would mean customer would have to configure the sending endpoint (monitoring instance) with the name of this queue, which we wanted to avoid.
```
jpalac, line 20: I wouldn't say "hardcoded" as that implies it can't be changed, whereas this value can be changed via the config file, so how about combining these two points into:

```suggestion
### Why is the "servicecontrol.throughput" queue not a sub queue of the Error instance?

The usage collection queue needs to be known to the Monitoring instance.

At the time of creating this feature it was decided that having a queue name that is not dependent on the name of the Error instance means less setup for majority of customers since the feature would "just work" out of the box. 

If the queue name was based on the Error instance name (i.e. "ErrorInstanceQueueName.throughput") then **every ** customer would have to make updates to their Monitoring instance config to set the correct queue name.

The decision favoured simplicity of upgrade over existing ServiceControl queue name conventions, keeping in line with the tech lead preferences for [software that "just works"](https://github.com/Particular/Strategy/blob/master/tech-lead-preferences/it-just-works.md#it-just-works) and [convenience](https://github.com/Particular/Strategy/blob/master/tech-lead-preferences/usability.md#convenience).

```

docs/throughput-collection.md
```diff
@@ -0,0 +1,27 @@
...
+
+#### But we have SCMU, can't that make sure the queue names match?
+SCMU is a Windows only tool, as well as, this would only work if they are installing all instances on the same machine.
+Same reasons why, we do not configure the remote audit instances.
+
+### Can the "servicecontrol.throughput" queue be renamed?
+Yes, see https://docs.particular.net/servicecontrol/monitoring-instances/configuration#usage-reporting
```
jpalac, line 27:
```suggestion
#### Why isn't SCMU used to ensure the names match?

SCMU is a Windows only tool, plus the install of the Monitoring instance is separate to that of the Error instance. 
Additionally, the Monitoring instance can be installed on a different server to the Error instance.
For similar reasons we do not configure the remote audit instances in SCMU.

### Can the "servicecontrol.throughput" queue be renamed?

Yes, the queue name can be changed via:

- the [LicensingComponent/ServiceControlThroughputDataQueue](https://docs.particular.net/servicecontrol/servicecontrol-instances/configuration#usage-reporting-when-using-servicecontrol-licensingcomponentservicecontrolthroughputdataqueue) setting on the Error instance
- the [Monitoring/ServiceControlThroughputDataQueue](https://docs.particular.net/servicecontrol/monitoring-instances/configuration#usage-reporting-monitoringservicecontrolthroughputdataqueue) on the Monitoring instance

These two settings must match for the usage reporting from Monitoring to work correctly.
```

jpalac (commented):
I have made some suggestions

jpalac (approved):

PR #5014, opened 2025-06-12
Fix race condition in RepeatedFailuresOverTimeCircuitBreaker

WilliamBZA:
A port of https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/1066

> Serializing the arm/disarm actions is not enough. The state transition preceding the arm/disarm action must be captured in the lock as well, otherwise the state transitions could happen properly but then the arm/disarm actions could be executed by the processor in the wrong order, leaving the logic of the circuit breaker itself intact but the consequences of the arm/disarm actions in an incorrect state.
> 
> I also added xdoc as a way to document that the arm/disarm actions are called within a lock, and to maybe pave the way for this to be used as a shared component.


src/ServiceControl.Infrastructure/RepeatedFailuresOverTimeCircuitBreaker.cs
```diff
@@ -0,0 +1,206 @@
+namespace NServiceBus;
```
danielmarbach, line 1: FYI You could add #nullable enable to that file only

afprtclr (approved):

hsvotwa (approved):

PR #5019, opened 2025-06-16
Convert SC primary instance loggers to ILogger

abparticular:


src/ServiceControl/CompositeViews/Messages/GetMessagesController.cs
```diff
@@ -161,6 +164,6 @@ public async Task<IList<MessagesView>> SearchByKeyword([FromQuery] PagingInfo pa
             return result.Results;
         }
 
-        static ILog logger = LogManager.GetLogger(typeof(GetMessagesController));
+        readonly ILogger<GetMessagesController> logger = logger;
```
PhilBastian, line 166: no need to have a readonly logger declared when using the primary constructor, unless it's a class made available to a third party where we're worried about it being overwritten externally

src/ServiceControl/CompositeViews/Messages/ScatterGatherApi.cs
```diff
@@ -120,19 +120,21 @@ async Task<QueryResult<TOut>> FetchAndParse(HttpClient httpClient, string pathAn
             catch (HttpRequestException httpRequestException)
             {
                 remoteInstanceSetting.TemporarilyUnavailable = true;
-                logger.Warn(
-                    $"An HttpRequestException occurred when querying remote instance at {remoteInstanceSetting.BaseAddress}. The instance at uri: {remoteInstanceSetting.BaseAddress} will be temporarily disabled.",
-                    httpRequestException);
+                logger.LogWarning(
+                    httpRequestException,
+                    "An HttpRequestException occurred when querying remote instance at {remoteInstanceSettingBaseAddress}. The instance at uri: {remoteInstanceSettingBaseAddress} will be temporarily disabled",
```
PhilBastian: it seems that this `remoteInstanceSettingBaseAddress` serves the same purpose as the `remoteBaseAddress` in GetMessagesController. I would therefore be inclined to name them the same, so that it's easier for a user to find all related messages

src/ServiceControl/Infrastructure/Metrics/MetricsReporterHostedService.cs
```diff
@@ -4,20 +4,24 @@
     using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.Extensions.Hosting;
-    using NServiceBus.Logging;
+    using Microsoft.Extensions.Logging;
 
     class MetricsReporterHostedService : IHostedService
     {
         readonly Metrics metrics;
+        readonly ILogger<Metrics> logger;
```
PhilBastian, line 12: interesting that the field declarations are at the top of the class here... I thought Particular "standard" (don't know if it's recorded anywhere) is to have them at the bottom

src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSender.cs
```diff
@@ -20,10 +20,8 @@ public virtual async Task HandleMessage(MessageContext message, IMessageDispatch
...
             byte[] body = null;
             var messageId = message.NativeMessageId;
-            if (Log.IsDebugEnabled)
-            {
-                Log.DebugFormat("{0}: Retrieving message body", messageId);
-            }
+
+            logger.LogDebug("{messageId}: Retrieving message body", messageId);
 
             if (outgoingHeaders.TryGetValue("ServiceControl.Retry.Attempt.MessageId", out var attemptMessageId))
```
PhilBastian, line 26: beneficial refactor here to extract "ServiceControl.Retry.Attempt.MessageId" to a variable or constant (I think there's a headers static class somewhere)

src/ServiceControl/SagaAudit/SagaUpdatedHandler.cs
```diff
@@ -28,7 +28,9 @@ public async Task Handle(SagaUpdatedMessage message, IMessageHandlerContext cont
             }
 
             var endpointName = context.MessageHeaders.TryGetValue(Headers.ReplyToAddress, out var val) ? val : "(Unknown Endpoint)";
-            log.ErrorFormat($"Received a saga audit message in the ServiceControl queue that should have been sent to the audit queue. This indicates that the endpoint '{endpointName}' using the SagaAudit plugin is misconfigured and should be changed to use the system's audit queue instead. The message has been forwarded to the audit queue, but this may not be possible in a future version of ServiceControl.");
+            logger.LogError("Received a saga audit message in the ServiceControl queue that should have been sent to the audit queue. " +
```
PhilBastian, line 31: why change to + concatenated string?

PhilBastian (approved):

src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSender.cs
abparticular: I agree that is a beneficial refactoring, but I don't think it belongs in the current scope of work.  There are many headers in this file (and others) that could be refactored that way.

PR #5106, opened 2025-08-29
[SPIKE] Using Postgress as the audit instance persister

johnsimons:
## Summary of the PostgreSQL spike

- We are storing an audit message in a single row, including the body (up to the [maximum body to store configuration](https://docs.particular.net/servicecontrol/audit-instances/configuration#performance-tuning-servicecontrol-auditmaxbodysizetostore)).
- We are using [jsonb columns](https://www.postgresql.org/docs/current/datatype-json.html) to store both headers and message_metadata.
- The [indexed columns are stored as native columns](https://github.com/Particular/ServiceControl/pull/5106/files#diff-3d0ca926a729c1c22dfea7b2f65068efa9b6ff8e7f4314c77b3f6215f0c7867fR36-R45) for performance reasons, so there is a bit of data duplication happening here, but according to my research indexing json stored in jsonb columns via a [GIN index](https://www.postgresql.org/docs/current/gin.html) is not as performance as a native column, example:
  - Native BOOLEAN: WHERE is_system_message = true → B-tree index, instant
  - JSON cast: WHERE (message_metadata->>'IsSystemMessage')::boolean = true → JSON extraction + cast + comparison
- Full-text search is on by default, and we are indexing both the body and the headers.
- The indexes are tuned to the pattern usage of ServicePulse v2, where we removed quite a few sorting options.
- `IFailedAuditStorage` is not implemented. I assume we need to create another table where we store messages that could not be imported.
- The saga snapshot is not really tuned and completed, at the moment we are storing all the "changes" in a jsonb column, but this design will not scale well for never ending sagas. So I think we would need to improve this and maybe only store a maximum of 50 changes.
- The clean-up based on configured retention is done but has not been properly tested, and if performance is an issue we would need to consider using partitioned tables based on either daily or hourly rotation.
- The know_endpoints storage is a bit more complex, because we need to update the `last_seen` column, what we found is that on a busy system, there would be too many deadlocks because of collision on the `id`. To mitigate this we ended up creating an insert only table and then have a background job that updates the "real" know_endpoints table. This design would also need to be considered for the saga snapshot if there are lots of concurrent updates to the same saga_id.

## Testing

We used a MBP with an Apple M3 Pro chip and 36GB of memory.
We used RabbitMQ and latest PostgreSQL database, all running in docker.
On a single Audit instance we were able to ingests on average 1400msg/s, this is just ingestion rate, no other queries or cleanup.
When scaled out to 4 instances, we could get the ingestion rate at 5000msg/s.

### DIFF
--- a/src/ServiceControl.Audit.Persistence.PostgreSQL/PostgreSQLPersistenceInstaller.cs
+++ b/src/ServiceControl.Audit.Persistence.PostgreSQL/PostgreSQLPersistenceInstaller.cs
@@ -188,13 +188,26 @@ class PostgreSQLPersistenceInstaller(DatabaseConfiguration databaseConfiguration
 
         // Create known_endpoints table
         await using (var cmd = new NpgsqlCommand(@"
+            CREATE TABLE IF NOT EXISTS known_endpoints_insert (
+                id BIGINT GENERATED ALWAYS AS IDENTITY PRIMARY KEY,
+                endpoint_id TEXT,
+                name TEXT,
+                host_id UUID,
+                host TEXT,
+                last_seen TIMESTAMPTZ
+            );", connection))
+        {
+            await cmd.ExecuteNonQueryAsync(cancellationToken);
+        }
+
+        // Create known_endpoints table
+        await using (var cmd = new NpgsqlCommand(@"
             CREATE TABLE IF NOT EXISTS known_endpoints (
                 id TEXT PRIMARY KEY,
                 name TEXT,
                 host_id UUID,
                 host TEXT,
-                last_seen TIMESTAMPTZ,
-                updated_at TIMESTAMPTZ NOT NULL DEFAULT now()
+                last_seen TIMESTAMPTZ
             );", connection))
         {
             await cmd.ExecuteNonQueryAsync(cancellationToken);
@@ -212,11 +225,6 @@ class PostgreSQLPersistenceInstaller(DatabaseConfiguration databaseConfiguration
             DROP TRIGGER IF EXISTS saga_snapshots_updated_at_trigger ON saga_snapshots;
             CREATE TRIGGER saga_snapshots_updated_at_trigger
             BEFORE UPDATE ON saga_snapshots
-            FOR EACH ROW EXECUTE FUNCTION update_updated_at_column();
-
-            DROP TRIGGER IF EXISTS known_endpoints_updated_at_trigger ON known_endpoints;
-            CREATE TRIGGER known_endpoints_updated_at_trigger
-            BEFORE UPDATE ON known_endpoints
             FOR EACH ROW EXECUTE FUNCTION update_updated_at_column();", connection))
         {
             await cmd.ExecuteNonQueryAsync(cancellationToken);
--- a/src/ServiceControl.Audit.Persistence.PostgreSQL/UnitOfWork/PostgreSQLAuditIngestionUnitOfWork.cs
+++ b/src/ServiceControl.Audit.Persistence.PostgreSQL/UnitOfWork/PostgreSQLAuditIngestionUnitOfWork.cs
@@ -129,15 +129,14 @@ class PostgreSQLAuditIngestionUnitOfWork : IAuditIngestionUnitOfWork
         // Insert KnownEndpoint into known_endpoints table
         var cmd = batch.CreateBatchCommand();
         cmd.CommandText = @"
-                INSERT INTO known_endpoints (
-                    id, name, host_id, host, last_seen
+                
+                INSERT INTO known_endpoints_insert (
+                    endpoint_id, name, host_id, host, last_seen
                 ) VALUES (
-                    @id, @name, @host_id, @host, @last_seen
-                )
-                ON CONFLICT (id) DO UPDATE SET
-                    last_seen = GREATEST(known_endpoints.last_seen, EXCLUDED.last_seen);";
+                    @endpoint_id, @name, @host_id, @host, @last_seen
+                );";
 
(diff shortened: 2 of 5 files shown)


src/ServiceControl.Audit.Persistence.PostgreSQL/BodyStorage/PostgreSQLAttachmentsBodyStorage.cs
johnsimons, line 1: This is not really required to be implemented, given that we are storing the message body with the message row.

src/ServiceControl.Audit.Persistence.PostgreSQL/PostgreSQLAuditDataStore.cs
```diff
@@ -0,0 +1,265 @@
...
+        cmd.Parameters.AddWithValue("message_id", messageId);
+        await using var reader = await cmd.ExecuteReaderAsync(System.Data.CommandBehavior.SequentialAccess, cancellationToken);
+        if (await reader.ReadAsync(cancellationToken))
+        {
+            var contentType = reader.GetFieldValue<Dictionary<string, string>>(reader.GetOrdinal("headers")).GetValueOrDefault(Headers.ContentType, "text/xml");
+            using var stream = await reader.GetStreamAsync(reader.GetOrdinal("body"), cancellationToken);
+            var responseStream = manager.GetStream();
+            await stream.CopyToAsync(responseStream, cancellationToken);
+            responseStream.Position = 0;
+            return MessageBodyView.FromStream(responseStream, contentType, (int)stream.Length, string.Empty);
```
johnsimons, line 41: I couldn't figure out a way to pass the stream down the pipeline without allocating.
It's possible that our abstraction is preventing it, or perhaps it's me.

@SzymonPobiega @danielmarbach any thoughts?

src/ServiceControl.Audit.Persistence.PostgreSQL/UnitOfWork/PostgreSQLAuditIngestionUnitOfWork.cs
```diff
@@ -0,0 +1,143 @@
...
+        // Insert KnownEndpoint into known_endpoints table
+        var cmd = batch.CreateBatchCommand();
+        cmd.CommandText = @"
+                INSERT INTO known_endpoints (
+                    id, name, host_id, host, last_seen
+                ) VALUES (
+                    @id, @name, @host_id, @host, @last_seen
+                )
+                ON CONFLICT (id) DO UPDATE SET
+                    last_seen = GREATEST(known_endpoints.last_seen, EXCLUDED.last_seen);";
```
johnsimons, line 132: After running a test scaling out the audit instance, this query caused too many deadlocks.
This would have to be redesigned to prevent that.

src/ServiceControl.Audit.Persistence.PostgreSQL/UnitOfWork/PostgreSQLAuditIngestionUnitOfWork.cs
```diff
@@ -0,0 +1,143 @@
...
+            sagaSnapshot.Endpoint
+        };
+
+        // Insert or update saga_snapshots table - add new change to the changes array
+        var cmd = batch.CreateBatchCommand();
+        cmd.CommandText = @"
+                INSERT INTO saga_snapshots (id, saga_id, saga_type, changes)
+                VALUES (@saga_id, @saga_id, @saga_type, @new_change)
+                ON CONFLICT (id) DO UPDATE SET
+                    changes = COALESCE(saga_snapshots.changes, '[]'::jsonb) || @new_change::jsonb;";
```
johnsimons, line 110: Probably same as https://github.com/Particular/ServiceControl/pull/5106/files#r2324099591

src/ServiceControl.Audit.Persistence.PostgreSQL/PostgreSQLPersistenceInstaller.cs
```diff
@@ -0,0 +1,238 @@
...
+
+        // Create saga_snapshots table
+        await using (var cmd = new NpgsqlCommand(@"
+            CREATE TABLE IF NOT EXISTS saga_snapshots (
+                id UUID PRIMARY KEY,
+                saga_id UUID,
+                saga_type TEXT,
+                changes JSONB,
+                updated_at TIMESTAMPTZ NOT NULL DEFAULT now()
+            );", connection))
```
johnsimons, line 175: The design of this table is incorrect, needs to be updated to the same design as the know_endpoints where we use an insert only table for the hot path and then we use a background job to reconcile the results in the more permanent table.

src/ServiceControl.Audit.Persistence.PostgreSQL/PostgreSQLPersistenceInstaller.cs
```diff
@@ -0,0 +1,238 @@
...
+            await cmd.ExecuteNonQueryAsync(cancellationToken);
+        }
+
+        // Create saga_snapshots table
+        await using (var cmd = new NpgsqlCommand(@"
+            CREATE TABLE IF NOT EXISTS saga_snapshots (
+                id UUID PRIMARY KEY,
+                saga_id UUID,
+                saga_type TEXT,
+                changes JSONB,
```
johnsimons, line 173: Also, this design is based on the Raven design where we append changes to the changes column.
But for never ending sagas this could get quite large and in essence not work.
I think we need to keep rotating changes where we keep a maximum of x changes that is it. Maybe a bit smarter so that we always keep the first change, ...

PR #5112, opened 2025-09-08
Fix `NullReferenceException` when connected to RavenDB cloud

ramonsmits:
Fix `NullReferenceException` when connected to RavenDB cloud which does not return memory information.


ramonsmits:
> Other than, maybe, more readable, is `is null` any better?

It is, `null ==` c habbit still haunts me after 30+ years ;-)

irinascurtu (approved):

PR #5139, opened 2025-10-16
Log email notification subject and body for all SendEmailNotificationHandler executions

soujay:
This PR  has the following changes,
- enhances the logging within SendEmailNotificationHandler by adding a single LogInformation statement at the start of the Handle method to log the subject and body of each email notification being processed. 
- removes a duplicate custom check

### DIFF
--- a/src/ServiceControl/Notifications/Email/SendEmailNotificationHandler.cs
+++ b/src/ServiceControl/Notifications/Email/SendEmailNotificationHandler.cs
@@ -36,7 +36,7 @@
 
             if (notifications == null || !notifications.Email.Enabled)
             {
-                logger.LogInformation("Skipping email sending. Notifications turned-off");
+                logger.LogDebug("Skipping email sending. Notifications turned-off");
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


tamararivera (approved):

src/ServiceControl/Notifications/Email/SendEmailNotificationHandler.cs
```diff
@@ -32,6 +32,8 @@ public async Task Handle(SendEmailNotification message, IMessageHandlerContext c
                 notifications = await manager.LoadSettings(cacheTimeout);
             }
 
+            logger.LogInformation("Processing email notification. Subject: {Subject}, Body: {Body}", message.Subject, message.Body);
+
             if (notifications == null || !notifications.Email.Enabled)
             {
                 logger.LogInformation("Skipping email sending. Notifications turned-off");
```
bording: I think we should remove this log entry.

src/ServiceControl/Notifications/Email/SendEmailNotificationHandler.cs
bording: While we're here, fix the formatting on this and put it back on a single line!

src/ServiceControl/Notifications/Email/SendEmailNotificationHandler.cs
```diff
@@ -32,6 +32,8 @@ public async Task Handle(SendEmailNotification message, IMessageHandlerContext c
                 notifications = await manager.LoadSettings(cacheTimeout);
             }
 
+            logger.LogInformation("Processing email notification. Subject: {Subject}, Body: {Body}", message.Subject, message.Body);
```
bording: Given that we know these custom checks also get surfaced up to ServicePulse, I'm wondering if we want this to get info level at all.

Maybe it would be better to be a Debug level entry?

PR #5149, opened 2025-10-23
Introduce an integration event for the Edit and Retry feature

cquirosj:
Introduce a ServiceControl integration event (`MessageEditedAndRetried`) that is dispatched every time an Edit and Retry takes place. With this event the user can programmatically track the status of the failed message (the fact it has been edited and retried).

### DIFF
--- a/src/Directory.Packages.props
+++ b/src/Directory.Packages.props
@@ -68,7 +68,7 @@
     <PackageVersion Include="RavenDB.Embedded" Version="6.2.6" />
     <PackageVersion Include="ReactiveUI.WPF" Version="20.1.63" />
     <PackageVersion Include="Seq.Extensions.Logging" Version="8.0.0" />
-    <PackageVersion Include="ServiceControl.Contracts" Version="5.0.0" />
+    <PackageVersion Include="ServiceControl.Contracts" Version="5.1.0" />
     <PackageVersion Include="System.Configuration.ConfigurationManager" Version="8.0.1" />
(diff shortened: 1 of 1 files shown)


src/ServiceControl/Operations/RetryConfirmationProcessor.cs
```diff
@@ -26,11 +27,11 @@ public async Task Process(List<MessageContext> contexts, IIngestionUnitOfWork un
             }
         }
 
-        public Task Announce(MessageContext messageContext)
+        public Task Announce(string failedMessageId)
```
bartekwasielak: This change creates a difference between this `Announce` and `ErrorProcessor.Announce`. I know that the latter uses much more information - would it be worth to keep these both consistent?

src/ServiceControl.AcceptanceTests/Recoverability/ExternalIntegration/When_a_failed_message_is_resolved_by_edit_and_retry.cs
```diff
@@ -0,0 +1,96 @@
+namespace ServiceControl.AcceptanceTests.Recoverability.ExternalIntegration
```
bartekwasielak: I feel like this should go to `ServiceControl.AcceptanceTests.Recoverability` (where other edit and retries tests are)

bartekwasielak (commented):
I appreciate the changes of the methods' names. This helps with understanding a difference between the business message and the edit request.

bartekwasielak (approved):

PR #5211, opened 2025-12-04
Programmatically register RavenDB license after the server starts instead of using an environment variable (which might be ignored)

mauroservienti:


mauroservienti:
Withdrawn in favor of https://github.com/Particular/ServiceControl/pull/5235

PR #5212, opened 2025-12-04
Update ServiceControl to use NServiceBus 10

danielmarbach:
Formatting errors started showing up

Plus X509 ctor is obsolete see https://learn.microsoft.com/en-us/dotnet/fundamentals/syslib-diagnostics/syslib0057

According to RavenDB documentation, it should be PKCS#12 for the client certificates 

https://docs.ravendb.net/7.1/server/security/authentication/client-certificate-usage/

Switched to aspnet:10.0-noble

### DIFF
--- a/src/ServiceControl.Transports.SQS/ServiceControl.Transports.SQS.csproj
+++ b/src/ServiceControl.Transports.SQS/ServiceControl.Transports.SQS.csproj
@@ -12,7 +12,6 @@
 
   <ItemGroup>
     <PackageReference Include="AWSSDK.CloudWatch" />
-    <!-- Required for IAM Roles for Service Accounts even though no API is added -->
     <PackageReference Include="NServiceBus.AmazonSQS" />
   </ItemGroup>
 


.github/workflows/build-containers.yml
```diff
@@ -75,7 +75,7 @@ jobs:
             org.opencontainers.image.created=${{ steps.date.outputs.date }}
             org.opencontainers.image.title=${{ matrix.title }}
             org.opencontainers.image.description=${{ matrix.description }}
-            org.opencontainers.image.base.name=mcr.microsoft.com/dotnet/aspnet:8.0-jammy-chiseled-composite-extra
+            org.opencontainers.image.base.name=mcr.microsoft.com/dotnet/aspnet:10.0-jammy-noble-composite-extra
```
DavidBoike: To go to noble, we'd need to validate that this error is no longer an issue when hosting in Azure.

https://github.com/Particular/ServiceControl/pull/4422

I want to say that we reported to Microsoft and they fixed it, but can't remember for sure. @bording do you remember?

danielmarbach: yeah that was me screwing up some edits to the file

PR #5227, opened 2025-12-12
[SPIKE] Host ServicePulse frontend code from ServiceControl

johnsimons:
This is another spike in the same league as https://github.com/Particular/ServiceControl/pull/5189

This spike uses the Particular.PlatformSample.ServicePulse package to extract the web resources from it and serve them as part of ServiceControl ASP.NET Core pipeline.
It also makes `app.constants.js` be served from the server, so that we can configure things.

In this spike, we have added the `embedded` flag to `app.constants.js` so that we can customise ServicePulse UI/X.

### DIFF
--- a/src/ServiceControl/Hosting/Commands/RunCommand.cs
+++ b/src/ServiceControl/Hosting/Commands/RunCommand.cs
@@ -1,17 +1,13 @@
 ﻿namespace ServiceControl.Hosting.Commands
 {
-    using System;
-    using System.IO;
-    using System.Reflection;
     using System.Threading.Tasks;
-    using Infrastructure.WebApi;
     using Microsoft.AspNetCore.Builder;
-    using Microsoft.Extensions.FileProviders;
     using NServiceBus;
     using Particular.ServiceControl;
     using Particular.ServiceControl.Hosting;
     using ServiceBus.Management.Infrastructure.Settings;
     using ServiceControl;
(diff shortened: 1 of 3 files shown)


src/ServiceControl/Hosting/Commands/RunCommand.cs
```diff
@@ -23,8 +27,18 @@ public override async Task Execute(HostArguments args, Settings settings)
             hostBuilder.AddServiceControl(settings, endpointConfiguration);
             hostBuilder.AddServiceControlApi();
 
+            var servicePulsePath = Path.Combine(AppContext.BaseDirectory, "platform", "servicepulse", "ServicePulse.dll");
+            var manifestEmbeddedFileProvider = new ManifestEmbeddedFileProvider(Assembly.LoadFrom(servicePulsePath), "wwwroot");
+            var fileProvider = new CompositeFileProvider(hostBuilder.Environment.WebRootFileProvider, manifestEmbeddedFileProvider);
+            var defaultFilesOptions = new DefaultFilesOptions { FileProvider = fileProvider };
+            var staticFileOptions = new StaticFileOptions { FileProvider = fileProvider };
+
```
mikeminutillo: It might be worth extracting all of this into a `.UseServicePulse()` extension to match how we're doing ServiceControl.

PR #5248, opened 2026-01-13
Instance log and diagnostics files not written to configured location

ramonsmits:
During smoketesting of v6.9.0, we noticed that after deployed on a Windows VM, the logs were not written to the expected location.

### Symptoms

The configured log path is not used, and logs are instead written to the installation directory of the instance

### Who's affected

Users hosting ServiceControl, Audit, or Monitoring as a Windows service

### Root cause

All instances invoke `ExeConfiguration.PopulateAppSettings` but that happens AFTER logging is initialized. This causes the internal `LogPath` setting to not be initialized correctly.

### Confirmed workarounds

None

## Resolution

Invoking `ExeConfiguration.PopulateAppSettings` before logging it initialized.

### DIFF
--- a/src/ServiceControl.Audit.UnitTests/PopulateAppSettingsTests.cs
+++ b/src/ServiceControl.Audit.UnitTests/PopulateAppSettingsTests.cs
@@ -3,6 +3,7 @@ namespace ServiceControl.UnitTests;
 using System;
 using System.Diagnostics;
 using System.IO;
+using System.Runtime.InteropServices;
 using System.Threading.Tasks;
 using NUnit.Framework;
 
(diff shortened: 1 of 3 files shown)


src/ServiceControl.Audit.UnitTests/PopulateAppSettingsTests.cs
```diff
@@ -0,0 +1,76 @@
...
+                      <configuration>
+                        <appSettings>
+                          <add key="ServiceControl.Audit/LogPath" value="{MagicValue}"   />
+                        </appSettings>
+                      </configuration>
+                      """;
+
+        await File.WriteAllTextAsync("ServiceControl.Audit.exe.config", config);
+
+#if WINDOWS
```
andreasohlund: Windows build is failing with

```
 Error: System.ComponentModel.Win32Exception : An error occurred trying to start process 'ServiceControl.Audit' with working directory 'D:\a\ServiceControl\ServiceControl\src\ServiceControl.Audit.UnitTests\bin\Release\net8.0'. The system cannot find the file specified.
```
so it seems like WINDOWS is not defined

PR #5274, opened 2026-01-22
The license status check fails when checking the license against a secured RavenDB instance 

mauroservienti:
The [recently introduced license status check](https://github.com/Particular/ServiceControl/pull/5235) doesn't account for connecting to secured RavenDB servers, such as RavenDB cloud.

This PR changes the way the license status check connects to the RavenDB server. Using the internal RavenDB HTTP Client, the call uses all the preconfigured connection settings, including any authentication configuration. Tested against RavenDB Cloud with certificate authentication.

### DIFF
--- a/src/ServiceControl.Persistence.RavenDB/LicenseStatusCheck.cs
+++ b/src/ServiceControl.Persistence.RavenDB/LicenseStatusCheck.cs
@@ -15,25 +15,31 @@ static class LicenseStatusCheck
         var ravenConfiguredHttpClient = documentStore.GetRequestExecutor().HttpClient;
         var licenseCheckUrl = documentStore.Urls[0].TrimEnd('/') + "/license/status";
 
-        // Not linking to the incoming cancellationToken to ensure no OperationCancelledException prevents the last InvalidOperationException to be thrown
-        using var cts = new CancellationTokenSource(30_000);
-        while (!cts.IsCancellationRequested)
+        using var cts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+        cts.CancelAfter(30_000);
+
+        try
         {
-            var httpResponse = await ravenConfiguredHttpClient.GetAsync(licenseCheckUrl, cancellationToken);
-            var licenseStatus = await httpResponse.Content.ReadFromJsonAsync<LicenseStatusFragment>(cancellationToken);
-            if (licenseStatus.Expired)
+            while (!cts.IsCancellationRequested)
             {
-                throw new InvalidOperationException("The current RavenDB license is expired. Please, contact support");
-            }
+                var httpResponse = await ravenConfiguredHttpClient.GetAsync(licenseCheckUrl, cancellationToken);
+                var licenseStatus = await httpResponse.Content.ReadFromJsonAsync<LicenseStatusFragment>(cancellationToken);
+                if (licenseStatus.Expired)
+                {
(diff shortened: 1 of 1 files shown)


src/ServiceControl.Audit.Persistence.RavenDB/LicenseStatusCheck.cs
```diff
@@ -1,27 +1,25 @@
...
 
-    public static async Task WaitForLicenseOrThrow(DatabaseConfiguration configuration, CancellationToken cancellationToken)
+    public static async Task WaitForLicenseOrThrow(IDocumentStore documentStore, DatabaseConfiguration configuration, CancellationToken cancellationToken)
     {
-        using var client = new HttpClient
-        {
-            BaseAddress = new Uri(configuration.ServerConfiguration.ConnectionString ?? configuration.ServerConfiguration.ServerUrl)
-        };
+        var ravenConfiguredHttpClient = documentStore.GetRequestExecutor().HttpClient;
+        var licenseCheckUrl = (configuration.ServerConfiguration.ConnectionString ?? configuration.ServerConfiguration.ServerUrl).TrimEnd('/') + "/license/status";
```
johnsimons: You could use `documentStore.Urls[0]`

mauroservienti: Is that set for the embedded instance as well? I'll double check

andreasohlund (approved):

src/ServiceControl.Audit.Persistence.RavenDB/LicenseStatusCheck.cs
```diff
@@ -1,27 +1,25 @@
...
+    public static async Task WaitForLicenseOrThrow(IDocumentStore documentStore, CancellationToken cancellationToken)
     {
-        using var client = new HttpClient
-        {
-            BaseAddress = new Uri(configuration.ServerConfiguration.ConnectionString ?? configuration.ServerConfiguration.ServerUrl)
-        };
+        var ravenConfiguredHttpClient = documentStore.GetRequestExecutor().HttpClient;
+        var licenseCheckUrl = documentStore.Urls[0].TrimEnd('/') + "/license/status";
 
         // Not linking to the incoming cancellationToken to ensure no OperationCancelledException prevents the last InvalidOperationException to be thrown
```
johnsimons: @mauroservienti @ramonsmits I don't really understand this comment, given that the `cancellationToken` is passed to the HTTP calls, and those will throw an `OperationCancelledException`.
So are you trying to prevent those exceptions?

ramonsmits: @johnsimons there is the CTS for a timeout on this operation, and there is the passsed CT.

When this CTS passes, we want to throw an informative exception. If the passed CT is  cancelled, we just want to have that be the reason for cancellation. 

If we link to the CTS we need to have a more complex catch to determine what is the source of cancellation, the passed CT, or the timeout/cancelled CTS.

To me that seems obvious but as I think @andreasohlund had a similar question maybe it is just better to have a solution with more code :-).

ramonsmits: @mauroservienti took the liberty to modify it in 9425a4e71e1b4a46031c459167c56036c55f7793

Didn't do a local test and wrote it fully in github editor (living on the edge..), might need a little fix

PR #5294, opened 2026-02-06
Add renovate config to restrict package version (#5291)

jasontaylordev:


warwickschroeder (approved):

warwickschroeder (changes requested):
Should not merge from master

PR #5353, opened 2026-03-10
[DO NOT MERGE] Allow disabling RabbitMQ broker requirement checks

dvdstelt:
## Motivation

In multi-tenant RabbitMQ deployments, granting management API access to individual tenants is a security risk. ServiceControl currently performs broker requirement checks that depend on management API access, with no way to disable them.

The RabbitMQ transport (v11.0.0) already supports `DisabledBrokerRequirementChecks`, but ServiceControl did not expose it.

### What this does

Adds DisableBrokerRequirementChecks=true as a connection string parameter:

`host=rabbitmq;username=tenant1;password=secret;DisableBrokerRequirementChecks=true`

When enabled, this:
- Disables broker version and stream feature checks at startup
- Disables delivery limit validation (also requires management API)
- Skips registration of `RabbitMQQuery` (broker throughput collection) and `QueueLengthProvider` (queue length monitoring), both of which require management API access
- Registers a `NoOpQueueLengthProvider` to satisfy DI requirements

### What still works without management API access

- Sending and receiving messages
- Usage reports and license validation (via audit and monitoring data sources)

### What becomes unavailable

- Queue length monitoring in ServicePulse
- Broker-sourced throughput data in usage reports

### DIFF
--- a/src/ServiceControl.Transports.RabbitMQ/NoOpQueueLengthProvider.cs
+++ b/src/ServiceControl.Transports.RabbitMQ/NoOpQueueLengthProvider.cs
@@ -2,14 +2,20 @@ namespace ServiceControl.Transports.RabbitMQ
 {
     using System.Threading;
     using System.Threading.Tasks;
+    using Microsoft.Extensions.Logging;
 
-    class NoOpQueueLengthProvider : IProvideQueueLength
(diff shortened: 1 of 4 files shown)


SzymonPobiega (approved):
LGTM

PR #5391, opened 2026-03-20
Add MCP endpoint for audit and failed messages

WilliamBZA:
Adds support for MCP servers within ServiceControl and ServiceControl.Audit


danielmarbach (approved):

dvdstelt (approved):

johnsimons:
@WilliamBZA, I have added a bit of telemetry so that we can make better decisions in the future, see https://github.com/Particular/ServiceControl/pull/5413

src/ServiceControl/Infrastructure/WebApi/HostApplicationBuilderExtensions.cs
```diff
@@ -9,18 +9,35 @@
...
             builder.Services.RegisterApiTypes(Assembly.GetExecutingAssembly());
 
             builder.AddServiceControlApis();
 
-            builder.Services.AddCors(options => options.AddDefaultPolicy(Cors.GetDefaultPolicy(corsSettings)));
+            if (settings.EnableMcpServer || settings.EnableMcpServerWriteMode)
+            {
+                var mcpBuilder = builder.Services
+                    .AddMcpServer()
+                    .WithHttpTransport()
```
johnsimons, line 28: @WilliamBZA, it seems that it is [recommended to explicitly enable stateless](https://csharp.sdk.modelcontextprotocol.io/concepts/stateless/stateless.html#stateless-mode-recommended) configuration for HTTP-based MCP servers.
Is this something you reviewed and decided not to enable?

src/ServiceControl/Infrastructure/WebApi/HostApplicationBuilderExtensions.cs
```diff
@@ -9,18 +9,35 @@
...
             // This registers concrete classes that implement IApi. Currently it is hard to find out to what
             // component those APIs should belong to so we leave it here for now.
             builder.Services.RegisterApiTypes(Assembly.GetExecutingAssembly());
 
             builder.AddServiceControlApis();
 
-            builder.Services.AddCors(options => options.AddDefaultPolicy(Cors.GetDefaultPolicy(corsSettings)));
+            if (settings.EnableMcpServer || settings.EnableMcpServerWriteMode)
+            {
+                var mcpBuilder = builder.Services
```
johnsimons, line 26: I was expecting to see some configuration here regarding authentication that is now supported.
Or does it just flow through? (From the reading I have done, it does not seem to)
Just in case, here is a [sample](https://github.com/modelcontextprotocol/csharp-sdk/blob/main/samples/ProtectedMcpServer/Program.cs) showing the configuration.

johnsimons (changes requested):
So that is not lost, I don't think we need the Audit instance MCP changes, given all API calls should flow via the error instance.
So the MCP server is hosted in the error instance, and the engine then uses the current flow to query data from all configured instances (using the scatter-gather approach).

src/ServiceControl/Infrastructure/WebApi/HostApplicationBuilderExtensions.cs
johnsimons: Actually, there is [this issue](https://github.com/modelcontextprotocol/csharp-sdk/issues/1494) that pretty much talks about our setup

johnsimons: I also found [this](https://github.com/modelcontextprotocol/csharp-sdk/issues/648) issue that may affect us.

PR #5416, opened 2026-04-13
RabbitMQ: Only set the AuthMechanisms when UseExternalAuthMechanism is set to true

danielmarbach:
Fixes:

- https://github.com/Particular/ServiceControl/issues/5417

### DIFF
--- a/src/ServiceControl.Transports.RabbitMQ/RabbitMQTransportExtensions.cs
+++ b/src/ServiceControl.Transports.RabbitMQ/RabbitMQTransportExtensions.cs
@@ -38,9 +38,8 @@ static class RabbitMQTransportExtensions
             transport.ValidateRemoteCertificate = !disableRemoteCertificateValidation;
         }
 
-        if (dictionary.TryGetValue("UseExternalAuthMechanism", out var useExternalAuthMechanismString))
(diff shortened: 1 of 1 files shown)


mauroservienti (approved):

src/ServiceControl.Transports.RabbitMQ/RabbitMQTransportExtensions.cs
```diff
@@ -38,9 +38,8 @@ public static void ApplySettingsFromConnectionString(this RabbitMQTransport tran
             transport.ValidateRemoteCertificate = !disableRemoteCertificateValidation;
         }
 
-        if (dictionary.TryGetValue("UseExternalAuthMechanism", out var useExternalAuthMechanismString))
+        if (dictionary.TryGetValue("UseExternalAuthMechanism", out var useExternalAuthMechanismString) && bool.TryParse(useExternalAuthMechanismString, out var useExternalAuthMechanism) && useExternalAuthMechanism)
```
ramonsmits, line 41: If parsing fails, there should be an exception as that means there was a typo unless the values in the dictionary have already been format validated.

danielmarbach: The rest of the code uses TryParse too. I'm trying to make a minimal fix here.

PR #5434, opened 2026-04-22
Support recoverability for Azure ServiceBus dead letter messages

andreasohlund:
With this change, [messages dead-lettered in AzureServiceBus](https://docs.particular.net/transports/azure-service-bus/configuration#dead-lettering) can [be auto-forwarded to the error queue](https://docs.particular.net/transports/azure-service-bus/configuration#dead-lettering-forward-dead-lettered-messages-to-the-error-queue) and managed in ServicePulse


andreasohlund:
Draft until I the new minor of ASB has been released

DavidBoike:
Looks like there's a problem in the ASB container with DetermineQueuesToCreate coming up with a duplicate for `Particular.ServiceControl`, and with When_ingesting_failed_message_with_missing_headers in both the Windows/Linux primary acceptance tests. I wonder if they're related?

andreasohlund:
I'm investigating

src/ServiceControl.Transports/TransportCustomization.cs
```diff
@@ -138,15 +138,22 @@ public virtual async Task ProvisionQueues(TransportSettings transportSettings, I
...
                     transportSettings.EndpointName,
                     new QueueAddress(transportSettings.EndpointName),
                     false,
                     false,
-                    transportSettings.ErrorQueue)};
+                    transportSettings.ErrorQueue)
+            };
+
+            var additionalQueuesToProvision = additionalQueues
+                .Distinct()
```
andreasohlund: @danielmarbach I wonder if this needs to be patched in the transport since it now requires the list of addresses not to have duplicates? (there was no tests for this so I'm on the fence) 

I think this change is still good, there should be no reason to call the transport with duplicates

andreasohlund:
Problems found and fixed

src/ServiceControl.UnitTests/Operations/EndpointDetailsParserTests.cs
```diff
@@ -0,0 +1,49 @@
...
+using NServiceBus.Faults;
+using NUnit.Framework;
+using ServiceControl.Contracts.Operations;
+using ServiceControl.Infrastructure;
+
+[TestFixture]
+public class EndpointDetailsParserTests
+{
+    [Test]
+    public void Receiving_endpoint_should_use_failed_queue_machine_when_host_is_missing()
```
danielmarbach: It is not clear to me what host is missing means in this case especially because the other test also doesn't seem to have a host.

All I see is a machine name being there or not after the at sign

src/ServiceControl.UnitTests/Recoverability/ExceptionTypeAndStackTraceFailureClassifierTest.cs
```diff
@@ -7,6 +7,8 @@
     [TestFixture]
     public class ExceptionTypeAndStackTraceFailureClassifierTest
     {
+        const string noStackTraceClassification = "exceptionType: No stacktrace available";
```
danielmarbach:
```suggestion
        const string noStackTraceClassification = "exceptionType: No stacktrace";
```

src/ServiceControl/Recoverability/Grouping/Groupers/ExceptionTypeAndStackTraceFailureClassifier.cs
```diff
@@ -54,10 +55,7 @@ public string ClassifyFailure(ClassifiableMessageDetails failure)
             return GetNonStandardClassification(exception.ExceptionType);
         }
 
-        static string GetNonStandardClassification(string exceptionType)
-        {
-            return exceptionType + ": 0";
-        }
+        static string GetNonStandardClassification(string exceptionType) => exceptionType + ": No stacktrace available";
```
danielmarbach:
```suggestion
        static string GetNonStandardClassification(string exceptionType) => exceptionType + ": No stacktrace";
```

danielmarbach (approved):

PR #5519, opened 2026-06-05
Make the name of the claim for roles configurable

WilliamBZA:


ramonsmits (approved):
Reverted return statement, as we always need a policy provider due to the Authorize attribute.

All test green

ramonsmits (approved):

PR #5520, opened 2026-06-05
Audit log for authorization decisions

ramonsmits:
Audit trail for authorization decisions, emitted as Elastic Common Schema (ECS) JSON for SIEM ingestion.

### Adds
- `IAuthorizationAuditLog` / `AuthorizationAuditLog` (Infrastructure.Auth) — one allow/deny entry per evaluation, serialized as an **ECS** document (`@timestamp`, `event.*`, `user.*`, and the app-specific `servicecontrol.*` namespace). Allow → `Information`, deny → `Warning`.
- `PermissionVerbHandler` emits an entry per check (subject id, display name, permission, resource, reason); subject claims are configurable.
- Dedicated NLog routing: the `ServiceControl.Audit` category goes to its own target, `Final` so it never mixes with the plain-text operational log. The schema is owned in the domain class; the target writes the pre-rendered ECS document verbatim. Captured from `Info` upward independent of the operational `LogLevel`, so lowering verbosity never drops audit entries.

### Sink behavior
- **Container** (`DOTNET_RUNNING_IN_CONTAINER=true`): JSON to **stdout** (runtime log driver ships it). No file.
- **Service / binary install**: rolling `audit.json` file (ship with Filebeat/Vector/etc.).

### Sample output (one ECS JSON object per line)
```json
{"@timestamp":"2026-06-15T08:19:46.2024106+00:00","event":{"kind":"event","category":["iam"],"type":["allowed"],"action":"error:messages:retry","outcome":"success"},"user":{"id":"alice-sub-001","name":"Alice Smith"},"servicecontrol":{"permission":"error:messages:retry","resource":"acme.sales","reason":"role:sc-operator matched"}}
{"@timestamp":"2026-06-15T08:19:46.2280169+00:00","event":{"kind":"event","category":["iam"],"type":["denied"],"action":"error:messages:retry","outcome":"failure"},"user":{"id":"bob-sub-002","name":"Bob Jones"},"servicecontrol":{"permission":"error:messages:retry","resource":null,"reason":"no matching role"}}
```

### Tests
Unit tests for the audit log, the JSON routing, and an end-to-end render asserted as ECS with `System.Text.Json` (+ `RecordingLoggerProvider` helper).

### Scope / follow-ups
Verb-level decisions only (resource is `null` today; populated once resource-scoped handlers land). `user.roles` / `user.email`, a RavenDB UI projection, and action/Tier-1 auditing are separate follow-ups.


src/ServiceControl.Hosting/Auth/PermissionVerbHandler.cs
```diff
@@ -1,42 +1,90 @@
...
+        // 401 because the policy also includes RequireAuthenticatedUser; skipping here keeps the
+        // audit log restricted to identified principals.
+        if (context.User.Identity?.IsAuthenticated != true)
+        {
+            return Task.CompletedTask;
+        }
+
+        var subjectId = RequireClaim(context.User, oidcSettings.SubjectIdClaim, "Authentication.SubjectIdClaim");
+        var subjectName = RequireClaim(context.User, oidcSettings.SubjectNameClaim, "Authentication.SubjectNameClaim");
+        var roles = context.User.FindAll(RoleClaimType).Select(claim => claim.Value).ToArray();
```
WilliamBZA:
```suggestion
        var roles = context.User.FindAll(ClaimTypes.Role).Select(claim => claim.Value).ToArray();
```


Rather use the standard `ClaimTypes.Role` instead of our own custom string.

src/ServiceControl.Hosting/Auth/RolesClaimsTransformation.cs
WilliamBZA:
```suggestion
            claims[i + 1] = new Claim(ClaimTypes.Role, roles[i]);
```

Rather use the standard `ClaimTypes.Role` instead of our own custom string.

WilliamBZA:
```suggestion
```

Rather use the standard `ClaimTypes.Role` instead of our own custom string.

src/ServiceControl.Hosting/Auth/PermissionVerbHandler.cs
```diff
@@ -1,42 +1,90 @@
...
+    OpenIdConnectSettings oidcSettings)
+    : AuthorizationHandler<PermissionRequirement>
 {
-    public PermissionVerbHandler(string rolesClaimName)
-    {
-        RoleClaimType = rolesClaimName;
-    }
+    // The per-IdP variability of the source claim is absorbed by RolesClaimsTransformation, which
+    // reads from the path configured in Authentication.RolesClaim and emits canonical "roles" claims.
+    const string RoleClaimType = "roles";
```
WilliamBZA:
```suggestion
```

Rather use the standard `ClaimTypes.Role` instead of our own custom string.

src/ServiceControl.Infrastructure/LoggingConfigurator.cs
```diff
@@ -28,6 +29,17 @@ public static void ConfigureLogging(LoggingSettings loggingSettings)
         }
 
         public static string ConfigureNLog(string logFileName, string logPath, LogLevel logLevel)
+        {
+            var nlogConfig = BuildConfiguration(logFileName, logPath, logLevel);
```
WilliamBZA, line 33: Wow we really need to open a follow-up issue to move away from NLog

WilliamBZA (approved):
Minor tweaks on switching to use `ClaimTypes.Role` instead of our own magic string, but that can be discussed/changed separately

src/ServiceControl.Hosting/Auth/PermissionVerbHandler.cs
ramonsmits: @WilliamBZA looking in the code that will use the hardcoded "microsoft standard" which is no standard but this is also a hardcoded string too. I'll replace it but likely we likely need to change this.

src/ServiceControl.Hosting/Auth/PermissionVerbHandler.cs
ramonsmits: Again, no standard... its a Microsoft "standard" the most IDPs do not support OOTB. I'll update it

WilliamBZA: It's not an IdP standard, it's a aspnet one

src/ServiceControl.Infrastructure/LoggingConfigurator.cs
ramonsmits: @WilliamBZA and replace it with what? Microsoft does not provide a file logger.

src/ServiceControl.Hosting/Auth/PermissionVerbHandler.cs
WilliamBZA: Let's change it in a separate PR to auth

fredricadell:
Looks good! All data needed is included.

PR #5522, opened 2026-06-08
Collect more historical data from cloud-based brokers

mikeminutillo:
- AWS 365 days
- Azure 90 days

Matches [a similar change made in the endpoint throughput tool](https://github.com/Particular/Particular.EndpointThroughputCounter/pull/1086).

### DIFF
--- a/src/Particular.LicensingComponent/BrokerThroughput/BrokerThroughputCollectorHostedService.cs
+++ b/src/Particular.LicensingComponent/BrokerThroughput/BrokerThroughputCollectorHostedService.cs
@@ -98,12 +98,7 @@ public class BrokerThroughputCollectorHostedService(
                 await dataStore.SaveEndpoint(endpoint, stoppingToken);
             }
 
-            var defaultStartDate = DateOnly.FromDateTime(timeProvider.GetUtcNow().DateTime).AddDays(-30);
-            var startDate = endpoint.LastCollectedDate < defaultStartDate
-                ? defaultStartDate
-                : endpoint.LastCollectedDate.AddDays(1);
-
-            await foreach (var queueThroughput in brokerThroughputQuery.GetThroughputPerDay(queueName, startDate, stoppingToken))
+            await foreach (var queueThroughput in brokerThroughputQuery.GetThroughputPerDay(queueName, endpoint.LastCollectedDate.AddDays(1), stoppingToken))
             {
                 try
                 {
--- a/src/ServiceControl.Transports.SQS/AmazonSQSQuery.cs
+++ b/src/ServiceControl.Transports.SQS/AmazonSQSQuery.cs
@@ -23,6 +23,8 @@ using Microsoft.Extensions.Logging;
 public class AmazonSQSQuery(ILogger<AmazonSQSQuery> logger, TimeProvider timeProvider, TransportSettings transportSettings)
     : BrokerThroughputQuery(logger, "AmazonSQS")
 {
+    const int MaxDaysToQuery = 365;
+
     AmazonCloudWatchClient? cloudWatch;
     AmazonSQSClient? sqs;
     string? prefix;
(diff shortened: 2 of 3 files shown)


src/Particular.LicensingComponent/BrokerThroughput/BrokerThroughputCollectorHostedService.cs
```diff
@@ -98,12 +98,7 @@ async Task Exec(IBrokerQueue queueName, string postfix)
                 await dataStore.SaveEndpoint(endpoint, stoppingToken);
             }
 
-            var defaultStartDate = DateOnly.FromDateTime(timeProvider.GetUtcNow().DateTime).AddDays(-30);
-            var startDate = endpoint.LastCollectedDate < defaultStartDate
-                ? defaultStartDate
-                : endpoint.LastCollectedDate.AddDays(1);
-
-            await foreach (var queueThroughput in brokerThroughputQuery.GetThroughputPerDay(queueName, startDate, stoppingToken))
+            await foreach (var queueThroughput in brokerThroughputQuery.GetThroughputPerDay(queueName, endpoint.LastCollectedDate.AddDays(1), stoppingToken))
```
PhilBastian, line 101: a side issue, but can we validate the `.AddDays(1)`? I remember looking at this a while back and working out that every time the ServiceControl service is restarted for any reason, we lose a day of data collection for some transports

src/ServiceControl.Transports.SQS/AmazonSQSQuery.cs
```diff
@@ -198,7 +200,13 @@ public override async IAsyncEnumerable<QueueThroughput> GetThroughputPerDay(IBro
...
     {
         var utcNow = timeProvider.GetUtcNow();
-        var endDate = DateOnly.FromDateTime(utcNow.DateTime).AddDays(-1); // Query date up to but not including today
+        var today = DateOnly.FromDateTime(utcNow.DateTime);
+        var earliestQueryDate = today.AddDays(-MaxDaysToQuery);
+        if (startDate < earliestQueryDate)
+        {
+            startDate = earliestQueryDate;
+        }
+        var endDate = today.AddDays(-1); // Query date up to but not including today
```
PhilBastian, line 209: this comment exists here but not on ASB, however it doesn't explain _why_

PhilBastian (approved):

src/Particular.LicensingComponent/BrokerThroughput/BrokerThroughputCollectorHostedService.cs
mikeminutillo: That would happen for RabbitMQ and SQL, which do not use this date at all. Their broker queries start when they are called and run for 24 hours before returning. If they stop mid-collection, all of the collected data is lost and the timer starts again.

For Azure and SQS (which do use this date), I believe the behavior is correct. Query for the last date we have data for and then start collection from the next day.

PhilBastian: that would still mean that if, on ASB or SQS, someone closed ServiceControl at 1AM on 1/1 and started it again 2 days later, we would lose 23 hours of data from 1/1 and only start again from 2/1

mikeminutillo: It shouldn't. The behavior of Azure and SQS is to collect data from the day after the last recorded date up to yesterday. We never get partial days of data. So in the example you gave, we would not have recorded anything for the day when the service stops at 1am. But when you restart 2 days later we'll query for that day and the day after.

mikeminutillo:
Closed in favor of https://github.com/Particular/ServiceControl/pull/5544 and https://github.com/Particular/ServiceControl/pull/5544

PR #5535, opened 2026-06-15
Change ASB usage data collection method to reduce required permissions

### LINKED ISSUE #5280: Monitoring Reader role lacks permissions for queue statistics after upgrade to 6.7.4
### Symptoms

After upgrading the ServiceControl to version 6.7.4 and following the "minimum permissions approach" as described in https://docs.particular.net/servicepulse/usage-config#connection-setup-azure-service-bus-minimum-permissions, we observe the following error:

```
Connection test to AzureServiceBus failed:
Service principal 'REDACTED' have no permission to access following resources: '/subscriptions/REDACTED/resourceGroups/Dev'. Please make sure grant 'Monitoring Reader' role permission to the subscription 'REDACTED' for the service principal 'REDACTED'
Status: 403 (Forbidden)
ErrorCode: Forbidden
```

### Who's affected

Azure Service Bus users upgrading ServiceControl to version 6.7.4+

### Root cause

The root cause is most likely this commit: https://github.com/Particular/ServiceControl/commit/8983a3f3c97b3ab4601092b80a6e8c3fc74ffffd

### Confirmed workarounds

Granting permissions at the resource group level solves the issue.

<details>

</details>

### Steps to reproduce

Upgrade the SC to version 6.7.4+
Grant minimal permissions according to the documentation
Run the SC
Notice the issue

### Relevant log output

```shell

```

#### Additional information

It could be the regression comes from

```csharp
var serviceBusNamespaceResource = await armClient
    .GetServiceBusNamespaceResource(resourceId).GetAsync(...)
```

and then the read of

```csharp
var regionName = serviceBusNamespaceResource.Value.Data.Location.Name;
```

to query then against a regionalized endpoint. This implies the new code requires to GET the service bus namespace resource from the control plane to read the location, which goes against the namespace ARM id which includes /resourceGroups/<rg> so any role setup that formerly worked without needing that GET can now fail.

mikeminutillo:
Fixes #5280 

Before this change ServiceControl was using the [Azure.Monitor.Query.Metrics](https://www.nuget.org/packages/Azure.Monitor.Query.Metrics) package. This package is designed to allow querying for metrics across mutliple resources in a single request, and requires the `Monitor Reader` role at the subscription level.

This change switches to the [Azure.ResourceManager.Monitor](https://www.nuget.org/packages/Azure.ResourceManager.Monitor) package, which only requires the `Monitor Reader` role at the Service Bus namespace level.

### DIFF
--- a/src/ServiceControl.Transports.ASBS/AzureQuery.cs
+++ b/src/ServiceControl.Transports.ASBS/AzureQuery.cs
@@ -12,9 +12,9 @@ using System.Threading.Tasks;
 using Azure.Core;
 using Azure.Core.Pipeline;
 using Azure.Identity;
-using Azure.Monitor.Query.Metrics;
-using Azure.Monitor.Query.Metrics.Models;
 using Azure.ResourceManager;
+using Azure.ResourceManager.Monitor;
+using Azure.ResourceManager.Monitor.Models;
 using Azure.ResourceManager.Resources;
 using Azure.ResourceManager.ServiceBus;
 using BrokerThroughput;
@@ -33,8 +33,6 @@ public class AzureQuery(ILogger<AzureQuery> logger, TimeProvider timeProvider, T
     TokenCredential? credential;
     ResourceIdentifier? resourceId;
     ArmEnvironment armEnvironment;
-    MetricsClientAudience metricsClientAudience;
-    MetricsClient? metricsClient;
 
     protected override void InitializeCore(ReadOnlyDictionary<string, string> settings)
     {
@@ -107,7 +105,7 @@ public class AzureQuery(ILogger<AzureQuery> logger, TimeProvider timeProvider, T
             Diagnostics.AppendLine("Client secret set");
         }
 
-        (armEnvironment, metricsClientAudience) = GetEnvironment();
+        armEnvironment = GetEnvironment();
 
         if (managementUrl == null)
         {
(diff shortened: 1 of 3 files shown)


src/ServiceControl.Transports.ASBS/AzureQuery.cs
```diff
@@ -241,85 +239,28 @@ public override async IAsyncEnumerable<QueueThroughput> GetThroughputPerDay(IBro
...
-        }
-
-        var timeSeries = metricResult.TimeSeries.SingleOrDefault();
-
-        if (timeSeries is null)
-        {
-            throw new Exception($"Metrics query result for '{metricResult.Name}' contained no time series");
+            foreach (var timeSeries in metric.Timeseries)
+            {
+                return timeSeries.Data;
```
PhilBastian: this doesn't seem right? only returning the first timeseries data for the first metric?

src/ServiceControl.Transports.ASBS/AzureQuery.cs
```diff
@@ -241,85 +239,28 @@ public override async IAsyncEnumerable<QueueThroughput> GetThroughputPerDay(IBro
...
+            }
         }
 
-        return timeSeries.Values.AsReadOnly();
+        // TODO: Better error handling
```
PhilBastian: TODO

src/ServiceControl.Transports.ASBS/AzureQuery.cs
mikeminutillo: I'll check but I believe there is only one metric, with one timeseries in it.

PhilBastian (approved):

PR #5536, opened 2026-06-15
Add service.name/version to OTLP log resource

ramonsmits:
## Summary

The OTLP **log** exporter was registered with a bare `AddOtlpExporter()` and no
OpenTelemetry resource, so exported logs arrived as `unknown_service` and couldn't be
attributed or correlated. This PR gives every exported log record a service identity for
all three instances (ServiceControl, Audit, Monitoring), matching the `service.name`
already used by the Audit **metrics** resource so logs and metrics correlate.

## What changed

`LoggerUtil.Initialize(serviceName, serviceVersion)` is called once at startup (in each
`Program.cs`, before any logger is created) and builds a single OpenTelemetry resource
shared by both the host logging pipeline and the static bootstrap loggers — so *all*
exported records are attributed, including early startup diagnostics.

| Attribute | Value |
|---|---|
| `service.name` | instance name (same as the metrics resource) |
| `service.version` | instance version |
| `service.instance.id` | auto-generated per process start |

## Operator configuration

Enable OTLP logging by adding `Otlp` to `LoggingProviders`; the endpoint comes from the
standard `OTEL_EXPORTER_OTLP_ENDPOINT`. The resource uses `ResourceBuilder.CreateDefault()`,
so operators can tag a deployment via `OTEL_RESOURCE_ATTRIBUTES` (these ride on every record):

```
OTEL_RESOURCE_ATTRIBUTES=deployment.environment=production,service.namespace=orders,acme.region=eu-west-1
```

Code-set `service.*` win; extra attributes are merged in. For a stable id across restarts,
set `service.instance.id` explicitly here.

## Testing

```yaml
# collector.yaml
receivers: { otlp: { protocols: { grpc: { endpoint: 0.0.0.0:4317 } } } }
exporters: { debug: { verbosity: detailed } }
service: { pipelines: { logs: { receivers: [otlp], exporters: [debug] } } }
```
```bash
podman run --rm -p 4317:4317 -v $PWD/collector.yaml:/etc/otelcol/config.yaml:z otel/opentelemetry-collector
# then start any instance with:  LoggingProviders=NLog,Otlp  OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317
```
The collector prints each record's `Resource attributes` with `service.name = <instance name>`.

### DIFF
--- a/src/ServiceControl.Audit/Program.cs
+++ b/src/ServiceControl.Audit/Program.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Diagnostics;
 using System.Reflection;
 using Microsoft.Extensions.Logging;
 using ServiceControl.Audit.Infrastructure.Hosting;
@@ -12,6 +13,13 @@ ILogger logger = null;
 try
 {
     ExeConfiguration.PopulateAppSettings(Assembly.GetExecutingAssembly());
+
+    // Establish the telemetry identity once, before any logger is created, so every logger — including the
+    // static bootstrap loggers — attributes exported OTLP logs to this instance. Mirrors the InstanceName
+    // resolution in Settings (InternalQueueName is the legacy fallback for the instance name).
+    var instanceName = SettingsReader.Read(Settings.SettingsRootNamespace, "InstanceName",
+        SettingsReader.Read(Settings.SettingsRootNamespace, "InternalQueueName", Settings.DEFAULT_INSTANCE_NAME));
+    LoggerUtil.Initialize(instanceName, FileVersionInfo.GetVersionInfo(Assembly.GetExecutingAssembly().Location).ProductVersion);
 
     var loggingSettings = new LoggingSettings(Settings.SettingsRootNamespace);
     LoggingConfigurator.ConfigureLogging(loggingSettings);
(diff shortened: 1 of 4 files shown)


src/ServiceControl.Audit/Program.cs
```diff
@@ -13,6 +14,13 @@
 {
     ExeConfiguration.PopulateAppSettings(Assembly.GetExecutingAssembly());
 
+    // Establish the telemetry identity once, before any logger is created, so every logger — including the
+    // static bootstrap loggers — attributes exported OTLP logs to this instance. Mirrors the InstanceName
+    // resolution in Settings (InternalQueueName is the legacy fallback for the instance name).
+    var instanceName = SettingsReader.Read(Settings.SettingsRootNamespace, "InstanceName",
```
ramonsmits: @WilliamBZA @dvdstelt this is really needed to ensure OTel related attributes are set.

The version could be extracted in the `LoggerUtil.Initialize` as that version is the same for all SC assemblies in the build.

However, the name is not, unfortnately although very similar for all 3 instances SettingsReader can be used in LoggerUtil but not `Settings` which is instance specific.

Any refactorings you can think of to maybe this a bit less fugly?

WilliamBZA (approved):

dvdstelt (approved):

PR #5538, opened 2026-06-16
Expose a per-instance my/routes authorization manifest

WilliamBZA:
## What

Adds `GET /api/my/routes`, served by every instance (Primary, Audit, Monitoring). It returns the concrete set of API routes the current token is allowed to call, as `[{ "method", "urlTemplate" }]` entries.

This is the authorization contract between ServiceControl and clients (ServicePulse): each instance advertises only the routes it actually serves and the caller is actually permitted to reach. The client matches its outgoing requests against that set and never learns ServiceControl's internal `instance:resource:action` permission vocabulary.

## Also in this PR

- Adds an **`admin` role** between `reader` and `writer`: read-all plus full management of config/admin-area resources (licensing, notifications, redirects, throughput, connections), but **not** the message-triage write actions (retry/edit/archive/restore). Roles are wildcard patterns expanded once into frozen sets for O(1), allocation-free per-request checks.


src/ServiceControl/Infrastructure/WebApi/MeController.cs
```diff
@@ -0,0 +1,91 @@
...
+    using System.Collections.Generic;
+    using System.Linq;
+    using Auth;
+    using Microsoft.AspNetCore.Authorization;
+    using Microsoft.AspNetCore.Mvc;
+    using ServiceBus.Management.Infrastructure.Settings;
+
+    [ApiController]
+    [Route("api")]
+    [Authorize]
```
ramonsmits: Will this work when auth is disabled? Will we assume that SP will not invoke this API when auth is disabled?

Instead we would use AllowAnonymous and have fallback logic.

src/ServiceControl/Infrastructure/WebApi/MeController.cs
```diff
@@ -0,0 +1,91 @@
...
+using ServiceBus.Management.Infrastructure.Settings;
+
+[ApiController]
+[Route("api")]
+[Authorize]
+public class MeController(Settings settings) : ControllerBase
+{
+    [HttpGet]
+    [Route("my/permissions/all")]
+    public ActionResult<PermissionsDescriptor> GetMyPermissions()
```
ramonsmits: Maybe use all as that is part of the route?
```suggestion
        public ActionResult<PermissionsDescriptor> GetAllPermissions()
```

src/ServiceControl/Infrastructure/WebApi/MeController.cs
```diff
@@ -0,0 +1,92 @@
...
+    {
+        var descriptor = new PermissionsDescriptor(
+            User.FindFirst("sub")?.Value ?? User.Identity?.Name ?? "unknown",
```
ramonsmits: What are we returning here? a user Id or display name? Can't we assume `sub` to always exist and not be null when a user is authenticated?


```suggestion
            User.FindFirst("sub"),
```

src/ServiceControl/Infrastructure/WebApi/MeController.cs
```diff
@@ -0,0 +1,92 @@
...
+            GrantedPermissions().OrderBy(p => p, StringComparer.Ordinal).ToList());
```
ramonsmits: Do we really need to order server side? Should we always order client side anyway?


```suggestion
            GrantedPermissions()
            ).ToList());
```

src/ServiceControl/Infrastructure/WebApi/MeController.cs
ramonsmits: @WilliamBZA in:

- https://github.com/Particular/ServiceControl/pull/5520

I added configurable settings for these keys

WilliamBZA: Will rebase once that's merged

src/ServiceControl/Infrastructure/WebApi/MeController.cs
WilliamBZA: If auth is disabled this returns 403 and the tab isn't displayed by SP. Why would allowanonymous work here instead? What permissions should we return if there is no authorization?

ramonsmits (approved):

PR #5555, opened 2026-06-24
Reduce SQL Server queue length monitoring query volume

ramonsmits:
Spike addressing high SQL query volume from queue-length monitoring (one `IF EXISTS` + `max-min(RowVersion)` statement per tracked queue, every 200ms).

### Changes
- **Single query.** Read approximate row counts for all tracked queues from `sys.partitions` (one query per catalog) instead of probing each queue table. No queue-table reads, no locks, only metadata visibility required. Accuracy is comparable to the existing `max-min(RowVersion)` estimate (both approximate; the catalog counter avoids identity-gap over-counting).
- **Concurrent pacing.** The poll interval now runs concurrently with the query, so the effective cadence is `max(interval, queryDuration)` instead of `interval + queryDuration`. That additive query-time term is what let the cadence drift past the 1s monitoring bucket and starve buckets of samples — the #4556 false-zero "sawtooth". Removing the drift supersedes the #4557 200ms oversampling workaround and lets the cadence return to a sane default.
- **Configurable interval** via `QueueLengthQueryDelayInterval` connection string part (default 1s, matching the finest monitoring bucket: 1-minute history / 60 = 1s per point), following the existing ASB convention.
- **Adaptive back-off** up to `QueueLengthQueryMaxDelayInterval` (default 10s) while every monitored queue is empty; now on by default. The base interval is always used while any queue has work, so the fix for #4556 is preserved. Set the max equal to the base to disable back-off.
- **`SqlTable` cleanup.** Identifier properties renamed to `Unquoted*` to make the quoted-vs-unquoted contract explicit, converted to a primary constructor, and the duplicated per-table length-query branches collapsed into helpers.

### Permissions
The bulk query reads only catalog views, so it needs **less** permission than the per-queue probe it replaces:
- Relies on standard metadata visibility — any permission the monitoring user holds on a queue table (e.g. `SELECT`) makes that table's row visible. No `VIEW DATABASE STATE` required (hence `sys.partitions` rather than the `sys.dm_db_partition_stats` DMV).
- The previous query read table data (`SELECT max([RowVersion]) … FROM <queue>`), which already requires `SELECT` on each queue table — strictly more — so any user that worked before continues to work.
- A queue table the user has no permission on simply returns no row and is treated as "does not exist" (same effective outcome as the old `IF EXISTS` → -1).

### DIFF
--- a/src/ServiceControl.Transports.SqlServer/ConnectionStringExtensions.cs
+++ b/src/ServiceControl.Transports.SqlServer/ConnectionStringExtensions.cs
@@ -1,5 +1,6 @@
 ﻿namespace ServiceControl.Transports.SqlServer
 {
+    using System;
     using System.Data.Common;
 
     static class ConnectionStringExtensions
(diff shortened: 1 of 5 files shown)


src/ServiceControl.Transports.SqlServer/SqlTable.cs
```diff
@@ -42,11 +51,56 @@ SELECT isnull(cast(max([RowVersion]) - min([RowVersion]) + 1 AS int), 0) FROM {_
         }
 
         readonly string _fullTableName;
+
+        // Unquoted identifier parts, used to group/match against the catalog views in the bulk query.
+        public string Name { get; }
+        public string Schema { get; }
+        public string? Catalog { get; }
```
mauroservienti: Wouldn't these be better named as `Unquoted*`?

mauroservienti (approved):
One minor comment, @ramonsmits. Otheriwise, it's good to go.

johnsimons (approved):
LGTM

mauroservienti (approved):

src/ServiceControl.Transports.SqlServer/ConnectionStringExtensions.cs
```diff
@@ -11,6 +12,39 @@ public static string RemoveCustomConnectionStringParts(this string connectionStr
...
+        {
+            interval = null;
+
+            var builder = new DbConnectionStringBuilder { ConnectionString = connectionString };
+
+            if (builder.TryGetValue(key, out var value))
+            {
+                if (!int.TryParse(value.ToString(), out var milliseconds) || milliseconds <= 0)
+                {
+                    throw new Exception($"Can't parse '{value}' as a valid {key} (expected a positive integer number of milliseconds).");
```
mauroservienti:
```suggestion
                    throw new Exception($"Can't parse '{value}' as a valid {key} (expected an integer number of milliseconds greater than zero).");
```

mauroservienti (approved):

PR #5569, opened 2026-07-02
Allow exclusions for finer grained control

WilliamBZA:


src/ServiceControl.Infrastructure/Auth/RolePermissions.cs
```diff
@@ -37,15 +37,17 @@ public static class RolePermissions
     static readonly Dictionary<string, string[]> RolePatterns = new(StringComparer.OrdinalIgnoreCase)
     {
         [Reader] = ["*:*:view"],
-        [Writer] = ["*:*:*"],
+        [Writer] =
+        [
+            "*:*:*",
+            "-error:licensing:*",
```
ramonsmits, line 43: I'm not a fan of denials. This means that when a user has both Writer+Admin roles that they now cannot access most configuration related views as **denials** must always be prioritized over **allows**.

ramonsmits:
Closed in favor of:

- https://github.com/Particular/ServiceControl/pull/5571

PR #5570, opened 2026-07-02
Audit trail for message actions (retry, archive, unarchive, edit)

ramonsmits:
Records every user-initiated **retry**, **edit**, **archive**, and **unarchive** action as a structured audit entry attributed to the initiating user (or `anonymous` when authentication is disabled). Each invocation produces one **operation** entry (who / what / scope) plus one **message** entry per message actually acted on. Message entries are emitted at execution time — after the message is really retried, edited, or archived — including for bulk/group operations resolved in the background and resumed after a restart. Both entry kinds share an operation id so an operation can be traced through to every message it affected. A group request that is ignored because the same operation is already in progress is not audited.

### Audit output

Entries are written to `audit.json` (one ECS-formatted JSON object per line) and to the console, on dedicated log categories `ServiceControl.Audit` (operations) and `ServiceControl.Audit.Messages` (per-message), so the audit stream can be shipped to a SIEM independently of the operational log. Successful actions log at Information, failures at Warning. The operation entry is written after the action is performed, so `outcome` records what actually happened — `failure` when the underlying send fails (the request then also fails).

Example — retrying all messages (operation entry followed by one entry per staged message):

```json
{"@timestamp":"2026-07-03T12:18:54.0412345+00:00","event":{"kind":"event","category":["configuration"],"type":["change"],"action":"error:messages:retry","outcome":"success"},"user":{"id":"d4c3b2a1-56f8-4c11-9a3e-2b7d0c4e8f01","name":"alice@example.com"},"servicecontrol":{"permission":"error:messages:retry","scope":"all","operation":{"id":"0HN7GK3M8QABC:00000042"}}}
{"@timestamp":"2026-07-03T12:18:54.0898765+00:00","event":{"kind":"event","category":["configuration"],"type":["change"],"action":"error:messages:retry","outcome":"success"},"user":{"id":"d4c3b2a1-56f8-4c11-9a3e-2b7d0c4e8f01","name":"alice@example.com"},"servicecontrol":{"permission":"error:messages:retry","scope":"all","message":{"id":"3f2504e0-4f89-11d3-9a0c-0305e82c3301"},"operation":{"id":"0HN7GK3M8QABC:00000042"}}}
```

- `event.type` is `change` for retry/edit/unarchive and `deletion` for archive; `outcome` is `success` or `failure`.
- Operation entries carry `scope` (single/batch/queue/endpoint/all/group/range), and where applicable `resource` and `count`; message entries carry `message.id` and the scope of the initiating operation. Null-valued fields are omitted.

### Settings

- **`LogPath`** — directory for `audit.json` (alongside the operational log). The file archives daily or above 30 MB, keeping the last 14 archives. In containers the audit trail goes to the console only (no file).
- **`LogLevel`** — controls operational verbosity only. The audit trail is always captured from Information upward, so lowering `LogLevel` never drops audit entries.

### Correlation

Each operation's id is the ASP.NET request id (`TraceIdentifier`) and is returned on every response as the **`Request-Id`** header (exposed via CORS on all instances). The same value appears in `servicecontrol.operation.id` on the operation entry and every per-message entry, so a caller can correlate a single API call to each affected message and to the request's other logs. For a retry forwarded to a remote instance (`instance_id`), the returned `Request-Id` is the remote instance's request id, matching the audit entries recorded on that instance.


fredricadell:
This looks very good I think. But as I understand it, if we activate the edit message function, just make sure that the information/data that's been edited is included in the audit. Otherwise I think this is fine!

WilliamBZA (approved):
Approval dependent on:

- https://github.com/Particular/ServiceControl/pull/5577/

src/ServiceControl.Infrastructure/Auth/MessageActionAuditLog.cs
```diff
@@ -0,0 +1,118 @@
...
+        _ => scope.ToString().ToLowerInvariant()
+    };
+
+    // Logged with the pre-rendered document as the state, not as a "{AuditEvent}" template parameter:
+    // a parameterized message is exported over OTLP with the literal "{AuditEvent}" placeholder as the
+    // record body and the JSON only in an attribute — backends that map body → message show the
+    // placeholder. A plain-string state exports the document exactly once, as the record body; NLog is
+    // unaffected either way and writes the same line to audit.json.
+    static readonly EventId OperationEventId = new(2001);
+    static readonly EventId MessageEventId = new(2002);
```
dvdstelt, line 116: Are failure and success the same event ID? In the branch this is merged into, we separate failure and success:
https://github.com/Particular/ServiceControl/pull/5518/changes#diff-3d7892db27fbec0b2e451ff32b2aa7c088cef3539c8939e4032ec60fcc8d0d5dR78-R82

To isolate failures and successes, we should create separate event IDs, I would say.

dvdstelt: If we keep this:
```
LogOperation         EventId = 2001  (Information)
LogOperationFailure  EventId = 2002  (Warning)
LogMessage           EventId = 2003  (Information)
LogMessageFailure    EventId = 2004  (Warning)
```

That keeps the `1xxx = authorization` / `2xxx = message actions` namespacing and makes "filter by EventId" behave the same across both.

src/ServiceControl.Infrastructure.Tests/Auth/CurrentUserAccessorTests.cs
```diff
@@ -0,0 +1,55 @@
...
+    public void Anonymous_when_not_authenticated()
+    {
+        Assert.That(Create().Resolve(new ClaimsPrincipal(new ClaimsIdentity())), Is.EqualTo(AuditUser.Anonymous));
+    }
+
+    [Test]
+    public void Anonymous_when_subject_claim_absent()
+    {
+        Assert.That(Create().Resolve(Authenticated(new Claim("preferred_username", "Alice"))), Is.EqualTo(AuditUser.Anonymous));
+    }
```
dvdstelt, line 54: This would also test if the claim is empty. Would that be valuable? Otherwise the user might be returned as anonymous.

```suggestion
    }
    
	[Test]
	public void Anonymous_when_subject_claim_is_empty()
	{
	    Assert.That(Create().Resolve(Authenticated(new Claim("sub", ""), new Claim("preferred_username", "Alice"))), Is.EqualTo(AuditUser.Anonymous));
	}
	
	[Test]
	public void Falls_back_to_id_when_name_claim_is_empty()
	{
	    var user = Create().Resolve(Authenticated(new Claim("sub", "alice-sub"), new Claim("preferred_username", "")));
	    Assert.That(user.Name, Is.EqualTo("alice-sub"));
	} 
```

dvdstelt (approved):

WilliamBZA (approved):

dvdstelt:
I created a card to look at my comments later with Ramon

PR #5571, opened 2026-07-02
Replace wildcard role-permission expansion with explicit lists

ramonsmits:
Alternative to:

- #5569

Removes the wildcard/exclusion pattern machinery (`Expand()`, segment matching) from `RolePermissions` and declares each role's permissions as explicit constant lists built from additive groups (`Read`, `ReadConfiguration`, `Operate`, `Configure`). The resulting per-role sets are identical to #5569: reader = all views, writer = operate but no licensing/notifications/redirects/throughput access, admin = everything.

A guard test fails when a new `Permissions` constant is not assigned to any role, so every new permission requires an explicit classification decision instead of being granted implicitly by a wildcard.


WilliamBZA (approved):
Much better!

PR #5575, opened 2026-07-06
Allow installing error instance from SCMU without a valid license

PhilBastian:
To allow for usage report generation in ServicePulse
- also fixed license expired popup message, which had its text binding broken on main


mikeminutillo (approved):

PhilBastian:
merge https://github.com/Particular/docs.particular.net/pull/8428 when released

PR #5579, opened 2026-07-07
Fix RolesClaim default docs and cover the RBAC deny path

dvdstelt:
Follow-up to the review of #5518. Two issues surfaced there that were not yet addressed.

## 1. `RolesClaim` default contradicted its own documentation

The setting defaults to the flat `roles` claim in code (and in the published docs), but two XML doc comments claimed the default was `realm_access.roles`. That is misleading: an operator following the comment for a stock Keycloak instance would expect roles to resolve out-of-box, when in fact the flat default yields no roles until `RolesClaim` is pointed at the nested path.

Corrected the comments in `OpenIdConnectSettings.cs` and `RolesClaimsTransformation.cs` to state the flat `roles` default and point operators at `realm_access.roles` (Keycloak out-of-box) or `cognito:groups` (AWS Cognito) as dotted-path options. No behavior change.

## 2. The RBAC deny path had no end-to-end coverage

The acceptance suite exercised authentication thoroughly (invalid, expired, wrong audience/issuer, and a reader accepted with 200) but never asserted the core authorization outcome: an authenticated user lacking the required role is forbidden. Three test files even carried a `// would be 403` comment for a case no test exercised.

Added two end-to-end tests against the Primary instance:

- a reader (view-only) POSTing to `errors/retry/all` is forbidden (403), not merely unauthenticated (401)
- `my/routes` projects to exactly what the role permits: the manifest advertises a view route the reader may call (`GET /api/errors`) and omits a retry route it may not (`POST /api/errors/retry/all`)

This pins the invariant the design relies on: the advertised manifest and the enforced policy read the same inputs and cannot drift apart.

### DIFF
--- a/src/ServiceControl.Hosting/Auth/RolesClaimsTransformation.cs
+++ b/src/ServiceControl.Hosting/Auth/RolesClaimsTransformation.cs
@@ -11,9 +11,10 @@ using ServiceControl.Infrastructure.Auth;
 /// <summary>
 /// Normalises per-IdP role claim shapes into a flat set of <c>roles</c> claims that
 /// <see cref="PermissionVerbHandler"/> can read directly. The source path is configured via
-/// <c>Authentication.RolesClaim</c> (default <c>realm_access.roles</c> — the Keycloak out-of-box
-/// shape). Flat claim names work too (<c>roles</c> for Keycloak with a "User Realm Role" mapper or
-/// Microsoft Entra ID app roles, <c>cognito:groups</c> for AWS Cognito).
+/// <c>Authentication.RolesClaim</c> (default <c>roles</c> — a flat top-level claim, as emitted by
+/// Microsoft Entra ID app roles or Keycloak with a "User Realm Role" mapper). A dotted path reaches
+/// into a nested JSON object claim: <c>realm_access.roles</c> for Keycloak's out-of-box shape, or
+/// <c>cognito:groups</c> for AWS Cognito.
 /// <para>
 /// ASP.NET may invoke <see cref="TransformAsync"/> multiple times for the same principal; a sentinel
 /// claim makes the transformation idempotent and returns the same principal on subsequent calls.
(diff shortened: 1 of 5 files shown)


src/ServiceControl.Hosting/Auth/RolesClaimsTransformation.cs
WilliamBZA, line 23: I actually have a feeling this entire class is not needed because we specify the `TokenValidationParameters.RoleClaimType` value in https://github.com/Particular/ServiceControl/blob/auth/src/ServiceControl.Hosting/Auth/HostApplicationBuilderExtensions.cs#L47

WilliamBZA: Nevermind, it is still needed.

WilliamBZA (approved):

src/ServiceControl.Hosting/Auth/RolesClaimsTransformation.cs
dvdstelt: This class is actually the load-bearing part here, and `RoleClaimType` is the redundant one. `TokenValidationParameters.RoleClaimType` doesn't create or rename claims to `ClaimTypes.Role`; it only tells the identity which claim `IsInRole()` / `[Authorize(Roles=…)]` should read. Our enforcement doesn't use those, it reads `context.User.FindAll(ClaimTypes.Role)` in both `PermissionVerbHandler` and `EffectivePermissions`. With only `RoleClaimType` set, that returns nothing:

```
RoleClaimType = "roles" on the identity
IsInRole("reader")             = True
FindAll(ClaimTypes.Role).Count = 0     ← what our handlers see
```

So dropping `RolesClaimsTransformation` would make every RBAC check find zero roles and deny everyone. It's also the only thing that can read a nested path like `realm_access.roles` (Keycloak out-of-box), which `RoleClaimType` can't express at all since it's a claim name, not a JSON path. The genuinely redundant line is `RoleClaimType = oidcSettings.RolesClaim` in `HostApplicationBuilderExtensions` — the transformation already normalizes everything to `ClaimTypes.Role`, so that setting is dead.

Issue #5592, opened 2026-07-09
TotalThroughput in the throughput report is the sum of per-endpoint peaks, not the peak daily total


rebchr:
#### Description

The throughput report generated by the licensing component reports a `TotalThroughput` that does not match the published definition of daily throughput. The [pricing definitions page](https://particular.net/pricing/definitions) defines *maximum message throughput per day* as:

> The maximum number of messages processed by all NServiceBus endpoints across the entire organization in a day.

That is: for each calendar day, sum the messages processed by all endpoints, then take the day with the highest total.

However, `GenerateThroughputReport` in `src/Particular.LicensingComponent/ThroughputCollector.cs` computes it the other way around: it first takes each endpoint's best single day (`Throughput = endpointData.ThroughputData.MaxDailyThroughput()`) and then sums those per-endpoint peaks (`TotalThroughput = queueThroughputs.Sum(q => q.Throughput ?? 0)`).

Unless every endpoint happens to peak on the same day, the sum of per-endpoint maxima is strictly greater than the maximum daily total. The more queues a system has and the longer the reporting window, the larger the overstatement — for a system with hundreds of queues reporting over many months, `TotalThroughput` can overstate the actual peak day severalfold. Since this report is used in licensing/tier conversations, the overstatement directly affects customers.

#### Example

Two endpoints, two days:

| | Day 1 | Day 2 | Endpoint max |
|---|---|---|---|
| Endpoint A | 100 | 10 | 100 |
| Endpoint B | 20 | 100 | 100 |
| **Daily total** | **120** | **110** | |

- Reported `TotalThroughput` today: 100 + 100 = **200**
- Per the published definition: max(120, 110) = **120**

#### Expected behavior

`TotalThroughput` equals the maximum over calendar days of the organization-wide daily sum, matching the published definition.

#### Versions

Observed in ServiceControl 6.18.0 (current `master` behaves the same; the unit test `Should_return_correct_throughput_in_report_when_data_only_from_one_source` encodes the sum-of-maxima expectation).

#### Additional context

I have a fix ready and will open a PR referencing this issue.

PhilBastian:
After some investigation we've discovered that this field, although it's inaccurate in its calculation, is not used as part of licensing. This is because this calculation includes all queues found on the broker, not just the ones which are involved in the NServiceBus system and contribute to endpoints where the throughput is counted for licensing purposes, unless the user has gone to the effort to mark each queue as "NotNServiceBus".

[...]

PR #5593, opened 2026-07-09
Calculate TotalThroughput as the maximum organization-wide daily total

rebchr:
Fixes:

- #5592

## What

`TotalThroughput` in the throughput report is now the maximum over calendar days of the organization-wide daily total, instead of the sum of each endpoint's individual best day. This matches the published definition of *maximum message throughput per day* at [particular.net/pricing/definitions](https://particular.net/pricing/definitions):

> The maximum number of messages processed by all NServiceBus endpoints across the entire organization in a day.

## Why

The previous calculation (`queueThroughputs.Sum(q => q.Throughput)`, where each queue's `Throughput` is its own peak day) counts every endpoint's peak as if they all occurred on the same day. Unless all endpoints peak simultaneously, this is strictly greater than the actual busiest day, and the overstatement grows with the number of queues and the length of the reporting window. Since the report is used in licensing conversations, the reported number should match the definition it is measured against.

Example — two endpoints, two days: A = [100, 10], B = [20, 100]. Previous `TotalThroughput`: 200. Actual busiest day: 120.

## How

- New `DailyThroughput()` extension yields an endpoint's per-day values; when multiple sources have data for the same day, the maximum value is used, consistent with the existing per-endpoint `MaxDailyThroughput()` semantics (all sources measure the same messages).
- `GenerateThroughputReport` accumulates daily totals across all endpoints in the existing loop and sets `TotalThroughput` to the highest daily total (0 when there is no data).
- The per-queue `Throughput` field is unchanged, so `TotalThroughput` remains ≥ every per-queue value. The set of queues contributing to the total is also unchanged.
- Added a test that encodes the definition (endpoints peaking on different days). Updated the one assertion and one approval file that encoded the previous sum-of-peaks behavior (195 → 185, 249 → 238).

All tests in `Particular.LicensingComponent.UnitTests` pass (75/75).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Particular.LicensingComponent.UnitTests/ApprovalFiles/ThroughputCollector_Report_Throughput_Tests.Should_generate_correct_report.approved.txt
+++ b/src/Particular.LicensingComponent.UnitTests/ApprovalFiles/ThroughputCollector_Report_Throughput_Tests.Should_generate_correct_report.approved.txt
@@ -141,7 +141,7 @@
         "DailyThroughputFromMonitoring": []
       }
     ],
-    "TotalThroughput": 249,
+    "TotalThroughput": 238,
     "TotalQueues": 5,
     "IgnoredQueues": [],
     "EnvironmentInformation": {
--- a/src/Particular.LicensingComponent.UnitTests/ThroughputCollector/ThroughputCollector_Report_Throughput_Tests.cs
+++ b/src/Particular.LicensingComponent.UnitTests/ThroughputCollector/ThroughputCollector_Report_Throughput_Tests.cs
@@ -147,8 +147,34 @@ class ThroughputCollector_Report_Throughput_Tests : ThroughputCollectorTestFixtu
 
         using (Assert.EnterMultipleScope())
         {
-            Assert.That(report.ReportData.TotalThroughput, Is.EqualTo(195), $"Incorrect TotalThroughput recorded");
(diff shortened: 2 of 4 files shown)


PhilBastian:
Thanks @rebchr, we're reviewing your PR. The `TotalThroughput` value in this report is not used by Particular for determining Tier, which is likely why it's never been picked up before that the calculation was incorrect. We perform a calculation from the raw daily values that looks very similar to what you've done here 😄.
I'm thinking that this field should be removed entirely...

rebchr:
Hi Phil, thanks for the quick response — good to hear the tier calculation is done from the raw daily values on your side, and that it aligns with what this PR does. Since the daily values for queues marked `NotNServiceBusEndpoint` are stripped from the report, we're assuming that calculation also only covers actual NServiceBus endpoints.

Removing the field entirely would work just as well for us. The reason we care is that the pre-computed total does end up in licensing conversations in practice: during our 2025 renewal the combined reports were quoted to us as "11+ million" throughput, while the actual busiest day computed from the daily values was ~8.4 million. Whichever way it's resolved — corrected or removed — the important thing is that the report no longer presents a total that contradicts the published definition.

PhilBastian:
@rebchr closing this PR with the intent to instead raise one to remove the value entirely. This won't happen immediately, since it need to be synchronised with the separate throughput tool and our internal processing of the generated file; we'll ping you on [the issue](https://github.com/Particular/ServiceControl/issues/5592) to let you know when this happens.

PR #5604, opened 2026-07-15
Handle ASB queue-length management throttling (pipeline detection)

### LINKED ISSUE #5606: Monitoring queue length checks can throttle the Azure Service Bus namespace with no log evidence
#### Description

The ServiceControl **Monitoring** instance's queue length monitoring routine periodically queries the Azure Service Bus namespace for queue runtime properties. On busy or large namespaces this generates enough management-plane requests to trigger Azure Service Bus **request throttling**. The throttling is visible on the Azure Service Bus namespace dashboard but produces **no corresponding error or warning in the ServiceControl logs**.

#### Expected behavior

Queue length monitoring should not drive the namespace into sustained throttling; and if throttling does occur, it should be observable in the ServiceControl logs so it can be diagnosed.

#### Actual behavior

- The Azure Service Bus namespace dashboard shows request throttling that begins when the Monitoring instance is started.
- Nothing is written to the ServiceControl Monitoring logs — no errors, warnings, or other indication.
- **Stopping the Monitoring instance resolves the throttling** (the dashboard shows it subside).

#### Versions

ServiceControl **6.18.2** (latest). All supported **6.x** versions are affected.

Transport: Azure Service Bus (`NetStandardAzureServiceBus`).

## Steps to reproduce

1. Configure a ServiceControl Monitoring instance on the Azure Service Bus transport, pointed at a namespace with many queues / active traffic.
2. Start the Monitoring instance.
3. Observe request throttling on the Azure Service Bus namespace metrics/dashboard.
4. Note that the Monitoring logs contain no related errors or warnings.
5. Stop the Monitoring instance and observe the throttling subside.

## Relevant log output

```
(none — the absence of any log evidence is part of the bug)
```

#### Workarounds

- Increase `QueueLengthQueryDelayInterval` (milliseconds) in the Monitoring instance's transport connection string to reduce polling frequency.
- Isolate the Monitoring instance to a dedicated namespace.

#### Possible solutions

- Make the throttling observable in the logs.
- Detect throttling and back off the query cadence.

#### Additional information (root-cause analysis)

- The queue length provider calls `GetQueuesRuntimeProperties` for all queues in the namespace at a default interval of 500 ms. These are management-plane operations, which Azure Service Bus throttles independently of message traffic.
- The Azure SDK retries throttled (HTTP 429) management requests internally and usually succeeds, so the throttling never surfaces as an exception — which is why the dashboard counts throttled requests while the logs stay silent.

ramonsmits:
- Resolves https://github.com/Particular/ServiceControl/issues/5606

Detects Azure Service Bus management throttling via a PerRetry HTTP pipeline policy (observes the 429s the SDK retries away) and reactively backs off the queue-length query interval, logging a warning once per throttling episode.

One of two alternative approaches for the same issue; the other (#5605) disables the SDK retries instead.

### DIFF
--- a/src/ServiceControl.Transports.ASBS.Tests/QueueLengthProviderBackoffTests.cs
+++ b/src/ServiceControl.Transports.ASBS.Tests/QueueLengthProviderBackoffTests.cs
@@ -1,0 +1,61 @@
+namespace ServiceControl.Transports.UnitTests.ASBS
+{
+    using System;
+    using NUnit.Framework;
+    using ServiceControl.Transports.ASBS;
+
+    [TestFixture]
+    class QueueLengthProviderBackoffTests
+    {
+        // Base is the configured QueueLengthQueryDelayInterval (default). Max is the internal reactive-backoff cap.
+        static readonly TimeSpan Base = TimeSpan.FromMilliseconds(500);
+        static readonly TimeSpan Max = TimeSpan.FromSeconds(60);
+
+        [Test]
+        public void Doubles_the_delay_when_throttled()
+        {
+            // A throttled cycle backs off so the provider stops making the throttling worse.
+            Assert.That(QueueLengthProvider.NextDelay(Base, Base, Max, throttled: true),
+                Is.EqualTo(TimeSpan.FromSeconds(1)));
+        }
+
+        [Test]
+        public void Caps_the_delay_at_max_when_throttled()
+        {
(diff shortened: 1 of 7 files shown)


johnsimons:
This PR was opened based on a support case. 
The solution for the support case was to control the queue length cycle delays. 
This is not necessary.

PR #5613, opened 2026-07-16
Enable EF database migrations and advanced persistence configuration

johnsimons:
Allows ServiceControl to automatically apply EF Core database migrations during the `setup` command. 
Configures EF Core to use retry-on-failure for transient errors, enhancing database connection stability. 
Adds an option to enable sensitive data logging for debugging purposes. 
Switches from `AddPooledDbContextFactory` to `AddDbContext` for direct `DbContext` access and improved flexibility.

### DIFF
--- a/src/ServiceControl.Persistence.EFCore.SqlServer/SqlServerDatabaseMigrator.cs
+++ b/src/ServiceControl.Persistence.EFCore.SqlServer/SqlServerDatabaseMigrator.cs
@@ -1,0 +1,23 @@
+namespace ServiceControl.Persistence.EFCore.SqlServer;
+
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.Logging;
+using ServiceControl.Persistence.EFCore.Abstractions;
+using ServiceControl.Persistence.EFCore.DbContexts;
+
+class SqlServerDatabaseMigrator(ServiceControlDbContext dbContext, ILogger<SqlServerDatabaseMigrator> logger) : IDatabaseMigrator
+{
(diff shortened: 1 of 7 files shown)


src/ServiceControl.Persistence.EFCore.SqlServer/SqlServerDatabaseMigrator.cs
```diff
@@ -0,0 +1,23 @@
...
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.Logging;
+using ServiceControl.Persistence.EFCore.Abstractions;
+using ServiceControl.Persistence.EFCore.DbContexts;
+
+class SqlServerDatabaseMigrator(ServiceControlDbContext dbContext, ILogger<SqlServerDatabaseMigrator> logger) : IDatabaseMigrator
+{
+    public async Task ApplyMigrations(CancellationToken cancellationToken = default)
+    {
+        logger.LogInformation("Starting SQL Server database migration");
```
rbev, line 12: Apart from the log messages these are identical - could they go in the shared code?
Don't the manifest settings end up in the container so you could format the log strings out of that?

rbev (approved):

src/ServiceControl.Persistence.EFCore.SqlServer/SqlServerDatabaseMigrator.cs
johnsimons: From my spike, the code here diverged; hence the copy-paste now, but we can always make it better as it takes shape. For now, I am happy with the way it is.

PR #5615, opened 2026-07-17
Integrate EF Core into ingestion unit of work

johnsimons:
The `IIngestionUnitOfWork` interface is updated to `IAsyncDisposable` to properly manage asynchronous resources like `ServiceControlDbContext` and `AsyncServiceScope`.

`EFIngestionUnitOfWork` now manages a dedicated `ServiceControlDbContext` instance within its own `AsyncServiceScope`, persisting changes via `SaveChangesAsync`. A new `IBodyStoragePersistence` interface is introduced for abstracting message body storage.

### DIFF
--- a/src/ServiceControl.Persistence.EFCore/Implementation/UnitOfWork/EFIngestionUnitOfWork.cs
+++ b/src/ServiceControl.Persistence.EFCore/Implementation/UnitOfWork/EFIngestionUnitOfWork.cs
@@ -1,21 +1,35 @@
 namespace ServiceControl.Persistence.EFCore.Implementation.UnitOfWork;
 
+using Microsoft.Extensions.DependencyInjection;
+using ServiceControl.Persistence.EFCore.Abstractions;
+using ServiceControl.Persistence.EFCore.DbContexts;
+using ServiceControl.Persistence.EFCore.Infrastructure;
 using ServiceControl.Persistence.UnitOfWork;
 
 public class EFIngestionUnitOfWork : IIngestionUnitOfWork
 {
+    readonly ServiceControlDbContext dbContext;
+    readonly AsyncServiceScope scope;
+
+    public EFIngestionUnitOfWork(AsyncServiceScope scope, ServiceControlDbContext dbContext, IBodyStoragePersistence storagePersistence, EFPersisterSettings settings)
+    {
+        this.scope = scope;
+        this.dbContext = dbContext;
+    }
+
     public IMonitoringIngestionUnitOfWork Monitoring =>
         throw new NotImplementedException();
 
     public IRecoverabilityIngestionUnitOfWork Recoverability =>
         throw new NotImplementedException();
 
-    public Task Complete(CancellationToken cancellationToken) =>
-        throw new NotImplementedException();
+    public Task Complete(CancellationToken cancellationToken) => dbContext.SaveChangesAsync(cancellationToken);
 
-    public void Dispose()
+    public async ValueTask DisposeAsync()
     {
-        // Nothing to dispose yet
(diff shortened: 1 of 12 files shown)


src/ServiceControl.Persistence.EFCore/Implementation/UnitOfWork/EFIngestionUnitOfWorkFactory.cs
```diff
@@ -1,12 +1,24 @@
...
+    IBodyStoragePersistence storagePersistence) : IIngestionUnitOfWorkFactory
 {
-    public ValueTask<IIngestionUnitOfWork> StartNew() =>
-        throw new NotImplementedException();
+    public ValueTask<IIngestionUnitOfWork> StartNew()
+    {
+        var scope = serviceProvider.CreateAsyncScope();
+        var dbContext = scope.ServiceProvider.GetRequiredService<ServiceControlDbContext>();
+        var settings = scope.ServiceProvider.GetRequiredService<EFPersisterSettings>();
+        var unitOfWork = new EFIngestionUnitOfWork(scope, dbContext, storagePersistence, settings);
```
rbev, line 19: Since you are passing the container reference into the unit of work it seems like you could just resolve everything inside it's own constructor

However I'd recommend instead passing it in as an (optional?) IDisposable since that's all it cares about. An added bonus here is that it also removes the depend on `Microsoft.Extensions.DI` and makes testing it not require a container.

rbev (approved):

src/ServiceControl.Persistence.EFCore/Implementation/UnitOfWork/EFIngestionUnitOfWorkFactory.cs
johnsimons: Good idea, I'll do this in the next PR

src/ServiceControl.Persistence.EFCore/Infrastructure/MessageBodyFileResult.cs
```diff
@@ -0,0 +1,11 @@
+namespace ServiceControl.Persistence.EFCore.Infrastructure;
+
+using System.IO;
+
+public class MessageBodyFileResult
```
warwickschroeder, line 5: Use the 'required' modifier for the props?

src/ServiceControl.Persistence.EFCore/Implementation/UnitOfWork/EFIngestionUnitOfWork.cs
```diff
@@ -1,21 +1,35 @@
...
 
-    public Task Complete(CancellationToken cancellationToken) =>
-        throw new NotImplementedException();
+    public Task Complete(CancellationToken cancellationToken) => dbContext.SaveChangesAsync(cancellationToken);
 
-    public void Dispose()
+    public async ValueTask DisposeAsync()
     {
-        // Nothing to dispose yet
+        await dbContext.DisposeAsync();
```
warwickschroeder, line 30: 'scope.dispose' on the next line should already take care of dbcontext

warwickschroeder (approved):

src/ServiceControl.Persistence.EFCore/Implementation/UnitOfWork/EFIngestionUnitOfWork.cs
rbev: If [this comment](https://github.com/Particular/ServiceControl/pull/5615#discussion_r3600976533) is implemented then you don't actually know that anymore

PR #5621, opened 2026-07-20
Fix flaky ASB tests

rbev:
This pins the az CLI version to address incompatibility with python 3.14 as seen in: https://github.com/Azure/azure-cli/issues/32980

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,6 +64,14 @@ jobs:
       - name: Smoke test PowerShell module import
         if: matrix.os-name == 'Windows'
         run: Import-Module ./deploy/PowerShellModules/Particular.ServiceControl.Management
+      # there is an issue with az cli and python 3.14, so for now we need to pin it
+      # once the issue is resolved it should be able to be re-floated
+      # https://github.com/Azure/azure-cli/issues/32980
+      - name: Pin Azure CLI
+        uses: azure/setup-azure-cli@v3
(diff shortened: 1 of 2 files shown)


johnsimons (approved):
I think this has to do for now.
But funny enough they say installing [via pip](https://github.com/Azure/azure-cli/issues/20476) is not supported either!

.github/workflows/ci.yml
```diff
@@ -64,6 +64,23 @@ jobs:
       - name: Smoke test PowerShell module import
         if: matrix.os-name == 'Windows'
         run: Import-Module ./deploy/PowerShellModules/Particular.ServiceControl.Management
+
+      # there is an issue with az cli and python 3.14, so for now we need to pin it
+      # once the issue is resolved it should be able to be re-floated
```
warwickschroeder, line 69: How would this be followed up on? Is there an eta, or can we create an issue to track it, and then resolve once fixed?

warwickschroeder (approved):

johnsimons:
It seems the github image is installing Python 3.12, not 3.14, see https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md, so this may come back!

.github/workflows/ci.yml
rbev: I created an issue on our board, we can create a full issue if the Microsoft changes take longer than we do.

rbev:
> It seems the github image is installing Python 3.12, not 3.14, see https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md, so this may come back!

From what I understand az cli ships with an embedded python (probably why pip installs are unsupported). I'm hoping they ship a compatible version and we can just remove this in a couple of weeks.

PR #5630, opened 2026-07-21
Run sql persister tests in CI

rbev:
Adds test categories for SQL server persistence into CI pipeline

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,20 +18,7 @@ jobs:
     strategy:
       matrix:
         os: [windows-latest, ubuntu-latest]
-        test-category: [ `
-          Default, 
-          SqlServer, 
-          SqlServerPersistence, 
-          AzureServiceBus, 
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -100,6 +107,14 @@ jobs:
           tag: ServiceControl
           registry-username: ${{ secrets.DOCKERHUB_USERNAME }}
           registry-password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Setup PostgreSQL persistence
```
johnsimons, line 110: I thought FTS is not an issue in postgreSQL?

johnsimons (approved):

.github/workflows/ci.yml
rbev: It's not - but these setup steps create the env var so the choice was to either

* install separately for transport/persistence runs
* install once, then re-write the env vars in an extra step

PR #5631, opened 2026-07-21
API endpoints for storing/retrieving license metadata

PhilBastian:
Also change max monthly usage to average monthly usage.


src/ServiceControl/Licensing/LicenseController.cs
```diff
@@ -44,29 +52,40 @@ public async Task<ActionResult<LicenseInfo>> License(bool refresh, string client
...
+        [Authorize(Policy = Permissions.ErrorLicensingManage)]
+        [HttpPost]
+        [Route("license/detailsUpload")]
+        public async Task UploadLicenseDetails([FromForm] IFormFile file, CancellationToken cancellationToken)
+        {
+            //perform date and license id checks
+            using var fileStream = file.OpenReadStream();
+            using var fileStreamReader = new StreamReader(fileStream);
+            var compressed = await fileStreamReader.ReadToEndAsync(cancellationToken);
+            var compressedData = Convert.FromBase64String(compressed);
```
mikeminutillo: We probably don't need the Base64 conversion when we aren't embedding details in the license file.

That means we could hook the BrotliStream directly to the fileStream.

It may be worth trying to pass that into the deserializer as well. That way we avoid creating big in-memory string objects.

mikeminutillo (approved):

PR #5632, opened 2026-07-21
Adding endpoint settings stores for Entity Framework

rbev:
* Implement `IQueueAddressStore` for EF persisters
* Implement `ITrialLicenseDataProvider` for EF persisters
* Implement `IEndpointSettingsStore` for EF persisters
* Move test-only error insertion into test project
* Improve test  coverage on `IQueueAddressStore`, `ITrialLicenseDataProvider`, `IEndpointSettingsStore`

### DIFF
--- a/src/ServiceControl.Persistence.EFCore/EntityConfigurations/EndpointSettingsConfiguration.cs
+++ b/src/ServiceControl.Persistence.EFCore/EntityConfigurations/EndpointSettingsConfiguration.cs
@@ -10,6 +10,6 @@ public class EndpointSettingsConfiguration : IEntityTypeConfiguration<EndpointSe
     {
         builder.ToTable("EndpointSettings");
         builder.HasKey(x => x.Name);
-        builder.Property(x => x.Name).HasMaxLength(200).IsRequired();
+        builder.Property(x => x.Name).HasMaxLength(ColumnLengths.ShortTextLength).IsRequired();
     }
 }
--- a/src/ServiceControl.Persistence.EFCore/EntityConfigurations/EndpointSettingsConfiguration.cs
+++ b/src/ServiceControl.Persistence.EFCore/EntityConfigurations/EndpointSettingsConfiguration.cs
@@ -4,11 +4,12 @@ using Entities;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Metadata.Builders;
 
-public class EndpointSettingsConfiguration :  IEntityTypeConfiguration<EndpointSettingsEntity>
+public class EndpointSettingsConfiguration : IEntityTypeConfiguration<EndpointSettingsEntity>
 {
     public void Configure(EntityTypeBuilder<EndpointSettingsEntity> builder)
     {
         builder.ToTable("EndpointSettings");
         builder.HasKey(x => x.Name);
+        builder.Property(x => x.Name).HasMaxLength(200).IsRequired();
     }
 }
--- a/src/ServiceControl.Persistence.EFCore/EntityConfigurations/EndpointSettingsConfiguration.cs
+++ b/src/ServiceControl.Persistence.EFCore/EntityConfigurations/EndpointSettingsConfiguration.cs
@@ -8,7 +8,6 @@ public class EndpointSettingsConfiguration : IEntityTypeConfiguration<EndpointSe
 {
     public void Configure(EntityTypeBuilder<EndpointSettingsEntity> builder)
     {
-        builder.ToTable("EndpointSettings");
         builder.HasKey(x => x.Name);
         builder.Property(x => x.Name).HasMaxLength(ColumnLengths.ShortTextLength).IsRequired();
     }
--- a/src/ServiceControl.Persistence.EFCore/Implementation/EndpointSettingsStore.cs
+++ b/src/ServiceControl.Persistence.EFCore/Implementation/EndpointSettingsStore.cs
@@ -25,20 +25,21 @@ public class EndpointSettingsStore(IServiceScopeFactory scopeFactory) : DataStor
                 await context.SaveChangesAsync(token);
                 return;
             }
-            catch (someexception e) when (condition)
+            catch (DbUpdateException)
             {
-                //this failed because of key conflict so it's a race!
+                //this probably failed because of key conflict so try again
             }
 
             await context.Entry(entity).ReloadAsync(token);
-            entity.TrackInstances = settings.TrackInstances;
-            await context.SaveChangesAsync(token);
         }
+
+        entity.TrackInstances = settings.TrackInstances;
+        await context.SaveChangesAsync(token);
     });
 
     public Task Delete(string name, CancellationToken cancellationToken) => ExecuteWithDbContext(async context =>
     {
-        context.KnownEndpoints.RemoveRange(context.KnownEndpoints.Where(x => x.Name == name));
+        context.EndpointSettings.RemoveRange(context.EndpointSettings.Where(x => x.Name == name));
         await context.SaveChangesAsync(cancellationToken);
     });
 }
--- a/src/ServiceControl.Persistence.EFCore/Implementation/EndpointSettingsStore.cs
+++ b/src/ServiceControl.Persistence.EFCore/Implementation/EndpointSettingsStore.cs
@@ -1,13 +1,44 @@
 namespace ServiceControl.Persistence.EFCore.Implementation;
 
-public class EndpointSettingsStore : IEndpointSettingsStore
+using Entities;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.DependencyInjection;
+
+public class EndpointSettingsStore(IServiceScopeFactory scopeFactory) : DataStoreBase(scopeFactory), IEndpointSettingsStore
 {
-    public IAsyncEnumerable<EndpointSettings> GetAllEndpointSettings(CancellationToken cancellationToken) =>
-        throw new NotImplementedException();
+    public IAsyncEnumerable<EndpointSettings> GetAllEndpointSettings(CancellationToken cancellationToken)
+        => ExecuteWithDbContext(context => context.EndpointSettings.Select(row => new EndpointSettings
+        {
+            Name = row.Name,
+            TrackInstances = row.TrackInstances
+        }).AsAsyncEnumerable(), cancellationToken);
 
-    public Task UpdateEndpointSettings(EndpointSettings settings, CancellationToken token) =>
-        throw new NotImplementedException();
+    public Task UpdateEndpointSettings(EndpointSettings settings, CancellationToken token) => ExecuteWithDbContext(async context =>
+    {
+        var entity = context.EndpointSettings.Find(settings.Name);
+        if (entity == null)
+        {
+            entity = new EndpointSettingsEntity() { Name = settings.Name, TrackInstances = settings.TrackInstances };
+            context.EndpointSettings.Add(entity);
+            try
+            {
+                await context.SaveChangesAsync(token);
+                return;
+            }
+            catch (someexception e) when (condition)
+            {
+                //this failed because of key conflict so it's a race!
+            }
 
-    public Task Delete(string name, CancellationToken cancellationToken) =>
-        throw new NotImplementedException();
+            await context.Entry(entity).ReloadAsync(token);
+            entity.TrackInstances = settings.TrackInstances;
+            await context.SaveChangesAsync(token);
+        }
+    });
+
+    public Task Delete(string name, CancellationToken cancellationToken) => ExecuteWithDbContext(async context =>
+    {
+        context.KnownEndpoints.RemoveRange(context.KnownEndpoints.Where(x => x.Name == name));
+        await context.SaveChangesAsync(cancellationToken);
+    });
 }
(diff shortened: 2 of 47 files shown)


johnsimons:
@rbev It may be good to do a rebase on master

src/ServiceControl.Persistence.EFCore/EntityConfigurations/EndpointSettingsConfiguration.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using Entities;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
+
+public class EndpointSettingsConfiguration : IEntityTypeConfiguration<EndpointSettingsEntity>
+{
+    public void Configure(EntityTypeBuilder<EndpointSettingsEntity> builder)
+    {
+        builder.ToTable("EndpointSettings");
```
johnsimons: We currently using conventionas for the table names

src/ServiceControl.Persistence.EFCore/EntityConfigurations/EndpointSettingsConfiguration.cs
```diff
@@ -0,0 +1,15 @@
...
+        builder.HasKey(x => x.Name);
+        builder.Property(x => x.Name).HasMaxLength(ColumnLengths.ShortTextLength).IsRequired();
+    }
```
johnsimons, line 13: Are you missing `TrackedInstances` definition here?

src/ServiceControl.Persistence.EFCore/Implementation/EndpointSettingsStore.cs
```diff
@@ -1,13 +1,45 @@
...
-        throw new NotImplementedException();
+        entity.TrackInstances = settings.TrackInstances;
+        await context.SaveChangesAsync(token);
+    });
 
-    public Task Delete(string name, CancellationToken cancellationToken) =>
-        throw new NotImplementedException();
+    public Task Delete(string name, CancellationToken cancellationToken) => ExecuteWithDbContext(async context =>
+    {
+        context.EndpointSettings.RemoveRange(context.EndpointSettings.Where(x => x.Name == name));
```
johnsimons: You made `Name` the id, whihc means you should be able to delete via that key? And there should only be one right?

src/ServiceControl.Persistence.EFCore/Implementation/EndpointSettingsStore.cs
```diff
@@ -1,13 +1,45 @@
...
+            entity = new EndpointSettingsEntity() { Name = settings.Name, TrackInstances = settings.TrackInstances };
+            context.EndpointSettings.Add(entity);
+            try
+            {
+                await context.SaveChangesAsync(token);
+                return;
+            }
+            catch (DbUpdateException)
+            {
+                //this probably failed because of key conflict so try again
```
johnsimons: Should you retry here?
See https://github.com/Particular/ServiceControl/blob/784e566f2b4ea2d5e9decfef2d94c13fbe0490fa/src/ServiceControl.Persistence.EFCore/Implementation/MonitoringDataStore.cs#L76-L86

src/ServiceControl.Persistence.EFCore/Implementation/EndpointSettingsStore.cs
```diff
@@ -1,13 +1,45 @@
...
+    public IAsyncEnumerable<EndpointSettings> GetAllEndpointSettings(CancellationToken cancellationToken)
+        => ExecuteWithDbContext(context => context.EndpointSettings.Select(row => new EndpointSettings
+        {
+            Name = row.Name,
+            TrackInstances = row.TrackInstances
+        }).AsAsyncEnumerable(), cancellationToken);
+
+    public Task UpdateEndpointSettings(EndpointSettings settings, CancellationToken token) => ExecuteWithDbContext(async context =>
+    {
+        var entity = context.EndpointSettings.Find(settings.Name);
```
johnsimons: We may want to discuss and agree on how to handle updates, becasue I did it ion a different style, see https://github.com/Particular/ServiceControl/blob/784e566f2b4ea2d5e9decfef2d94c13fbe0490fa/src/ServiceControl.Persistence.EFCore/Implementation/MonitoringDataStore.cs#L56

src/ServiceControl.Persistence.EFCore/Implementation/QueueAddressStore.cs
```diff
@@ -1,13 +1,32 @@
...
 using ServiceControl.Persistence.Infrastructure;
 
-public class QueueAddressStore : IQueueAddressStore
+public class QueueAddressStore(IServiceScopeFactory scopeFactory) : DataStoreBase(scopeFactory), IQueueAddressStore
 {
     public Task<QueryResult<IList<QueueAddress>>> GetAddresses(PagingInfo pagingInfo) =>
-        throw new NotImplementedException();
+        GetAddressesBySearchTerm(string.Empty, pagingInfo);
 
     public Task<QueryResult<IList<QueueAddress>>> GetAddressesBySearchTerm(string search, PagingInfo pagingInfo) =>
```
johnsimons: I noticed that there are no usages of the search action which is the thing that calls this method, https://github.com/Particular/ServiceControl/blob/784e566f2b4ea2d5e9decfef2d94c13fbe0490fa/src/ServiceControl/MessageFailures/Api/QueueAddressController.cs#L31
in either SP or SI!
Whihc makes me wonder if we should be indexing the name field at all!

src/ServiceControl.Persistence.EFCore/Implementation/QueueAddressStore.cs
```diff
@@ -1,13 +1,32 @@
...
-        throw new NotImplementedException();
-}
+        ExecuteWithDbContext(async context =>
+        {
+            var query = context.FailedMessages
+                .Where(fm => string.IsNullOrWhiteSpace(search) || fm.FailingEndpointAddress.StartsWith(search))
+                .Select(fm => new { fm.UniqueMessageId, fm.FailingEndpointAddress })
```
johnsimons: Why do we need to do a unique on `UniqueMessageId`, don't u just need a distinct on `FailingEndpointAddress`?
Also, we are doing a croup and a count here, but in SP , we are not using the count at all!
I think this migration needs to be rechecked all against what we currently support/need.

src/ServiceControl.Persistence.EFCore/Implementation/TrialLicenseDataProvider.cs
```diff
@@ -1,10 +1,23 @@
...
 
-    public Task StoreTrialEndDate(DateOnly trialEndDate, CancellationToken cancellationToken) =>
-        throw new NotImplementedException();
-}
+    public Task StoreTrialEndDate(DateOnly trialEndDate, CancellationToken cancellationToken)
+        => ExecuteWithDbContext(async context =>
+        {
+            var trialMetadata = await context.TrialMetadata.SingleAsync(t => t.Id == TrialMetadataEntity.TrialMetadataId, cancellationToken);
+            trialMetadata.TrialEndDate = trialEndDate;
+            await (Task)context.SaveChangesAsync(cancellationToken);
```
johnsimons: Wht not use `ExecuteUpdateAsync`, a single call vs 2 calls?

src/ServiceControl.Persistence.Tests/TrialLicenseDataProviderTests.cs
```diff
@@ -0,0 +1,33 @@
...
+using ServiceControl.Persistence;
+
+class TrialLicenseDataProviderTests : PersistenceTestBase
+{
+    [Test]
+    public async Task GetTrialEndDate_returns_null_by_default()
+    {
+        var trialLicenseDataProvider = ServiceProvider.GetRequiredService<ITrialLicenseDataProvider>();
+
+        var trialEndDate = await trialLicenseDataProvider.GetTrialEndDate(default);
```
johnsimons: This may be a discussion but, should we hardcode `default` or actually pass a test cancealtion token?

src/ServiceControl.Persistence.Tests/EndpointSettingsStoreTests.cs
```diff
@@ -0,0 +1,55 @@
...
+            Assert.That(settings, Has.Count.EqualTo(1));
+            Assert.That(settings.Single().Name, Is.EqualTo("Shipping"));
+            Assert.That(settings.Single().TrackInstances, Is.True);
+        }
+    }
+
+    async Task<IReadOnlyList<EndpointSettings>> GetAllEndpointSettings()
+    {
+        var settings = new List<EndpointSettings>();
+        await foreach (var setting in EndpointSettingsStore.GetAllEndpointSettings())
```
johnsimons: Could you just use `ToArrayAsync()`?

src/ServiceControl.Persistence.EFCore/Implementation/EndpointSettingsStore.cs
rbev: You still have to do a read/remove/save in EF, but yeah this is an artifact from before I made that change

src/ServiceControl.Persistence.EFCore/Implementation/QueueAddressStore.cs
rbev: Yeah you are right this was written before the final shape of `FailedMessages` was checked in, probably need to add an index but it should just be simplified to: 

```sql
select FailingEndpointAddress, count(UniqueMessageId) 
group by FailingEndpointAddress`
```

src/ServiceControl.Persistence.Tests/TrialLicenseDataProviderTests.cs
rbev: I'd lean towards putting timeouts on all the tests so we have one, but yes we should discuss.

src/ServiceControl.Persistence.EFCore/Implementation/EndpointSettingsStore.cs
johnsimons: not really @rbev https://learn.microsoft.com/en-us/ef/core/saving/execute-insert-update-delete#executedelete

rbev: Updated

src/ServiceControl.Persistence.EFCore/Implementation/EndpointSettingsStore.cs
rbev: I'll make sure this path is tested

src/ServiceControl.Persistence.Tests/TrialLicenseDataProviderTests.cs
johnsimons: @warwickschroeder @abparticular any thoughts?

src/ServiceControl.Persistence.EFCore/Implementation/EndpointSettingsStore.cs
rbev: I've double checked that this strategy definitely works just fine, I'm going to raise a separate PR to add an extension method to encapsulate it

src/ServiceControl.Persistence.EFCore/Implementation/QueueAddressStore.cs
```diff
@@ -1,13 +1,30 @@
...
+                .OrderBy(failuresByEndpoint => failuresByEndpoint.Key)
+                .Select(failuresByEndpoint => new QueueAddress
+                {
+                    PhysicalAddress = failuresByEndpoint.Key,
+                    FailedMessageCount = failuresByEndpoint.Count()
+                });
+
+            var items = await query.Skip(pagingInfo.Offset).Take(pagingInfo.PageSize).ToListAsync();
+
+            return new QueryResult<IList<QueueAddress>>(items, new QueryStatsInfo("", query.Count(), false));
```
rbev: This needs a real Etag generated

src/ServiceControl.Persistence.EFCore/Implementation/EndpointSettingsStore.cs
rbev: Updated with shared upsert extension method

src/ServiceControl.Persistence.EFCore/Implementation/QueueAddressStore.cs
rbev: Raised https://github.com/Particular/ServiceControl/pull/5657 to remove

src/ServiceControl.Persistence.EFCore/Implementation/EndpointSettingsStore.cs
rbev: Updated with shared extension method

src/ServiceControl.Persistence.EFCore/Implementation/QueueAddressStore.cs
rbev: done

src/ServiceControl.Persistence.Tests/TrialLicenseDataProviderTests.cs
rbev: done

src/ServiceControl.Persistence.EFCore/EntityConfigurations/EndpointSettingsConfiguration.cs
rbev: The defaults include all properties with default settings, you don't have to configure anything if those are acceptable. Note that this field is present in the migrations as you would expect.

johnsimons (approved):

warwickschroeder (approved):

PR #5638, opened 2026-07-23
EF persistence for failed messages and streamline known endpoints

johnsimons:
This change introduces the core EF Core schema and entities for storing failed messages, including their associated groups and retry information. It also sets up full-text search capabilities for failed message headers, bodies, and message types in both PostgreSQL and SQL Server.

The separate 'insert-only' table and background reconciler for KnownEndpoints have been removed, simplifying how known endpoints are managed. A new `MaxBodySizeToStore` setting is added to control message body storage.


src/ServiceControl.Persistence.EFCore/Abstractions/EFPersistenceConfigurationBase.cs
```diff
@@ -1,23 +1,29 @@
...
 {
     const string ConnectionStringKey = "Database/ConnectionString";
     const string CommandTimeoutKey = "Database/CommandTimeout";
     const string MessageBodyStoragePathKey = "MessageBody/StoragePath";
     const string MinBodySizeForCompressionKey = "MessageBody/MinCompressionSize";
+    const string MaxBodySizeToStoreKey = "MaxBodySizeToStore";
     const string ErrorRetentionPeriodKey = "ErrorRetentionPeriod";
     const string EnableFullTextSearchOnBodiesKey = "EnableFullTextSearchOnBodies";
 
+    const int MaxBodySizeToStoreDefault = 102400; //100 kb
```
warwickschroeder: Should we extract this magic number into a constant? Its also used to set the default in `EFPersistenceSettings`.

src/ServiceControl.Persistence.EFCore/Abstractions/EFPersistenceConfigurationBase.cs
```diff
@@ -1,23 +1,29 @@
...
+
     public PersistenceSettings CreateSettings(SettingsRootNamespace settingsRootNamespace)
     {
         var settings = CreateSettings(GetRequiredSetting<string>(settingsRootNamespace, ConnectionStringKey));
 
         settings.CommandTimeout = SettingsReader.Read(settingsRootNamespace, CommandTimeoutKey, 30);
         settings.MessageBodyStoragePath = SettingsReader.Read<string>(settingsRootNamespace, MessageBodyStoragePathKey);
         settings.MinBodySizeForCompression = SettingsReader.Read(settingsRootNamespace, MinBodySizeForCompressionKey, 4096);
```
warwickschroeder: `4096` is also used in `EFPersistenceSettings` - extract the default to a constant? Same with CommandTimeout

docs/error-ingestion-design.md
```diff
@@ -0,0 +1,186 @@
...
+`FailedMessages` table. There is deliberately **no attempts history table**. A message that fails
+repeatedly keeps a single row that records:
+
+- the **last** processing attempt (every denormalized and payload column comes from it),
+- the **number of distinct attempts** (`NumberOfProcessingAttempts`),
+- the **failure window** (`FirstTimeOfFailure`, `LastTimeOfFailure`).
+
+This is a deliberate difference from the document-database persister, which retained an array of
+attempts. The read side only ever consumed the last attempt plus the count, so storing the full
+history earned nothing and cost write amplification. One consequence: `NumberOfProcessingAttempts`
```
rbev: Rephrase this as being an improvement - raven's implementation would stop counting at 10 this one will show the true number

src/ServiceControl.Persistence.EFCore/EntityConfigurations/FailedMessageConfiguration.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        builder.Property(e => e.Status).IsRequired();
+        builder.Property(e => e.StatusChangedAt).IsRequired();
+        builder.Property(e => e.LastModified).IsRequired();
+        builder.Property(e => e.NumberOfProcessingAttempts).IsRequired();
+        builder.Property(e => e.FirstTimeOfFailure).IsRequired();
+        builder.Property(e => e.LastTimeOfFailure).IsRequired();
+        builder.Property(e => e.LastAttemptedAt).IsRequired();
+
+        builder.Property(e => e.MessageId).HasMaxLength(ShortTextLength);
```
warwickschroeder: Do the KnownEndpoints have a MaxLength? If not they should also be given one

docs/error-ingestion-design.md
```diff
@@ -0,0 +1,186 @@
...
+
+The instance runs a single ingestion loop today, but the write path is built so that multiple
+instances could ingest against the same database (for example to scale out under load). That means
+two transactions can try to write the **same** `UniqueMessageId` at the same time. Two hazards
+follow:
+
+- **Insert races.** Both writers see the row as absent and both insert, colliding on the primary
+  key.
+- **Lost updates.** A read-modify-write done in application code (read the row, add to the count,
+  write it back) can lose one writer's increment.
```
rbev: Doesn't EF have optimistic concurrency that would roll back if another writer updated the row? i.e. this isn't actually a problem that forces you to not use the EF write abstractions

src/ServiceControl.Persistence.EFCore.SqlServer/FullTextSearchSql.cs
```diff
@@ -0,0 +1,52 @@
...
+/// Server, so it is applied by the AddFullTextSearch migration. The statements live here, and not
+/// in the migration itself, so that regenerating the migrations with the dotnet-ef CLI only costs
+/// a one line migration body.
+/// </summary>
+static class FullTextSearchSql
+{
+    const string CatalogName = "ServiceControlFullTextCatalog";
+
+    // Both statements are guarded: an instance without Full-Text Search installed still migrates,
+    // it just has no full text index. They also cannot run inside a transaction, so the migration
```
rbev, line 14: do we actually need to support instances without FTS?
If so we probably need to surface that as an error and have a way to recreate this index so that there is an upgrade path.

src/ServiceControl.Persistence.EFCore/Abstractions/BasePersistence.cs
```diff
@@ -32,6 +32,9 @@ protected static void RegisterDataStores(IServiceCollection services)
         services.AddSingleton<IExternalIntegrationRequestsDataStore>(p => p.GetRequiredService<ExternalIntegrationRequestsDataStore>());
         services.AddHostedService(p => p.GetRequiredService<ExternalIntegrationRequestsDataStore>());
 
+        services.AddSingleton<RetentionSweeper>();
+        services.AddHostedService(p => p.GetRequiredService<RetentionSweeper>());
```
rbev: `AddHostedService<T>` [adds the singleton for you](https://github.com/dotnet-architecture/Messaging/blob/5c337696349679136ebdd6b6fd425535fa78b25a/src/Microsoft.Extensions.Hosting/ServiceCollectionExtensions.cs#L10)

rbev (approved):
Looks good

src/ServiceControl.Persistence.EFCore/Abstractions/BasePersistence.cs
johnsimons: @rbev But it is registered as `IHostedService` correct?

src/ServiceControl.Persistence.EFCore.SqlServer/FullTextSearchSql.cs
johnsimons: I am not sure, I remember as part of the spikes we allowed customers to configure it.
@warwickschroeder any thoughts?

docs/error-ingestion-design.md
```diff
@@ -0,0 +1,186 @@
...
+
+- text at or under the cap: stored inline in `BodyText`, nothing external.
+- text over the cap: the full body goes to external storage, and a **search prefix** of at most
+  the cap, cut on a valid UTF-8 boundary, is kept inline in `BodyText` so search still works.
+- binary, or not strictly UTF-8 decodable, or containing a NUL byte: external storage only,
+  `BodyText` is null, regardless of size.
+
+When `BodyStoredExternally` is true the external copy is authoritative and `BodyText` is a
+search aid only; it must never be served as the body. `BodySize` is always the true original
+size. External writes happen before the row that points at them is committed.
```
warwickschroeder, line 60: For storage cost reasons, could we give the option to only store in cheap external storage and not in the DB at all? This would mean giving up Full Text Search, but maybe thats less important for error messages and more important for auditing, as auditing would typically have alot more messages to search through.

docs/error-ingestion-design.md
johnsimons: Rephrased

docs/error-ingestion-design.md
johnsimons: Rephrased

src/ServiceControl.Persistence.EFCore.SqlServer/FullTextSearchSql.cs
johnsimons: We can make a decision, later on this?
Without FTS, they would get no search at all.
I don't think we want to support `LIKE %blah%` !
From our spikes FTS was not an issue at all, so i think we can make it a required thing.

src/ServiceControl.Persistence.EFCore/Abstractions/BasePersistence.cs
rbev: That's correct, but there are currently no places where it gets resolved as itself from the container.

warwickschroeder (commented):
Review so far. Not finished yet. Will review the rest tonight/tomorrow.

src/ServiceControl.Persistence.EFCore/Abstractions/BasePersistence.cs
johnsimons: I'll have a look

src/ServiceControl.Persistence.EFCore.SqlServer/FullTextSearchSql.cs
warwickschroeder: Not having it as an option gives the user the option to not having to store the body in the DB and only use cheap external storage. That may be worth it?

docs/error-ingestion-design.md
johnsimons: But that would be weird, given that then we could search audits but not errors, and specially now that we have an agregate list of messages (All messages).
It is possible that we are trying to optimise things too early.

src/ServiceControl.Persistence.EFCore/Abstractions/BasePersistence.cs
johnsimons: fixed

src/ServiceControl.Persistence.EFCore/Abstractions/EFPersistenceConfigurationBase.cs
johnsimons: done

src/ServiceControl.Persistence.EFCore/EntityConfigurations/FailedMessageConfiguration.cs
johnsimons: fixed

src/ServiceControl.Persistence.EFCore/Abstractions/EFPersistenceConfigurationBase.cs
johnsimons: done

docs/error-ingestion-design.md
warwickschroeder: Yeh. Maybe that option is a bit premature.

src/ServiceControl.Persistence.Tests.RavenDB/ServiceControl.Persistence.Tests.RavenDB.csproj
```diff
@@ -25,6 +25,9 @@
 
   <ItemGroup>
     <Compile Include="..\ServiceControl.Persistence.Tests\**\*.cs" LinkBase="Shared" />
+
+    <!-- Tests of behaviour specific to the EF Core persisters, shared by their two test projects only -->
+    <Compile Remove="..\ServiceControl.Persistence.Tests\EFCore\**\*.cs" />
```
DavidBoike, line 30: * I'm guessing this is why the change in https://github.com/Particular/repo-integrity-action/pull/153 was made, which I don't disagree with, but does this indicate that the tests involved should be further split into different sources so a Compile Include+Remove pattern is not needed?

johnsimons: @DavidBoike the reason to go with this pattern is so we can share tests between both PostgreSQL and SqlServer tests.
What do you mean by further split?

PR #5646, opened 2026-07-24
Implement configurable message body storage

johnsimons:
This change introduces a pluggable message body storage mechanism, replacing the previous fake implementation. It enables configuring the storage type (e.g., FileSystem, Azure Blob, S3) via settings.


src/ServiceControl.Persistence.EFCore/Abstractions/EFPersisterSettings.cs
```diff
@@ -11,7 +11,19 @@ public abstract class EFPersisterSettings : PersistenceSettings
     public required string ConnectionString { get; set; }
     public int CommandTimeout { get; set; } = DefaultCommandTimeout;
     public TimeSpan ErrorRetentionPeriod { get; set; }
+    public BodyStorageType BodyStorageType { get; set; }
     public string? MessageBodyStoragePath { get; set; }
+    public string? AzureBlobConnectionString { get; set; }
+    public string? AzureBlobServiceUri { get; set; }
+    public string? AzureBlobManagedIdentityClientId { get; set; }
+    public string? AzureBlobAuthorityHost { get; set; }
+    public string AzureBlobContainerName { get; set; } = "error-bodies";
```
warwickschroeder: nit: maybe group some of these product related settings into their own class?

src/ServiceControl.Persistence.Tests/EFCore/AzureBlobBodyStorageTests.cs
```diff
@@ -0,0 +1,142 @@
...
+    AzuriteContainer azurite;
+    AzureBlobBodyStoragePersistence store;
+
+    [OneTimeSetUp]
+    public async Task StartAzurite()
+    {
+        azurite = new AzuriteBuilder("mcr.microsoft.com/azure-storage/azurite:latest")
+            // We need this because the Azurite image is not yet compatible with the latest Azure SDK, 
+            // see https://github.com/Azure/Azurite/issues/2623
+            .WithCommand("--skipApiVersionCheck")
```
warwickschroeder, line 27: how do we track this, or manage this in the future? I dont think we want to leave it like this and it may happen again. Can we pin the azurite container tag?

src/ServiceControl.Persistence.EFCore/Implementation/BodyStorage/FileSystemBodyStoragePersistence.cs
```diff
@@ -0,0 +1,146 @@
...
+using System.IO.Compression;
+using System.Text;
+using ServiceControl.Persistence.EFCore.Abstractions;
+using ServiceControl.Persistence.EFCore.Infrastructure;
+
+// Bodies are immutable and keyed by bodyId alone, so a re-failure resolves to the same file and an
+// existing one is left untouched. 
+public class FileSystemBodyStoragePersistence(EFPersisterSettings settings) : IBodyStoragePersistence
+{
+    const int FormatVersion = 1;
```
warwickschroeder, line 17: Its not clear to me how this is meant to be used. So if we change the body format, we increase this version number. Does that then mean users with error messages in the queue and body in external storage, retrieving these messages will just throw? 

1. What's does that experience look like for the user?
2. Maybe add some comments to explain it? Making it clear what the impact of updating this actually is.
3. We should probably create a gate of some sort, maybe a test, to ensure this version is incremented.

Applies to the other seams as well

src/ServiceControl.Persistence.EFCore/Implementation/BodyStorage/AzureBlobBodyStoragePersistence.cs
```diff
@@ -0,0 +1,98 @@
...
+        var blob = container.GetBlobClient(bodyId);
+
+        try
+        {
+            var content = (await blob.DownloadStreamingAsync(cancellationToken: cancellationToken)).Value;
+            try
+            {
+                var metadata = content.Details.Metadata;
+
+                if (metadata.TryGetValue("FormatVersion", out var version) && version != FormatVersion)
```
warwickschroeder, line 65: Seems a missing `FormatVersion` is tolerated in Azure and AWS, but not file storage. Is there a reason for this?

src/ServiceControl.Persistence.EFCore/Implementation/FileSystemBodyStoragePersistence.cs
```diff
@@ -0,0 +1,146 @@
...
+            }
+
+            var contentType = reader.ReadString();
+            var bodySize = reader.ReadInt32();
+            var isCompressed = reader.ReadBoolean();
+
+            // The returned stream owns fileStream and is disposed by the caller.
+            Stream bodyStream = isCompressed
+                ? new BrotliStream(fileStream, CompressionMode.Decompress, leaveOpen: false)
+                : fileStream;
```
warwickschroeder: AI found the below:

When we save a body to disk we write a small header first and then the body. On the way back out we read the header and hand the rest of the file to ASP.NET to send to the browser.

The problem is that we hand over the file handle itself. ASP.NET asks it how big it is, and it answers with the size of the *whole file*, header included. But we've already read past the header, so only the body bytes actually get sent. We promise 38 bytes and send 4. Kestrel spots the mismatch and kills the response, so the download fails.

```suggestion
            Stream bodyStream = isCompressed
                ? new ExpectedLengthStream(new BrotliStream(fileStream, CompressionMode.Decompress, leaveOpen: false), bodySize)
                : new ExpectedLengthStream(fileStream, bodySize);
```

src/ServiceControl.Persistence.EFCore/Abstractions/EFPersistenceConfigurationBase.cs
```diff
@@ -19,19 +31,115 @@ public PersistenceSettings CreateSettings(SettingsRootNamespace settingsRootName
...
+        }
+
+        return raw;
+    }
+
+    static BodyStorageType ReadBodyStorageType(SettingsRootNamespace settingsRootNamespace)
+    {
+        var raw = SettingsReader.Read<string>(settingsRootNamespace, BodyStorageTypeKey);
+
+        if (string.IsNullOrWhiteSpace(raw))
```
rbev, line 150: should this error or default to filesystem in some predictable location instead of failing?

src/ServiceControl.Persistence.EFCore/Abstractions/EFPersistenceConfigurationBase.cs
```diff
@@ -19,19 +31,115 @@ public PersistenceSettings CreateSettings(SettingsRootNamespace settingsRootName
...
+        if (hasConnectionString == hasServiceUri)
+        {
+            throw new Exception($"Azure Blob body storage requires exactly one of {AzureConnectionStringKey} (shared key / SAS) or {AzureServiceUriKey} (managed identity). {(hasConnectionString ? "Both were set." : "Neither was set.")}");
+        }
+
+        settings.AzureBlobConnectionString = hasConnectionString ? connectionString : null;
+        settings.AzureBlobServiceUri = hasServiceUri ? serviceUri : null;
+        settings.AzureBlobManagedIdentityClientId = SettingsReader.Read<string>(settingsRootNamespace, AzureManagedIdentityClientIdKey);
+        settings.AzureBlobAuthorityHost = ReadAzureAuthorityHost(settingsRootNamespace);
+        settings.AzureBlobContainerName = SettingsReader.Read(settingsRootNamespace, AzureContainerNameKey, settings.AzureBlobContainerName);
```
rbev: this value should probably be mandatory if you are using Azure Blob

src/ServiceControl.Persistence.EFCore/Implementation/BodyStorage/FileSystemBodyStoragePersistence.cs
johnsimons: The version marker is insurance rather than something we plan to use. It exists so that if the layout ever does change, a body written under the old one fails loudly instead of being silently misread.

To your first question: if we bumped it today with no other change, every body already in storage would throw out of `ReadBody` and the message body panel would return a 500 until those messages aged past `ErrorRetentionPeriod`. That is exactly why the rule has to be that we never bump it on its own. Any future format change has to ship with read support for every version still within the retention window, so the bump becomes additive and no existing body breaks.

On the gate: agreed in principle, but there is nothing to gate yet while there is only one version. I would rather add the pinned-layout test at the point we actually introduce a v2, when we will know what shape the dispatch needs to take. Happy to add a comment stating the contract now if that is useful.

src/ServiceControl.Persistence.EFCore/Implementation/BodyStorage/AzureBlobBodyStoragePersistence.cs
johnsimons: Not a deliberate difference. The file header is positional so the field is present by construction and absence is unrepresentable, whereas the cloud stores do a dictionary lookup that can miss, and the miss path just falls through. In practice it is not load-bearing: we always write the metadata, and this ships new so there are no pre-existing objects without it.

Happy to tighten Azure and S3 to require it if you would prefer them consistent, it is a two-line change.

src/ServiceControl.Persistence.EFCore/Implementation/BodyCompression.cs
```diff
@@ -0,0 +1,36 @@
...
+static class BodyCompression
+{
+    // Returns the Brotli-compressed bytes, or null when compression fails and the caller should
+    // store the body uncompressed.
+    public static byte[]? TryCompress(ReadOnlyMemory<byte> body)
+    {
+        var buffer = ArrayPool<byte>.Shared.Rent(BrotliEncoder.GetMaxCompressedLength(body.Length));
+        try
+        {
+            return BrotliEncoder.TryCompress(body.Span, buffer, out var written, quality: 1, window: 22)
```
rbev: Should there be documentation around the decision to use Brotli over something else?

src/ServiceControl.Persistence.EFCore/Implementation/BodyStorage.cs
```diff
@@ -1,9 +1,97 @@
...
+using System.Text;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.DependencyInjection;
 using ServiceControl.Operations.BodyStorage;
+using ServiceControl.Persistence.EFCore.DbContexts;
+using ServiceControl.Persistence.EFCore.Entities;
+using ServiceControl.Persistence.EFCore.Infrastructure;
 
-public class BodyStorage : IBodyStorage
+// A body is stored inline in BodyText (small text) or in external storage (binary, or large text
```
rbev: should this be `///` doc comments?

src/ServiceControl.Persistence.Tests/EFCore/AzureBlobBodyStorageTests.cs
johnsimons: I have pinned it and added better comment

src/ServiceControl.Persistence.EFCore/Implementation/FileSystemBodyStoragePersistence.cs
rbev: Would it be better to store the metadata beside the body which I believe we do for Learning Transport, or will that bloat the disk space and operations too much?

src/ServiceControl.Persistence.EFCore/Implementation/BodyStorage/FileSystemBodyStoragePersistence.cs
```diff
@@ -0,0 +1,146 @@
...
+                    writer.Write(FormatVersion);
+                    writer.Write(contentType);
+                    writer.Write(body.Length);
+                    writer.Write(shouldCompress);
+                    writer.Flush();
+                }
+
+                if (shouldCompress)
+                {
+                    var brotliStream = new BrotliStream(fileStream, CompressionLevel.Fastest, leaveOpen: true);
```
rbev, line 51: Should the Brotli part of this be using the BodyCompression class?

src/ServiceControl.Persistence.EFCore/Implementation/BodyStorage/FileSystemBodyStoragePersistence.cs
```diff
@@ -0,0 +1,146 @@
...
+            {
+                throw new InvalidOperationException($"Unsupported body file format version {formatVersion} for {bodyId}.");
+            }
+
+            var contentType = reader.ReadString();
+            var bodySize = reader.ReadInt32();
+            var isCompressed = reader.ReadBoolean();
+
+            // The returned stream owns fileStream and is disposed by the caller.
+            Stream bodyStream = isCompressed
```
rbev, line 112: Should the Brotli part of this be using the BodyCompression class?

src/ServiceControl.Persistence.EFCore/Implementation/BodyStorage/S3BodyStoragePersistence.cs
```diff
@@ -0,0 +1,113 @@
...
+                if (!string.IsNullOrEmpty(version) && version != FormatVersion)
+                {
+                    throw new InvalidOperationException($"Unsupported object format version {version} for {bodyId}.");
+                }
+
+                var bodySize = int.TryParse(metadata["body-size"], out var size) ? size : 0;
+                var isCompressed = bool.TryParse(metadata["is-compressed"], out var compressed) && compressed;
+                var contentType = response.Headers.ContentType ?? "application/octet-stream";
+                Stream stream = isCompressed
+                    ? new ExpectedLengthStream(new BrotliStream(response.ResponseStream, CompressionMode.Decompress), bodySize)
```
rbev, line 74: Should the Brotli part of this be using the BodyCompression class?

src/ServiceControl.Persistence.EFCore/Implementation/BodyStorage/S3ClientFactory.cs
```diff
@@ -0,0 +1,39 @@
...
+{
+    public static IAmazonS3 Create(EFPersisterSettings settings)
+    {
+        var config = new AmazonS3Config();
+
+        var serviceUrl = settings.S3ServiceUrl;
+        if (!string.IsNullOrEmpty(serviceUrl))
+        {
+            config.ServiceURL = serviceUrl;
+            config.ForcePathStyle = true; // Required for S3-compatible endpoints (MinIO, LocalStack).
```
rbev, line 18: What are the implications of this option when not on the non-AWS stacks?

src/ServiceControl.Persistence.EFCore/Abstractions/EFPersisterSettings.cs
johnsimons: done

src/ServiceControl.Persistence.Tests/EFCore/AzureBlobBodyStorageTests.cs
johnsimons: done

src/ServiceControl.Persistence.EFCore/Implementation/FileSystemBodyStoragePersistence.cs
johnsimons: fixed and added test

rbev (approved):

warwickschroeder (approved):

src/ServiceControl.Persistence.EFCore/Implementation/BodyCompression.cs
johnsimons: Is there something else that we should use instead?
Does it matter?

src/ServiceControl.Persistence.EFCore/Implementation/FileSystemBodyStoragePersistence.cs
johnsimons: @rbev that would mean 2 reads instead of just one, not sure if it is worth it.
It also means compression needs to be done 2x.

src/ServiceControl.Persistence.EFCore/Implementation/BodyStorage/FileSystemBodyStoragePersistence.cs
johnsimons: No, I tried that, and trying to reuse that part of the code got too complex and therefore opted this way.
The complexity is becasue the filesystem has the file has the cache, so there is no in memory data.

src/ServiceControl.Persistence.EFCore/Abstractions/EFPersistenceConfigurationBase.cs
johnsimons: There's no defensible default path that holds across SCMU installs and containers, so any default we picked would be wrong somewhere. More importantly, silently writing bodies to local disk on an instance the operator intended to point at blob storage is a worse failure than refusing to start.
So yes it is a deliberate choice.

src/ServiceControl.Persistence.EFCore/Implementation/BodyCompression.cs
johnsimons: Anyway, I have added more docs about it.

src/ServiceControl.Persistence.EFCore/Implementation/BodyStorage.cs
johnsimons: fixed

src/ServiceControl.Persistence.EFCore/Implementation/BodyStorage/S3ClientFactory.cs
johnsimons: It's only set when a custom ServiceUrl is configured, so it never applies against real AWS, where the SDK uses virtual-host-style addressing. The flag is there for S3-compatible endpoints like LocalStack and MinIO, which can't do virtual-host addressing without wildcard DNS.

src/ServiceControl.Persistence.EFCore/Abstractions/EFPersistenceConfigurationBase.cs
johnsimons: For this setting we actually have a default of "error-bodies", so that is why we left it as not required

PR #5649, opened 2026-07-24
Move all ravendb specific ID generation out of the shared project

rbev:
* Moved singleton document ids into their respective RavenDB repositories
* Moved Collection name constants into the RavenDB repositories
* Moved ID formatters into the RavenDB repositories
* Removed some unused code


src/ServiceControl.Persistence.RavenDB/Editing/NotificationsManager.cs
```diff
@@ -7,13 +7,14 @@
 
     class NotificationsManager(IAsyncDocumentSession session) : AbstractSessionManager(session), INotificationsManager
     {
+        const string SingleDocumentId = "NotificationsSettings/All";
```
johnsimons, line 10: Should these hardcoded ids be something like?

```suggestion
        const string SingleDocumentId = $"{nameof(NotificationsSettings)}/All";
```

src/ServiceControl.Persistence.RavenDB/ErrorMessagesDataStore.cs
```diff
@@ -172,6 +173,7 @@ public async Task<FailedMessage[]> FailedMessagesFetch(Guid[] ids)
         public async Task StoreFailedErrorImport(FailedErrorImport failure)
         {
             using var session = await sessionProvider.OpenSession();
+            failure.Id = MakeDocumentId(failure.Id);
```
johnsimons: This looks weird

src/ServiceControl.Persistence/Recoverability/ReclassifyErrorSettings.cs
johnsimons, line 1: No longer needed?

src/ServiceControl.Persistence/UnitOfWork/UnitOfWorkServiceCollectionExtensions.cs
```diff
@@ -6,31 +6,10 @@
...
         /// </summary>
         public static void AddUnitOfWorkFactory<T>(this IServiceCollection serviceCollection)
             where T : class, IIngestionUnitOfWorkFactory
             => serviceCollection.AddSingleton<IIngestionUnitOfWorkFactory, T>();
-
-        /// <summary>
-        /// Register an ingestion unit of work factory that only implements some of the components.
-        /// The rest will be handled by RavenDb.
-        /// </summary>
-        public static void AddPartialUnitOfWorkFactory<T>(this IServiceCollection serviceCollection)
```
johnsimons, line 19: Not used?

src/ServiceControl/Operations/ErrorIngestionFaultPolicy.cs
```diff
@@ -57,7 +57,7 @@ async Task Handle(ErrorContext errorContext, CancellationToken cancellationToken
                     Body = errorContext.Body.ToArray()
                 },
                 ExceptionInfo = errorContext.Exception.ToFriendlyString(),
-                Id = FailedErrorImport.MakeDocumentId(Guid.NewGuid())
+                Id = Guid.NewGuid().ToString()
```
johnsimons: If things like this end-up being indexed fields, would it make sense to use the new `Guid. CreateVersion7()` because it is sequential?

johnsimons (approved):
Looks reasonable to me.
I have added a few comments along the way.

src/ServiceControl.Persistence.RavenDB/ErrorMessagesDataStore.cs
rbev: Yeah this is the only one that isn't trivial. Would it read better if i did this?

```cs
if (!failure.Id.StartsWith("Collection/")) {
   failure.Id = MakeDocumentId(failure.Id);
}
```

src/ServiceControl.Persistence.RavenDB/Editing/NotificationsManager.cs
rbev: I don't think so.

Those types are defined in the persistence seam's API, if they get tied to database names in a specific persistence then it opens up potential pain if the typename needs to change for some reason.

src/ServiceControl.Persistence/Recoverability/ReclassifyErrorSettings.cs
rbev: I couldn't find any references it it anywhere except old PRs

src/ServiceControl.Persistence.RavenDB/ErrorMessagesDataStore.cs
johnsimons: Maybe an explanation of why we need to do this.

src/ServiceControl.Persistence.RavenDB/ErrorMessagesDataStore.cs
```diff
@@ -172,6 +173,13 @@ public async Task<FailedMessage[]> FailedMessagesFetch(Guid[] ids)
         public async Task StoreFailedErrorImport(FailedErrorImport failure)
         {
             using var session = await sessionProvider.OpenSession();
+            // This object's ID is generated externally, but is not in the RavenDB format
+            // Check that's true to make sure that if it already is that it doesn't get double-formatted
+            if (!failure.Id.StartsWith(CollectionName))
+            {
+                failure.Id = MakeDocumentId(failure.Id);
+            }
+            failure.Id = MakeDocumentId(failure.Id);
```
warwickschroeder:
```suggestion
```

warwickschroeder (approved):

PR #5653, opened 2026-07-27
Create upsert utility extension method

rbev:
To avoid a lot of repeated implementations when upserting low-volume records create a central extension method that encapsulates the read/insert/error/retry logic.


johnsimons (approved):
LGTM

PR #5655, opened 2026-07-28
Event log persistence for the SQL Server and PostgreSQL persisters

warwickschroeder:
The event log is the list ServicePulse shows on its **Events** page. It worked on RavenDB and threw `NotImplementedException` on both EF Core persisters, so nothing could be written or read. This implements it.

Both EF persisters still log *"is still under development and is not ready for use"* at startup, so no customer is affected today.

## What changed

- **`EventLogItemEntity`, its configuration, and one migration per provider.** `EventLogDataStore` now implements both methods.
- **The write moved to the interface that owns it.** `AuditEventLogWriter` called `IErrorMessageDataStore.StoreEventLogItem` while `IEventLogDataStore.Add` sat unused. Now it calls `Add`, and the stray member is gone from `IErrorMessageDataStore` and both implementations.
- **Event log items expire on the EF side too.** `RetentionSweeper` sweeps them on its existing hourly timer against a separate `EventsRetentionPeriod`, independent of `ErrorRetentionPeriod`.
- **Conditional GET.** `GetEventLogItems` takes an optional `knownVersion`. On a match the persister skips the page query and `EventLogApiController` answers `304`, with `ETag` and `Total-Count` still sent.
- **`knownVersion` drops a query per poll on the EF persisters.** A read is three queries: count, newest `RaisedAt`, then the page. When the caller's version matches, the page query is skipped and the two that remain are satisfied by `IX_EventLogItems_RaisedAt_Id` without reading row data. ServicePulse polls this every five seconds, so on an idle instance that is the common case rather than the exception. RavenDB saves no database work, because its version is the query's own `ResultEtag` and does not exist until the query has run; there the saving is only the serialised response body.
- **`IEventLogDataStore` now documents retention**, which was previously invisible at the seam despite each persister enforcing it differently.

## Tests

21 new tests plus 3 added to `RetentionSweepTests`, and one acceptance test for the `304`. The 14 in `EventLogDataStoreTests` are the valuable ones: they live in the shared root of `ServiceControl.Persistence.Tests`, so they run as conformance tests against RavenDB, SQL Server and PostgreSQL. Round-tripping, ordering, paging boundaries and all four `knownVersion` cases have to behave identically on all three.


docs/testing.md
```diff
@@ -33,6 +33,28 @@ Transport tests are done by executing the transport test suite for each transpor
...
+Register both sources once, from an **elevated** PowerShell prompt:
+
+```powershell
+[System.Diagnostics.EventLog]::CreateEventSource('ServiceControl', 'Application')
+[System.Diagnostics.EventLog]::CreateEventSource('ServiceControl.Audit', 'Application')
+```
+
+Use the .NET API as shown rather than `New-EventLog`, which is not available in PowerShell 7. Afterwards the tests run normally without elevation, because the lookup finds the registered source before it needs to read `Security`.
+
+Despite mentioning "event logs", this is unrelated to ServiceControl's own event log feature.
```
johnsimons: This does not seem to be related to this PR?

src/ServiceControl.AcceptanceTests/EventLogs/When_the_event_log_is_polled_with_an_etag.cs
johnsimons, line 1: this is a production issue correct?

src/ServiceControl.Persistence.EFCore/Entities/EventLogItemEntity.cs
```diff
@@ -0,0 +1,24 @@
+namespace ServiceControl.Persistence.EFCore.Entities;
+
+using ServiceControl.EventLog;
+
+public class EventLogItemEntity
+{
+    public long Id { get; set; }
+
+    // The API-visible identity, assigned by EventLogMappingDefinition as
+    // "EventLogItem/{Category}/{EventType}/{guid}"
```
johnsimons: This feels a lot like Raven Ids!

src/ServiceControl.Persistence.EFCore/Infrastructure/RetentionSweeper.cs
```diff
@@ -57,6 +59,52 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
+                .Take(BatchSize)
+                .Select(eventLogItem => eventLogItem.Id)
+                .ToListAsync(cancellationToken);
+
+            if (expired.Count == 0)
+            {
+                break;
+            }
+
+            await dbContext.EventLogItems
```
rbev: Instead of query/delete you can just call .ExecuteDelete() instead of `.ToListAsync()`, then use the returned deleted row count to exit the loop.

src/ServiceControl.Persistence.EFCore/Entities/EventLogItemEntity.cs
```diff
@@ -0,0 +1,24 @@
...
+    public required string EventLogItemId { get; set; }
```
johnsimons: Do we ever retrieve a event log entry by id?
If we don't, does this even need to be a string or assigned, could it be a long auto assigned?

src/ServiceControl.Persistence.EFCore/Entities/EventLogItemEntity.cs
```diff
@@ -0,0 +1,24 @@
...
+
+    public required string Description { get; set; }
+
+    public Severity Severity { get; set; }
+
+    public DateTime RaisedAt { get; set; }
+
+    public List<string> RelatedTo { get; set; } = [];
```
johnsimons, line 17: is EF smart to serialize this as xml/json?

src/ServiceControl.Persistence.RavenDB/EventLogDataStore.cs
```diff
@@ -6,16 +6,22 @@
...
+    class EventLogDataStore(IRavenSessionProvider sessionProvider, ExpirationManager expirationManager) : IEventLogDataStore
     {
         public async Task Add(EventLogItem logItem)
         {
             using var session = await sessionProvider.OpenSession();
             await session.StoreAsync(logItem);
+
+            // Retention on RavenDB is per-document expiry metadata stamped at write time, not a
+            // sweep. It has to be set here, on the only write path, or items never expire.
+            expirationManager.EnableExpiration(session, logItem);
```
rbev, line 24: This is modifying RavenDB behaviour should it be lifted out into it's own PR?

rbev (approved):

src/ServiceControl.Persistence.EFCore/EntityConfigurations/EventLogItemConfiguration.cs
```diff
@@ -0,0 +1,29 @@
...
+class EventLogItemConfiguration : IEntityTypeConfiguration<EventLogItemEntity>
+{
+    public void Configure(EntityTypeBuilder<EventLogItemEntity> builder)
+    {
+        builder.ToTable("EventLogItems");
+        builder.HasKey(e => e.Id);
+        builder.Property(e => e.Id).ValueGeneratedOnAdd();
+        // Long enough for "EventLogItem/{Category}/{EventType}/{guid}" form. Deliberately not
+        // indexed and not unique. Nothing looks an item up by it, and each id embeds a fresh Guid.
+        builder.Property(e => e.EventLogItemId).IsRequired().HasMaxLength(600);
```
johnsimons: See comment above.
It makes me wonder why we need `EventLogItemId` at all

src/ServiceControl.Persistence.EFCore/EntityConfigurations/EventLogItemConfiguration.cs
```diff
@@ -0,0 +1,29 @@
...
+        builder.Property(e => e.Description).IsRequired();
+        builder.Property(e => e.Severity).IsRequired().HasConversion<string>().HasMaxLength(50);
```
johnsimons: I thought we can configure conventions like enums serialization globally?
I actually think i have done it already 🤔

src/ServiceControl.Persistence.EFCore/EntityConfigurations/EventLogItemConfiguration.cs
```diff
@@ -0,0 +1,29 @@
...
+        builder.Property(e => e.RaisedAt).IsRequired();
+        // RelatedTo is only ever returned to the API, never queried, so it does not warrant a child table.
+        builder.Property(e => e.RelatedTo).IsRequired();
+        builder.Property(e => e.Category).IsRequired().HasMaxLength(255);
+        builder.Property(e => e.EventType).IsRequired().HasMaxLength(255);
+        // Every read is "order by RaisedAt descending" plus paging. The key is included as a
+        // tiebreaker so that items sharing a RaisedAt do not shuffle between pages, and so that
+        // MAX(RaisedAt) for the ETag is an index seek.
```
johnsimons, line 23: is this something we need to tune for all list queries?

src/ServiceControl.Persistence.EFCore/Implementation/ErrorMessagesDataStore.cs
```diff
@@ -108,9 +108,6 @@ public Task<string[]> GetRetryPendingMessages(DateTime from, DateTime to, string
     public Task<byte[]> FetchFromFailedMessage(string uniqueMessageId) =>
         throw new NotImplementedException();
 
-    public Task StoreEventLogItem(EventLogItem logItem) =>
```
johnsimons: Nice refactor

src/ServiceControl.Persistence.EFCore/Implementation/EventLogDataStore.cs
```diff
@@ -1,13 +1,74 @@
...
+        ExecuteWithDbContext(async dbContext =>
+        {
+            dbContext.EventLogItems.Add(new EventLogItemEntity
+            {
+                EventLogItemId = logItem.Id,
+                Description = logItem.Description,
+                Severity = logItem.Severity,
+                RaisedAt = logItem.RaisedAt,
+                // The column is non-null. The API model allows null and the SignalR broadcast path
+                // deliberately empties it, so normalise rather than reject.
```
johnsimons: Not really adding much value, consider removing

src/ServiceControl.Persistence.EFCore/Implementation/EventLogDataStore.cs
```diff
@@ -1,13 +1,74 @@
...
+            if (knownVersion is not null && knownVersion == version)
+            {
+                return (null, total, version);
+            }
+
+            var items = await query
+                // The key breaks ties so that items sharing a RaisedAt cannot shuffle between
+                // pages. IX_EventLogItems_RaisedAt_Id is declared in exactly this order.
+                .OrderByDescending(e => e.RaisedAt)
+                .ThenByDescending(e => e.Id)
```
johnsimons, line 61: does this make sense, why not just by order of creation?

src/ServiceControl.Persistence.RavenDB/EventLogDataStore.cs
warwickschroeder: Same behaviour, just moved location.

src/ServiceControl.Persistence.EFCore/Implementation/EventLogDataStore.cs
```diff
@@ -1,13 +1,74 @@
...
+                    Category = e.Category,
+                    EventType = e.EventType
+                })
+                .ToListAsync();
+
+            return (items, total, version);
+        });
+
+    // Raven returns the query's ResultEtag but there is no relational equivalent, so it is synthesised
+    // using total count and the newest item's RaisedAt timestamp.
```
johnsimons: I don't think we need be mention Raven here.

src/ServiceControl.Persistence.EFCore/Implementation/EventLogDataStore.cs
```diff
@@ -1,13 +1,74 @@
...
+    static string Version(long total, DateTime? newest) =>
+        DeterministicGuid.MakeId($"{total}|{newest?.Ticks ?? 0}").ToString();
```
johnsimons: This feels like a pattern that we will need to use again 😄

src/ServiceControl.Persistence.RavenDB/EventLogDataStore.cs
johnsimons, line 24: So, if I correctly read this, the RavenDb was never setting the expiry on logitems and therefore they are never cleaned up?

johnsimons: and if this is true, should this be a PR on its own?

src/ServiceControl.Persistence.Tests/EFCore/EventLogDataStoreEFTests.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.Threading.Tasks;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.DependencyInjection;
+using NUnit.Framework;
+using ServiceControl.EventLog;
+using ServiceControl.Persistence.EFCore.DbContexts;
+using ServiceControl.Persistence.Infrastructure;
+
+// Behaviour only the EF Core persisters guarantee, so it cannot live in the shared root folder.
+// RavenDB orders event log items by RaisedAt alone.
```
johnsimons: Not adding much, consider removing?

src/ServiceControl.Persistence.Tests/EFCore/EventLogKnownVersionSkipsPageQueryTests.cs
```diff
@@ -0,0 +1,94 @@
...
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Diagnostics;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Logging.Abstractions;
+using NUnit.Framework;
+using ServiceControl.EventLog;
+using ServiceControl.Persistence.Infrastructure;
+
+// EF only. RavenDB gets documents, total and etag from a single query, so it has nothing to skip
```
johnsimons: Well that is already stated by the folder that is in 😉

src/ServiceControl.Persistence.Tests/EventLogWriterTests.cs
```diff
@@ -0,0 +1,71 @@
...
+using ServiceControl.Infrastructure.DomainEvents;
+using ServiceControl.Infrastructure.SignalR;
+using ServiceControl.Persistence.Infrastructure;
+
+// This exercises writer -> store -> row -> readable on every persister, using a test-local domain
+// event and mapping rather than borrowing a production contract, so it pins the writer mechanism
+// rather than the shape of MessageFailed.
+class EventLogWriterTests : PersistenceTestBase
+{
+    public EventLogWriterTests() => RegisterServices = services => services.AddSignalR();
```
johnsimons: Why do we need SignalR?

src/ServiceControl.Persistence.Tests/EventLogWriterTests.cs
```diff
@@ -0,0 +1,71 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using NUnit.Framework;
+using ServiceControl.EventLog;
+using ServiceControl.Infrastructure.DomainEvents;
+using ServiceControl.Infrastructure.SignalR;
+using ServiceControl.Persistence.Infrastructure;
+
+// This exercises writer -> store -> row -> readable on every persister, using a test-local domain
+// event and mapping rather than borrowing a production contract, so it pins the writer mechanism
+// rather than the shape of MessageFailed.
```
johnsimons: But why?

src/ServiceControl/EventLog/EventLogApiController.cs
```diff
@@ -16,14 +18,25 @@ public class EventLogApiController(IEventLogDataStore logDataStore) : Controller
...
             Response.WithEtag(version);
 
-            return results;
+            // Null items means "you already hold this version"
+            if (results is null)
+            {
+                return StatusCode((int)HttpStatusCode.NotModified);
+            }
+
+            return Ok(results);
```
johnsimons: This feels weird.
My thought is that we have middleware that handles it for us

src/ServiceControl.Persistence.Tests/EFCore/EventLogKnownVersionSkipsPageQueryTests.cs
```diff
@@ -0,0 +1,94 @@
...
+        Id = $"EventLogItem/Recoverability/MessageFailed/{Guid.NewGuid()}",
+        Category = "Recoverability",
+        EventType = "MessageFailed",
+        Description = "failed",
+        Severity = Severity.Info,
+        RaisedAt = new DateTime(2026, 7, 27, 12, 0, 0, DateTimeKind.Utc),
+        RelatedTo = []
+    };
+
+    sealed class CommandCounter : ILoggerProvider
```
johnsimons: This feels quite fragile, assertions on log messages!

src/ServiceControl.Persistence.EFCore/Infrastructure/RetentionSweeper.cs
johnsimons: no you cannot because we do it in batches

docs/testing.md
warwickschroeder: Its not. It was something I ran into and thought it would be useful to document.

src/ServiceControl.AcceptanceTests/EventLogs/When_the_event_log_is_polled_with_an_etag.cs
warwickschroeder: Yes its related, but also related to this PR. It tests the knownVersion change on the Etag side of things. Theres a few more tests created to test the DB query improvements

src/ServiceControl.Persistence.EFCore/Entities/EventLogItemEntity.cs
warwickschroeder: It sure is, which is why I havent made this the PK and introduced a long for that. Ill take another look into if we need this field at all in EF.

src/ServiceControl.Persistence.EFCore/Entities/EventLogItemEntity.cs
warwickschroeder: As above, will look into ifnwe need an event id in this format at all. If not, we can probably remove

src/ServiceControl.Persistence.EFCore/Entities/EventLogItemEntity.cs
warwickschroeder: In SQL server, yes. It maps to a json array string. In Postrgres, maps to a a native text[] type

src/ServiceControl.Persistence.EFCore/EntityConfigurations/EventLogItemConfiguration.cs
warwickschroeder: Ok, will double check

src/ServiceControl.Persistence.EFCore/EntityConfigurations/EventLogItemConfiguration.cs
warwickschroeder: Maybe, it makes sense when theres pages of data that potentially are created very close in time. Viewing lists of messages could potentially benefit from this. Ill create a board item to investigate

src/ServiceControl.Persistence.EFCore/Implementation/EventLogDataStore.cs
warwickschroeder: This is the change for this: 

https://github.com/Particular/ServiceControl/pull/5655#discussion_r3662145255

Adding the Id as a sort after the timestamp helps the flakey sort on pages. Becomes a very deterministic sort.

src/ServiceControl.Persistence.EFCore/Implementation/EventLogDataStore.cs
warwickschroeder: Making a deterministic Id is pre-existing. As for the version id, maybe, but it wont be used for RavenDB as it has this built in. Currently, its solely for EF. Maybe if we introduce more persistence later we could lift it.

src/ServiceControl.Persistence.RavenDB/EventLogDataStore.cs
warwickschroeder: It was setting this. This was just moved here as part of the event log data store refactor

src/ServiceControl/EventLog/EventLogApiController.cs
warwickschroeder: We do have middleware for this, but the problem with that is that it was still querying the database. This skips the query if the version tag matches the header

src/ServiceControl.Persistence.Tests/EFCore/EventLogKnownVersionSkipsPageQueryTests.cs
warwickschroeder: Yeh maybe. Probably dont need a test for this to be honest.. will see if its flakey, but EF seems to work with this

src/ServiceControl.Persistence.EFCore/Entities/EventLogItemEntity.cs
johnsimons: I just looked in SP and it seems this is the only place we are using the "related_to" prop, see https://github.com/Particular/ServicePulse/blob/5f0ede166f9a1209a288659d4f8f0c59067525ea/src/Frontend/src/components/EventLogItem.vue#L30-L33 and it seems we are only looking at the first item!

src/ServiceControl.Persistence.Tests/EventLogWriterTests.cs
warwickschroeder: So, looks like SignalR is still fully wired up and broadcasting 25 domain event types in ServiceControl... just to nobody it seems! 

https://github.com/Particular/ServiceControl/blob/ecbad0a6ca19a53485902b2d9b5f57dd1b64dedd/src/ServiceControl/Infrastructure/WebApi/HostApplicationBuilderExtensions.cs#L42

I did a search in ServicePulse and ServiceInsight and neither are clients of it. Integration events is something seperate and dont use SignalR. I dont know what else to check to make sure this isnt being used. Ill add something to our board to confirm and potentially clean up.

rbev: If there's no clients signalR doesn't actually "broadcast" anything, the only cost would be if there's a distributed backplane set up but looking at that it's only using the in-memory one.

src/ServiceControl.Persistence.EFCore/Implementation/EventLogDataStore.cs
```diff
@@ -1,13 +1,74 @@
...
+            await dbContext.SaveChangesAsync();
+        });
+
+    public Task<(IList<EventLogItem>? items, long total, string version)> GetEventLogItems(
+        PagingInfo pagingInfo, string? knownVersion = null) =>
+        ExecuteWithDbContext<(IList<EventLogItem>? items, long total, string version)>(async dbContext =>
+        {
+            var query = dbContext.EventLogItems.AsNoTracking();
+
+            var total = await query.LongCountAsync();
```
rbev: are these data pages really so big that doing two database roundtrips is faster than just going to get the data?

src/ServiceControl.Persistence.EFCore/DbContexts/ServiceControlDbContext.cs
```diff
@@ -24,5 +25,6 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
         modelBuilder.ApplyConfiguration(new FailedMessageConfiguration());
         modelBuilder.ApplyConfiguration(new FailedMessageGroupConfiguration());
         modelBuilder.ApplyConfiguration(new FailedMessageRetryConfiguration());
+        modelBuilder.ApplyConfiguration(new EventLogItemConfiguration());
```
rbev, line 34: Should put these in alphabetical order

src/ServiceControl.Persistence.EFCore/Infrastructure/RetentionSweeper.cs
rbev: Does this not work? 

```cs
await dbContext.EventLogItems
                .AsNoTracking()
                .Where(eventLogItem => eventLogItem.RaisedAt < cutoff)
                .OrderBy(eventLogItem => eventLogItem.RaisedAt)
                .Take(BatchSize)
                .ExecuteDeleteAsync(cancellationToken);

```

src/ServiceControl/EventLog/EventLogApiController.cs
```diff
@@ -16,14 +18,25 @@ public class EventLogApiController(IEventLogDataStore logDataStore) : Controller
...
+
+            // Passing knownVersion lets the persister skip work it would otherwise waste
+            var (results, totalCount, version) = await logDataStore.GetEventLogItems(pagingInfo, knownVersion);
 
             Response.WithPagingLinksAndTotalCount(pagingInfo, totalCount);
             Response.WithEtag(version);
 
-            return results;
+            // Null items means "you already hold this version"
+            if (results is null)
```
rbev:
```suggestion
            if (results is null && totalCount > 0)
```

src/ServiceControl.Persistence.Tests/EventLogWriterTests.cs
johnsimons: The only product that consumed SignalR was SP a long long time ago, so I think is time to remove it

src/ServiceControl.Persistence.EFCore/DbContexts/ServiceControlDbContext.cs
johnsimons: OCD 😉

src/ServiceControl.Persistence.Tests/EventLogWriterTests.cs
warwickschroeder: Once (if) SignalR is removed, I will remove it from this test too

src/ServiceControl.Persistence.Tests/EventLogWriterTests.cs
warwickschroeder: > using a test-local domain event and mapping rather than borrowing a production contract

Just to simplify the assertions and what we actually want to test, the actual Event Mapping

src/ServiceControl.Persistence.EFCore/DbContexts/ServiceControlDbContext.cs
warwickschroeder: Is there a good reason to put it in alphabetical order, other than preference?

src/ServiceControl.Persistence.EFCore/Implementation/EventLogDataStore.cs
warwickschroeder: To reliably create a version id, we need to count all rows. There could be a lot of rows in the event logs. Maybe this would work? AI recons that would be 1 round trip, matching (and improving) what Raven is currently doing as its 1 trip, but also returning data.

```
var stats = await query
    .GroupBy(_ => 1)
    .Select(g => new { Total = g.LongCount(), Newest = g.Max(e => (DateTime?)e.RaisedAt) })
    .FirstOrDefaultAsync();
```

src/ServiceControl.Persistence.Tests/EventLogWriterTests.cs
johnsimons: But the code reads well, there is no need for this comment, or am I missing something here?
You obviously are testing things in isolation and therefore you are doing some setup, but that is nothing new.

src/ServiceControl/EventLog/EventLogApiController.cs
warwickschroeder: This suggestion doesnt look right. `results` being null means that the `If-None-Match` matched the current version of the data. I dont think we want to check for actual data

src/ServiceControl.Persistence.EFCore/Infrastructure/RetentionSweeper.cs
johnsimons: Maybe 😄

src/ServiceControl.Persistence.EFCore/Entities/EventLogItemEntity.cs
warwickschroeder: I think we should keep the API-visible identity for each event log item the same. The database ID field used by the persistence doesn't need to be this though.

src/ServiceControl.Persistence.EFCore/Infrastructure/RetentionSweeper.cs
warwickschroeder: Ill test this now

warwickschroeder: Works great. Tests all pass ✅. Will update.

src/ServiceControl.Persistence.EFCore/DbContexts/ServiceControlDbContext.cs
rbev: Apart from easier merges and general tidiness, no.

src/ServiceControl.Persistence.EFCore/EntityConfigurations/EventLogItemConfiguration.cs
warwickschroeder: I think storing as int is the global default in EF. Raven stores this field as an int. FailedMessageStatus is stored as an int, so ill update to do the same. The API turns it back into a string on the way out anyway.

src/ServiceControl.Persistence.Tests/EventLogWriterTests.cs
warwickschroeder: Oh the comment iteself. I thought you were talking about what the comment was referring to. I prefer to be quite explicit with commenting. I feel having a short summary of what the tests are doing helps new comers, but I can remove if you think its not adding any value .

src/ServiceControl.Persistence.EFCore/Entities/EventLogItemEntity.cs
warwickschroeder: Yeh looks that way, only the leading `/message/` link. Ill add a board card to discuss.

src/ServiceControl.Persistence.EFCore/Implementation/EventLogDataStore.cs
warwickschroeder: Yep that worked. 1 round trip now.

src/ServiceControl.Persistence.EFCore/Entities/EventLogItemEntity.cs
johnsimons: I'm not sure I understand what you mean

warwickschroeder: This [commit](https://github.com/Particular/ServiceControl/pull/5655/changes/fb53f38ab2dbd3a5b1a10732cf0bcf3bed3db4e3) should move the Id generation into the persistence seams and out of the abstraction layer.

src/ServiceControl.Persistence.EFCore/Entities/EventLogItemEntity.cs
```diff
@@ -0,0 +1,25 @@
...
+using ServiceControl.EventLog;
+
+public class EventLogItemEntity
+{
+    // The physical key only. Sequential so inserts stay at the tail of the index, and narrow so it
+    // is a cheap tiebreaker for RaisedAt paging.
+    public long Id { get; set; }
+
+    // A globally unique identity common between persisters (e.g. used in migrations).
+    public Guid UniqueEventId { get; set; }
```
johnsimons: It is unclear to me why we're justifying this extra id becasue of migrations, the way I look at it is, that even if we need to do migrations on the data we would not do it by ids, we probably would do it via severity, category, eventtype, but not by id!

johnsimons: Also, events should never be migrated, this is past history 😉

src/ServiceControl.Persistence.EFCore/EntityConfigurations/EventLogItemConfiguration.cs
```diff
@@ -0,0 +1,28 @@
...
+
+class EventLogItemConfiguration : IEntityTypeConfiguration<EventLogItemEntity>
+{
+    public void Configure(EntityTypeBuilder<EventLogItemEntity> builder)
+    {
+        builder.ToTable("EventLogItems");
+        builder.HasKey(e => e.Id);
+        builder.Property(e => e.Id).ValueGeneratedOnAdd();
+        builder.Property(e => e.UniqueEventId).IsRequired();
+        builder.HasIndex(e => e.UniqueEventId).IsUnique();
```
johnsimons: why do we need this index, do we ever retrieve records by this id?

src/ServiceControl.Persistence.EFCore/Implementation/EventLogDataStore.cs
```diff
@@ -1,13 +1,87 @@
...
+            // single row, and an empty table yields no rows at all, hence the null coalescing.
+            var stats = await query
+                .GroupBy(_ => 1)
+                .Select(g => new
+                {
+                    Total = g.LongCount(),
+                    Newest = g.Max(e => (DateTime?)e.RaisedAt),
+                    HighestId = g.Max(e => (long?)e.Id)
+                })
+                .FirstOrDefaultAsync();
```
johnsimons, line 43: Nice trick, we probably need to reuse it in other areas too

src/ServiceControl/EventLog/AuditEventLogWriter.cs
```diff
@@ -28,15 +29,16 @@ public async Task Handle(IDomainEvent message, CancellationToken cancellationTok
             }
 
             var logItem = mappings.ApplyMapping(message);
+            var eventId = Guid.CreateVersion7();
```
johnsimons: I am wondering what would happened if the id is not generated here at all, and instead is created at the DB side. Would there be any downsides?

src/ServiceControl/EventLog/EventLogApiController.cs
```diff
@@ -16,14 +18,24 @@ public class EventLogApiController(IEventLogDataStore logDataStore) : Controller
...
         [HttpGet]
-        public async Task<IList<EventLogItem>> Items([FromQuery] PagingInfo pagingInfo)
+        public async Task<ActionResult<IList<EventLogItemView>>> Items([FromQuery] PagingInfo pagingInfo)
         {
-            var (results, totalCount, version) = await logDataStore.GetEventLogItems(pagingInfo);
 
-            Response.WithPagingLinksAndTotalCount(pagingInfo, totalCount);
-            Response.WithEtag(version);
+            // The Trim handles both ETag formats (quoted and unquoted) deliberately.
+            var knownVersion = Request.Headers.IfNoneMatch.FirstOrDefault()?.Trim('"');
```
johnsimons: This should be abstracted away?

johnsimons (approved):
I am still a bit confused about the need for `UniqueEventId` but we can finetune it in future PRs

src/ServiceControl.Persistence.EFCore/Entities/EventLogItemEntity.cs
warwickschroeder: As discussed, ill remove the global ID as its dead weight right now. If we decided its needed later we can add it back in. I have a card on our board for this.

src/ServiceControl.Persistence.EFCore/EntityConfigurations/EventLogItemConfiguration.cs
warwickschroeder: As above, will remove

src/ServiceControl/EventLog/AuditEventLogWriter.cs
warwickschroeder: As above, will remove

PR #5656, opened 2026-07-28
Add persistence and replay for failed error imports

johnsimons:
Introduces a new persistence mechanism for error messages that failed during their initial ingestion into ServiceControl. This ensures that these messages are reliably stored and can be re-processed at a later time.

The new `FailedErrorImports` table stores the failed messages, including their headers, body, and exception information. A deduplication mechanism ensures that repeated failures of the same message only update the existing record with the latest details.


src/ServiceControl.Persistence.EFCore/Implementation/ErrorMessagesDataStore.cs
```diff
@@ -32,8 +44,67 @@ public Task FailedMessageMarkAsArchived(string failedMessageId) =>
...
+                Body = storedBody,
+                BodyStoredExternally = storeExternally,
+                ExceptionInfo = failure.ExceptionInfo
+            });
+
+            try
+            {
+                await dbContext.SaveChangesAsync();
+            }
+            catch (DbUpdateException)
```
rbev: there is now a mechanism for checking if this is duplicate key exception via `dbContext`

src/ServiceControl.Persistence.EFCore/Implementation/FailedErrorImportDataStore.cs
```diff
@@ -1,12 +1,159 @@
...
+            return [];
+        }
+
+        await using var stream = stored.Stream;
+        using var buffer = new MemoryStream(stored.BodySize);
+        await stream.CopyToAsync(buffer, cancellationToken);
+        return buffer.ToArray();
+    }
+
+    // The row is removed before its external body: a surviving row with a missing body would replay
```
rbev, line 127: If you wanted to avoid this leak you could change it to flag the row as replayed and then delete in the other order.

rbev (approved):

src/ServiceControl.Persistence.EFCore/Implementation/ErrorMessagesDataStore.cs
johnsimons: Changed to use new upset

src/ServiceControl.Persistence.EFCore/Implementation/FailedErrorImportDataStore.cs
```diff
@@ -1,12 +1,159 @@
...
+        };
+    }
+
+    async Task<byte[]> ReadExternalBody(Guid uniqueMessageId, CancellationToken cancellationToken)
+    {
+        var stored = await bodyStorage.ReadBody(FailedErrorImportEntity.ExternalBodyId(uniqueMessageId), cancellationToken);
+
+        if (stored is null)
+        {
+            return [];
```
warwickschroeder: Should we be throwing here? If we try import a message that is expecting to find an external body file, but it doesnt, is that a "failed to import" flag? Returning an empty body would successfully import the message with no body.

johnsimons: Fixed, good catch

warwickschroeder (approved):

src/ServiceControl.Persistence.EFCore/Implementation/FailedErrorImportDataStore.cs
johnsimons: I am not sure this is going to be an issue at all.
I am going to leave it for now.

PR #5659, opened 2026-07-28
Remove SignalR

johnsimons:
This came out of the EF persistence work. Adding event log persistence for the SQL Server and PostgreSQL persisters (#5655) meant a new `EventLogWriterTests` had to stand up SignalR infrastructure just to construct `AuditEventLogWriter`, which takes a `GlobalEventHandler` purely so it can broadcast. That prompted
the question: why do we need SignalR at all?
https://github.com/Particular/ServiceControl/pull/5655#discussion_r3662205823

We don't. Neither ServicePulse nor ServiceInsight use it, and neither has for several releases. ServicePulse dropped it as part of the migration to Vue.js.  
Nothing has connected to the `/api/messagestream` hub since, so this is dead weight that new work still has to carry. Audit and Monitoring were never involved.

Removed:

- `MessageStreamerHub` and the `/api/messagestream` endpoint. Alongside pushing notifications out, the hub exposed a `SendMessage` method that let clients dispatch arbitrary NServiceBus messages into ServiceControl's input queue, so that goes too.
- `GlobalEventHandler`, `Envelope` and `ServicePulseNotifier`.
- `AddSignalR()`, its JSON protocol configuration, and the `AddServicePulseSignalRNotifier()` registration.
- `IUserInterfaceEvent`, the marker that selected which domain events got broadcast, and the 25 classes that implemented it.
- `EventLogItemAdded`, which was only ever constructed to be broadcast. `AuditEventLogWriter` now just maps and stores the log item, which is what `/api/eventlogitems/` reads.
- The `Microsoft.AspNetCore.SignalR.Client` package reference, the acceptance tests that used it, and the `HttpMessageHandlerFactory` test infrastructure those tests were the only consumers of.


danielmarbach (approved):
At the time of the .NET migration, we wanted to drop it but weren't sure whether the minimum required SC version of Pulse would make sure no old SP instance that would require SignalR would be supported anymore. If that is the case, I think we can and should drop it at this stage.

rbev (approved):

PR #5660, opened 2026-07-28
.NET runtime updates (July 2026) 

bartekwasielak:


rbev:
Currently global.json has roll forward enabled to the latest patch version, this means that when Microsoft updates default SDK in the github runner agents the CI builds will only install a newer runtime and the builds pinned to this runtime will begin failing. 

To avoid this we should also update global.json to use the [matching SDK version](https://github.com/dotnet/core/blob/main/release-notes/10.0/10.0.10/10.0.10.md) instead of the latest patch version.

WilliamBZA (approved):

PR #5663, opened 2026-07-29
Add EF Core-backed subscription storage for ServiceControl

johnsimons:
Introduces a new persistence mechanism for NServiceBus subscriptions using Entity Framework Core. This enables ServiceControl to track and retrieve subscriber information from a relational database.

### DIFF
--- a/src/ServiceControl.Persistence.EFCore/Implementation/SubscriptionStorage.cs
+++ b/src/ServiceControl.Persistence.EFCore/Implementation/SubscriptionStorage.cs
@@ -1,20 +1,122 @@
 namespace ServiceControl.Persistence.EFCore.Implementation;
 
+using System.Collections.Concurrent;
+using Abstractions;
+using Entities;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.DependencyInjection;
+using NServiceBus;
 using NServiceBus.Extensibility;
+using NServiceBus.Settings;
 using NServiceBus.Unicast.Subscriptions;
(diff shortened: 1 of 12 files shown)


src/ServiceControl.Persistence.EFCore/Implementation/SubscriptionStorage.cs
```diff
@@ -1,20 +1,122 @@
...
+        return subscribers.ToArray();
+    }
+
+    void InvalidateCache(string typeName)
+    {
+        foreach (var entry in cache)
+        {
+            if (entry.Value.TypeNames.Contains(typeName, StringComparer.Ordinal))
+            {
+                cache.TryRemove(entry.Key, out _);
```
rbev, line 109: Does this cause a collection changed exception in the enumerator?

rbev (approved):

src/ServiceControl.Persistence.EFCore/Implementation/SubscriptionStorage.cs
johnsimons: Nope, and is tested in Subscribing_evicts_the_cached_lookup_for_that_message_type.

The long response is:

ConcurrentDictionary is explicitly documented to allow this: its enumerator walks the buckets lock-free with no version check, so it never throws InvalidOperationException for concurrent (or same-thread) mutation. From the GetEnumerator docs:

> The enumerator returned from the dictionary is safe to use concurrently with reads and writes to the dictionary, however it does not represent a moment-in-time snapshot of the dictionary.

PR #5670, opened 2026-07-30
Remove decommissioned endpoints from usage reports

### LINKED ISSUE #5517: Remove decommissioned endpoints from future reports

Currently, when a message comes into ServiceControl from either:

- Heartbeats Plugin
- Audit instance
- Monitoring instance

the endpoint name associated with the message is saved against the EndpointDocuments collection. Likewise, if the transport/broker is queryable, every queue found as part of the throughput hosted service is saved against the EndpointDocuments collection. When any of these actions occur, a new timeseries record is written.

When the throughput report is generated, the entire list of endpoints in the EndpointDocuments collection is read, along with the last 14 months of timeseries data for that endpoint. Even if the resulting timeseries data is empty, the endpoint is still written to the throughput report. This means that if an endpoint is decommissioned, and the associated queues are removed or no longer process any messages, the queue will still appear in the throughput report indefinitely.

Only active queues should be written to the throughput report. If no activity has been seen related to a queue for the entire reporting period then it should be removed from the EndpointDocuments collection
        
### Additional Context

The the Licensing Component was introduced in [v5.4.0](https://github.com/Particular/ServiceControl/releases/tag/5.4.0))

PhilBastian:
remove any EndpointDocuments that haven't seen any throughput data for the reporting period. Handles send-only endpoints (reported by Audit instances) by reporting a throughput of 0 daily.

fixes:

- #5517


mikeminutillo (approved):
Approved, except as discussed, this may remove information coming from Audit that we have about send-only endpoints. When a send-only endpoint sends a message it stamps the outgoing message with a sender tag, and the audit instance treats this as a known endpoint, even though it has knowledge of the endpoint processing messages. These should probably report as `0`s from the Audit instance and stored in the licensing store so they are not deleted.

src/Particular.LicensingComponent/AuditThroughput/AuditThroughputCollectorHostedService.cs
```diff
@@ -78,11 +78,15 @@ async Task GatherThroughput(CancellationToken cancellationToken)
 
             var auditCounts = (await auditQuery.GetAuditCountForEndpoint(knownEndpointsLookup[endpointId].UrlName, cancellationToken)).ToList();
 
-            if (endpoint == null)
+            if (endpoint == null && auditCounts.Count > 0)
             {
                 endpoint = ConvertToEndpoint(knownEndpointsLookup[endpointId]);
                 await dataStore.SaveEndpoint(endpoint, cancellationToken);
             }
+            else if (endpoint is null)
```
mikeminutillo: There is a mix here of `endpoint == null` and `endpoint is null` that is confusing on a casual read. This whole thing could be refactored as:

```cs
if(endpoint is null)
{
  if (auditCounts.Count <= 0)
  {
    continue;
  }

  endpoint = ConvertToEndpoint(knownEndpointsLookup[endpointId]);
  await dataStore.SaveEndpoint(endpoint, cancellationToken);
}
```

src/ServiceControl.Audit.Persistence.RavenDB/RavenAuditDataStore.cs
```diff
@@ -138,6 +138,9 @@ public async Task<QueryResult<IList<AuditCount>>> QueryAuditCounts(string endpoi
                 .OrderBy(m => m.ProcessedAt)
                 .FirstOrDefaultAsync(token: cancellationToken);
 
+            var hasSent = await session.Query<MessagesViewIndex.SortAndFilterOptions>(indexName)
```
mikeminutillo: We only need to do this query if we have no evidence of receive, right? Could this be moved to the `else` clause?

mikeminutillo (approved):

PR #5671, opened 2026-07-31
Add EF Core-backed failed message queries

johnsimons:
Implements IFailedMessageQueryDataStore for the EF Core persisters, covering the failed message lists, the summary and the single-message lookups. Exception source and stack trace, and the Edited and EditOf markers, are read from the stored headers rather than duplicated into columns.


src/ServiceControl.Persistence.EFCore.SqlServer/SqlServerIngestionSqlDialect.cs
```diff
@@ -109,7 +109,7 @@ static async Task Execute(ServiceControlDbContext dbContext, string sql, IEnumer
     // The columns the newer attempt wins wholesale
     static readonly string[] PayloadColumns =
     [
-        "MessageId", "MessageType", "TimeSent", "ConversationId", "QueueAddress",
+        "MessageId", "MessageType", "TimeSent", "ConversationId", "FailingEndpointAddress",
```
rbev, line 112: I notice that FailingEndpointAddress was added in a different PR. If this list having correct coverage of the entity is important there should be a test added for it.

src/ServiceControl.Persistence.EFCore/Implementation/FailedMessageQueryDataStore.cs
```diff
@@ -1,30 +1,148 @@
...
+            sortInfo));
 
-    public Task<QueryStatsInfo> GetFailedMessagesStats(string status, string modified, string queueAddress) =>
-        throw new NotImplementedException();
+    public Task<QueryStatsInfo> GetFailedMessagesStats(string? status, string? modified, string? queueAddress) =>
+        ExecuteWithDbContext(dbContext => dbContext.FailedMessages
+            .AsNoTracking()
+            .FilterByStatus(status)
+            .FilterByLastModifiedRange(modified)
+            .FilterByQueueAddress(queueAddress)
```
rbev, line 30: Does this need to be added to the `Status, Modified` index?

src/ServiceControl.Persistence.EFCore/Implementation/FailedMessageQueryDataStore.cs
```diff
@@ -1,30 +1,148 @@
...
+                .FilterByQueueAddress(queueAddress),
+            pagingInfo,
+            sortInfo));
 
-    public Task<QueryStatsInfo> GetFailedMessagesStats(string status, string modified, string queueAddress) =>
-        throw new NotImplementedException();
+    public Task<QueryStatsInfo> GetFailedMessagesStats(string? status, string? modified, string? queueAddress) =>
+        ExecuteWithDbContext(dbContext => dbContext.FailedMessages
+            .AsNoTracking()
+            .FilterByStatus(status)
```
rbev, line 28: When status is null I'm pretty sure this will miss the index as the `Where` doesn't get added

src/ServiceControl.Persistence.EFCore/Infrastructure/FailedMessageQueryFilters.cs
```diff
@@ -0,0 +1,146 @@
...
+                    excludes.Add(excluded);
+                }
+
+                continue;
+            }
+
+            if (Enum.TryParse<FailedMessageStatus>(filter, true, out var included))
+            {
+                includes.Add(included);
+            }
```
rbev: `else` instead of `continue` might read better.

```suggestion
            }
            else
            {
                if (Enum.TryParse<FailedMessageStatus>(filter, true, out var included))
                {
                    includes.Add(included);
                }
            }
```

src/ServiceControl.Persistence.EFCore/Infrastructure/FailedMessageQueryFilters.cs
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        if (includes.Count > 0)
+        {
+            source = source.Where(message => includes.Contains(message.Status));
+        }
+
+        foreach (var exclude in excludes)
+        {
+            // Captured per iteration so each exclusion closes over its own value.
```
rbev: Could it just be `!excludes.Contains(message.Status)` instead of this?

src/ServiceControl.Persistence.EFCore/Infrastructure/FailedMessageQueryFilters.cs
```diff
@@ -0,0 +1,146 @@
...
+    public static IQueryable<FailedMessageEntity> FilterByQueueAddress(this IQueryable<FailedMessageEntity> source, string? queueAddress)
+    {
+        if (string.IsNullOrWhiteSpace(queueAddress))
+        {
+            return source;
+        }
+
+        var address = queueAddress.ToLowerInvariant();
+
+        // The ToLower() here causes a full table scan!
```
rbev, line 96: Is there a way to avoid this?

src/ServiceControl.Persistence.EFCore/Infrastructure/FailedMessageQueryFilters.cs
```diff
@@ -0,0 +1,146 @@
...
+        return source.Where(message => message.FailingEndpointAddress != null && message.FailingEndpointAddress.ToLower() == address);
+    }
+
+    public static IQueryable<FailedMessageEntity> Sort(this IQueryable<FailedMessageEntity> source, SortInfo? sortInfo)
+    {
+        var descending = sortInfo?.Direction != "asc";
+        var sort = sortInfo?.Sort;
+
+        if (sort == null || !SortInfo.AllowedSortOptions.Contains(sort))
```
rbev: is this check redundant given the fallback in the switch statement?

src/ServiceControl.Persistence.Tests/EFCore/ErrorIngestionTests.cs
```diff
@@ -31,7 +31,7 @@ public async Task First_failure_stores_the_message()
             Assert.That(row.MessageType, Is.EqualTo(failure.MessageType));
             Assert.That(row.TimeSent, Is.EqualTo(failure.TimeSent));
             Assert.That(row.ConversationId, Is.EqualTo(failure.ConversationId));
-            Assert.That(row.QueueAddress, Is.EqualTo(failure.QueueAddress));
+            Assert.That(row.FailingEndpointAddress, Is.EqualTo(failure.QueueAddress));
```
rbev: this assert looks a bit sus

src/ServiceControl.Persistence.EFCore/Implementation/FailedMessageQueryDataStore.cs
```diff
@@ -1,30 +1,148 @@
...
+        {
+            var endpoints = await CountBy(dbContext, message => message.ReceivingEndpointName);
+            var hosts = await CountBy(dbContext, message => message.ReceivingEndpointHost);
+            var messageTypes = await CountBy(dbContext, message => message.MessageType);
 
-    public Task<FailedMessageView> GetLatestFailedMessageView(string failedMessageId) =>
-        throw new NotImplementedException();
+            return (IDictionary<string, object>)new Dictionary<string, object>
+            {
+                ["Endpoints"] = endpoints,
```
rbev: Where do these magic strings come from?
Are they magic strings on the raven side too?

rbev (approved):
Approved with a few suggestions

src/ServiceControl.Persistence.EFCore.SqlServer/SqlServerIngestionSqlDialect.cs
johnsimons: Added test

src/ServiceControl.Persistence.EFCore/Infrastructure/FailedMessageQueryFilters.cs
johnsimons: Fixed

src/ServiceControl.Persistence.EFCore/Infrastructure/FailedMessageQueryFilters.cs
johnsimons: Fixed

src/ServiceControl.Persistence.EFCore/Infrastructure/FailedMessageQueryFilters.cs
johnsimons: Fixed

src/ServiceControl.Persistence.Tests/EFCore/ErrorIngestionTests.cs
johnsimons: Fixed

src/ServiceControl.Persistence.EFCore/Implementation/FailedMessageQueryDataStore.cs
johnsimons: Yes they are, I have made them more explicit and sharing between Raven and EF

src/ServiceControl.Persistence.EFCore/Infrastructure/FailedMessageQueryFilters.cs
johnsimons: `queueAddress` only arrives from the Pending Retries screen, which does poll every 5 seconds, so this isn't a cold path. But the predicate always runs after the status and modified filters, so it's a residual test over retry-issued messages in the selected range rather than a scan of the whole table, and that set is normally small. Not worth a schema change on that evidence. Worth noting the `ToLower()` is redundant on SQL Server anyway, since the default collation is already case-insensitive, and only PostgreSQL genuinely needs it, so if this ever does show up in a profile the fix is a normalised column rather than a new index.

src/ServiceControl.Persistence.EFCore/Implementation/FailedMessageQueryDataStore.cs
johnsimons: `FailingEndpointAddress` already has its own index, but the `ToLower()` below stops it being used, so adding the column to the composite wouldn't buy anything until that's resolved. See https://github.com/Particular/ServiceControl/pull/5671#discussion_r3701433956

src/ServiceControl.Persistence.EFCore/Implementation/FailedMessageQueryDataStore.cs
johnsimons: I looked at this and it seems the only place this method is called is from the https://github.com/Particular/ServiceControl/blob/68e5784037aedbc967f1bc6d4073d8aed3a8e4d0/src/ServiceControl/MessageFailures/Api/GetAllErrorsController.cs#L37 which is never called from ServicePulse so I am hesitant in adding another index to the hot path ingestion table!

PR #5681, opened 2026-07-31
Better naming of methods

johnsimons:
The two by-id lookups differed only in that one reads unresolved groups and the other archived ones, which neither the names nor the signatures showed.
The four status-scoped members are now named after what they return: GetUnresolvedGroupsByClassifier, GetArchivedGroupsByClassifier, GetUnresolvedGroup and GetArchivedGroup.

GetUnresolvedGroup returns a single view like its archived twin, so neither controller picks the first result off a list any more.

GetCurrentForwardingBatch moves to IRetryDocumentDataStore, where the NowForwarding pointer it reads already lives. QueryFailureGroupViewOnGroupId is deleted because it ran the same query as GetUnresolvedGroup, so RetryAllInGroupHandler calls that instead.

Both group-by-id endpoints now return 404 for an unknown group instead of an empty 204, matching GetErrorByIdController. ServicePulse reads these only to label a group, and its client already fails on the empty 204, so this gives it a status it can actually handle.


rbev (approved):

johnsimons:
This PR was just renames + refactors, that are now in master

PR #5683, opened 2026-08-02
Store message redirects

johnsimons:
Message redirect persistence for the SQL Server and PostgreSQL persisters.

## Changes

- `MessageRedirects` table keyed on `FromPhysicalAddress`, since `MessageRedirectId` is a hash of it, with migrations for both providers.
- `IMessageRedirectsDataStore` exposes `AddRedirect`, `UpdateRedirect` and `RemoveRedirect` rather `Save(collection)`, since every caller changes a single redirect. `GetRedirects` still all of them for the API's duplicate and dependents validation.
- The Raven specific parts move into the Raven project: the stored document type, its tick based and the change vector handling. Storage there is unchanged. The shared model keeps the itself, with lookups as extension methods.
- `IRetryBatchesManager.GetOrCreateMessageRedirectsCollection` duplicated the store's read, so it removed and `RetryProcessor` takes `IMessageRedirectsDataStore`.

## Behaviour changes

- ETag value are computed deterministically from the payload. But this are is going to be revisited.
- Concurrency is now per redirect rather than across all of them. RavenDB keeps every redirect in a document, so two admins editing at the same time contend on that document even when they changing different redirects, and one of them fails. With a row per redirect, edits to redirects no longer interfere at all, and two edits to the same redirect are last write instead of a failure.

### DIFF
--- a/src/ServiceControl.Persistence.EFCore.PostgreSql/Migrations/20260802023138_AddMessageRedirects.Designer.cs
+++ b/src/ServiceControl.Persistence.EFCore.PostgreSql/Migrations/20260802023138_AddMessageRedirects.Designer.cs
@@ -3,6 +3,7 @@ using System;
 using System.Collections.Generic;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Migrations;
 using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
 using Npgsql.EntityFrameworkCore.PostgreSQL.Metadata;
 using ServiceControl.Persistence.EFCore.PostgreSql;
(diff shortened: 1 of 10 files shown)


src/ServiceControl.Persistence.RavenDB/MessageRedirects/MessageRedirectsDataStore.cs
rbev, line 1: We should be doing RavenDB/API refactorings in standalone PRs so that it's easier on anyone doing a SC release.

rbev (changes requested):

src/ServiceControl.Persistence.RavenDB/MessageRedirects/MessageRedirectsDataStore.cs
johnsimons: this is now done

rbev (approved):

PR #5689, opened 2026-08-04
Reshape the retry consumer contract away from document semantics

johnsimons:
Groundwork for the EF Core retry consumer. The contract it would have to implement is by RavenDB rather than by what the retry does, so it is worth fixing before there a second persister bound to it.

## What is document oriented about it

`IRetryBatchesManager` is an `IAsyncDocumentSession` with the names changed. `RetryProcessor` documents, mutates them, and relies on `SaveChanges` noticing by object identity; 
`Evict` is the session cache and `CancelExpiration` strips the `@expires` metadata. On a store none of that exists: expiry is a status and a timestamp, so cancelling it implicit, and mutate-then-save means keeping a domain-object to row identity map and back by hand.

`RetryBatch` has the same problem one level down. It is both the contract and the stored, so callers receive `FailureRetries`, the document's own membership list, when the thing anyone reads is `.Count`. The two persisters were already filling it with
different values.

## What replaces it

`IRetryStagingStore` states the operations `RetryProcessor` performs, each atomic on its own, `Evict`, `Store`, the `Delete` overloads and `SaveChanges` all disappear. A pass forwards stages but never both, and every write that shared a `SaveChanges` still shares a, so the crash windows are unchanged. The stored `RetryBatch` moves into the project keeping its class name, and the contract becomes a read model with `MessageCount`, which EF answers with a `COUNT` over claim rows.

## Behaviour changes

A staging batch with nothing left to stage is discarded. Previously only the "all claimed by earlier batch" case deleted it; when the claims existed but the messages were gone, the went to `Forwarding` with no pointer and stayed there.

A batch that fails to reach the transport logs one warning naming the batch, rather than one message carrying the same exception. The next attempt dispatches one at a time and logs failure.

## Moving the stored types

`RavenPersistedTypes.Verify` fires by design: it pins the assembly-qualified name of every an index references. Collections come from the class name, which is unchanged.
`RetryDocumentCompatibilityTests` pins the three collection names and reads documents the pre-move `Raven-Clr-Type` through the store, saving one of them, which is where changed collection would surface.


src/ServiceControl.Persistence.RavenDB/RetryStagingStore.cs
```diff
@@ -0,0 +1,207 @@
...
+
+            var messages = await session.LoadAsync<FailedMessage>(claims.Select(claim => claim.FailedMessageId));
+
+            return
+            [
+                .. claims
+                    .Select(claim => new { Claim = claim, Message = messages[claim.FailedMessageId] })
+                    .Where(row => row.Message != null)
+                    .Select(row => ToStagingMessage(row.Message, row.Claim.StageAttempts))
+            ];
```
rbev: Seems a bit unnecessary to use this syntax

```suggestion
             claims
                    .Select(claim => new { Claim = claim, Message = messages[claim.FailedMessageId] })
                    .Where(row => row.Message != null)
                    .Select(row => ToStagingMessage(row.Message, row.Claim.StageAttempts))
                    .ToArray();
```

src/ServiceControl.Persistence/RetryBatch.cs
```diff
@@ -1,30 +1,32 @@
...
+        public string Originator { get; init; }
+        public string Classifier { get; init; }
+        public DateTime StartTime { get; init; }
+        public DateTime? Last { get; init; }
+        public string RequestId { get; init; }
+        public int InitialBatchSize { get; init; }
+        public RetryType RetryType { get; init; }
+        public RetryBatchStatus Status { get; init; }
+
+        // The messages the batch still holds, which is what a forwarded batch is counted against.
```
rbev: Should these comments be promoted to  `/// <summary>` or `/// <remarks>`

rbev (approved):

src/ServiceControl.Persistence.RavenDB/RetryStagingStore.cs
johnsimons: fixed

src/ServiceControl.Persistence/RetryBatch.cs
johnsimons: fixed

PR #5692, opened 2026-08-04
Implement the retry consumer on EF Core

johnsimons:
The point of reshaping that contract was that a document session does not survive the
translation to a relational store. With operations instead of a session, the implementation
is mostly statements: staging is a join from claims to messages, handing a batch to the
forwarder is one transaction, and the batch's message count is a `COUNT` over claims rather
than a list it carries around.

No schema change. `RetryBatches`, the single row `RetryBatchNowForwarding` and
`FailedMessageRetries` all landed with the producer side, and the claim row is the membership.

Worth knowing while reading it:

- `GetMessagesToStage` projects the five columns staging uses, so a claim whose message is
  gone drops out of the join instead of being loaded and discarded, and message bodies are
  never read to stage a batch.
- `MarkBatchAsForwarding` is one transaction: the batch to `Forwarding` with its staging id,
  claims trimmed to what was staged, those messages to `RetryIssued`, and the pointer set.
  There is no "cancel expiration" step, since retention only sweeps `Resolved` and `Archived`,
  so the status write is the cancellation.
- `CompleteForwarding` deliberately leaves the claims behind: they are what stops a message
  being staged again before its retry is confirmed. `DiscardBatch` does delete them, because a
  discarded batch only ever holds claims whose messages are gone.

### DIFF
--- a/src/ServiceControl.Persistence.EFCore/Implementation/RetryStagingStore.cs
+++ b/src/ServiceControl.Persistence.EFCore/Implementation/RetryStagingStore.cs
@@ -22,11 +22,12 @@ public class RetryStagingStore(IServiceScopeFactory scopeFactory, TimeProvider t
             return batch?.ToRetryBatch(await CountMessages(dbContext, batch.Id));
         });
 
-    public Task<StagingMessage[]> GetMessagesToStage(string batchId) =>
-        ExecuteWithDbContext(async dbContext =>
-        {
-            var batch = ParseBatchId(batchId);
+    public Task<StagingMessage[]> GetMessagesToStage(string batchId)
+    {
+        var batch = ParseBatchId(batchId);
 
+        return ExecuteWithDbContext(async dbContext =>
+        {
             // A message claimed by an earlier batch is not claimed by this one, and a claim whose
             // message is gone drops out of the join, which is what leaves it out of the staging.
             var rows = await dbContext.FailedMessageRetries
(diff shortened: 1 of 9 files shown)


src/ServiceControl.Persistence.EFCore/Implementation/RetryStagingStore.cs
```diff
@@ -1,34 +1,231 @@
...
+                .SingleOrDefaultAsync();
+
+            return nowForwarding?.RetryBatchId.ToString();
+        });
 
     public Task<RetryBatch?> GetBatch(string batchId, CancellationToken cancellationToken) =>
-        throw new NotImplementedException();
+        ExecuteWithDbContext(async dbContext =>
+        {
+            var batch = ParseBatchId(batchId);
```
rbev: This class is inconsistent between doing the argument parsing/validation upfront or inside `ExecuteWithDbContext`

src/ServiceControl.Persistence.EFCore/Implementation/RetryStagingStore.cs
```diff
@@ -1,34 +1,231 @@
...
+        }
 
-    public Task IncrementStagingAttempts(string uniqueMessageId) =>
-        throw new NotImplementedException();
+        return parsed;
+    }
 
-    public Task RemoveFromBatch(string uniqueMessageId) =>
-        throw new NotImplementedException();
+    // Batch ids only ever come from CreateBatch, so anything else is a programming error.
```
rbev: Should this be a `///<summary>`?

src/ServiceControl.Persistence.EFCore/Implementation/RetryStagingStore.cs
```diff
@@ -1,34 +1,231 @@
...
+        dbContext.Database.CreateExecutionStrategy().ExecuteAsync(async () =>
+        {
+            await using var transaction = await dbContext.Database.BeginTransactionAsync();
+
+            await operations();
+
+            await transaction.CommitAsync();
+        });
+
+    // Message ids reach this from the API, so an id that is not a message id cannot match a stored
```
rbev: Should this be a `///<summary>`?

rbev (approved):

PR #5705, opened 2026-08-06
Remove RavenDB specific ID from Notifications persisence contract

rbev:
**Description**

The current Notifications data store is stored as a singleton document in RavenDB, and is returned directly from the INotificationsManager. The Id field on this document is not used or relevant to consumers of this interface.

To remove this field from the contract this change implements a new internal class for Raven to persist, and removes the ID field from the returned object.

**Testing Improvements:**

* Added a new test class `NotificationsDataStoreTests` with comprehensive tests to verify default values, round-trip persistence, field updates, and field preservation for notification settings. (`src/ServiceControl.Persistence.Tests/NotificationsDataStoreTests.cs`)
* Removed inclusion of the new test file from SQL Server and PostgreSQL test project files as they will be implemented in a future change.

### DIFF
--- a/src/ServiceControl.Persistence.EFCore/Implementation/EditFailedMessagesManager.cs
+++ b/src/ServiceControl.Persistence.EFCore/Implementation/EditFailedMessagesManager.cs
@@ -20,9 +20,10 @@ public class EditFailedMessagesManager : IEditFailedMessagesManager
     public Task SaveChanges() =>
         throw new NotImplementedException();
 
-    public void Dispose()
+    public ValueTask DisposeAsync()
     {
         // Nothing to dispose yet
         GC.SuppressFinalize(this);
+        return ValueTask.CompletedTask;
     }
 }
--- a/src/ServiceControl.Persistence.EFCore/Implementation/EditFailedMessagesManager.cs
+++ b/src/ServiceControl.Persistence.EFCore/Implementation/EditFailedMessagesManager.cs
@@ -20,10 +20,5 @@ public class EditFailedMessagesManager : IEditFailedMessagesManager
     public Task SaveChanges() =>
         throw new NotImplementedException();
 
-    public ValueTask DisposeAsync()
-    {
-        // Nothing to dispose yet
-        GC.SuppressFinalize(this);
-        return ValueTask.CompletedTask;
-    }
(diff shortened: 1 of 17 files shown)


src/ServiceControl.Persistence.RavenDB/Editing/NotificationSettingsDocument.cs
```diff
@@ -0,0 +1,9 @@
+namespace ServiceControl.Persistence.RavenDB.Editing;
+
+using Notifications;
+
+class NotificationsSettingsDocument
+{
+    public string Id { get; set; }
+    public EmailNotifications Email { get; set; } = new();
```
johnsimons, line 8: Should Raven have its own version of `EmailNotifications`?

johnsimons (approved):
LG but Raven entity is still using the model, not sure if that is ok.

src/ServiceControl.Persistence.RavenDB/Editing/NotificationSettingsDocument.cs
rbev: That depends, if we are going to just serialise `EmailNotifications` as the type in the SQL settings table then they will effectively be the same.
I'm not opposed to it, but EmailNotifications is a pretty simple POCO and unlikely to change significantly so I don't know how much value will be gained from replicating it in each persistence along with object-object mapping.

johnsimons: fair enough

src/ServiceControl.Persistence.EFCore/Implementation/EditFailedMessagesManager.cs
```diff
@@ -20,9 +20,10 @@ public Task SetFailedMessageAsResolved() =>
     public Task SaveChanges() =>
         throw new NotImplementedException();
 
-    public void Dispose()
+    public ValueTask DisposeAsync()
```
johnsimons: Should this be `public ValueTask DisposeAsync() => default;`
Therre is no need for `GC.SuppressFinalize(this);`

johnsimons (approved):

PR #5707, opened 2026-08-06
Replace TrialMetadata table with a generic Settings key/value table

johnsimons:
Replaces the single row `TrialMetadata` table with a generic `Settings(Key, Value)` table that stores JSON documents, keyed by entries in `SettingKeys`.

This is groundwork for the EF licensing data store, which needs to persist `BrokerMetadata`, `AuditServiceMetadata`, `ReportMasks` and `LicensedEndpointDetails`: all singleton documents in RavenDB, all read and written whole and never queried by their contents. Rather than six near identical tables, they share one.

`TrialLicenseDataProvider` now goes through `GetSetting`/`StoreSetting`, with writes using the existing `UpsertAsync` extension. Migrations for both providers drop `TrialMetadata` and create `Settings`. Nothing EF related has shipped, so no data is copied.


rbev (approved):
Looks good
My only suggestion is that it might be worth writing the type name to a column for debugging/admin purposes.

PR #5708, opened 2026-08-07
Refactor message body full-text search configuration and tests

johnsimons:
The `EnableFullTextSearchOnBodies` setting is only honored by the RavenDB persister, which uses it to control whether message bodies are indexed for full-text search. EF Core persisters, by design, always index message bodies.

This change removes the redundant `EnableFullTextSearchOnBodies` setting from the EF Core persistence configuration. Corresponding acceptance tests are updated: a new test is added specifically for RavenDB to verify that the setting correctly disables body search, while the previously shared test for disabled body search is removed as it's not applicable to EF Core.

### DIFF
--- a/src/ServiceControl.Persistence.EFCore/Infrastructure/FailedMessageQueryFilters.cs
+++ b/src/ServiceControl.Persistence.EFCore/Infrastructure/FailedMessageQueryFilters.cs
@@ -113,6 +113,55 @@ static class FailedMessageQueryFilters
         };
     }
 
+    public static IQueryable<FailedMessageEntity> FilterBySentTimeRange(this IQueryable<FailedMessageEntity> source, DateTimeRange? timeSentRange)
+    {
+        if (timeSentRange?.From is { } from)
+        {
+            source = source.Where(message => message.TimeSent >= from);
+        }
+
+        if (timeSentRange?.To is { } to)
+        {
+            source = source.Where(message => message.TimeSent <= to);
+        }
+
+        return source;
+    }
+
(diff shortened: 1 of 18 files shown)


src/ServiceControl.Persistence.EFCore/Infrastructure/FailedMessageQueryFilters.cs
```diff
@@ -113,6 +113,55 @@ public static IQueryable<FailedMessageEntity> Sort(this IQueryable<FailedMessage
...
+        // zero and sorting by them is meaningless here. RavenDB behaves the same way, its index
+        // fields for them are always null.
+        return sortInfo?.Sort switch
+        {
+            "id" or "message_id" => source.OrderBy(message => message.MessageId, descending),
+            "message_type" => source.OrderBy(message => message.MessageType, descending),
+            "processed_at" => source.OrderBy(message => message.LastAttemptedAt, descending),
+            // Ordering follows the status the view reports, not the one the column stores.
+            "status" => source.OrderBy(message =>
+                message.Status == FailedMessageStatus.Resolved
```
rbev, line 153: Can this be a switch statement instead of nested ternary?

rbev (approved):

src/ServiceControl.Persistence.EFCore/Infrastructure/FailedMessageQueryFilters.cs
johnsimons: No. Inside a LINQ expression tree, a switch expression is a compile error: CS8514 "An expression tree may not contain a switch expression."

rbev: :cry:

PR #5714, opened 2026-08-07
Implement the EF licensing data store

johnsimons:
RavenDB stored throughput as an incremental time series keyed by a case-insensitive document id. The relational equivalent is a row per endpoint per day, with lowercased name columns to keep endpoint identity and sanitized name matching case-insensitive on PostgreSQL. Recording adds to the day's total via ExecuteUpdate so concurrent writers queue on the row rather than overwriting each other, and the four singleton documents share the Settings table.

### DIFF
--- a/src/ServiceControl.Persistence.EFCore/Implementation/LicensingDataStore.cs
+++ b/src/ServiceControl.Persistence.EFCore/Implementation/LicensingDataStore.cs
@@ -1,23 +1,362 @@
+namespace ServiceControl.Persistence.EFCore.Implementation;
+
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.DependencyInjection;
 using Particular.LicensingComponent.Contracts;
 using Particular.LicensingComponent.Persistence;
+using ServiceControl.Persistence.EFCore.DbContexts;
+using ServiceControl.Persistence.EFCore.Entities;
+using ServiceControl.Persistence.EFCore.Infrastructure;
 
-class LicensingDataStore : ILicensingDataStore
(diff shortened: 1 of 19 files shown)


src/ServiceControl.Persistence.EFCore/Implementation/LicensingDataStore.cs
```diff
@@ -1,23 +1,362 @@
...
-    public Task SaveAuditServiceMetadata(AuditServiceMetadata auditServiceMetadata, CancellationToken cancellationToken) => throw new NotImplementedException();
-    public Task SaveBrokerMetadata(BrokerMetadata brokerMetadata, CancellationToken cancellationToken) => throw new NotImplementedException();
-    public Task SaveEndpoint(Particular.LicensingComponent.Contracts.Endpoint endpoint, CancellationToken cancellationToken) => throw new NotImplementedException();
-    public Task SaveLicensedEndpointDetails(LicensedEndpointDetails result, CancellationToken cancellationToken) => throw new NotImplementedException();
-    public Task SaveReportMasks(List<string> reportMasks, CancellationToken cancellationToken) => throw new NotImplementedException();
-    public Task UpdateUserIndicatorOnEndpoints(List<UpdateUserIndicator> userIndicatorUpdates, CancellationToken cancellationToken) => throw new NotImplementedException();
-}
\ No newline at end of file
+    // A report covers the last 14 months, so older throughput is never read.
+    const int ReportedMonths = 14;
```
rbev: Should these constants be shared across Raven and SQL?

rbev (approved):

src/ServiceControl.Persistence.EFCore/Implementation/LicensingDataStore.cs
johnsimons: fixed

PR #5722, opened 2026-08-10
Use setup-ibmmq-action for IBM MQ setup

danielmarbach:
Replaced manual IBM MQ setup with a GitHub Action for better maintainability.


danielmarbach:
FYI the status check is not yet required. Currently the new persistence steps also don't have a required status check. I assume this is a transient thing that will be resolved // cc @johnsimons

johnsimons:
This was folded into #5715

PR #5744, opened 2026-08-11
Propagate cancellation tokens through data store persistence

johnsimons:
Propagate cancellation tokens through licensing data store persistence.
Propagate cancellation tokens through remaining persistence data stores and lifecycle operations
Propagate cancellation tokens through endpoint settings, subscriptions and trial license persistence

### DIFF
--- a/src/ServiceControl.Persistence.EFCore/Infrastructure/RetentionSweeper.cs
+++ b/src/ServiceControl.Persistence.EFCore/Infrastructure/RetentionSweeper.cs
@@ -24,13 +24,13 @@ public class RetentionSweeper(
     static readonly TimeSpan InitialDelay = TimeSpan.FromMinutes(1);
     static readonly TimeSpan BatchPause = TimeSpan.FromSeconds(1);
 
-    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
+    protected override async Task ExecuteAsync(CancellationToken cancellationToken = default)
     {
         logger.LogInformation("Starting retention sweep");
 
         try
         {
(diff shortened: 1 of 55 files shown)


src/ServiceControl.Persistence.EFCore/Infrastructure/RetentionSweeper.cs
```diff
@@ -24,29 +24,32 @@ public class RetentionSweeper(
...
             do
             {
                 try
                 {
-                    await Sweep(pace: true, stoppingToken);
+                    await Sweep(pace: true, cancellationToken);
                 }
+#pragma warning disable PS0019 // The filter already excludes OperationCanceledException, so
+                // cancellation propagates; PS0019 only recognises a cancellationToken guard.
                 catch (Exception ex) when (ex is not OperationCanceledException)
```
rbev, line 45: Could you avoid the pragma by doing 

```suggestion
                catch (Exception ex) when (ex is not OperationCanceledException && cancellationToken.IsCancellationRequested)
```

Also it probably wants to log cancellation exceptions that are not related to the current token

rbev (approved):

src/ServiceControl.Persistence.EFCore/Infrastructure/RetentionSweeper.cs
johnsimons: Our analyser seems to be very picky and always wants `catch (OperationCanceledException ex) when (cancellationToken.IsCancellationRequested)`.
So any deviation from that causes a warning.

PR #5759, opened 2026-08-13
Refactor body storage retrieval to use explicit result states

johnsimons:
Update IBodyStorage and its implementations to return a result object that distinguishes between not found, empty, and unavailable bodies. This allows the API and retry logic to handle these scenarios more accurately and provide better feedback.

### DIFF
--- a/src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSender.cs
+++ b/src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSender.cs
@@ -2,14 +2,15 @@ namespace ServiceControl.Recoverability
 {
     using System;
     using System.Collections.Generic;
+    using System.IO;
     using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.Extensions.Logging;
     using NServiceBus.Routing;
     using NServiceBus.Transport;
-    using ServiceControl.Persistence;
+    using Operations.BodyStorage;
 
-    class ReturnToSender(IFailedMessageRetryDataStore errorMessageStore, ILogger<ReturnToSender> logger)
+    class ReturnToSender(IBodyStorage bodyStorage, ILogger<ReturnToSender> logger)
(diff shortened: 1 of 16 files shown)


src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSender.cs
```diff
@@ -56,21 +57,30 @@ public virtual async Task HandleMessage(MessageContext message, IMessageDispatch
             logger.LogDebug("{MessageId}: Forwarded message to {RetryTo}", messageId, retryTo);
         }
 
-        async Task<byte[]> FetchFromFailedMessage(Dictionary<string, string> outgoingHeaders, string messageId, string attemptMessageId, CancellationToken cancellationToken)
+        async Task<byte[]> FetchFromFailedMessage(Dictionary<string, string> outgoingHeaders, string messageId, CancellationToken cancellationToken)
         {
             var uniqueMessageId = outgoingHeaders["ServiceControl.Retry.UniqueMessageId"];
-            byte[] body = await errorMessageStore.GetFailedMessageBody(uniqueMessageId, cancellationToken);
+            var result = await bodyStorage.TryFetch(uniqueMessageId, cancellationToken);
```
rbev, line 63: Was this the only usage of `GetFailedMessageBody`? if so it should be removed now that it's directly using body storage.

rbev (approved):

src/ServiceControl/Recoverability/Retrying/Infrastructure/ReturnToSender.cs
johnsimons: I thought we are doing that in another PR?

rbev: https://github.com/Particular/ServiceControl/pull/5773

PR #5765, opened 2026-08-13
Implement external integration dispatch requests on EF Core

abparticular:
Implements `IExternalIntegrationRequestsDataStore` for the SQL Server and PostgreSQL persisters, replacing the no-op stub. This is ServiceControl's outbox for publishing NServiceBus events to external subscribers on domain events (message failed, custom check failed, etc.): `StoreDispatchRequest` persists a batch, a background loop wakes up, hands dispatched contexts to the subscribed callback, and deletes the rows once the callback succeeds.

Storage. New `ExternalIntegrationDispatchRequests` table (auto-increment Id for FIFO ordering). DispatchContext's concrete type is only known at runtime and lives in ServiceControl.csproj, which the EF Core project doesn't reference, so there's no way to type this at compile time - a new `DispatchContextSerializer` stores it as a versionless assembly-qualified type name alongside a System.Text.Json blob, and resolves it back via Type.GetType(...) on read. New migrations for both SQL Server and PostgreSQL.

Notification. SQL has no equivalent of RavenDB's server-push Changes API, so the dispatch loop is woken by an in-process signal (SemaphoreSlim) on every `StoreDispatchRequest`, with a 30s PeriodicTimer as a safety net in case a signal is ever missed. This assumes single-process-per-role deployment, same as every other EF Core data store.

Batch dispatch. RavenDB does select → callback → delete inside one document session. There's no relational equivalent, so this is three separate phases, each its own DbContext scope (same shape as RetentionSweeper.Sweep), so no DB transaction is held open across the network call to publish on the bus. Delete only happens if the callback returns without throwing — matching RavenDB's actual behavior (verified by reading TryDispatchEventBatch: it has no try/catch around the callback, so a throw skips the deletes too). This is at-least-once with redelivery on callback failure, not at-most-once.

Settings. Added `ExternalIntegrationsDispatchingBatchSize` (default 100) to EFPersisterSettings, read from the same config key the RavenDB backend already exposes, so switching backends doesn't silently change behavior.

Testing. New shared cross-backend tests (ExternalIntegrationRequestsDataStoreTests, run against RavenDB/SqlServer/PostgreSql) covering store→dispatch→delete, callback-throws-then-retries, batching over ExternalIntegrationsDispatchingBatchSize, and concurrent stores. EF-only tests (ExternalIntegrationRequestsDataStoreEFTests, DispatchContextSerializerTests) exercise the new DispatchNow test hook directly rather than polling on timers, plus a serializer round-trip test.

### DIFF
--- a/src/ServiceControl.Persistence.EFCore/Abstractions/EFPersistenceConfigurationBase.cs
+++ b/src/ServiceControl.Persistence.EFCore/Abstractions/EFPersistenceConfigurationBase.cs
@@ -27,6 +27,7 @@ public abstract class EFPersistenceConfigurationBase : PersistenceConfiguration,
     const string ErrorRetentionPeriodKey = "ErrorRetentionPeriod";
     const string EventsRetentionPeriodKey = "EventsRetentionPeriod";
     const string SubscriptionCacheDurationKey = "SubscriptionCacheDuration";
+    const string ExternalIntegrationsDispatchingBatchSizeKey = "ExternalIntegrationsDispatchingBatchSize";
 
     public PersistenceSettings CreateSettings(SettingsRootNamespace settingsRootNamespace)
     {
@@ -38,6 +39,7 @@ public abstract class EFPersistenceConfigurationBase : PersistenceConfiguration,
         settings.ErrorRetentionPeriod = GetRequiredSetting<TimeSpan>(settingsRootNamespace, ErrorRetentionPeriodKey);
         settings.EventsRetentionPeriod = SettingsReader.Read(settingsRootNamespace, EventsRetentionPeriodKey, EFPersisterSettings.DefaultEventsRetentionPeriod);
         settings.SubscriptionCacheDuration = SettingsReader.Read(settingsRootNamespace, SubscriptionCacheDurationKey, EFPersisterSettings.DefaultSubscriptionCacheDuration);
+        settings.ExternalIntegrationsDispatchingBatchSize = SettingsReader.Read(settingsRootNamespace, ExternalIntegrationsDispatchingBatchSizeKey, 100);
 
         return settings;
     }
--- a/src/ServiceControl.Persistence.EFCore/Abstractions/EFPersistenceConfigurationBase.cs
+++ b/src/ServiceControl.Persistence.EFCore/Abstractions/EFPersistenceConfigurationBase.cs
@@ -39,7 +39,7 @@ public abstract class EFPersistenceConfigurationBase : PersistenceConfiguration,
         settings.ErrorRetentionPeriod = GetRequiredSetting<TimeSpan>(settingsRootNamespace, ErrorRetentionPeriodKey);
         settings.EventsRetentionPeriod = SettingsReader.Read(settingsRootNamespace, EventsRetentionPeriodKey, EFPersisterSettings.DefaultEventsRetentionPeriod);
         settings.SubscriptionCacheDuration = SettingsReader.Read(settingsRootNamespace, SubscriptionCacheDurationKey, EFPersisterSettings.DefaultSubscriptionCacheDuration);
-        settings.ExternalIntegrationsDispatchingBatchSize = SettingsReader.Read(settingsRootNamespace, ExternalIntegrationsDispatchingBatchSizeKey, 100);
+        settings.ExternalIntegrationsDispatchingBatchSize = ReadExternalIntegrationsDispatchingBatchSize(settingsRootNamespace);
 
         return settings;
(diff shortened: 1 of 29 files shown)


johnsimons (approved):

warwickschroeder:
Do these need to be removed?

https://github.com/Particular/ServiceControl/blob/acdce9f5bc0dc5481e02879c4a39c5569281006f/src/ServiceControl.AcceptanceTests.SqlServer/ServiceControl.AcceptanceTests.SqlServer.csproj#L38-L54

src/ServiceControl.Persistence.EFCore/Implementation/ExternalIntegrationRequestsDataStore.cs
```diff
@@ -1,15 +1,207 @@
...
+                    // Shutting down while backing off after a failure - nothing more to do.
+                    break;
+                }
+
+                continue;
+            }
+
+            try
+            {
+                await Task.WhenAny(signal.WaitAsync(cancellationToken), Task.Delay(PollingInterval, timeProvider, cancellationToken));
```
warwickschroeder: Theres something smelly here..

Task.WhenAny will complete if the delay wins, but that doesnt cancel the `signal.WaitAsync`, so that stays queued inside the semaphore. So on next `signal.Release` the wake up goes to that abandoned waiter rather than the new dispatch request?

src/ServiceControl.Persistence.EFCore/Implementation/DispatchContextSerializer.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        var typeName = $"{type.FullName}, {type.Assembly.GetName().Name}";
+        var json = JsonSerializer.Serialize(dispatchContext, type);
+
+        return (typeName, json);
+    }
+
+    public static object Deserialize(string typeName, string json)
+    {
+        var type = Type.GetType(typeName, throwOnError: true)!;
```
warwickschroeder, line 21: Maybe wrap this in a try..catch. Looks like the decoding happens all at once in a batch so one bad decode can block everything behind it.

src/ServiceControl.Persistence.EFCore/Abstractions/EFPersistenceConfigurationBase.cs
```diff
@@ -38,6 +39,7 @@ public PersistenceSettings CreateSettings(SettingsRootNamespace settingsRootName
         settings.ErrorRetentionPeriod = GetRequiredSetting<TimeSpan>(settingsRootNamespace, ErrorRetentionPeriodKey);
         settings.EventsRetentionPeriod = SettingsReader.Read(settingsRootNamespace, EventsRetentionPeriodKey, EFPersisterSettings.DefaultEventsRetentionPeriod);
         settings.SubscriptionCacheDuration = SettingsReader.Read(settingsRootNamespace, SubscriptionCacheDurationKey, EFPersisterSettings.DefaultSubscriptionCacheDuration);
+        settings.ExternalIntegrationsDispatchingBatchSize = SettingsReader.Read(settingsRootNamespace, ExternalIntegrationsDispatchingBatchSizeKey, 100);
```
warwickschroeder: If this is set to zero, it stops dispatching but not the loop, so the table keeps growing. Should either enforce > 0 or disable the loop if 0 is ok.

abparticular: Fixed by https://github.com/Particular/ServiceControl/pull/5765/commits/fbdf1cd05e08f0141e16c93480dbb39b7eb66aa9

src/ServiceControl.Persistence.EFCore/Implementation/DispatchContextSerializer.cs
abparticular: Objects that can't get deserialized are now discarded and an error is logged https://github.com/Particular/ServiceControl/pull/5765/commits/94306036dd575057017563014afde64a037b677b

src/ServiceControl.Persistence.EFCore/Implementation/ExternalIntegrationRequestsDataStore.cs
abparticular: The pending task waiting on the signal is now reused https://github.com/Particular/ServiceControl/pull/5765/commits/e4cd063a38358c79004d69fb7d5a4edde13f4c48

warwickschroeder (approved):

PR #5766, opened 2026-08-13
Fix flaky ETag validation in event log acceptance tests

johnsimons:
The test could fail if new events were written to the log between the initial request and the conditional poll, resulting in a 200 OK instead of a 304 Not Modified. The validation now accounts for this by checking if the ETag has changed when a 200 OK is received.

### DIFF
--- a/src/ServiceControl.AcceptanceTests/EventLogs/When_the_event_log_is_polled_with_an_etag.cs
+++ b/src/ServiceControl.AcceptanceTests/EventLogs/When_the_event_log_is_polled_with_an_etag.cs
@@ -52,12 +52,9 @@ namespace ServiceControl.AcceptanceTests.EventLogs
                         return false;
                     }
 
-                    var current = await Poll(currentEtag);
+                    var current = await PollUntilTheLogGoesQuiet(currentEtag);
 
-                    // The endpoint keeps writing startup events, so a 200 here means either the
-                    // validator went stale between the two requests, or the server ignored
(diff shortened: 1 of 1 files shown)


src/ServiceControl.AcceptanceTests/EventLogs/When_the_event_log_is_polled_with_an_etag.cs
```diff
@@ -45,21 +45,25 @@ await Define<ScenarioContext>()
...
                     }
 
-                    etag = values.FirstOrDefault();
+                    var current = await Poll(currentEtag);
 
-                    if (string.IsNullOrEmpty(etag))
+                    // The endpoint keeps writing startup events, so a 200 here means either the
+                    // validator went stale between the two requests, or the server ignored
+                    // If-None-Match. Only the second is a failure, and the ETag tells them apart.
+                    if (current.StatusCode == HttpStatusCode.OK && ReadEtag(current) != currentEtag)
```
rbev: Does the 200 with a new e-tag make the test inconclusive?
Shouldn't the Poll() be retrying with the new e-tag until the server goes quiet enough to get a conclusive pass or fail?

rbev (approved):

src/ServiceControl.AcceptanceTests/EventLogs/When_the_event_log_is_polled_with_an_etag.cs
johnsimons: Fixed

PR #5788, opened 2026-08-18
Add acceptance test guidelines and the first two scenarios written to them

johnsimons:
Two tests in this suite were recently found to be testing nothing: a double registered against its concrete type, never resolved, passing for years. This writes down how to avoid that, and adds two tests in the shape it argues for.

## Guidelines

`docs/writing-acceptance-tests.md`, linked from `docs/testing.md`.

## The two scenarios

`When_creating_a_usage_report_on_a_non_broker_transport` and `..._on_a_broker_transport`, which
between them cover all eight `api/licensing` routes. Those had no acceptance coverage in any
suite. Two tests rather than one because LearningTransport registers no `IBrokerThroughputQuery`,
so each branch needs its own, and the names say which is which.

## Worth a closer look

Three changes that are neither docs nor new tests:

- **Throughput data is now isolated per test.** `ThroughputDatabaseName` defaults to a fixed name
  and the Raven acceptance storage config set only `DatabaseName`, so every acceptance test was
  sharing one throughput database. The EF suites were unaffected. This is the highest risk change
  here, since it touches every Raven acceptance test.
- **`When_a_critical_error_is_triggered` was not overriding the registration its comment claimed
  it did.** The check ran on its 60 second production interval and the test passed anyway, about
  a minute slower than intended. Now 6 seconds instead of 65.
- **New `CustomizeHostBuilderBeforeServiceControl` hook**, for the narrow case of registrations
  that `AddServiceControl` reacts to. The MultiInstance runner is updated for the new parameter.

## Plan

`docs/acceptance-test-review-plan.md` carries the review this came from, including the confirmed
list of routes with no acceptance test. 40 of 85 had none; this closes 8, leaving 32.

### DIFF
--- a/src/ServiceControl.AcceptanceTests/Licensing/When_creating_a_usage_report_on_a_broker_transport.cs
+++ b/src/ServiceControl.AcceptanceTests/Licensing/When_creating_a_usage_report_on_a_broker_transport.cs
@@ -1,0 +1,193 @@
+namespace ServiceControl.AcceptanceTests.Licensing
+{
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using System.IO.Compression;
+    using System.Linq;
+    using System.Text.Json;
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using AcceptanceTesting.EndpointTemplates;
+    using Microsoft.Extensions.DependencyInjection;
+    using Microsoft.Extensions.Hosting;
+    using Microsoft.Extensions.Logging;
+    using NServiceBus;
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.Routing;
+    using NServiceBus.Transport;
+    using NUnit.Framework;
+    using Particular.LicensingComponent.BrokerThroughput;
+    using Particular.LicensingComponent.Contracts;
+    using Particular.LicensingComponent.MonitoringThroughput;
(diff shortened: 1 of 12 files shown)


src/ServiceControl.AcceptanceTests/Licensing/When_creating_a_usage_report_on_a_broker_transport.cs
```diff
@@ -0,0 +1,193 @@
...
+
+        static JsonDocument ReadReport(byte[] archive)
+        {
+            using var zip = new ZipArchive(new MemoryStream(archive), ZipArchiveMode.Read);
+            using var entry = zip.Entries.Single().Open();
+
+            return JsonDocument.Parse(entry);
+        }
+
+        const string SalesQueue = "Contoso/Sales";
```
rbev, line 155: Using "Contoso" for test strings isn't something we do anywhere else (it's a Microsoft convention), consider changing them throughout this PR.

```suggestion
        const string SalesQueue = "Sales";
```

rbev (approved):

src/ServiceControl.AcceptanceTests/Licensing/When_creating_a_usage_report_on_a_broker_transport.cs
johnsimons: I have another PR to address this

Issue #5028, opened 2025-06-30
Deploy only one ServiceControl container image with roles support


mauroservienti:
#### Is your improvement related to a problem? Please describe.

Today, each instance type is shipped into a different ServiceControl container image. It would be helpful to ship only a single container image with all the required components to run all instances and support the concept of `Roles`. When starting the container, an environment variable, such as `SERVICE_CONTROL_ROLE=Audit`, determines how the container image behaves.

The advantage could be that for small shops or testing purposes, roles can be combined to reduce the topology complexity/footprint by doing something similar to: `SERVICE_CONTROL_ROLE=Primary, Monitoring`

rbev:
A spike creating single entry point for the platform was created as part of discussing potential way forward on this issue: https://github.com/Particular/ServiceControl/pull/5791

PR #5791, opened 2026-08-18
[SPIKE] Single container platform

rbev:
Related to: https://github.com/Particular/ServiceControl/issues/5028

This spike shows a potential solution for building a unified container for ServiceControl instances, it demonstrates a single entrypoint that can run more than one role at once.
Further work will build this into a single container image as the entrypoint


rbev:
Closing.

Can be picked up again in future when we decide to implement https://github.com/Particular/ServiceControl/issues/5028

PR #5794, opened 2026-08-19
Refactor and improve data versioning / etags

warwickschroeder:
# Context

An HTTP server can tell a client "nothing has changed since you last asked" instead of resending the whole answer. It does that by stamping each response with a short opaque label, an *entity-tag*, and the client sends that label back on its next request. If the label still matches, the server answers `304 Not Modified` with no body at all.

Before this branch, ServiceControl built those labels ad hoc: loose strings threaded through the persistence layer, plus a helper (`EtagHelper`) that glued a few fields together with a `StringBuilder`. Different stores disagreed about what an absent label looked like, and the empty string was used to mean "no label", which is dangerous because the empty string matches itself.

## Size

**76% tests and docs, 24% production code**, by lines changed across the whole PR. Even before this PR the test coverage for ETags and data versioning was lacking heavily. By adding these tests, several real defects were descovered and resolved.

## Design

See `docs/data-versioning-design.md`

### Shared

- **One type for a version.** `DataVersion` composes from a backend token, named terms, per-row terms, or several instances combined. Terms are length-prefixed, so user text containing a delimiter cannot make two different results digest identically. Replaces `EtagHelper` and `WithDeterministicEtag`, which hashed on the way out so the store never recognised its own version coming back. Rule and factory guide in `docs/data-versioning-design.md`.
- **Archived groups and retry history return a version.** Both returned bare lists, so the controller invented one. Retry history's named only the historic request ids, so acknowledging an operation changed the body without moving the validator and a dismissed operation came back.
- **Correct conditional GET.** `If-None-Match` now uses RFC 9110 weak comparison rather than `EntityTagHeaderValue.Equals`, which compares strength too. The header is read through typed headers, so a comma-separated list is no longer treated as one malformed value, and `*` is handled. The `304` decision reads the action result's status code, not `Response.StatusCode`, which is not set yet at that point. A `FileStreamResult` replaced by a `304` is registered for disposal.
- **Multi-instance results combine properly.** `DataVersion.Combine` over instance and version pairs replaces sorting and concatenating raw etags, and reports nothing when any instance did not supply one, so a response never claims to cover data it could not version.
- **Endpoints with no honest version say so.** The known-endpoints list is built in memory with no store version behind it, so it now states that rather than emitting an empty validator that can never match.

### RavenDB

- **Paged and filtered reads name their query.** The version was the index etag alone, so page 2 could carry page 1's validator. It now covers the index etag, the total, the page, sort and filters, and the id of every row rendered.
- **Counts name their filters too.** `GetGroupErrorsCount` returned the bare index etag, so the unresolved count and the archived count of one group shared a validator.
- **The overload that omitted the terms is gone.** `ToPagedQueryStatsInfo` and `QueryResultConvert` now require an id selector and the query terms, so a new call site cannot quietly skip them.

### EF Core

- **Paged and filtered reads name their query**, through the same shared `QueryNarrowing.Terms` as Raven. Applied in the custom check, queue address, failure group and failed message query helpers.
- **Custom checks had no validator at all.** The store returned an empty string, so ServicePulse re-downloaded the list on every poll and the acceptance test covering this was excluded on both EF backends.
- **Message bodies revalidate.** The validator was the message id, on the assumption a body never changes. Ingestion upserts the existing row, so a replaced body kept the old validator. It now covers the id and the row's last-modified stamp.
- **Retry history rows are ordered.** Rows are named by position, so an unordered query versioned row order rather than data: historic operations now order by completion time, unacknowledged ones by key.

### Not about data versioning

- **Custom checks dropped their originating endpoint.** The projection never selected the endpoint name, host or host id, so ServicePulse showed a check with no endpoint against it. EF only, found while verifying the above.
- **Custom checks reported the page size as the total count.** The paging links were wrong on every page. Same query, same fix.
- **`QueryStatsInfo.Fresh`.** EF reads cannot be stale, and the alternative was repeating `isStale: false` at every new call site.
- **`GetAuditCountsForEndpointApi` moved to `ScatterGatherRemoteOnly`.** It had to source its version from remotes only, since its own instance holds none of the data. That removed a never-implemented local query and an unused store dependency.
- **Ingestion and clock test helpers moved to the shared test project.** The new store tests need to ingest data and control time on every backend, not just one.
- **Etag acceptance test re-enabled on SQL Server and PostgreSQL.** It was excluded with a note deferring it to a separate PR. This is that PR.

### Outside the error instance

- **Audit RavenDB persister.** All five paged message queries returned the bare index etag, so two pages, two searches or two endpoints shared a validator. `/api/messages` combines the primary's validator with audit's, so fixing only the primary would leave that endpoint wrong. Audit keeps its own `string ETag` type; only the composition changed, mirroring the primary's formatting including ticks-precision timestamps.

## Test coverage

- **The version type.** Every factory, collisions, precision, missing instances.
- **HTTP.** Weak comparison, wildcards, malformed and unknown validators, non-success responses, stream disposal.
- **Stores.** Per-endpoint version tests, plus a conformance suite proving two queries never share a version, on RavenDB, SQL Server and PostgreSQL. Half its cases return nothing, where only the query terms separate them.
- **Message bodies.** Moves when the body is replaced, holds when it is not.
- **Audit.** The same conformance shape over pages, searches and endpoints.
- **End to end.** Etag acceptance tests on all three backends.


src/ServiceControl.Persistence.EFCore/Implementation/BodyStorage/BodyStorage.cs
```diff
@@ -28,12 +29,15 @@ public async Task<MessageBodyResult> TryFetch(string bodyId, CancellationToken c
             return MessageBodyResult.NotFound();
         }
 
-        // Bodies are immutable per message, so the id is a stable ETag.
-        var uniqueMessageId = row.UniqueMessageId.ToString();
+        // Ingestion updates the existing row rather than adding one, so the message id is unchanged
+        // and cannot serve as a version alone. LastModified is written on every upsert.
+        var version = DataVersion.Compose(
```
johnsimons, line 34: @warwickschroeder I haven't done a thorough review yet, but this one seems wrong. We don't need to add another column to this table, as we previously mentioned; bodies are immutable. In other words, the body would only change if a different `UniqueMessageId` was issued.

docs/data-versioning-design.md
```diff
@@ -0,0 +1,64 @@
...
+
+`OverRows` names rows by position, so a caller whose query has no `ORDER BY` has to sort them first or the validator churns.
+
+## Absence
+
+`DataVersion.None` is `default`, and it **matches nothing, not even itself**. Two parties that both know nothing have not established that nothing changed, so an empty-string validator matching itself would answer `304` for every request.
+
+Absence propagates in the safe direction. `WithEtag` writes no header for `None`, so no header means no `If-None-Match`, which means the full body. `Combine` returns `None` as soon as any instance reports none, rather than quietly claiming to cover an instance it could not see.
+
+## Two comparisons, and they are not the same question
```
johnsimons: This is confusing to me and not sure what value it adds?

src/ServiceControl.Persistence.EFCore/Implementation/BodyStorage/BodyStorage.cs
warwickschroeder: @johnsimons from what I'm reading, it doesnt seem to be immutable. The failed message is set via an [upsert](https://github.com/Particular/ServiceControl/blob/53244b541a567a927892a3a9c6ce3cb5b21dd5f9/src/ServiceControl.Persistence.EFCore.SqlServer/SqlServerFailedMessageIngestionSqlDialect.cs#L15), which also [includes the body text](https://github.com/Particular/ServiceControl/blob/53244b541a567a927892a3a9c6ce3cb5b21dd5f9/src/ServiceControl.Persistence.EFCore.SqlServer/SqlServerFailedMessageIngestionSqlDialect.cs#L105). 

It looks like if an already retried message is edited and retried again, the body would be updated due to it [checking for the originals header](https://github.com/Particular/ServiceControl/blob/53244b541a567a927892a3a9c6ce3cb5b21dd5f9/src/ServiceControl.Persistence/Infrastructure/TransportMessageExtensions.cs#L42)?

johnsimons: An edit creates a brand new message.
I am 100% sure that bodies are immutable.
If we are updating the body as part of an upsert, we should not.

johnsimons: I just reviewed the upsert, I think because we don't know whether it is going to be an insert or update, we still need to send the body regardless, but we could skip updating the body if it is an update.

src/ServiceControl.Audit.Persistence.RavenDB/Extensions/RavenQueryStatisticsExtensions.cs
```diff
@@ -1,13 +1,70 @@
...
+        /// cannot: a page with no rows contributes no row terms, so without it two filters that both match
+        /// nothing, and a page past the end, all share one value.
+        /// </summary>
+        public static QueryStatsInfo ToPagedQueryStatsInfo<TRow>(this QueryStatistics stats, IEnumerable<TRow> page, Func<TRow, string> id, params (string Name, object Value)[] query)
         {
-            return new QueryStatsInfo($"{stats.ResultEtag}", stats.TotalResults);
+            // No index etag means no version at all, as on the primary side. The rows here carry only ids,
+            // so the etag is the only term covering a change to a field a row renders; without it a
+            // validator would stand still while that field moved.
+            if (stats.ResultEtag is not { } resultEtag)
```
johnsimons: Would Raven ever return a null for this?

src/ServiceControl.Persistence.EFCore/Implementation/EventLogDataStore.cs
```diff
@@ -43,13 +43,18 @@ public Task<QueryResult<IList<EventLogItemView>>> GetEventLogItems(
...
+                ("total", total),
+                ("newest", stats?.Newest),
+                ("highestId", stats?.HighestId),
+                ("page", pagingInfo.Page),
+                ("pageSize", pagingInfo.PageSize));
+            var queryStats = QueryStatsInfo.Fresh(version, total);
 
             // The point of knownVersion. Everything above is index work.
             // If the caller already has the latest version, skip the rest of the query.
             // No database round trip is needed. No response body is needed.
```
johnsimons: Probably no longer needed

src/ServiceControl.Persistence.EFCore/Implementation/QueueAddressStore.cs
```diff
@@ -20,8 +21,9 @@ public Task<QueryResult<IList<QueueAddress>>> GetAddresses(PagingInfo pagingInfo
                 });
 
             var items = await query.Skip(pagingInfo.Offset).Take(pagingInfo.PageSize).ToListAsync(token);
-            var eTag = DeterministicGuid.MakeId($"{items.Count}|{string.Join(",", items.Select(x => x.PhysicalAddress))}").ToString();
+            var addressCount = await query.CountAsync(token);
```
johnsimons, line 24: Is this right? Wouldn't we need to check whether the list is the same?

src/ServiceControl.Persistence.EFCore/Implementation/RetryBatchStore.cs
```diff
@@ -111,7 +111,8 @@ public Task<QueryResult<IList<RetryBatch>>> GetOrphanedBatches(string retrySessi
 
             IList<RetryBatch> batches = [.. orphaned.Select(batch => batch.ToRetryBatch(messageCounts.GetValueOrDefault(batch.Id)))];
 
-            return new QueryResult<IList<RetryBatch>>(batches, new QueryStatsInfo(string.Empty, batches.Count, false));
+            // No version: orphaned batches are consumed by the retry session, never by a caching client.
+            return new QueryResult<IList<RetryBatch>>(batches, QueryStatsInfo.Fresh(DataVersion.None, batches.Count));
```
johnsimons: This is what we talked about @warwickschroeder, this one is used for internal consumption so, it is not required

src/ServiceControl.Persistence.EFCore/Infrastructure/CustomCheckQueries.cs
```diff
@@ -0,0 +1,18 @@
...
+
+static class CustomCheckQueries
+{
+    /// <summary>
+    /// Every field of every check the body shows, plus the total. OriginatingEndpoint has no term of its
+    /// own and does not need one: Id is a deterministic hash of the endpoint name, its host id and the
+    /// check id, so naming Id covers all three, and the host string is written once on insert and never
+    /// updated.
+    /// </summary>
+    public static QueryStatsInfo ToQueryStatsInfo(this IReadOnlyCollection<CustomCheck> page, long totalCount, params (string Name, object? Value)[] query) =>
```
johnsimons: This is only referenced in one place. Do we really need to have a class on its own?

src/ServiceControl.Persistence.EFCore/Infrastructure/QueueAddressQueries.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using ServiceControl.MessageFailures;
+using ServiceControl.Persistence.Infrastructure;
+
+static class QueueAddressQueries
+{
+    /// <summary>
+    /// Both fields of every address the body shows, plus the total behind Total-Count.
+    /// </summary>
+    public static QueryStatsInfo ToQueryStatsInfo(this IReadOnlyCollection<QueueAddress> page, long totalCount, params (string Name, object? Value)[] query) =>
```
johnsimons: same as before, only referenced in one place

src/ServiceControl.Persistence.EFCore/Infrastructure/RetryHistoryQueries.cs
```diff
@@ -0,0 +1,35 @@
+namespace ServiceControl.Persistence.EFCore.Infrastructure;
+
+using ServiceControl.Persistence.Infrastructure;
+using ServiceControl.Recoverability;
+
+static class RetryHistoryQueries
```
johnsimons: only referenced by one class,

src/ServiceControl.Persistence.RavenDB/Recoverability/GroupsDataStore.cs
```diff
@@ -66,7 +68,7 @@ public async Task<QueryResult<FailureGroupView>> GetUnresolvedGroup(string group
                 .FilterByLastModifiedRange(modified)
                 .FirstOrDefaultAsync(cancellationToken);
 
-            return new QueryResult<FailureGroupView>(document, stats.ToQueryStatsInfo());
```
johnsimons, line 69: i don't entirely understand why we need to change Raven, given Raven has its own etag buildt into it?

johnsimons (commented):
I'm halfway through reviewing this PR.
One thing that isn't sitting well with me is that I don't think the data storage layer should need to be aware of ETags.

Imagine an internal service that needs data from two different data storage implementations. That service doesn't care about ETags. The only thing that really cares about them is the HTTP layer.
So in my opinion, the data stores shouldn't have to return `QueryResult` at all. They should return just the data, and something sitting between the controller/action and the data storage should be responsible for adding the ETag information.

I can understand how this separation can be hard to achieve given RavenDB mixes both worlds.

src/ServiceControl/CompositeViews/Messages/ScatterGatherApi.cs
```diff
@@ -103,12 +98,30 @@ internal QueryResult<TOut> AggregateResults(TIn input, QueryResult<TOut>[] resul
 
         protected abstract TOut ProcessResults(TIn input, QueryResult<TOut>[] results);
 
-        protected virtual QueryStatsInfo AggregateStats(TIn input, IEnumerable<QueryResult<TOut>> results, TOut processedResults)
+        protected virtual QueryStatsInfo AggregateStats(TIn input, IEnumerable<QueryResult<TOut>> results, TOut processedResults) =>
+            Aggregate(results);
+
+        /// <summary>
+        /// For an API whose own instance holds none of the data. Its local result carries no version, and
```
rbev: I'm not clear on why this is treated differently.

If a remote instance has no data and forces it to none why should a local instance be excluded from that?

src/ServiceControl.Persistence.RavenDB/RavenCustomCheckDataStore.cs
```diff
@@ -57,7 +57,8 @@ public async Task<QueryResult<IList<CustomCheck>>> GetStats(PagingInfo paging, s
                 .Paging(paging)
                 .ToListAsync(cancellationToken);
 
-            return new QueryResult<IList<CustomCheck>>(results, new QueryStatsInfo($"{stats.ResultEtag}", stats.TotalResults, stats.IsStale));
+            return new QueryResult<IList<CustomCheck>>(results,
+                stats.ToPagedQueryStatsInfo(results, check => check.Id, ("status", status), ("page", paging.Page), ("pageSize", paging.PageSize)));
```
rbev: Shouldn't this just return `stats.ResultsEtag` in the version struct?

src/ServiceControl.Persistence.RavenDB/RavenQueryStatisticsExtensions.cs
```diff
@@ -1,18 +1,36 @@
...
     using ServiceControl.Persistence.Infrastructure;
 
     static class RavenQueryStatisticsExtensions
     {
-        public static QueryStatsInfo ToQueryStatsInfo(this QueryStatistics stats)
-        {
-            return new QueryStatsInfo($"{stats.ResultEtag}", stats.TotalResults, stats.IsStale);
-        }
+        /// <summary>
+        /// For a paged query. The index etag covers whether the data moved, and the row ids cover which
```
rbev: I'm confused as to the reason this needs to be done, are you saying the etag in raven is based on the whole table? 
My understanding is that it was already aggregated for you from the result set it returned

src/ServiceControl.Persistence.Tests/VersionAssert.cs
```diff
@@ -0,0 +1,46 @@
...
+        using (Assert.EnterMultipleScope())
+        {
+            Assert.That(one.HasValue, Is.True, "the first query produced no version to compare");
+            Assert.That(other.HasValue, Is.True, "the second query produced no version to compare");
+            Assert.That(other.Matches(one), Is.False, because);
+        }
+    }
+
+    /// <summary>Nothing changed, so a caller holding the earlier version still holds the current one.</summary>
+    public static void Held(DataVersion first, DataVersion second, string because)
```
rbev: Should this be `VersionAssert.Matches()`? "Held" seems like how this assertion is used rather than what it's asserts

src/ServiceControl.Persistence.Tests/VersionAssert.cs
```diff
@@ -0,0 +1,46 @@
...
+            Assert.That(after.HasValue, Is.True, "the version went missing rather than moving");
+            Assert.That(after.Matches(before), Is.False, because);
+        }
+    }
+
+    /// <summary>
+    /// Two different queries, answered at the same instant. Neither describes the other, so a caller
+    /// holding one must never be told the other is current.
+    /// </summary>
+    public static void Distinct(DataVersion one, DataVersion other, string because)
```
rbev: This is the same code as `Moved` with different messages, does it need to be it's own assertion?

src/ServiceControl.Persistence/Infrastructure/QueryNarrowing.cs
```diff
@@ -0,0 +1,22 @@
+namespace ServiceControl.Persistence.Infrastructure
+{
+    /// <summary>
+    /// The page, ordering and filters a read was narrowed by, expressed as version terms.
+    /// <para>
+    /// A version over a list normally tells two queries apart by the rows it returns. A query that matches
+    /// nothing returns no rows and so contributes no terms, which leaves every empty view of the same data
```
rbev: I thought that blank results were meant to return an `DataVersion.None` anyway so that they weren't flagged as unchanged.

src/ServiceControl.UnitTests/ScatterGather/MessageView_ScatterGatherTest.cs
```diff
@@ -39,7 +39,7 @@ QueryResult<IList<MessagesView>> GetPage(IEnumerable<MessagesView> source, strin
 
             return new QueryResult<IList<MessagesView>>(
                 pageOfResults,
-                new QueryStatsInfo(etag, allResults.Count, isStale: false))
+                QueryStatsInfo.Fresh(DataVersion.FromToken(etag), allResults.Count))
```
rbev: Given the "isStale" flag is exposing a ravenDb concept directly could it be replaced by returning ETags correcly? I know that the EF implementation always just returns `false`.

Or is it used downstream for something else?

docs/data-versioning-design.md
warwickschroeder: Removed

src/ServiceControl.Persistence.EFCore/Implementation/EventLogDataStore.cs
warwickschroeder: After looking into if this would benefit any other endpoints (it does not) I agree we should remove this for the sake of simplicity.

src/ServiceControl.Persistence.EFCore/Infrastructure/CustomCheckQueries.cs
warwickschroeder: Its following the pattern of `FailureGroupQueries`, but I could consolidate all of these helper functions into a single class?

src/ServiceControl.Persistence.EFCore/Implementation/BodyStorage/BodyStorage.cs
warwickschroeder: Ok, I can look into doing that so it is 100% immutable. Then I can remove the additional field.

src/ServiceControl.Persistence.RavenDB/RavenQueryStatisticsExtensions.cs
warwickschroeder: You are right, I'm going to revert this. I was thinking too much from a "backend" PoW and forgot about what these versions would be used for and who will be using it. An identical etag for page 1 and page 2 in ServicePulse is fine as they have different URLs, and SP doesnt do anything with etags iteself either. This may not be the case for other clients depending on how they consume and use the etag, but we dont have those, so ill simplify.

src/ServiceControl.Persistence.EFCore/Implementation/QueueAddressStore.cs
warwickschroeder: We are. `ToQueryStatsInfo` is an extension method off `List<QueueAddress>`, so it does check the items too.

src/ServiceControl.Audit.Persistence.RavenDB/Extensions/RavenQueryStatisticsExtensions.cs
warwickschroeder: It is nullable so possibly, and if it is null we should be sending a `DataVersion.None` rather than a `null` or empty string. This entire file has been refactored quite a bit to simplify however.

src/ServiceControl.Persistence.EFCore/Implementation/RetryBatchStore.cs
warwickschroeder: Simplified

src/ServiceControl.Persistence.EFCore/Infrastructure/CustomCheckQueries.cs
warwickschroeder: I've consolidated all these into an extensions file.

src/ServiceControl.Persistence.RavenDB/Recoverability/GroupsDataStore.cs
warwickschroeder: Reverted to just using ravens `ResultEtag`

src/ServiceControl.Persistence.RavenDB/RavenCustomCheckDataStore.cs
warwickschroeder: Reverted. No longer taking paging into account.

src/ServiceControl.Persistence.Tests/VersionAssert.cs
warwickschroeder: Removed

src/ServiceControl.Persistence/Infrastructure/QueryNarrowing.cs
warwickschroeder: Removed the entire file. No longer needed, as discussed in another comment

src/ServiceControl.UnitTests/ScatterGather/MessageView_ScatterGatherTest.cs
warwickschroeder: The only places that its used is in `RetryDocumentManager.cs`, which sits downstream. Its only that one place that needs it though so I've taken IsStale out of the QueryStatsInfo class (which is used almost everywhere) and created an `OrphanedBatches` object to use instead only in the Retry logic

src/ServiceControl.Persistence.RavenDB/Recoverability/RetryHistoryDataStore.cs
```diff
@@ -3,20 +3,32 @@
...
+    using ServiceControl.Persistence.Infrastructure;
     using ServiceControl.Recoverability;
 
     class RetryHistoryDataStore(IRavenSessionProvider sessionProvider) : IRetryHistoryDataStore
     {
         const string DocumentId = "RetryOperations/History";
 
-        public async Task<RetryHistory> GetRetryHistory(CancellationToken cancellationToken = default)
+        // Before the first operation completes there is no document and so no change vector, but an
+        // empty history still has to be cacheable.
```
rbev: nitpick: 

Why does an empty history need to be cacheable? If we are still hitting the DB to figure out that it's empty to check the ETag then the difference between returning an empty array payload and having that response cached against an ETag is marginal.

rbev (approved):

src/ServiceControl.Persistence.RavenDB/Recoverability/RetryHistoryDataStore.cs
warwickschroeder: Sure - I don't mind either way to be honest. I guess by treating an empty body the same as if it wasn't empty keeps things consistent though. Ill update the comment to clarify this.

src/ServiceControl/CompositeViews/Messages/ScatterGatherApi.cs
warwickschroeder: A remote instance with no data still runs the query and still returns a version, so it never forces `None`. `None` only comes from a remote that could not be reached, and the local instance (error instance here) is excluded because for these endpoints it holds none of the audit data and shouldnt be queried at all.

warwickschroeder:
> I'm halfway through reviewing this PR. One thing that isn't sitting well with me is that I don't think the data storage layer should need to be aware of ETags.
> 
> Imagine an internal service that needs data from two different data storage implementations. That service doesn't care about ETags. The only thing that really cares about them is the HTTP layer. So in my opinion, the data stores shouldn't have to return `QueryResult` at all. They should return just the data, and something sitting between the controller/action and the data storage should be responsible for adding the ETag information.
> 
> I can understand how this separation can be hard to achieve given RavenDB mixes both worlds.

@johnsimons - So the database implementations dont know about ETags, but they do "version" their own data in the most efficient way they can. Currently the versioning approach differ between SQL and Raven. Raven have a native way to generate a version off its index state, but SQL doesnt. So there are essentially 2 options here:

1. Leave as is. Raven versions its data in its own way, and SQL/postgres does it there way. This approach we require an object wrapper where we can store both the result and its version. Ravens version is [requested off the DB query](https://github.com/Particular/ServiceControl/blob/1d559be21d510355d3ad25075d37e91a41484e7f/src/ServiceControl.Persistence.RavenDB/EventLogDataStore.cs#L37).
2. Forgo the native raven versioning and standardise on [`DataVersion.Compose`](https://github.com/Particular/ServiceControl/blob/1d559be21d510355d3ad25075d37e91a41484e7f/src/ServiceControl.Persistence/Infrastructure/DataVersion.cs#L41) which creates a deterministic guid from an incremental hash of the data. If we did this we could include this process as a filter in the web pipeline, and just have the datastores return pure data.

johnsimons (approved):

PR #5811, opened 2026-08-21
Create aspire harness for testing error instance SQL migration

rbev:
This PR contains a harness that was used to test that the migration flow to a nee primary instance [documented here](https://docs.particular.net/servicecontrol/migrations/replacing-error-instances/).


rbev:
Closing this as it has been decided that there is not a good case for keeping this in the codebase for future testing.

PR #5824, opened 2026-08-25
Switch to standard OpenTelemetry environment variables for OTLP configuration

johnsimons:
Removes the custom OtlpEndpointUrl setting in favor of standard environment variables such as OTEL_EXPORTER_OTLP_ENDPOINT and OTEL_EXPORTER_OTLP_METRICS_ENDPOINT. This ensures compatibility with native OpenTelemetry SDK configuration, supports both gRPC and HTTP protocols, and simplifies instance configuration.

This came from https://github.com/Particular/ServiceControl/pull/5816#discussion_r3840896578

### DIFF
--- a/src/ServiceControl.Audit/Infrastructure/Settings/Settings.cs
+++ b/src/ServiceControl.Audit/Infrastructure/Settings/Settings.cs
@@ -148,7 +148,6 @@
         public int Port { get; set; }
 
         public bool PrintMetrics => SettingsReader.Read<bool>(SettingsRootNamespace, "PrintMetrics");
-        public string OtlpEndpointUrl { get; set; } = SettingsReader.Read<string>(SettingsRootNamespace, nameof(OtlpEndpointUrl));
         public string Hostname { get; private set; }
         public string VirtualDirectory => SettingsReader.Read(SettingsRootNamespace, "VirtualDirectory", string.Empty);
 
--- a/src/ServiceControl.Infrastructure/OtlpEndpoint.cs
+++ b/src/ServiceControl.Infrastructure/OtlpEndpoint.cs
@@ -1,0 +1,15 @@
+namespace ServiceControl.Infrastructure;
+
+using System;
+
+public static class OtlpEndpoint
+{
+    public static string MetricsEndpointFromEnvironment() =>
+        Read("OTEL_EXPORTER_OTLP_METRICS_ENDPOINT") ?? Read("OTEL_EXPORTER_OTLP_ENDPOINT");
+
+    static string Read(string variable)
+    {
+        var value = Environment.GetEnvironmentVariable(variable);
+        return string.IsNullOrWhiteSpace(value) ? null : value;
+    }
+}
--- a/src/ServiceControl.Infrastructure/OtlpEndpoint.cs
+++ b/src/ServiceControl.Infrastructure/OtlpEndpoint.cs
@@ -1,15 +1,28 @@
 namespace ServiceControl.Infrastructure;
 
 using System;
+using Microsoft.Extensions.Configuration;
 
 public static class OtlpEndpoint
 {
-    public static string MetricsEndpointFromEnvironment() =>
-        Read("OTEL_EXPORTER_OTLP_METRICS_ENDPOINT") ?? Read("OTEL_EXPORTER_OTLP_ENDPOINT");
+    // This is the default environment variable name used by the OpenTelemetry .NET SDK to configure the OTLP exporter endpoint
+    // as specified in https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md.
+    const string EndpointKey = "OTEL_EXPORTER_OTLP_ENDPOINT";
 
-    static string Read(string variable)
+    public static Uri Read(IConfiguration configuration)
     {
-        var value = Environment.GetEnvironmentVariable(variable);
-        return string.IsNullOrWhiteSpace(value) ? null : value;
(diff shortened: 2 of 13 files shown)


src/ServiceControl.Audit/Infrastructure/Settings/Settings.cs
```diff
@@ -148,7 +148,6 @@ public string RootUrl
         public int Port { get; set; }
 
         public bool PrintMetrics => SettingsReader.Read<bool>(SettingsRootNamespace, "PrintMetrics");
-        public string OtlpEndpointUrl { get; set; } = SettingsReader.Read<string>(SettingsRootNamespace, nameof(OtlpEndpointUrl));
```
warwickschroeder, line 151: Is this not a potential breaking change? Should a start up check be create or use `SettingInfo.RemovedFrom`?

src/ServiceControl.Monitoring/App.config
```diff
@@ -19,9 +19,11 @@ These settings are only here so that we can debug ServiceControl while developin
...
     <!--<add key="Monitoring/TransportType" value="SQLServer" />-->
 
-	<!-- options are any comma separated combination of NLog,Seq,Otlp -->
-	<add key="Monitoring/LoggingProviders" value="NLog,Seq"/>
-	<add key="Monitoring/SeqAddress" value="http://localhost:5341"/>
+    <!-- options are any comma separated combination of NLog,Seq,Otlp -->
+    <add key="Monitoring/LoggingProviders" value="NLog,Seq"/>
+    <add key="Monitoring/SeqAddress" value="http://localhost:5341"/>
+	
+    <!-- OTLP logs and metrics are configured with the standard OpenTelemetry environment variables, e.g. OTEL_EXPORTER_OTLP_ENDPOINT -->
```
warwickschroeder, line 26: Does monitoring emit OTLP metrics? There wasnt any setting for it.

src/ServiceControl.Infrastructure/OtlpEndpoint.cs
```diff
@@ -0,0 +1,15 @@
+namespace ServiceControl.Infrastructure;
+
+using System;
+
+public static class OtlpEndpoint
+{
+    public static string MetricsEndpointFromEnvironment() =>
+        Read("OTEL_EXPORTER_OTLP_METRICS_ENDPOINT") ?? Read("OTEL_EXPORTER_OTLP_ENDPOINT");
```
warwickschroeder: It looks like `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` isnt supported when using `AddOtlpExporter`  - https://github.com/open-telemetry/opentelemetry-dotnet/blob/core-1.17.0/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/README.md?plain=1#L520-L530

src/ServiceControl/HostApplicationBuilderExtensions.cs
```diff
@@ -152,14 +152,11 @@ public static void AddIngestionMetrics(this IHostApplicationBuilder hostBuilder,
         {
             hostBuilder.Services.AddSingleton<IngestionMetrics>();
 
-            if (string.IsNullOrEmpty(settings.OtlpEndpointUrl))
-            {
-                return;
-            }
+            var otlpEndpoint = OtlpEndpoint.MetricsEndpointFromEnvironment();
 
-            if (!Uri.TryCreate(settings.OtlpEndpointUrl, UriKind.Absolute, out var otlpEndpoint))
```
warwickschroeder, line 160: Have you confirmed that the OpenTelemetry library does indeed fail an invalid URL?

src/ServiceControl.Infrastructure/OtlpEndpoint.cs
```diff
@@ -0,0 +1,15 @@
...
+
+    static string Read(string variable)
+    {
+        var value = Environment.GetEnvironmentVariable(variable);
```
warwickschroeder: Should we read through `IConfiguration` rather direct from env var? This would provide more flexibility to the user in how they set this.

src/ServiceControl.Audit/Infrastructure/Settings/Settings.cs
johnsimons: It is, but I have looked at our docs, and this setting is not documented at all. I also remember talking to Andreas, and it seems this was added so that we could more easily diagnose performance issues when customers are facing them. So it is a tool that we use for diagnostics. Hence, the reason it is not publicly documented.

src/ServiceControl.Monitoring/App.config
johnsimons: No, it does not, but I left it in mostly for consistency, given it does not cause any issues.

src/ServiceControl.Infrastructure/OtlpEndpoint.cs
johnsimons: I removed that and mentioned it in the docs

src/ServiceControl/HostApplicationBuilderExtensions.cs
johnsimons: It now fails if the URL is invalid, so it no longer defaults.

src/ServiceControl.Infrastructure/OtlpEndpoint.cs
johnsimons: done

warwickschroeder (approved):

PR #5827, opened 2026-08-26
Simplify and clarify Platform Install experience on SCMU

jpalac:
Reduce the friction for the minimal install of the platform needed to provide the usage report, by making the intent of what is being installed clear from the start.

Changes based on the below mockup with a few improvements.

image

### DIFF
--- a/src/ServiceControl.Config/UI/Shell/NewInstancePopup.xaml.cs
+++ b/src/ServiceControl.Config/UI/Shell/NewInstancePopup.xaml.cs
@@ -29,9 +29,9 @@ namespace ServiceControl.Config.UI.Shell
 
             CbServiceControl.IsChecked = true;
             CbServicePulse.IsChecked = true;
+            CbServicePulse.Opacity = 1.0;
             CbAudit.IsChecked = true;
 
-            CbServicePulse.Opacity = 1.0;
             NextButton.IsEnabled = true;
 
             suppressPresetEvents = true;
@@ -48,7 +48,7 @@ namespace ServiceControl.Config.UI.Shell
 
         void Preset_Checked(object sender, RoutedEventArgs e)
         {
-            if (CbServiceControl == null || suppressPresetEvents)
+            if (!IsLoaded || suppressPresetEvents)
             {
                 return;
             }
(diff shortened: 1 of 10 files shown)


andreasohlund (approved):

src/ServiceControl.Config/UI/Shell/NewInstancePopup.xaml.cs
```diff
@@ -1,13 +1,208 @@
...
+
+            suppressCheckBoxEvents = true;
+
+            CbMonitoring.IsChecked = false;
+            MonitoringInfoBox.Visibility = Visibility.Collapsed;
+            SetServiceControlCheckboxesEnabled(true);
+
+            var (sc, sp, audit) = sender switch
+            {
+                _ when sender == PresetMinimal => (true, true, false),
```
rbev, line 64: these presets are defined twice, it might be nice to centralise them even if it means losing the switch statement in SyncPresetSelection

rbev (approved):

ramonsmits:
@jpalac When I look at that image.. I don't seem to understand it.. why is monitoring not added in a full setup?

Also, I don't understand the -- OR INSTALL SEPARATELY --

From an UX perspective, most users likely don't know at all what it all means. Why not list the features as big buttons where some are already selected by default?

- User interface (web application for users)
- Recoverability (ability to retry failed / dead lettered messages)
- Endpoint metrics (message processing metrics per endpoint)
- Auditing (Rotating message store there succesfully processed message are stored for a configurable retention window)

I very much like these top buttons! However, FULL implies everything to me that feels it should be fully to the right? Maybe minimal, audit only, default, full). Just not sure if that really clear to our users.

Take minimal... what would that then be? UI + primary instance? I guess for most users a UI is really required.

I guess the installer could also detect if SP is already installed and based on that suggest that the UI is not required OR after installation indicate that the exiting SP install can be removed or even better an option to uninstall it?

jpalac:
@ramonsmits Thanks for your input. We have taken some of that into consideration and are closing this PR in favour of https://github.com/Particular/ServiceControl/pull/5837.
That hopefully makes the screen more user friendly. This is being done as part of a multi-topic so we don't want to rock the boat too much - we are not looking to redo the whole screen, but rather make some improvements to it.

PR #5828, opened 2026-08-26
Report deployment and storage environment data in the usage report

johnsimons:
The usage report already carries throughput, versions and whether audit and monitoring are in
use. It says nothing about how the instance is actually deployed or what it stores data in, so
questions like "how many customers run in containers", "who has moved off embedded RavenDB" or
"is anyone using S3 body storage with an IAM role" cannot be answered from the reports we receive.

This adds environment data covering the host, the persistence, the body storage and the
instance's configuration posture.

## Keys

| Group | Keys |
| --- | --- |
| Host | `Host.Model` (Container / WindowsService / Console), `Host.Orchestrator`, `Host.OSPlatform`, `Host.OSVersion`, `Host.Architecture`, `Host.RuntimeVersion`, `Host.ProcessorCount`, `Host.AvailableMemoryGB` |
| Persistence | `Persistence.Type`, `Persistence.RavenServer` (Embedded / External), `Persistence.Hosting`, `Persistence.ServerVersion`, `Persistence.FullTextSearch` |
| Body storage | `Persistence.BodyStorage.Type`, `Persistence.BodyStorage.Auth`, `Limits.MaxBodySizeToStore` |
| Posture | `Security.Authentication`, `Security.RoleBasedAuthorization`, `Security.Https`, `Features.MessageEditing`, `Features.ExternalIntegrationsPublishing`, `Features.ForwardErrorMessages`, `Features.EmailNotifications`, `Retention.ErrorHours`, `Retention.AuditHours`, `Retention.EventsHours` |

Keys are dotted strings, following the `Features.IntegratedServicePulse` key that was already
there. `EnvironmentDataType` is left alone rather than grown, so adding a datum no longer means
changing a contract.

## No customer data

Every value is a fixed enum member, a boolean, a count or a version number. Nothing derived from
a host name, URL, connection string, bucket or container name, managed identity client id, file
path, machine name or endpoint name is emitted. `RuntimeInformation.OSDescription` is
deliberately not used because on Linux it carries a distro build string that identifies a
specific image. An acceptance test asserts the report carries no value matching the machine name,
and a unit test asserts no S3 bucket name or access key reaches the report.

## Cloud database detection

`Persistence.Hosting` is resolved by asking the engine where it can, and by classifying the
configured host name where it cannot.

SQL Server answers definitively through `SERVERPROPERTY('EngineEdition')`, where 5, 8 and 11 mean
Azure SQL Database, Managed Instance and Synapse. PostgreSQL has no equivalent, so the probe
checks for the administrative roles that each managed offering creates and a self-hosted server
does not: `azure_pg_admin`, `rds_superuser` and `cloudsqlsuperuser`. Both probes run with a five
second command timeout and fall back to host suffix matching on any failure, so a report is never
held up by a database that will not answer.

`Persistence.ServerVersion` is the engine major version only. SQL Server's minor is always zero,
and Azure SQL Database reports major 12 regardless of the engine actually running, so
`major.minor` would add a digit that carries no information. Patch level, if it is ever wanted,
belongs in a separate key sourced from `ProductUpdateLevel`.

### DIFF
--- a/src/ServiceControl.Persistence.EFCore.PostgreSql/PostgreSqlDatabaseHostingProbe.cs
+++ b/src/ServiceControl.Persistence.EFCore.PostgreSql/PostgreSqlDatabaseHostingProbe.cs
@@ -34,7 +34,9 @@ class PostgreSqlDatabaseHostingProbe(PostgreSqlPersisterSettings settings, IServ
                 return HostingFromConnectionString();
             }
 
-            return new DatabaseHosting(HostingFromRoles(reader), MajorVersion(reader));
+            var hosting = HostingFor(reader.GetBoolean(1), reader.GetBoolean(2), reader.GetBoolean(3), ConfiguredHost);
+
+            return new DatabaseHosting(hosting, MajorVersion(reader), DatabaseHostingSource.Probe);
         }
         catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
         {
@@ -48,21 +50,32 @@ class PostgreSqlDatabaseHostingProbe(PostgreSqlPersisterSettings settings, IServ
         }
     }
 
-    // PostgreSQL has no equivalent of SQL Server's EngineEdition, but every managed offering creates
-    // a distinctive administrative role that a self-hosted server does not have.
-    string HostingFromRoles(DbDataReader reader)
+    /// <summary>
+    /// PostgreSQL has no equivalent of SQL Server's EngineEdition, but every managed offering
+    /// creates a distinctive administrative role that a self-hosted server does not have. A server
+    /// that answered and has none of them is self-hosted, unless its host name names a managed
+    /// service that does not announce itself this way.
+    /// </summary>
+    internal static string HostingFor(bool azure, bool rds, bool cloudSql, string? host)
     {
-        if (reader.GetBoolean(1))
+        if (azure)
         {
             return "AzurePostgres";
         }
 
-        if (reader.GetBoolean(2))
+        if (rds)
         {
             return "AwsRds";
         }
 
-        return reader.GetBoolean(3) ? "GoogleCloudSql" : HostingFromConnectionString().Hosting;
+        if (cloudSql)
+        {
+            return "GoogleCloudSql";
+        }
+
+        var classified = DatabaseHostClassifier.Classify(host);
+
+        return classified == DatabaseHostClassifier.Unknown ? DatabaseHostClassifier.SelfHosted : classified;
     }
 
     static string MajorVersion(DbDataReader reader) =>
@@ -70,17 +83,27 @@ class PostgreSqlDatabaseHostingProbe(PostgreSqlPersisterSettings settings, IServ
 
     DatabaseHosting HostingFromConnectionString()
     {
-        try
-        {
-            var host = new NpgsqlConnectionStringBuilder(settings.ConnectionString).Host;
+        var host = ConfiguredHost;
 
-            return new DatabaseHosting(DatabaseHostClassifier.Classify(host), DatabaseHostClassifier.Unknown);
-        }
-        catch (Exception e)
-        {
-            logger.LogDebug(e, "Could not classify the configured PostgreSQL host");
+        return host is null
+            ? DatabaseHosting.Unclassified
+            : new DatabaseHosting(DatabaseHostClassifier.Classify(host), DatabaseHostClassifier.Unknown, DatabaseHostingSource.ConnectionString);
+    }
 
-            return DatabaseHosting.Unavailable;
+    string? ConfiguredHost
+    {
+        get
+        {
+            try
+            {
+                return new NpgsqlConnectionStringBuilder(settings.ConnectionString).Host;
+            }
+            catch (Exception e)
+            {
+                logger.LogDebug(e, "Could not read the configured PostgreSQL host");
+
+                return null;
+            }
         }
     }
 
--- a/src/ServiceControl.Persistence.EFCore.PostgreSql/PostgreSqlDatabaseHostingProbe.cs
+++ b/src/ServiceControl.Persistence.EFCore.PostgreSql/PostgreSqlDatabaseHostingProbe.cs
@@ -1,0 +1,95 @@
+namespace ServiceControl.Persistence.EFCore.PostgreSql;
+
+using System.Data.Common;
+using System.Globalization;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Npgsql;
+using ServiceControl.Persistence;
+using ServiceControl.Persistence.EFCore.DbContexts;
+using ServiceControl.Persistence.EFCore.Infrastructure;
+
+class PostgreSqlDatabaseHostingProbe(PostgreSqlPersisterSettings settings, IServiceScopeFactory scopeFactory, ILogger<PostgreSqlDatabaseHostingProbe> logger) : IDatabaseHostingProbe
+{
+    public string StorageName => "PostgreSQL";
+
+    public async Task<DatabaseHosting> Probe(CancellationToken cancellationToken = default)
+    {
+        try
+        {
+            await using var scope = scopeFactory.CreateAsyncScope();
+            var dbContext = scope.ServiceProvider.GetRequiredService<ServiceControlDbContext>();
+
+            await using var command = dbContext.Database.GetDbConnection().CreateCommand();
+            command.CommandText = ProbeSql;
+            command.CommandTimeout = ProbeTimeoutSeconds;
+
+            await dbContext.Database.OpenConnectionAsync(cancellationToken);
+
+            await using var reader = await command.ExecuteReaderAsync(cancellationToken);
+
+            if (!await reader.ReadAsync(cancellationToken))
+            {
+                return HostingFromConnectionString();
+            }
+
+            return new DatabaseHosting(HostingFromRoles(reader), MajorVersion(reader));
+        }
+        catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested)
+        {
+            throw;
+        }
+        catch (Exception e)
+        {
+            logger.LogDebug(e, "Could not ask PostgreSQL how it is hosted, falling back to the connection string");
+
+            return HostingFromConnectionString();
+        }
+    }
+
+    // PostgreSQL has no equivalent of SQL Server's EngineEdition, but every managed offering creates
+    // a distinctive administrative role that a self-hosted server does not have.
+    string HostingFromRoles(DbDataReader reader)
+    {
+        if (reader.GetBoolean(1))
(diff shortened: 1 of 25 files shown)


src/ServiceControl.Persistence/DatabaseHostClassifier.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            if (normalized.EndsWith(suffix, StringComparison.Ordinal) || normalized.Contains($"{suffix},", StringComparison.Ordinal))
+            {
+                return hosting;
+            }
+        }
+
+        // Cloud SQL is reached either through a host name or through a unix socket directory named
+        // after the instance, so neither end of the value is a reliable place to look.
+        return normalized.Contains("cloudsql", StringComparison.Ordinal) ? "GoogleCloudSql" : SelfHosted;
```
warwickschroeder: If the DB query checks in `PostgreSqlDatabaseHostingProbe` or `SqlServerDatabaseHostingProbe` fail for any reason; db down, firewall, wrong password, permissions, etc, it will fallback to checking the connection string. The problem with checking the connection string is that it could be a private DNS name, which if they architectured their system correctly, really should be. So we'll be getting "SelfHosted" rather than "Unknown" or "Unavailable".

Maybe in `PostgreSqlDatabaseHostingProbe` or `SqlServerDatabaseHostingProbe`, if it couldnt get the info for whatever reason, it might be better to just return "Unavailable" rather than trying via the connection string. Or if we keep the fallback, perhaps add a 3rd value that tells us where the info came from. The probe or the fallback.

src/ServiceControl.Persistence.EFCore.PostgreSql/PostgreSqlDatabaseHostingProbe.cs
```diff
@@ -0,0 +1,95 @@
...
+        {
+            return "AzurePostgres";
+        }
+
+        if (reader.GetBoolean(2))
+        {
+            return "AwsRds";
+        }
+
+        return reader.GetBoolean(3) ? "GoogleCloudSql" : HostingFromConnectionString().Hosting;
```
warwickschroeder: If the probe query succeeded, and we get to here, and its not azure, aws, or google, is this not evidence that its self hosted?

src/ServiceControl/ServiceControlErrorInstanceEnvironmentDataProvider.cs
```diff
@@ -1,13 +1,51 @@
...
 using ServiceBus.Management.Infrastructure.Settings;
 
-class ServiceControlErrorInstanceEnvironmentDataProvider(Settings settings) : IEnvironmentDataProvider
+class ServiceControlErrorInstanceEnvironmentDataProvider(Settings settings, INotificationsDataStore notificationsDataStore) : IEnvironmentDataProvider
 {
-    public IEnumerable<(string key, string value)> GetData()
+    public async Task<IEnumerable<(string key, string value)>> GetData(CancellationToken cancellationToken = default)
     {
-        yield return ("Features.IntegratedServicePulse", settings.EnableIntegratedServicePulse ? "Enabled" : "Disabled");
+        var notificationsSettings = await notificationsDataStore.LoadSettings(cancellationToken);
```
warwickschroeder: If this throws, we loose all the other telemetry items. Scope the email notification setting to its own try..catch and return "Unavailable" or something if the DB query throws.

src/ServiceControl/ServiceControlErrorInstanceEnvironmentDataProvider.cs
```diff
@@ -1,13 +1,51 @@
...
+            ("Security.Authentication", Toggle(settings.OpenIdConnectSettings.Enabled)),
+            ("Security.RoleBasedAuthorization", Toggle(settings.OpenIdConnectSettings.RoleBasedAuthorizationEnabled)),
+            ("Security.Https", Toggle(settings.HttpsSettings.Enabled)),
+            ("Features.IntegratedServicePulse", Toggle(settings.EnableIntegratedServicePulse)),
+            ("Features.MessageEditing", Toggle(settings.AllowMessageEditing)),
+            ("Features.ExternalIntegrationsPublishing", Toggle(!settings.DisableExternalIntegrationsPublishing)),
+            ("Features.ForwardErrorMessages", Toggle(settings.ForwardErrorMessages)),
+            ("Features.EmailNotifications", EmailNotifications(notificationsSettings)),
+            ("Retention.ErrorHours", Hours(settings.ErrorRetentionPeriod)),
+            ("Retention.AuditHours", Hours(settings.AuditRetentionPeriod)),
```
warwickschroeder: This setting is coming from the error instance. Is it used?

src/ServiceControl.Persistence.EFCore.SqlServer/SqlServerDatabaseHostingProbe.cs
```diff
@@ -0,0 +1,85 @@
...
+
+    public async Task<DatabaseHosting> Probe(CancellationToken cancellationToken = default)
+    {
+        try
+        {
+            await using var scope = scopeFactory.CreateAsyncScope();
+            var dbContext = scope.ServiceProvider.GetRequiredService<ServiceControlDbContext>();
+
+            await using var command = dbContext.Database.GetDbConnection().CreateCommand();
+            command.CommandText = "SELECT SERVERPROPERTY('EngineEdition'), SERVERPROPERTY('ProductMajorVersion')";
```
warwickschroeder: Should `SERVERPROPERTY('ProductMajorVersion')` be `SERVERPROPERTY('ProductVersion')`? I dont thing `ProductMajorVersion` can be used for anything other than SQL Server (e.g. azure sql database would return `null`)- https://learn.microsoft.com/en-us/sql/t-sql/functions/serverproperty-transact-sql?view=sql-server-ver17#propertyname

src/ServiceControl.Persistence.EFCore.SqlServer/SqlServerDatabaseHostingProbe.cs
```diff
@@ -0,0 +1,85 @@
...
+            logger.LogDebug(e, "Could not ask SQL Server how it is hosted, falling back to the connection string");
+
+            return HostingFromConnectionString();
+        }
+    }
+
+    // EngineEdition is the authoritative answer: the server itself reports which Azure service it is,
+    // where a host name suffix is only a guess. Anything on-premises falls back to the suffix.
+    string HostingFromEngineEdition(DbDataReader reader) =>
+        reader.IsDBNull(0) ? HostingFromConnectionString().Hosting : Convert.ToInt32(reader.GetValue(0), CultureInfo.InvariantCulture) switch
```
warwickschroeder: Are we missing some versions? See `EngineEdition` at https://learn.microsoft.com/en-us/sql/t-sql/functions/serverproperty-transact-sql?view=sql-server-ver17#propertyname

warwickschroeder (commented):
I think, if you havent already, we should test the probes using real cloud resources to ensure we arent getting false information back. It would be better to get an "Unavailable" or "Unknown" that a false "Azure SQL Database", for example.

src/ServiceControl.Persistence/DatabaseHostClassifier.cs
johnsimons: Good catch, fixed. The classifier no longer defaults to SelfHosted on a suffix miss: it returns SelfHosted only for values that are demonstrably on this machine (localhost, 127.0.0.1, (local), (localdb), named pipes, unix sockets) and Unknown for everything else, so a private DNS name is now Unknown.

I also took your second suggestion and added Persistence.HostingSource, with values Probe, Configuration, ConnectionString or None, so a suffix-derived guess can be told apart from an answer the server gave us.

src/ServiceControl.Persistence.EFCore.PostgreSql/PostgreSqlDatabaseHostingProbe.cs
johnsimons: Yes, agreed, and that is now explicit. The probe distinguishes "answered and found no managed fingerprint", which reports SelfHosted with source=Probe, from "could not answer", which falls back to the host name and reports Unknown rather than SelfHosted. The host name still gets checked first in the answered case, otherwise a managed service that creates no distinctive role would be misreported.

src/ServiceControl/ServiceControlErrorInstanceEnvironmentDataProvider.cs
johnsimons: Fixed, and generalised. Isolation moved into the collector and is now per key rather than per provider: a provider lists its keys with a func to read each one, and the collector catches around each read, so one failure costs one key. No provider contains a try/catch any more. A failed read is recorded as "ReadFailed" rather than "Unavailable", since "Unavailable" reads like a valid state and already means one elsewhere in the codebase.

src/ServiceControl/ServiceControlErrorInstanceEnvironmentDataProvider.cs
johnsimons: It is not, so I have removed the key.

src/ServiceControl.Persistence.EFCore.SqlServer/SqlServerDatabaseHostingProbe.cs
johnsimons: You are right, fixed. The docs list ProductMajorVersion as "Applies to: SQL Server" only, so it comes back null on Azure SQL Database, Managed Instance and Synapse, which are exactly the platforms this is meant to identify. Now using ProductVersion and taking the segment before the first dot.

src/ServiceControl.Persistence.EFCore.SqlServer/SqlServerDatabaseHostingProbe.cs
johnsimons: Fixed. Mapped: 5 AzureSql, 8 AzureSqlManagedInstance, 9 AzureSqlEdge. Synapse (6, 11) and Fabric (12) are deliberately left unmapped since ServiceControl does not run on them, and an edition we do not map falls to the host name rather than being taken as evidence of an ordinary SQL Server.

Editions 1 to 4 are now positive evidence of a regular SQL Server, but the host name is checked first, otherwise RDS for SQL Server (which reports Standard or Enterprise) would be misreported as self-hosted. The mapping is a pure function now, so there are table-driven tests covering it.

johnsimons:
> using real cloud resources to ensure we aren't getting false information back

I am spinning up temporary cloud infrastructure so we can check each probe against a real Azure SQL, Azure PostgreSQL, RDS and Cloud SQL instance before this ships.

warwickschroeder (approved):

johnsimons:
I have tested this against both Azure and AWD infrastructure:
- Azure SQL Database
- Azure PostgreSQL Flexible
- AWS Aurora PostgreSQL 
- AWS RDS SQL Server
- AWS RDS PostgreSQL

PR #5831, opened 2026-08-28
Add retention sweep metrics for cycle duration, rows deleted, and consecutive failures

johnsimons:


src/ServiceControl.Persistence.EFCore/Infrastructure/RetentionSweeper.cs
warwickschroeder: If the failed messages sweep throws, the others just dont run, and they arent marked as failed in this cycle. I think the retention metrics still report for all three though? So they would be reporting no failures, but no data. It might be worth adding a domain event mapping for the retention to make any retention sweep failures very visible to customers. If rows arent being cleaned up for "reasons", the DB would just fill up.

src/ServiceControl.Persistence.EFCore/Infrastructure/RetentionMetrics.cs
```diff
@@ -0,0 +1,120 @@
...
+    public void Complete() => completed = true;
+
+    public void Dispose()
+    {
+        if (cancellationToken.IsCancellationRequested)
+        {
+            return;
+        }
+
+        metrics.RecordCycle(entity, stopwatch.Elapsed, completed);
```
warwickschroeder: We still want to store the cycle, if it completes, before shutdown.

```suggestion
        metrics.RecordCycle(entity, stopwatch.Elapsed,
          completed ? CycleOutcome.Success
          : cancellationToken.IsCancellationRequested ? CycleOutcome.Cancelled
          : CycleOutcome.Failed);
```

src/ServiceControl.Persistence.EFCore/Infrastructure/RetentionSweeper.cs
warwickschroeder, line 161: Should we be recording "zero rows deleted" about this breakout, rather than recording no data for the sweep? EventLogs and GroupComments both record data (or zeros) rather than no data.

warwickschroeder: Are we sure a "missing" body will ever hit this? e.g. Azure blob body storage [handles](https://github.com/Particular/ServiceControl/blob/31d5150be28ae100503a5ced540c757647806496/src/ServiceControl.Persistence.EFCore/Implementation/BodyStorage/AzureBlobBodyStoragePersistence.cs#L97) this already, which means this catches legit infra failures such as expired tokens etc. Should we have metrics for orphaned bodies, and potentially raise an event for it for the customer?

warwickschroeder (commented):
Does `docs/telemetry.md` need updating?

src/ServiceControl.Persistence.EFCore/Infrastructure/RetentionMetrics.cs
```diff
@@ -0,0 +1,120 @@
...
+
+    const string MeterVersion = "0.1.0";
+    const string InstrumentPrefix = "sc.retention";
+}
+
+/// <summary>
+/// One pass of the retention sweep. A pass interrupted by shutdown is not a measurement of
+/// anything, so a cancelled cycle records neither a duration nor a failure.
+/// </summary>
+public sealed class RetentionCycleMetrics : IDisposable
```
rbev: Should these classes be in separate files, and possibly in their own namespace/folder?

src/ServiceControl.Persistence.EFCore/Infrastructure/RetentionMetrics.cs
```diff
@@ -0,0 +1,120 @@
...
+{
+    internal RetentionCycleMetrics(RetentionMetrics metrics, RetentionEntity entity, CancellationToken cancellationToken)
+    {
+        this.metrics = metrics;
+        this.entity = entity;
+        this.cancellationToken = cancellationToken;
+    }
+
+    public void Complete() => completed = true;
```
rbev: I know it's probably not material but shouldn't this also freeze the timer in case there is time between the `Complete()` and `Dispose()`?

rbev (approved):

src/ServiceControl.Persistence.EFCore/Infrastructure/RetentionSweeper.cs
johnsimons: This is now fixed, and we have added a new card to report failures to the user as a domain event

src/ServiceControl.Persistence.EFCore/Infrastructure/RetentionMetrics.cs
johnsimons: We are

src/ServiceControl.Persistence.EFCore/Infrastructure/RetentionSweeper.cs
johnsimons: I have redone this area so that if an infrastructure issue occurs when deleting a body, it fails the batch

src/ServiceControl.Persistence.EFCore/Infrastructure/RetentionMetrics.cs
johnsimons: done

src/ServiceControl.Persistence.EFCore/Infrastructure/RetentionSweeper.cs
johnsimons: Agreed, fixed. The pass records a zero on that break now, so all three report on every run rather than the failed messages series only showing up once something is actually deleted.

src/ServiceControl.Persistence.EFCore/Infrastructure/RetentionMetrics.cs
johnsimons: Good call, done. `Complete()` stops the stopwatch and `Dispose()` records the frozen value.

warwickschroeder (approved):

PR #5832, opened 2026-08-28
Disable RavenDB storage prefetching by default

ramonsmits:
During load testing we notices that RavenDB does a lot of read IO. This is caused due to prefetching. We tested with disabling prefetching and index rate remained pretty much identical, ingestion rate slightly improved but read IO was reduced over 5x.

Storage.EnablePrefetching is a server-wide-only RavenDB setting that defaults to true. ServiceControl now disables it by default:

- Embedded mode (Primary and Audit): --Storage.EnablePrefetching=false is passed via ServerOptions.CommandLineArgs
- Container image (servicecontrol-ravendb): RAVEN_Storage_EnablePrefetching=false is set as an image ENV, which users can still override


mauroservienti (approved):

ramonsmits:
Local docker image build tested with Podman and confirmed this in RavenDB studio server settings.

PR #5834, opened 2026-08-28
Spike: shrink audit full-text search index

ramonsmits:
## Spike: shrink the audit `MessagesView` full-text index

Draft/spike — measuring on a real audit database before deciding whether to ship.

### Problem

The `Query` search field of both `MessagesViewIndex` and `MessagesViewIndexWithFullTextSearch` indexes *every* metadata value and *every* header value. A large share of those are timestamps, durations, sizes, booleans and identifiers. Nobody searches for a timestamp or a `ProcessingTime`, and because each of those values is unique per message they never share dictionary terms across documents, so they inflate the term dictionary far more than the shared tokens (endpoint names, message types, host names) do.

### Changes

1. **Exclude temporal, numeric and boolean values from `Query`** — metadata `TimeSent`, `CriticalTime`, `ProcessingTime`, `DeliveryTime`, `ContentLength`, `BodyUrl`, `IsSystemMessage`, `IsRetried`, `BodyNotStored`, `OriginatesFromSaga` and headers `TimeSent`, `ProcessingStarted`, `ProcessingEnded`, `DeliverAt`, `Timeout.Expire`, `Retries.Timestamp`, `ExceptionInfo.TimeOfFailure`, `TimeOfFailure`, `NonDurableMessage`, `TimeToBeReceived`. Header values are indexed individually rather than as one joined string; null/empty values dropped; `Distinct()` removes metadata/header duplicates.
2. **Match identifiers exactly instead of via full-text** — `/messages/search/{keyword}` now ORs `MessageId = keyword` and `ConversationId = keyword` with the full-text search, and ids (plus their `NServiceBus.MessageId/ConversationId/CorrelationId/RelatedTo` header copies) are dropped from `Query`. The OR group is built with the DocumentQuery API because the LINQ provider does not parenthesize `Search(..., SearchOptions.Or)`, which would bind the endpoint/time-range filters to the last OR term only (verified by inspecting the generated RQL).
3. Analyzer referenced by its short name `StandardAnalyzer`, as the old comment suggested doing whenever the index has to rebuild anyway.
4. New acceptance test `When_processed_message_searched_by_conversationid`.

### Estimated effect

Representative `OrderPlaced` audit message (21 headers, no body), StandardAnalyzer-style tokenization:

| | before | after |
|---|---|---|
| tokens / message | 83 | 35 |
| distinct terms / message | 42 | 22 |
| per-message-unique terms | 18 | 0 |

Real numbers to be measured on an existing audit database (side-by-side trial indexes filtered on `ProcessedAt`, old vs new `Query` projection over the same document set).

### Behaviour changes / open questions

- Any deployment picking this up rebuilds both `MessagesView` indexes on startup (unavoidable for any map change).
- `CorrelationId` and `RelatedTo` are no longer searchable (no exact-match field). Correlation id usually equals message id. Acceptable?
- The EF/SQL persisters still index ids and timestamps as free text — same pattern, follow-up candidate.

### Primary instance

The error `MessagesViewIndex` and `ErrorMessagesDataStore` get the same treatment (third commit): identical `Query` exclusions (minus audit-only metadata keys), exact-match OR on `MessageId`/`ConversationId`, headers indexed individually with `Distinct()`. Exception details (`NServiceBus.ExceptionInfo.ExceptionType/Message/StackTrace`) and `MsgFullText` remain searchable; only `ExceptionInfo.TimeOfFailure` is dropped. This index also rebuilds on deployment.

### Verification

- `ServiceControl.Audit.Persistence.Tests.RavenDB` 41/41; `ServiceControl.Persistence.Tests.RavenDB` 326/326
- `ServiceControl.Audit.AcceptanceTests.RavenDB` and `ServiceControl.Audit.AcceptanceTests` search tests 5/5 each; `ServiceControl.AcceptanceTests.RavenDB` search/queried tests 8/8

### DIFF
--- a/src/ServiceControl.Audit.AcceptanceTests/Auditing/When_processed_message_searched_by_conversationid.cs
+++ b/src/ServiceControl.Audit.AcceptanceTests/Auditing/When_processed_message_searched_by_conversationid.cs
@@ -1,0 +1,59 @@
+﻿namespace ServiceControl.Audit.AcceptanceTests.Auditing
+{
+    using System.Threading.Tasks;
+    using AcceptanceTesting;
+    using AcceptanceTesting.EndpointTemplates;
+    using Audit.Auditing.MessagesView;
+    using NServiceBus;
+    using NServiceBus.AcceptanceTesting;
+    using NServiceBus.AcceptanceTesting.Customization;
+    using NServiceBus.Settings;
+    using NUnit.Framework;
+
+    class When_processed_message_searched_by_conversationid : AcceptanceTest
+    {
+        [Test]
+        public async Task Should_be_found() =>
+            await Define<MyContext>()
+                .WithEndpoint<Sender>(b => b.When((bus, c) => bus.Send(new MyMessage())))
+                .WithEndpoint<Receiver>()
+                .Done(async c => c.ConversationId != null && await this.TryGetMany<MessagesView>("/api/messages/search/" + c.ConversationId))
+                .Run();
+
+        public class Sender : EndpointConfigurationBuilder
+        {
+            public Sender() =>
+                EndpointSetup<DefaultServerWithoutAudit>(c =>
+                {
+                    var routing = c.ConfigureRouting();
+                    routing.RouteToEndpoint(typeof(MyMessage), typeof(Receiver));
+                });
+        }
+
+        public class Receiver : EndpointConfigurationBuilder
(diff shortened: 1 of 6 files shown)


mauroservienti:
Closing, raised:

- https://github.com/Particular/ServiceControl/issues/5852

PR #5836, opened 2026-09-01
Add OpenTelemetry metrics for retry and archive operations

johnsimons:
Adds telemetry for error management, the follow-up to the ingestion (#5816) and retention (#5831) metrics. Retry and archive were the last long-running operations in the error instance with no signal at all: a retry of a large group can run for hours across four stages and two polling loops, and until now the only way to see it was the log.

## What is instrumented

**`sc.retry.*`** covers the whole retry pipeline: an end-to-end operation duration from request to completion, a duration histogram per stage (prepare, stage, forward), a message counter by outcome, an in-progress gauge, and the depth of the bulk request queue. The `abandoned` message outcome is the one worth alerting on, since it counts messages that hit the staging retry limit and were dropped from their batch, which previously existed only as a log line. The in-progress gauge is what surfaces a stuck operation, because an operation that never completes never records a duration.

**`sc.archive.*`** covers group archive and unarchive: whole-operation duration, per-batch duration, a message counter, and an in-progress gauge. The instrumentation lives once in the shared `InMemoryArchive`/`InMemoryUnarchive` state machines, but only the EF persistence registers `ArchiveMetrics`, so on RavenDB the family does not exist rather than being half-populated. No RavenDB file is touched.

The host also gains the standard ASP.NET Core, HTTP client and runtime instrumentation packages, behind the existing `OTEL_EXPORTER_OTLP_ENDPOINT` gate. That covers the read APIs per route, the scatter-gather calls to remote instances, and GC/thread pool context. `AddIngestionMetrics` is renamed `AddTelemetry` now that it wires more than ingestion.

All tag values come from enums, so series counts are bounded; group ids, endpoint names and queue addresses stay out and are recovered from the audit log when a specific operation needs investigating. The full instrument contract is documented in `docs/telemetry.md`.

### DIFF
--- a/src/ServiceControl/Recoverability/Retrying/Metrics/RetryDurationScope.cs
+++ b/src/ServiceControl/Recoverability/Retrying/Metrics/RetryDurationScope.cs
@@ -1,0 +1,61 @@
+namespace ServiceControl.Recoverability.Retrying.Metrics;
+
+using System;
+using System.Diagnostics;
+using System.Diagnostics.Metrics;
+using System.Threading;
+
+/// <summary>
+/// One timed stretch of the retry pipeline. A stretch that finished counts as its outcome even if
+/// shutdown has since been requested; one that shutdown cut short is recorded as cancelled rather
+/// than as a failure.
+/// </summary>
+public sealed class RetryDurationScope : IDisposable
+{
+    internal RetryDurationScope(Histogram<double> histogram, TimeProvider timeProvider, TagList tags, CancellationToken cancellationToken)
+    {
+        this.histogram = histogram;
+        this.timeProvider = timeProvider;
+        this.tags = tags;
+        this.cancellationToken = cancellationToken;
+        start = timeProvider.GetTimestamp();
+    }
+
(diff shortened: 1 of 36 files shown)


src/ServiceControl/Recoverability/Retrying/Metrics/RetryDurationScope.cs
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    static string ResultTag(ScopeOutcome outcome) => outcome switch
+    {
+        ScopeOutcome.Success => "success",
+        ScopeOutcome.Empty => "empty",
+        ScopeOutcome.Cancelled => "cancelled",
+        ScopeOutcome.Failed => "failed",
+        _ => "failed"
+    };
```
rbev, line 51: Not necessarily better but you could avoid all the string constants using enum.ToString()
```suggestion
    static string ResultTag(ScopeOutcome outcome) => (Enum.IsDefined(outcome) ? outcome : ScopeOutcome.Failed).ToString().ToLower();
```

rbev (approved):

PR #5839, opened 2026-09-01
Replace CombGuid with Guid.CreateVersion7 for message ID generation

johnsimons:
UUIDv7 is time-ordered and monotonically increasing, providing the same database-friendly ordering properties as the COMB algorithm without requiring a custom implementation. This removes the locally maintained CombGuid class in favour of the built-in .NET API.


src/ServiceControl/Recoverability/Editing/EditHandler.cs
```diff
@@ -80,10 +79,10 @@ await domainEvents.Raise(new MessageEditedAndRetried
 
         OutgoingMessage BuildMessage(EditAndSend message)
         {
-            var messageId = CombGuid.Generate().ToString();
+            var messageId = Guid.CreateVersion7().ToString("D");
             var headers = HeaderFilter.RemoveErrorMessageHeaders(message.NewHeaders);
             corruptedReplyToHeaderStrategy.FixCorruptedReplyToHeader(headers);
-            headers[Headers.MessageId] = Guid.NewGuid().ToString("D");
```
johnsimons, line 86: I also updated this, which was wrong!

rbev (approved):

PR #5848, opened 2026-09-03
Terminate message queries that exceed a configurable time limit

ramonsmits:
Customer debug packages showed audit message queries against `MessagesViewIndexWithFullTextSearch` running for up to ~65 minutes. RavenDB's own `Databases.QueryTimeoutInSec` (default 300) does not stop them because the server renews the operation deadline while a query keeps making progress, so large scans/sorts run for hours and their sorted results spill into the server's Temp folder, filling the disk.

## Query time limit

The message view and search queries now run under a linked `CancellationTokenSource` that cancels after a configurable time limit, default 1 minute, maximum 1 hour. Cancelling the client request aborts the call to the database server, which does terminate the query server-side. Invalid values fall back to the default.

One `QueryTimeLimit` helper in `ServiceControl.Infrastructure` serves all persisters. Its rule is that whatever exception surfaces after the deadline fired is the timeout: RavenDB and Npgsql raise `OperationCanceledException`, Microsoft.Data.SqlClient raises `SqlException("Operation cancelled by user")`. The SQL Server path is covered by a test against the real container that slows the command down server-side.

The limit applies to exactly these queries:

- Audit `RavenAuditDataStore`: `GetMessages`, `QueryMessages`, `QueryMessagesByReceivingEndpoint`, `QueryMessagesByReceivingEndpointAndKeyword`, `QueryMessagesByConversationId`, read from `ServiceControl.Audit/QueryTimeoutInSeconds`. `QueryAuditCounts` (licensing throughput) and `QuerySagaHistoryById` are not under the limit.
- Primary `IMessagesViewDataStore` on RavenDB, SQL Server and PostgreSQL: `GetAllMessages`, `GetAllMessagesForEndpoint`, `GetAllMessagesByConversation`, `GetAllMessagesForSearch`, `SearchEndpointMessages`, read from `ServiceControl/QueryTimeoutInSeconds`. On SQL Server and PostgreSQL these queries also use the limit as their per-command timeout, so `Database/CommandTimeout` cannot undercut it.

Other paged index queries (`GetFailedMessages`, groups, event log, ...) are unchanged.

## What a caller sees

- A query that runs out of its time is answered with `504 Gateway Timeout` and a `application/problem+json` body whose detail names the setting, on both hosts.
- The primary's scatter-gather keeps partial data: a local timeout is absorbed the way a remote failure already was, a remote `504` or error is a missing instance rather than an instance with no data. The response lists every missing instance in the `X-Particular-Incomplete-Results` header (`instanceId:timeout|unavailable|error`, exposed through CORS) and carries no `ETag`. Only when no queried instance answered and one of them timed out does the composite itself answer `504`.
- The in-process audit counts used for licensing throughput fail when any instance is missing, instead of recording a partial sum as the day's throughput.
- The remote instance `HttpClient` timeout is `ServiceControl/QueryTimeoutInSeconds` plus 30 seconds instead of the 100 second default, so raising the limit on both instances does not make the primary give up on the audit instance first.

ServicePulse follow-up: the audit list already treats a non-2xx answer (the new `504`) as a failed query; reading `X-Particular-Incomplete-Results` on a `200` is still to be added.

Docs: Particular/docs.particular.net#8486

### DIFF
--- a/src/ServiceControl.Audit.Persistence.RavenDB/RavenAuditDataStore.cs
+++ b/src/ServiceControl.Audit.Persistence.RavenDB/RavenAuditDataStore.cs
@@ -1,4 +1,4 @@
-﻿namespace ServiceControl.Audit.Persistence.RavenDB
+namespace ServiceControl.Audit.Persistence.RavenDB
 {
     using System;
     using System.Collections.Generic;
@@ -11,6 +11,7 @@
     using Raven.Client.Documents;
     using ServiceControl.Audit.Auditing;
     using ServiceControl.Audit.Infrastructure;
+    using ServiceControl.Infrastructure;
     using ServiceControl.SagaAudit;
     using Transformers;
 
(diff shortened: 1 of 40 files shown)


src/ServiceControl.Audit.Persistence.RavenDB/RavenAuditDataStore.cs
```diff
@@ -123,69 +129,94 @@ public async Task<MessageBodyView> GetMessageBody(string messageId, Cancellation
...
-                        UtcDate = DateTime.UtcNow.Date,
-                        Count = 0
-                    });
+                    // The caller cancelled (e.g. the HTTP request was aborted), not the query timeout
+                    throw;
                 }
-            }
 
-            return new QueryResult<IList<AuditCount>>(results, QueryStatsInfo.Zero);
+                throw new TimeoutException($"The query did not complete within the allowed query time of {databaseConfiguration.QueryTimeout.TotalSeconds:0} seconds and was cancelled. The '{RavenPersistenceConfiguration.QueryTimeoutInSecondsKey}' setting can be used to change the allowed query time.", e);
```
mauroservienti:
```suggestion
                throw new TimeoutException($"The query did not complete within the allowed query time of {databaseConfiguration.QueryTimeout.TotalSeconds:0} seconds and was cancelled. The '{RavenPersistenceConfiguration.QueryTimeoutInSecondsKey}' setting can be used to change the database query timeout.", e);
```

src/ServiceControl.Persistence.RavenDB/ErrorMessagesDataStore.cs
```diff
@@ -30,126 +30,155 @@ class ErrorMessagesDataStore(
...
+            catch (OperationCanceledException e) when (timeoutTokenSource.Token.IsCancellationRequested)
+            {
+                if (cancellationToken.IsCancellationRequested)
+                {
+                    // The caller cancelled (e.g. the HTTP request was aborted), not the query timeout
+                    throw;
+                }
 
-            return new QueryResult<IList<MessagesView>>(results, stats.ToQueryStatsInfo());
+                throw new TimeoutException($"The query did not complete within the allowed query time of {settings.QueryTimeout.TotalSeconds:0} seconds and was cancelled. The '{RavenPersisterSettings.QueryTimeoutInSecondsKey}' setting can be used to change the allowed query time.", e);
```
mauroservienti:
```suggestion
                throw new TimeoutException($"The query did not complete within the allowed query time of {settings.QueryTimeout.TotalSeconds:0} seconds and was cancelled. The '{RavenPersisterSettings.QueryTimeoutInSecondsKey}' setting can be used to change the allowed database query timeout.", e);
```

src/ServiceControl.Persistence.EFCore/Implementation/MessagesViewDataStore.cs
```diff
@@ -3,47 +3,72 @@ namespace ServiceControl.Persistence.EFCore.Implementation;
...
+        }
+        catch (OperationCanceledException e) when (timeoutTokenSource.Token.IsCancellationRequested)
+        {
+            if (cancellationToken.IsCancellationRequested)
+            {
+                // The caller cancelled (e.g. the HTTP request was aborted), not the query timeout
+                throw;
+            }
+
+            throw new TimeoutException($"The query did not complete within the allowed query time of {settings.QueryTimeout.TotalSeconds:0} seconds and was cancelled. The '{EFPersisterSettings.QueryTimeoutInSecondsKey}' setting can be used to change the allowed query time.", e);
```
mauroservienti:
```suggestion
            throw new TimeoutException($"The query did not complete within the allowed query time of {settings.QueryTimeout.TotalSeconds:0} seconds and was cancelled. The '{EFPersisterSettings.QueryTimeoutInSecondsKey}' setting can be used to change the allowed database query timeout.", e);
```

mauroservienti (changes requested):

mauroservienti (approved):

mauroservienti:
@ramonsmits, I have not finished reviewing this yet. However, we also have this [CustomCheck](https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/Operations/CheckRemotes.cs) that probably needs to be adapted to incorporate the query timeouts. Also, we have this concept of an instance being [temporarily unavailable](https://github.com/Particular/ServiceControl/blob/e74a45b691ae626b5ad0f46e88634099a5e00859/src/ServiceControl/Infrastructure/Settings/RemoteInstanceSetting.cs#L22) if it fails to respond to the mentioned custom check. When that property is `true`, the ScatterGather API [skips the unavailable instance until it comes back](https://github.com/Particular/ServiceControl/blob/e74a45b691ae626b5ad0f46e88634099a5e00859/src/ServiceControl/CompositeViews/Messages/ScatterGatherApi.cs#L62-L70).

src/ServiceControl/Infrastructure/WebApi/HttpResponseExtensions.cs
```diff
@@ -11,6 +11,42 @@ namespace ServiceControl.Infrastructure.WebApi
...
+            }
+
+            response.WithHeader(IncompleteResultsHeader, string.Join(", ", incomplete.Select(instance => $"{instance.InstanceId}:{instance.Reason switch
+            {
+                QueryFailure.TimedOut => "timeout",
+                QueryFailure.Unavailable => "unavailable",
+                QueryFailure.Failed => "error",
+                _ => "error"
+            }}")));
+        }
```
mauroservienti, line 48: I feel that the `TemporaryUnavailable` instances, I mentioned, should be counted here

ramonsmits: They are. `ScatterGatherApi.Execute` adds every skipped remote to the results with `QueryFailure.Unavailable`, `AggregateResults` turns those into `IncompleteInstances`, and this header renders them as `instanceId:unavailable`. `IncompleteResultsTests.A_temporarily_unavailable_remote_is_reported_as_missing` pins it.

ramonsmits:
Looked at `CheckRemotes` against the query timeouts. The check probes `/api`, which runs no query, so a remote whose queries time out never fails the health check, and that is intended: a 504 means the remote is up, so `TemporarilyUnavailable` stays false and the next query asks it again (`A_remote_query_timeout_keeps_the_local_data_and_reports_the_remote_instance` asserts that).

There was a real interaction though, caused by this PR: the probe used the remote's named `HttpClient`, whose timeout is now the query time limit. With a short limit a healthy but slow remote would fail the health check, and the failure message would still claim 10 seconds. While in there I also fixed a pre-existing problem: the probe's own timeout never applied, because its cancellation was swallowed by a no-op branch, so a hanging remote passed the check and was never disabled, and a shutdown waited out the full budget.

79d19e913 gives the probe its own budget independent of the query limit, makes a remote that doesn't answer in time fail the check and become temporarily unavailable, and lets a shutdown abort the probe without a verdict. `CheckRemotesTests` covers the four cases.

Since your last pass I also dropped the 30 second margin (22e610973): the remote client now times out at this instance's limit, so `ServiceControl/QueryTimeoutInSeconds` bounds the whole composite. An audit instance that is slow, hung, or configured with a larger limit is reported as missing rather than stretching the response. The docs PR is updated accordingly.

mauroservienti (approved):
This is good to go, @ramonsmits 🚀

PR #5862, opened 2026-09-08
Add Database/Schema support to SQL Server and PostgreSQL persisters

johnsimons:
Each test now takes its own schema in one shared database rather than its own database, which exercises the same Database/Schema setting offered to customers. Migrations are stamped with the configured schema at generation time via a custom IMigrationsSqlGenerator, raw SQL in the dialects is qualified using the schema-aware table name helper, and the migrators require the schema to exist rather than silently creating it. SchemaName validates the value before it reaches any DDL interpolation.

### DIFF
--- a/src/ServiceControl.Persistence.Tests.SqlServer/SchemaMustExistTests.cs
+++ b/src/ServiceControl.Persistence.Tests.SqlServer/SchemaMustExistTests.cs
@@ -1,0 +1,37 @@
+// ReSharper disable once CheckNamespace
+namespace ServiceControl.Persistence.Tests;
+
+using System;
+using System.IO;
+using System.Threading.Tasks;
+using EFCore.SqlServer;
+using Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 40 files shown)


src/ServiceControl.Persistence.Tests.SqlServer/SchemaMustExistTests.cs
```diff
@@ -0,0 +1,37 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using NUnit.Framework;
+using ServiceControl.Persistence.EFCore.Abstractions;
+
+[TestFixture]
+class SchemaMustExistTests
+{
+    [Test]
+    public async Task Migration_fails_when_the_configured_schema_does_not_exist()
```
rbev, line 17: can this test be written in the shared folder instead of duplicated?

rbev (approved):

PR #5863, opened 2026-09-08
Add GitHub Actions workflows and provisioning scripts for cloud databases

johnsimons:
Introduces two workflows and supporting PowerShell scripts to run the persistence and acceptance test suites against managed cloud databases (Azure SQL, Azure Database for PostgreSQL, Aurora PostgreSQL, RDS SQL Server).

The test workflow provisions a fresh database per run, executes the relevant test projects against it, and tears it down regardless of outcome.

A nightly cleanup workflow sweeps any resources tagged `sc-cloud-test` that are older than four hours.


.github/workflows/cloud-database-tests.yml
```diff
@@ -0,0 +1,138 @@
...
+  DOTNET_NOLOGO: true
+  # Its own region rather than the AWS_REGION secret the SQS tests use: that one is at its VPC
+  # ceiling, so RDS has no default VPC to place a publicly accessible instance in. us-east-2 already
+  # has one, has VPC headroom, and offers both instance classes these targets ask for.
+  AWS_DATABASE_REGION: us-east-2
+defaults:
+  run:
+    shell: pwsh
+concurrency:
+  # Deliberately not cancel-in-progress: cancelling a run mid-flight risks skipping the teardown
```
rbev, line 27: There are absolutely still cases where this can and will happen such as an agent failing and timing out the run. 

Generally you would just make sure the resources are somewhere predictable with a `DeleteAfter` tag on them so a daily cleanup can clear them out.

rbev (approved):

.github/workflows/cloud-database-tests.yml
johnsimons: Yes, you are right, and we are dealing with that situation too.
But in a successful run we want to clean-up even if another commit is pushed

PR #5870, opened 2026-09-09
Bound acceptance tests by the test's CancelAfter budget instead of the fixed 90 s Done limit

ramonsmits:
NServiceBus.AcceptanceTesting caps `Done` at 90 seconds unless `Run` receives a cancellable token. The journey tests added in #5810 (`When_deleted_messages_are_restored`, `When_a_failing_custom_check_is_dismissed`) regularly exceed that on Windows runners and account for most Windows-Raven failures across all branches over the past week.

`NServiceBusAcceptanceTest` now carries `[CancelAfter(180_000)]`, which NUnit applies to every derived fixture, and `Define` wraps the scenario so that `Run()` without a token uses `TestContext.CurrentContext.CancellationToken`. A test that already declares `[CancelAfter]` keeps its own value. The runsettings `DefaultTimeout` route does not work here: NUnit3TestAdapter only forwards it under `#if NET462`.

Verified locally with a throwaway test: a method-level `[CancelAfter(5_000)]` on a never-completing scenario fails after 5.9 s with `The scenario did not complete within the test's CancelAfter budget of 5000 ms.`, and existing Primary, Audit and Monitoring tests pass unchanged.

Pairs with #5872: without it a timed-out test reports the `TaskCanceledException` from the aborted host shutdown instead of the timeout.

### DIFF
--- a/src/ServiceControl.AcceptanceTesting/NServiceBusAcceptanceTest.cs
+++ b/src/ServiceControl.AcceptanceTesting/NServiceBusAcceptanceTest.cs
@@ -13,6 +13,7 @@
     /// </summary>
     [TestFixture]
     [FixtureLifeCycle(LifeCycle.InstancePerTestCase)]
+    [CancelAfter(180_000)]
     public abstract partial class NServiceBusAcceptanceTest
     {
         [SetUp]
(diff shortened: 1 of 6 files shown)


mauroservienti (approved):

rbev (approved):

rbev (approved):

danielmarbach:
Should we try to improve this in Core instead? I'm worried because it adds extra API surface the Core acceptance test infrastructure might need to have anyway and then things need to be kept in sync

danielmarbach:
See:

- https://github.com/Particular/NServiceBus/pull/7940

ramonsmits:
Decided to move this to core:

- https://github.com/Particular/NServiceBus/pull/7940

PR #5871, opened 2026-09-09
Bound acceptance tests by the test's CancelAfter budget via an explicit CancellationToken

ramonsmits:
Alternative to #5870 without the scenario wrapper. Same motivation: NServiceBus.AcceptanceTesting caps `Done` at 90 seconds unless `Run` receives a cancellable token, and the journey tests from #5810 regularly exceed that on Windows runners.

`NServiceBusAcceptanceTest` carries `[CancelAfter(180_000)]`, which NUnit applies to every derived fixture. Every test that runs a scenario now takes a trailing `CancellationToken cancellationToken = default`, which NUnit fills from that budget (also for `[TestCase]` methods), and passes it to `Run`. The rewrite is mechanical: 282 methods across 195 files, plus forwarding the token to the few `ReadAsStringAsync` and `AssertAuthConfigurationResponse` calls CA2016 then flags. A test that already declares `[CancelAfter]` keeps its own value.

NUnit1027 is suppressed in the acceptance projects: the analyzer only looks for `[CancelAfter]` on the method and its direct containing class, so it reports the token parameter as unsupplied even though NUnit honours the inherited attribute. Reported upstream as nunit/nunit.analyzers#1019, fix in nunit/nunit.analyzers#1020.

Verified locally: a `[CancelAfter(5_000)]` test with a never-completing `Done` is cancelled after 5 s, and Primary, Audit and Monitoring fixtures pass unchanged.

Pairs with #5872: without it a timed-out test reports the `TaskCanceledException` from the aborted host shutdown instead of the timeout.

### DIFF
--- a/src/ServiceControl.AcceptanceTesting/NServiceBusAcceptanceTest.cs
+++ b/src/ServiceControl.AcceptanceTesting/NServiceBusAcceptanceTest.cs
@@ -13,6 +13,7 @@
     /// </summary>
     [TestFixture]
     [FixtureLifeCycle(LifeCycle.InstancePerTestCase)]
+    [CancelAfter(180_000)]
     public abstract partial class NServiceBusAcceptanceTest
     {
         [SetUp]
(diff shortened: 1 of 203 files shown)


rbev (approved):

danielmarbach (approved):

ramonsmits:
Decided to move this to core:

- https://github.com/Particular/NServiceBus/pull/7940

PR #5874, opened 2026-09-09
Keep index building out of the retention window in the audit retention tests

ramonsmits:
`AuditMessageRetention` and `SagaSnapshotRetention` failed four times in the last 60 CI runs, always on the query that should still see the document. The two second retention window starts at ingestion, and on a busy Windows runner building the fresh database's indexes ate the whole window before the query ran.

The tests now wait for the indexes before ingesting, use a three second window with a one second expiration pass, and poll for expiry instead of sleeping a fixed four seconds. Passes three runs in a row locally at 7 s and 4 s.

### DIFF
--- a/src/ServiceControl.Audit.Persistence.Tests.RavenDB/RetentionTests.cs
+++ b/src/ServiceControl.Audit.Persistence.Tests.RavenDB/RetentionTests.cs
@@ -2,8 +2,10 @@
 {
     using System;
     using System.Collections.Generic;
+    using System.Threading;
     using System.Threading.Tasks;
     using Auditing;
+    using Auditing.MessagesView;
     using NServiceBus;
     using NUnit.Framework;
     using SagaAudit;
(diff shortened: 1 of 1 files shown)


src/ServiceControl.Audit.Persistence.Tests.RavenDB/RetentionTests.cs
```diff
@@ -55,20 +60,42 @@ await IngestSagaAudits(
...
+                DataStore.QuerySagaHistoryById(sagaId, cancellationToken);
+        }
+
+        // The retention window starts at ingestion, so building the indexes of the fresh database must
+        // not happen inside it: on a busy runner that alone has taken longer than the window.
+        Task WaitForIndexesToBeBuilt() => configuration.CompleteDBOperation();
+
+        // Polls instead of sleeping for a fixed time, so the test takes as long as expiration actually
+        // needs and still tolerates a slow expiration pass.
+        static async Task<T> WaitUntil<T>(Func<CancellationToken, Task<T>> query, Func<T, bool> condition)
```
danielmarbach, line 82: So we have no existing helper that already does that?

mauroservienti (approved):

PR #5881, opened 2026-09-10
Improve index performance on Entity Framework persisters

rbev:
This pull request introduces new and optimized database indexes to both the SQL Server and PostgreSQL EF Core persistence layers, aiming to improve the performance of queries on failed messages and failure groups. The changes include adding covering indexes (with INCLUDE columns) for key aggregate queries, introducing new indexes to support sorting and filtering in the UI, and enforcing indexable column lengths for message types. These enhancements are reflected in both the migration files and the EF Core model configuration.

**Database Indexing Improvements**

* Added covering indexes (with INCLUDE columns) to the `Status, LastModified` index on `FailedMessages` and the `Type, GroupId` index on `FailedMessageGroups` in both SQL Server and PostgreSQL providers, improving index-only scan performance for group aggregate queries. [[1]](diffhunk://#diff-063a268d799dca158ee75f0bab71ba62fa12fd7100f348fed4c55d34cdf73584R1-R56) [[2]](diffhunk://#diff-e142e2e0b6f66e98b0205ab4f529be48f73e70814d265230f1a2169e3275b4deR1-R56) [[3]](diffhunk://#diff-c813275abd68774e4d5714c89d2b6b18b57acd0e9df30b5b5b56ea0623cce6e6R354-R359) [[4]](diffhunk://#diff-c813275abd68774e4d5714c89d2b6b18b57acd0e9df30b5b5b56ea0623cce6e6R387-R388) [[5]](diffhunk://#diff-378c2c7dde62b490d9780110ffbf887f9f508452cea8bcc60ad7752ffa96595bR444-R451) [[6]](diffhunk://#diff-378c2c7dde62b490d9780110ffbf887f9f508452cea8bcc60ad7752ffa96595bR486-R487) [[7]](diffhunk://#diff-0d340659c2cd295979859d2f6ca1b777f429a418b6b44eefbf11435fe135ac7cR24-R40) [[8]](diffhunk://#diff-4e546b0425c509c9b53dccf42ca59fe92ff1c10a75cd3b622c94a91d093283deR26-R37)

[...]

PR #5889, opened 2026-09-14
Truncate message type from the start to maintain the useful type name.

rbev:
https://github.com/Particular/ServiceControl/pull/5881 was auto merged, this PR contains the fixes requested from review.

* Make fields in query type required
* Truncate type name field from the start to maintain more useful type information.


johnsimons (approved):

src/ServiceControl.Persistence.EFCore/Implementation/UnitOfWork/EFRecoverabilityIngestionUnitOfWork.cs
```diff
@@ -70,13 +70,10 @@ public Task RecordSuccessfulRetry(string retriedMessageUniqueId, DateTime succee
...
-    // above, because it is an EFCore storage limit the persister owns rather than a shared
-    // ingestion rule. EnclosedMessageTypes' first comma token is a type's full name, so the cap can
-    // only ever bite on pathological generic names, where a truncated sort key still sorts and
-    // groups consistently.
-    static string? TruncateForColumn(string? value) =>
-        value is { Length: > ColumnLengths.ShortTextLength } ? value[..ColumnLengths.ShortTextLength] : value;
+    // an index key serving sort=message_type. This field contains a type name, so it's more useful to 
+    // truncate from the start of the name instead of the end.
+    static string? TruncateTypeName(string? value) =>
+        value is { Length: > ColumnLengths.ShortTextLength } ? value[^ColumnLengths.ShortTextLength..] : value;
```
warwickschroeder, line 76: Should we log this as a warning?

johnsimons: We can, but if a user reads that warning, what can they do?

rbev: Also if their types are this long potentially it's going to log on every single error because their namespaces are out of control everywhere.

warwickschroeder: Hmm yeh thats true. Its something we doing that may surprise the user with no explanation. Maybe its worth something in documentation then. Also, could there be a situation where the truncation results in messages being incorrectly grouped?

rbev: Added a followup issue to document this somewhere

PR #5891, opened 2026-09-14
Load the HTTPS certificate while validating settings instead of at Kestrel bind time

ramonsmits:

`AddServiceControlHttps` loads the certificate inside the `ConfigureHttpsDefaults` callback, which Kestrel invokes when binding endpoints — the last step of startup. An unusable certificate therefore fails only after RavenDB, the transport and every hosted service have started and have to be torn down again. The reported error is .NET's own text, which names neither the file nor the setting that supplies the password.

Two further certificate problems were not reported at all or only at bind time:

- A PFX without a private key loads fine and Kestrel binds it, but every TLS handshake then fails. The only symptom is that clients cannot connect.
- A certificate whose Extended Key Usage (EKU) extension excludes Server Authentication is rejected by Kestrel, but again only when the endpoint is bound.

The certificate is now loaded and validated while `HttpsSettings` is constructed, before the host is built, and `AddServiceControlHttps` uses the loaded instance. Validation rejects, in order:

1. A file that cannot be loaded. The message names the file, its size and last-modified time, whether a password was configured (never the password itself, since .NET collapses several unrelated causes into "the password may be incorrect"), and the root cause via `GetBaseException()`. For an empty file that exposes the underlying `EndOfStreamException` that the outer `CryptographicException` hides; for the other cases both messages are identical.
2. A certificate without a private key.
3. A certificate with an EKU extension that lacks Server Authentication (OID 1.3.6.1.5.5.7.3.1). This mirrors Kestrel's own rule, including accepting a certificate that has no EKU extension at all, so nothing that binds today is refused.

Every message ends with the same escape hatch: set `Https.Enabled` to `false` to start without HTTPS while investigating.

The stale ".pfx or .pem" wording in the property documentation and the required-path message was replaced with PKCS#12 / .pfx, which is what the loader accepts.

#### Describe alternatives you've considered

Improving only the message and leaving the load where it is: keeps the discarded startup work and the ungraceful teardown that follows a failed `Host.StartAsync`.

Leaving the EKU rule to Kestrel: it is the last certificate failure that would still surface only at bind time, and the check is a few lines that follow Kestrel's rule exactly.

#### Additional context

The test fixture used an empty `Path.GetTempFileName()` as a stand-in certificate, which only worked because nothing opened it; it now generates real self-signed PFX files, optionally with a password or a specific EKU. Tests added for loading, a wrong password, and a client-authentication-only EKU. The private key check has no dedicated test since it only forwards `X509Certificate2.HasPrivateKey`.


src/ServiceControl.Infrastructure/HttpsSettings.cs
```diff
@@ -103,6 +110,29 @@ void ValidateCertificateConfiguration()
             logger.LogCritical(message);
             throw new InvalidOperationException(message);
         }
+
+        // Loaded here rather than when Kestrel binds its endpoints: an unusable certificate is a
+        // configuration error, and binding happens only after every hosted service has started.
+        try
+        {
+            Certificate = string.IsNullOrEmpty(CertificatePassword)
```
warwickschroeder, line 118: Another scenario to consider is that the certificate file has no private key. This currently wont throw but just not allow any conections. Could we also check for `!Certificate.HasPrivateKey` and throw if false?

src/ServiceControl.Infrastructure/HttpsSettings.cs
```diff
@@ -103,6 +110,29 @@ void ValidateCertificateConfiguration()
...
+        }
+        catch (Exception ex)
+        {
+            // .NET reports several unrelated causes as "the password may be incorrect", so describe
+            // the file itself too. Never the password, only whether one was configured.
+            var file = new FileInfo(CertificatePath);
+            var message = $"The HTTPS certificate could not be loaded, so this instance cannot start. " +
+                          $"Https.CertificatePath: '{CertificatePath}' ({file.Length} bytes, last modified {file.LastWriteTimeUtc:u}). " +
+                          $"Https.CertificatePassword configured: {!string.IsNullOrEmpty(CertificatePassword)}. " +
+                          $"{ex.GetType().Name}: {ex.Message} " +
```
warwickschroeder, line 130: Would this hide the actual issue? Should we use `ex.GetBaseException()`?

warwickschroeder (approved):

src/ServiceControl.Infrastructure/HttpsSettings.cs
ramonsmits: @warwickschroeder In the logs it showed:

```txt
2026-09-14 12:27:48.2887|00:00:12.500|14|Error|Microsoft.Extensions.Hosting.Internal.Host|Hosting failed to start|System.Security.Cryptography.CryptographicException: The certificate data cannot be read with the provided password, the password may be incorrect.
```

We should be good but I'll test just to be sure.

src/ServiceControl.Infrastructure/HttpsSettings.cs
ramonsmits: That makes sense indeed as we will be using the private key.

src/ServiceControl.Infrastructure/HttpsSettings.cs
ramonsmits: @warwickschroeder checked, exception message are correct, for a empty file this shows `0 bytes` and "Error occurred during a cryptographic operation.", using `GetBaseException` that would be " EndOfStreamException: "Unable to read beyond the end of the stream."" which think is not providing more useful response.

src/ServiceControl.Infrastructure/HttpsSettings.cs
ramonsmits: Check is added, also added an EKU check which Claude suggested as that is check Kestrel does internally.

> Kestrel's HttpsConnectionMiddleware constructor calls its own certificate check when the HTTPS endpoint is bound. If the certificate carries an Extended Key Usage extension that lacks Server Authentication, OID 1.3.6.1.5.5.7.3.1, Kestrel throws InvalidOperationException with the message "cannot be used as an SSL server certificate". A certificate with no EKU extension at all is accepted.

warwickschroeder (approved):

PR #5892, opened 2026-09-15
[Spike] Run audit ingestion in the primary instance

johnsimons:
> [!IMPORTANT]
> **Draft until the RFC concludes.** Whether this ships is being decided in [the Collaboration RFC](https://github.com/Particular/CloudPlatformExperience/blob/main/research/cloudxp-832/rfc-audit-ingestion-in-the-primary.md), which reverses the 2019 decision to give audit its own process. Review the code here; argue the direction there. If the RFC concludes against it, this branch is abandoned.

## What it does

A persister declares whether it can hold audit data alongside the primary's own. SQL Server and PostgreSQL do; RavenDB does not, and registers nothing. The line is drawn on write concurrency: both relational engines admit many concurrent writers to one database, so audit and error ingestion contend only at the row level, while Voron admits one writer per database and would put both streams behind a single write transaction.

Three deployment options, all the same executable and the same persister, differing only in which database each process gets and the mode it starts in:

| Option | Processes | For |
| --- | --- | --- |
| Shared database | `ServiceControl.exe` alone | The default. Removes a process, a container image and a database from a typical deployment |
| Scaled out audit ingestion | plus `ServiceControl.exe --audit-ingestion-only` | Audit throughput beyond one process, same database |
| Dedicated audit database | plus `ServiceControl.exe --audit-instance` | Audit volume that would degrade the error side. Reached through the existing scatter gather |

Every database has exactly one owner, which runs setup, migrations, retention and the API. Everything else on that database is a worker: no installers, no API, no retention, no licensing.

## Reviewing this

205 files is a lot. The commits are ordered so they can be read in sequence, and each is self-contained:

1. **Contracts and capability model** (`d2c6525b1`) - how a persister declares audit support.
2. **Audit runtime into the primary** (`959ecd765`, `c33fea60b`, `0a1eb0126`, `399c823b3`) - the receiver, the ingestion only command and its startup guards, the composed host, platform connection and licensing.
3. **EF audit schema and write path** (`694bd892f`, `84fba2d79`, `ffaaa67d7`, `457f9867d`) - tables, the shared ingestion helpers moved out of the error pipeline folder, the write path, schema-aware partitioning.
4. **Retention** (`7711eb5d5`) - hourly expiry, partition lifecycle, the session scoped retention lock.
5. **Queries** (`d3cc6277b`, `6ead2274f`) - audit counts, saga history, and the combined message views as one SQL statement over both tables.
6. **Failed audit imports** (`c0b48ca7f`).
7. **Enabling it** (`623acf0f1`) - the persisters advertise audit support.
8. **Dedicated audit host** (`f78e6cb04`) - `--audit-instance`.

The highest value files to review are the ingestion write path, the combined message view SQL, and the host startup guards.

## Testing

27 new test files:

- 14 EF audit persistence suites: write path, retention and partition lifecycle, message views, audit counts, saga history, failed imports.
- 3 audit ingestion acceptance tests, including the ingestion only and dedicated audit host compositions.
- Host startup guard unit tests for the mode combinations that are refused.
- Persistence manifest tests pinning that RavenDB does not advertise audit support and both relational persisters do.

Sustained local load, one process ingesting both queues into one PostgreSQL database over RabbitMQ: **~1,200 audit/s alongside ~500 errors/s**. The error rate is deliberately unrealistic; the question was whether audit ingestion starves error ingestion when both are under pressure, and it does not. Single run, developer hardware, no comparison against the separate audit instance.

## Known gaps, to be closed before this leaves draft

These are decisions taken during the RFC that the code has not caught up with yet:

- [ ] **`IngestAuditMessages` precedence rule.** The setting currently defaults to `true` unconditionally, and a primary with both local audit ingestion and remotes refuses to start. Both must change: resolve by precedence (explicit setting wins, otherwise off when remotes are configured, otherwise on), and allow local plus remotes, since that is the supported migration configuration.
- [ ] **Two way schema probe.** `DatabaseSchemaProbe` only detects the database being *behind* the host. An old worker against a newly migrated database sees nothing pending and starts. It needs to detect the database being ahead too, and stop the host if the schema moves while it is running.
- [ ] **The probe should run in every host**, not only the ingestion only ones. A primary upgraded but not set up currently fails on its first write.
- [ ] **Co-located worker identity.** Two workers on one machine sharing an instance name overwrite each other's custom check rows. Documented; decide whether to leave it documented or derive host id from something process unique.
- [ ] **Spike plan documents.** `src/audit-ingestion-in-primary-plan.md` and `src/audit-ef-persistence-plan.md` are working documents from the spike and overlap `docs/audit-ingestion-in-the-primary.md`. Fold anything still current into the design document and delete them.

## Not in scope

- RavenDB, which keeps its own audit instance, executable, settings, API and installers.
- The monitoring instance.
- One primary process holding audit and error in two different databases. Considered and declined in the RFC, with a telemetry criterion that would reopen it.

## Design

`docs/audit-ingestion-in-the-primary.md`, added by this PR, is the design of record: deployment options, settings and the precedence rule, retiring an audit instance without migrating data, queue ownership, schema upgrades, body storage, health endpoints, retention, querying and telemetry.


docs/audit-ingestion-in-the-primary.md
```diff
@@ -0,0 +1,202 @@
...
+
+Storage that advertises `SupportsAuditIngestion` in its `persistence.manifest` can hold audit data
+alongside the primary's own data, which lets the primary ServiceControl process ingest the audit
+queue itself instead of relying on a separate ServiceControl.Audit instance.
+
+The standalone RavenDB audit instance is unaffected. RavenDB does not advertise audit support, does
+not gain combined hosting, and keeps its own executable, settings, API and installers.
+
+The SQL Server and PostgreSQL persisters advertise audit support, so a primary on either ingests
+the audit queue by default. On RavenDB the audit component registers nothing and behavior is
```
rbev, line 13: I think this ingestion should be an explicit opt-in (defaulted on for new installs) since a migrating user might be caught off guard by their error instance suddenly consuming their audits.

docs/audit-ingestion-in-the-primary.md
```diff
@@ -0,0 +1,202 @@
...
+| `ServiceControl/MessageBody/FileSystem/PathIsShared` | `false` | Required by both ingestion only modes when body storage is the file system |
+
+### Setting collisions
+
+`ServiceControl` and `ServiceControl.Audit` settings can both be set by bare environment variable
+name, and `ServiceBus/AuditQueue` is literally the same key for both processes. A combined primary
+and a standalone audit instance sharing one environment file therefore collide on
+`INGESTAUDITMESSAGES`, `AUDITRETENTIONPERIOD`, `FORWARDAUDITMESSAGES` and `SERVICEBUS_AUDITQUEUE`.
+
+That combination is unsupported. A primary that ingests audit into its own database and also lists
```
rbev, line 94: If we aren't migrating audit then doesn't this break the migration path into a single Sql server from Error+Audit on raven?

How would you go from this:
```
Error -> RavenEmbedded
Audit -> RavenEmbedded
```
To this, if the error refuses to start with remotes? 
```
Error With Audit ingestion -> Sql
Audit (No ingestion, decomission when empty) -> RavenEmbedded

File: docs/audit-ingestion-in-the-primary.md [hunk 3]
```diff
@@ -0,0 +1,202 @@
...
+
+## Queue ownership
+
+The setup path of a database's owner creates the audit queue, and the audit forwarding queue when
+forwarding is enabled. Ingestion only workers run no installers: they never create queues, never
+apply database migrations and never provision body storage. Run setup from the owner before starting
+any worker; a worker started against a database the owner has not migrated refuses to start and says
+so, rather than failing on its first write.
+
+The primary and a dedicated audit database run the same migrations, so the owners are upgraded
```
rbev, line 106: The paragraph above says that it guards against startup and never upgrades the db, then this paragraph says they can be upgraded in any order. Seems like these are incompatible constraints/requirements.

docs/audit-ingestion-in-the-primary.md
```diff
@@ -0,0 +1,202 @@
...
+Endpoints detected from audit headers are written straight to the `KnownEndpoints` table of the
+database being ingested into, through the ingestion unit of work. On a dedicated audit database they
+are additionally reported to the primary's input queue, since that table is not the primary's.
+
+## Body storage
+
+Audit and failed message bodies share one store, and each owns a prefixed keyspace, so an edited
+message's failed body and its audited body do not collide. Audit bodies are keyed
+`audit/{ingestion hour}/{unique message id}`, which is what lets retention drop an hour's bodies in
+one operation. `IBodyStorage.TryFetch` resolves in a fixed order: failed message by
```
rbev, line 127: It seems weird that the audit and error bodies are merged together at read time despite being explicitly stored separately

docs/audit-ingestion-in-the-primary.md
```diff
@@ -0,0 +1,202 @@
...
+`UniqueMessageId`, then failed message by `MessageId`, then audit message by `UniqueMessageId`.
+
+Hosts on one database share one body store: the audit host and its workers share the audit store,
```
rbev, line 130: Do we need to allow a single primary instance do audit/error on different databases? It seems like it's a bit of an uneccessary limitation/cost to require a separate instance to allow that scenario. 
In that case would they still share a body storage?

docs/audit-ingestion-in-the-primary.md
```diff
@@ -0,0 +1,202 @@
...
+the primary and its workers share the primary's.
+
+Every ingesting process must write bodies somewhere every host can read. Blob and S3 storage
+qualify. File system storage qualifies only if the path is a shared mount, which nothing in the
+settings can detect, so both ingestion only modes refuse to start unless
+`ServiceControl/MessageBody/FileSystem/PathIsShared` asserts it.
+
+## Health endpoints
```
rbev, line 138: we should answer the question about how these can appear in the platform health page, that currently relies on the remotes list.

rbev (commented):
Design doc comments/questions

src/ServiceControl.AcceptanceTests/RootControllerTests.cs
```diff
@@ -19,6 +19,8 @@ public async Task Should_gather_remote_data()
 
             SetSettings = settings =>
             {
+                // A primary with audit remotes holds no audit data of its own.
```
rbev, line 22: Shouldn't it be possible to have both? 
e.g. migrating from single primary to error + audit instance to move audit data onto a different cluster, there will be a cutover where there's still local data aging out.

docs/audit-ingestion-in-the-primary.md
johnsimons: How is this different from today using scatter-gather 😉.  We are doing exactly this, except it is worse, becasue we do it all in memory!

src/ServiceControl.Persistence.EFCore/Implementation/UnitOfWork/EFIngestionUnitOfWork.cs
```diff
@@ -1,38 +1,47 @@
...
-// add to thread safe collections. Every database call happens in Complete, on one thread.
+// The Record methods run concurrently across the batch, so they only add to thread safe
+// collections. Every database call happens in Complete, on one thread, inside one transaction
+// shared by the failed message, known endpoint and audit rows of the batch.
 public class EFIngestionUnitOfWork : IIngestionUnitOfWork
 {
     readonly ServiceControlDbContext dbContext;
     readonly IAsyncDisposable scope;
     readonly IFailedMessageIngestionSqlDialect dialect;
+    readonly IAuditIngestionSqlDialect auditDialect;
```
rbev: What's the rationale behind combining audit and error ingestion units of work into one?

src/ServiceControl.AcceptanceTests.RavenDB/ServiceControl.AcceptanceTests.RavenDB.csproj
```diff
@@ -42,6 +42,10 @@
 
     <!-- RavenDB is the one persister that supports maintenance mode; StartupModeTests covers it. -->
     <Compile Remove="..\ServiceControl.AcceptanceTests\MaintenanceModeTests.cs" />
+
+    <Compile Remove="..\ServiceControl.AcceptanceTests\Auditing\When_composing_audit_ingestion_in_the_primary.cs" />
```
rbev, line 46: should this be a wildcard exclude since raven will never support primary audit ingestion?

src/ServiceControl.Persistence.EFCore.PostgreSql/AuditPartitioningSql.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            var name = Qualify(schema, table);
+            var clone = Qualify(schema, $"{table}_tmp");
+
+            return $"""
+                CREATE TABLE {clone} (LIKE {name} INCLUDING ALL);
+                DROP TABLE {name};
+                CREATE TABLE {name} (LIKE {clone} INCLUDING ALL){partitionBy};
+                DROP TABLE {clone};
+
```
rbev, line 54:
```suggestion
```

src/ServiceControl.Persistence.EFCore/Abstractions/EFPersisterSettings.cs
```diff
@@ -8,6 +8,10 @@ public abstract class EFPersisterSettings : PersistenceSettings
     public const int DefaultExternalIntegrationsDispatchingBatchSize = 100;
     public static readonly TimeSpan DefaultEventsRetentionPeriod = TimeSpan.FromDays(14);
 
+    // The same default the management utility and the container image apply, rather than the
```
rbev, line 11: Why wouldn't this mirror the audit instance defaults?

src/ServiceControl.Persistence.EFCore/Implementation/Audit/IAuditPartitionManager.cs
```diff
@@ -0,0 +1,38 @@
...
+    Task<HourDrop> DropHour(ServiceControlDbContext dbContext, DateTime hour, int batchSize, CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// When the newest provisioned partition ends, or null where the provider does not partition
+    /// and so can never run out of provisioned hours.
+    /// </summary>
+    Task<DateTime?> NewestProvisionedHourEnd(ServiceControlDbContext dbContext, CancellationToken cancellationToken = default);
+}
+
+public readonly record struct HourDrop(int RowsDeleted, bool Completed);
```
rbev, line 38: this return type is a bit cryptic on it's own, the documentation for what "Complete" means is on the method that returns it rather than this record property.

PR #5894, opened 2026-09-15
[Migration Engine Part 1] Implement the foundation for migration engine

warwickschroeder:
## What this adds

The foundation for moving an error instance's data from RavenDB to SQL Server or PostgreSQL. **Nothing is copied yet**, and the migration is not switched on anywhere.

- **A read-only RavenDB source.** `RavenMigrationSource` opens both RavenDB databases, embedded or external, through a client that refuses anything outside a read allow-list (`OnBeforeRequest`, non-tracking sessions), so a patch, a bulk insert or a HiLo range throws in-process rather than reaching the database. It can describe itself and count every collection; `Count`, `Read` and `ReadBody` for a category all throw `NotSupportedException`, so no reader exists yet.
- **`--migration-source-report`.** Prints the source's RavenDB version, where it is, both database names with the setting each came from, and a row count per collection. It reads the instance's own RavenDB settings; `ServiceControl/Migration/SourcePersistenceType` defaults to `RavenDB`, and `Settings.MigrationSourcePersistenceType` now appears in the approved platform sample settings.
- **Two persisters in one process.** `PersistenceFactory.CreateMigrationSource` loads the source persister in its own `AssemblyLoadContext`, beside the configured one, and `OpenMigrationSource` opens it.
- **The migration engine, against fakes.** Store-neutral seams in `ServiceControl.Persistence/DataMigration` (`IMigrationSource`, `IMigrationTarget`, `IMigrationCheckpointStore`, the category registry) and `MigrationEngine`, which copies a category in batches with checkpoint resume, ordering between categories, body read retries, throttling, skip reasons and a halt threshold. A checkpoint carries a `Version` and the target extends it inside the transaction that writes the rows, so a save that loses a race throws `MigrationCheckpointConflictException` and leaves the other writer's row standing rather than halting the category. There is no SQL target yet; it runs against in-memory fakes.
- **Docs and CI.** `docs/migration/` holds the overview, a system design diagram, and short instructions for what can be run today, linked from the README. `.github/workflows/ci.yml` gains a `Migration` test category, which also downloads the RavenDB server.

## Tests

- `ServiceControl.UnitTests/Migration`: engine, registry, options and the fakes.
- `ServiceControl.UnitTests/Hosting`: `--migration-source-report` argument parsing.
- `ServiceControl.UnitTests/ApprovalFiles`: the platform sample settings, which gained the new setting.
- `ServiceControl.Persistence.Tests.RavenDB/DataMigration`: the read-only source lifecycle.
- `ServiceControl.Migration.Tests` (new): the source report command, and both persisters loaded in one process.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
         # Categories are declared by the <TestCategory> property in each test project. A category can
         # span several projects that share infrastructure, so that the container is provisioned once
         # and the union of their build closures is compiled once.
-        test-category: [ DefaultCore, DefaultMonitoring, Raven, SqlServer, PostgreSql, RabbitMQ, AzureServiceBus, AzureStorageQueues, MSMQ, SQS, IBMMQ ]
+        test-category: [ DefaultCore, DefaultMonitoring, Migration, Raven, SqlServer, PostgreSql, RabbitMQ, AzureServiceBus, AzureStorageQueues, MSMQ, SQS, IBMMQ ]
         include:
           - os: windows-latest
             os-name: Windows
@@ -87,7 +87,7 @@ jobs:
         id: select
         run: ./tools/select-test-projects.ps1 -Category ${{ matrix.test-category }}
       - name: Download RavenDB Server
-        if: matrix.test-category == 'DefaultCore'
+        if: contains(fromJSON('["DefaultCore", "Migration"]'), matrix.test-category)
         run: ./tools/download-ravendb-server.ps1
       - name: Build
         id: build
@@ -128,14 +128,14 @@ jobs:
       # purpose: with the VM and the build competing for the runner's memory, jobs have died with "The
       # hosted runner lost communication with the server".
       - name: Setup WSL
-        if: contains(fromJSON('["RabbitMQ", "SqlServer", "PostgreSql", "IBMMQ"]'), matrix.test-category)
+        if: contains(fromJSON('["RabbitMQ", "SqlServer", "PostgreSql", "IBMMQ", "Migration"]'), matrix.test-category)
         uses: Particular/setup-wsl-action@v1.2.0
         with:
           # The action defaults to 4GB. The runner has 16GB, so give the VM real headroom.
           memory: 8GB
       - name: Setup SQL Server
         uses: Particular/install-sql-server-action@v3.0.0
-        if: matrix.test-category == 'SqlServer'
+        if: contains(fromJSON('["SqlServer", "Migration"]'), matrix.test-category)
         with:
           connection-string-env-var: ServiceControl_Persistence_SqlServer_ConnectionString
           catalog: ServiceControl
--- a/docs/ravendb-to-sql-migration-overview.md
+++ b/docs/ravendb-to-sql-migration-overview.md
@@ -1,0 +1,246 @@
+# Moving data from RavenDB to SQL
+
+## Problem
+
+A customer can already point ServiceControl at SQL Server or PostgreSQL. They cannot bring their existing data with them.
+
+## Strategy
+
+- Switch over first, and copy only what has to be copied. Retention does most of the work: error retention is between 5 and 45 days and event retention is shorter, so most of the source ages out on its own within weeks. That is why archived and resolved messages are optional rather than required. Retention would have deleted them anyway.
+- The required set is small because unresolved failures are the only category a customer can act on, and its assumed that customers keep that number low by resolving and archiving.
+- Anything not selected simply ages out of RavenDB, and the customer deletes the old database when they are ready.
+
+## Goals
+
+- **Minimal downtime**. Only the required data copies with ServiceControl closed. Optional data copies in the background while it serves traffic.
+- **All three RavenDB sources are supported**. Embedded, a container, or RavenDB Cloud, on one code path rather than three.
+- **No writes through the client**. Nothing the copier does changes the source, so the old database stays a valid fallback.
+- **Abandonable up to a known point**. Free while ServiceControl is closed. After it opens, abandoning loses whatever SQL has ingested since.
+- **No duplicates and no gaps**. Rows and the resume cursor commit in one transaction, so a crash needs no reconciliation.
+- **Every identifier anything depends on is carried across**. The event log and historic retry operations are renumbered, because nothing references their keys.
+- **Refuse rather than half-migrate**. Every check runs before the first row moves, and a failure is a host that will not start.
+- **No silent loss**. A migration cannot end with a selected category incomplete, and skipped rows are counted and reported.
+- **Bounded impact on a live instance**. Throttled behind normal ingestion and streamed, so memory does not track the size of the database.
+- **Known before it starts, visible while it runs**. A dry run reports what will move and how long ServiceControl is closed, and every category transition is reported as it happens.
+- **Use existing functionality where possible**. Progress goes through custom checks and the activity feed, so no new client or screen is needed.
+
+## Nice to have, but not planned in this initial design (phase 1)
+
+- **Zero downtime.** The required data is copied with ServiceControl closed, so there is a real, if short, outage.
+- **Reversible once ServiceControl opens.** Nothing copies SQL rows back to RavenDB.
+- **Steerable while running.** No pause, resume, or abort. Changing anything means editing configuration and restarting.
+- **A general-purpose migration tool.** The source is always RavenDB and the target is always a ServiceControl EF Core persister, both at versions this build can read.
+- **Custom migration UI via ServicePulse.** Custom checks and the event log will be used for progress reporting, but migration configuration and migration engine control will not be available via the UI.
+
+## Supported migration scenarios
+
+The copier runs inside the ServiceControl host, so every row and every message body travels from the source, through ServiceControl, to the target. There is no database-to-database transfer, no backup and restore, and no replication. Whether a combination works therefore comes down to whether the ServiceControl host can reach both ends at once, and how long it takes comes down to how far the data has to travel.
+
+**Supported locations:**
+
+- **The RavenDB source**: embedded on the ServiceControl host, self-hosted on the same network in a container, VM or bare metal, or RavenDB Cloud
+- **The SQL target**: SQL Server or PostgreSQL on the ServiceControl host, elsewhere on the same network, in a container, or as a managed cloud service such as Azure SQL, Amazon RDS or Google Cloud SQL
+- **Any combination of the two**, subject to the requirements below
+
+**By where the data has to travel:**
+
+- **On-prem to on-prem.** The common case and the fastest. Embedded or self-hosted RavenDB to SQL on the same host or the same network.
+- **On-prem to cloud.** RavenDB on the network, managed SQL in a cloud. Works, but each batch is one round trip, so write latency multiplies by the number of batches rather than being amortised away.
+- **Cloud to on-prem.** RavenDB Cloud down to local SQL. Works, and the customer pays egress on everything copied, most of which is archived messages and their bodies.
+- **Cloud to cloud.** Works, and is only sensible when ServiceControl runs alongside one of them. A host sitting on-prem between two clouds pulls every byte down and pushes it straight back up.
+- **The message body store is a third location.** Bodies come out of RavenDB attachments and go wherever the target is configured to put them: a filesystem, Azure Blob or S3, with small text bodies kept inline in the database. That decision is made at the same time as the database move.
+
+**Infrastructure requirements:**
+
+- The ServiceControl host needs network access to the RavenDB source, the SQL target and the body store simultaneously
+- Both RavenDB databases, primary and throughput, on one server or cluster
+- A SQL Server target must have Full-Text Search installed. `--setup` checks `SERVERPROPERTY('IsFullTextInstalled')` and fails if it is absent, because message search is not optional. A stock SQL Server container image does not include it. PostgreSQL needs nothing extra, since its index is a GIN over `to_tsvector`
+- A managed target's transient failures are already survivable: retry on failure is on by default and there is no setting to turn it off
+
+**Not supported:**
+
+- Any server-to-server copy: no backup and restore, no RavenDB ETL or replication into SQL, no external data pipeline
+- A host that can reach only one of the two databases at a time, so no staged move by way of an offline copy
+- Primary and throughput RavenDB databases in different locations
+- 🍅 Merging two ServiceControl instances into one SQL database, or splitting one instance across several
+- Anything but RavenDB as the source, or anything but a ServiceControl EF Core persister as the target
+
+🍅 Unverified: a source whose RavenDB licence has expired. The licence check inside `DatabaseSetup.Execute` is RavenDB's own, polling the server's `/license/status` and refusing an expired one, and the read-only lifecycle never calls it. So the one thing that would have blocked this is removed by design. Whether the RavenDB server itself still serves reads on an expired licence is a question about RavenDB's behaviour and licence terms that reading our code cannot answer, and one test would settle it. A customer whose licence has expired is exactly the one most likely to be leaving RavenDB.
+
+## Migration workflow
+
+1. Upgrade ServiceControl as normal, still on RavenDB.
+2. Set four things in configuration: the new `PersistenceType`, its connection string, `MigrationMode=true`, and which [optional data](#data-to-be-migrated) they want copied.
+3. Run `--setup` to creates the SQL schema. It fails against a SQL Server instance without Full-Text Search installed.
+4. Run the [dry run](#dry-run). It reports what it resolved as a source, what each category holds, 🍅 and how long ServiceControl will be closed, so nobody starts a copy without knowing what it will do.
+5. Start ServiceControl (`MigrationMode=true`).
+6. Every check runs before a single row moves. If one fails the host does not start and names which, having copied nothing, so a wrong database name or unconfigured body storage costs a restart rather than a half-finished migration.
+7. The copying of [required data](#required) starts, with ServiceControl still closed. This is assumed to be a small amount of data.
+8. ServiceControl opens, and whatever [optional data](#optional) they asked for is copied in the background while the instance runs normally. They can watch it from ServicePulse custom checks and events, but not steer it.
+9. They run the verification pass once the background job has completed, which reports row counts on both sides category by category, accounting for deliberate skips so a difference is explained rather than reported as a fault, then set `MigrationMode=false` and restart.
+10. RavenDB data can be removed.
+
+- If `MigrationMode=false` is set while a selected category is still incomplete, the host refuses to start and names exactly what is outstanding.
+- A category that ended *complete with errors* counts as complete and does not block, though its skipped count is printed so the loss is stated rather than silent.
+- An explicit override exists for a customer who has changed their mind and accepts leaving data behind.
+
+## Architecture
+
+```mermaid
+flowchart TB
+    cfg["Configuration + restart<br/>the only way to change anything"]
+    checks["Custom checks + activity feed<br/>progress, with no new client needed"]
+
+    subgraph host["One ServiceControl host process, started with MigrationMode = true"]
+        direction LR
+        raven["RavenDB persister<br/>own AssemblyLoadContext<br/>new read-only lifecycle"]
+        engine["MigrationEngine<br/>categories, throttle,<br/>dry run, verification"]
+        target["EF Core persister<br/>SQL Server or PostgreSQL<br/>own AssemblyLoadContext"]
+        raven -->|"IMigrationSource"| engine
+        engine -->|"IMigrationTarget"| target
+    end
+
+    old[("Old RavenDB<br/>read only, never written to")]
+    sql[("SQL Server or PostgreSQL<br/>plus a new checkpoint table")]
+    bodies[("Message body store<br/>filesystem, Azure Blob or S3")]
+
+    cfg --> host
+    host --> checks
+    old --> raven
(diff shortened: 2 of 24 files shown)


docs/migration/ravendb-to-sql-migration-instructions.md
```diff
@@ -0,0 +1,54 @@
...
+| Setting | Environment variable | What it is |
+| --- | --- | --- |
+| `ServiceControl/RavenDB/ConnectionString` | `SERVICECONTROL_RAVENDB_CONNECTIONSTRING` | An external RavenDB server. Leave unset for an embedded database |
+| `ServiceControl/DbPath` | `SERVICECONTROL_DBPATH` | The embedded database's data directory |
+| `ServiceControl/RavenDB/DatabaseName` | `SERVICECONTROL_RAVENDB_DATABASENAME` | The primary database, `primary` by default |
+| `LicensingComponent/RavenDB/ThroughputDatabaseName` | `LICENSINGCOMPONENT_RAVENDB_THROUGHPUTDATABASENAME` | The throughput database, `throughput` by default |
+| `ServiceControl/RavenDB/ClientCertificatePath` or `ServiceControl/RavenDB/ClientCertificateBase64`, with `ServiceControl/RavenDB/ClientCertificatePassword` | `SERVICECONTROL_RAVENDB_CLIENTCERTIFICATEPATH` and so on | A secured external server's client certificate |
+| `ServiceControl/ErrorRetentionPeriod` | `SERVICECONTROL_ERRORRETENTIONPERIOD` | Required. Don't change it during the move |
+
+`ServiceControl/Migration/SourcePersistenceType` defaults to `RavenDB` and needs no setting.
```
rbev, line 21: So given the abstractions does this technically allow two way migration? or will that be missing the migrationsource implementation?

If this defaults to the only permissible value should it even be a configuration knob?

docs/migration/ravendb-to-sql-migration-overview.md
```diff
@@ -0,0 +1,294 @@
...
+- Primary and throughput RavenDB databases in different locations
+- Anything but RavenDB as the source, or anything but a ServiceControl EF Core persister as the target
+
+## Migration workflow
+
+1. Upgrade ServiceControl as normal, still on RavenDB.
+2. Set four things in configuration: the new `PersistenceType`, its connection string, `MigrationMode=true`, and which [optional data](#data-to-be-migrated-categories) they want copied.
+3. Run `--setup` to create the SQL schema. It fails against a SQL Server instance without Full-Text Search installed.
+4. Run the [dry run](#dry-run). It reports what it resolved as a source, what each category holds, and an estimate of how long ServiceControl will be closed. Read [what the dry run reports](#dry-run) before booking an outage around its estimate.
```
rbev, line 75: Should make sure that dry-run checks that the sql schema has been created, just for completeness given item 3

docs/migration/ravendb-to-sql-migration-overview.md
```diff
@@ -0,0 +1,294 @@
...
+- `UniqueMessageId` keeps its value, but converts type: the source holds a string and the target column is a `uniqueidentifier`. It is the primary key, the ServicePulse URL, the retry correlation key and the body lookup key at once.
+- `StatusChangedAt` is reconstructed from `@expires` for resolved and archived messages, which is the only place RavenDB sets it. Unresolved and retry-issued messages have no `@expires`, so the copier uses the newest processing attempt's timestamp. The column is `NOT NULL`, so it cannot be left empty, but the value is harmless for those two: the retention sweep only considers resolved and archived rows, so an unresolved message never ages out whatever is written here.
+- Message bodies go through `IBodyStoragePersistence`, which owns the compression threshold and the choice of filesystem, Azure Blob or S3. The separate 102,400-byte inline threshold is not there: it lives on the ingestion path, so the copier has to apply it rather than inherit it.
+- Throughput rows are written directly rather than through the collector, and the write sets each day's count rather than adding to it.
```
rbev: Is throughput copied after SC starts up?
Overwrite rather than merge seems problematic here if it is, if not then it needs to be called out

docs/migration/ravendb-to-sql-migration-overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+**A category can finish with a small amount of loss and still count as complete.** A few skipped rows in a large table leave the category in a *complete with errors* state, which blocks nothing. Its skipped count is printed and the ids of the skipped rows are written to the log, so while the RavenDB database still exists you can go and look at exactly what did not make it.
+
+## The one point you can go back
+
+While ServiceControl is closed and the required copy is running, nothing except the copier has written to SQL, and the migration has written nothing to RavenDB, which is still authoritative. RavenDB's own expiration still runs, though: unless you disabled it, it keeps deleting expired failed messages and event log items, as [Goals](#goals) describes. Back up both RavenDB databases, or disable expiration on them, before you start. If you need your instance back, set `MigrationMode=false`, point `PersistenceType` back at RavenDB, and start. You lose the copy, not your data, and you can start again later.
```
rbev: How does the count work when ravenDB is actively deleting rows from underneath you? Do they just end up in a "skipped" bucket with an unknown reason because you ran off the end of the work with less rows than you expected?

src/ServiceControl.Persistence/DataMigration/IMigrationTarget.cs
```diff
@@ -0,0 +1,25 @@
...
+
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+/// <summary>Implemented by a persister that can be the new database a migration copies into.</summary>
+public interface IMigrationTarget
+{
+    /// <summary>How many rows to read per batch for this category. The target picks it because its own database sets the limits.</summary>
+    int BatchSizeFor(MigrationCategory category);
```
rbev, line 11: Should this be `MaximumBatchSize` so it's explicit that it's a limit rather than a target to the consumer?

src/ServiceControl.Persistence/DataMigration/IMigrationTarget.cs
```diff
@@ -0,0 +1,25 @@
...
+
+    /// <summary>Saves the batch's rows and checkpointAfterBatch in one go, so progress never gets ahead of the data. Save the checkpoint exactly as given, without adding counts to it.</summary>
```
rbev:
```suggestion
    /// <summary>Saves the batch's rows and checkpointAfterBatch in one transaction, so progress is consistent with the data. Save the checkpoint exactly as given, without adding counts to it.</summary>
```

src/ServiceControl.Persistence/DataMigration/MigrationEngine.cs
```diff
@@ -0,0 +1,265 @@
...
+                {
+                    Cursor = batch.Cursor,
+                    CopiedCount = checkpoint.CopiedCount + batchToWrite.Rows.Count,
+                    SkippedCount = checkpoint.SkippedCount + bodySkips,
+                    SkipReasons = AddSkipReasons(checkpoint.SkipReasons, bodySkips == 0 ? null : new Dictionary<string, long> { [nameof(MigrationSkipReason.BodyUnreadable)] = bodySkips })
+                };
+
+                var result = await target.Write(category, batchToWrite, checkpointAfterBatch, cancellationToken);
+
+                checkpoint = checkpointAfterBatch with
```
rbev: If the `Write` call writes `checkpointAfterBatch` as given (as described in the interface specification) then how do these extra counts that are being added on from the `result` survive a crash/reboot before the next batch commits?

src/ServiceControl.Persistence/DataMigration/MigrationEngine.cs
```diff
@@ -0,0 +1,265 @@
...
+
+                foreach (var id in result.SkippedIds)
+                {
+                    logger.LogWarning("Skipped {SourceId} in category {CategoryId}", id, category.Id);
+                }
+
+                processedThisRun += bodySkips + result.Copied + result.Skipped + result.AlreadyPresent;
+                skippedThisRun += bodySkips + result.Skipped;
+
+                if (HaltThreshold.Exceeded(skippedThisRun, processedThisRun, options.HaltThresholdPercent, options.HaltThresholdMinimum))
```
rbev, line 141: Halt threshold is based on percent of progress rather than percent of total, does this cause a problem if there are a small number of invalid records at the start of the migration that would otherwise be ok?

rbev (commented):
Partial review - adding as comments for now.

docs/migration/ravendb-to-sql-migration-instructions.md
warwickschroeder: The abstraction is designed to allow for possible "any direction" migrations, but there will not be an implementation for SQL as the source or Raven as the target for this version.

Yep - I'll remove that setting as it is redundant at the moment. I did think about removing this..

docs/migration/ravendb-to-sql-migration-overview.md
warwickschroeder: Part of the design is that there will be a bunch of source and target checks (implemented at that persistence layer) that run before migration starts. Checking the schema can be part of the SQL target checks.

src/ServiceControl.UnitTests/Migration/MigrationEngineCategorySelectionTests.cs
```diff
@@ -0,0 +1,56 @@
...
+    static MigrationEngine BuildEngine(IReadOnlyCollection<string> selectedOptionalIds, out InMemoryMigrationCheckpointStore checkpointStore)
+    {
+        checkpointStore = new InMemoryMigrationCheckpointStore();
+        var target = new InMemoryMigrationTarget(checkpointStore);
+        var options = new MigrationEngineOptions(TimeSpan.FromSeconds(1), 5, 100, selectedOptionalIds);
+        return new MigrationEngine(new InMemoryMigrationSource(), target, checkpointStore, new FakeTimeProvider(), options, NullLogger<MigrationEngine>.Instance);
+    }
+
+    [Test]
+    public void All_twelve_required_categories_are_always_selected()
```
rbev: this should be an approval test so that ordering can be validated

src/ServiceControl.UnitTests/Migration/MigrationEngineCategorySelectionTests.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+        var engine = BuildEngine([], out _);
+
+        Assert.That(engine.SelectCategories(MigrationCategoryKind.Required), Has.Count.EqualTo(12));
+    }
+
+    [Test]
+    public void Only_configured_optional_categories_are_selected()
+    {
+        var engine = BuildEngine(["EventLog"], out _);
```
rbev: This should also be an approval test

src/ServiceControl.Persistence/DataMigration/IMigrationTarget.cs
warwickschroeder: Have update the comment for it. Its a ceiling for SQL server, but more of a target for postgres. The consumer never sees this anyway

docs/migration/ravendb-to-sql-migration-instructions.md
```diff
@@ -0,0 +1,54 @@
...
+
+| Setting | Environment variable | What it is |
+| --- | --- | --- |
+| `ServiceControl/RavenDB/ConnectionString` | `SERVICECONTROL_RAVENDB_CONNECTIONSTRING` | An external RavenDB server. Leave unset for an embedded database |
+| `ServiceControl/DbPath` | `SERVICECONTROL_DBPATH` | The embedded database's data directory |
+| `ServiceControl/RavenDB/DatabaseName` | `SERVICECONTROL_RAVENDB_DATABASENAME` | The primary database, `primary` by default |
+| `LicensingComponent/RavenDB/ThroughputDatabaseName` | `LICENSINGCOMPONENT_RAVENDB_THROUGHPUTDATABASENAME` | The throughput database, `throughput` by default |
+| `ServiceControl/RavenDB/ClientCertificatePath` or `ServiceControl/RavenDB/ClientCertificateBase64`, with `ServiceControl/RavenDB/ClientCertificatePassword` | `SERVICECONTROL_RAVENDB_CLIENTCERTIFICATEPATH` and so on | A secured external server's client certificate |
+| `ServiceControl/ErrorRetentionPeriod` | `SERVICECONTROL_ERRORRETENTIONPERIOD` | Required. Don't change it during the move |
```
johnsimons, line 19: So are these exclusively being used for the migration? And if so, should then be something like `ServiceControl/Migration\....`?

docs/migration/ravendb-to-sql-migration-overview.md
```diff
@@ -0,0 +1,294 @@
...
+- Failed error imports, the record of errors that could not be ingested
+- Group comments, **copied last of everything**, after archived and resolved messages. A comment survives only once the failed messages its group is built from have arrived, so on a large archive the comments are the last thing to appear. An empty comment field partway through a migration is the copy still running, not data loss
+- Failed message edits
+
+### Not migrated
+
+- The fifteen RavenDB index definitions, which map to a much smaller set of ordinary SQL indexes, and two of which are dead already
+- The transient in-flight collections, which are empty when nothing is running: `RetryBatches`, `RetryBatchNowForwardings`, `FailedMessageRetries`, `ArchiveOperations` and `UnarchiveOperations`
+- `ArchiveBatches` and `UnarchiveBatches`, which exist only because of how RavenDB works
+- `ConnectedApplications`, which only versions 6.0 and 6.1 wrote and nothing has read since
```
johnsimons, line 178: This is used for when the user is using the MassTransitConnector and this data is passed to ServicePulse to toggle features on/off

docs/migration/ravendb-to-sql-migration-overview.md
```diff
@@ -0,0 +1,294 @@
...
+- Archived and resolved failed messages: the biggest category by far, and most of the copying time
+- The event log, without which the ServicePulse activity feed starts empty
+- Custom checks, which cost almost nothing to skip because every check re-reports on its next interval
+- Failed error imports, the record of errors that could not be ingested
+- Group comments, **copied last of everything**, after archived and resolved messages. A comment survives only once the failed messages its group is built from have arrived, so on a large archive the comments are the last thing to appear. An empty comment field partway through a migration is the copy still running, not data loss
+- Failed message edits
+
+### Not migrated
+
+- The fifteen RavenDB index definitions, which map to a much smaller set of ordinary SQL indexes, and two of which are dead already
```
johnsimons: pointless mentioning?

docs/migration/ravendb-to-sql-migration-overview.md
```diff
@@ -0,0 +1,294 @@
...
+- The transient in-flight collections, which are empty when nothing is running: `RetryBatches`, `RetryBatchNowForwardings`, `FailedMessageRetries`, `ArchiveOperations` and `UnarchiveOperations`
+- `ArchiveBatches` and `UnarchiveBatches`, which exist only because of how RavenDB works
+- `ConnectedApplications`, which only versions 6.0 and 6.1 wrote and nothing has read since
+- Integration events still waiting to be sent when you switch over are never sent
```
johnsimons, line 179: anyway to mitigate this?

docs/migration/ravendb-to-sql-migration-overview.md
```diff
@@ -0,0 +1,294 @@
...
+- Throughput history
+- Retry operations, unacknowledged and historic. One category, because RavenDB holds both lists in a single document
+- Licensing report masks
+- The uploaded licensed endpoint details file, which nothing recomputes: skipping it means the customer re-downloads it from the licence portal and uploads it again
+- Subscriptions
+
+### Optional
+
+- Archived and resolved failed messages: the biggest category by far, and most of the copying time
+- The event log, without which the ServicePulse activity feed starts empty
```
johnsimons, line 167: I think I still see beneficial of bringing in at least the last 7 days or something like that

docs/migration/ravendb-to-sql-migration-overview.md
```diff
@@ -0,0 +1,294 @@
...
+- Custom checks, which cost almost nothing to skip because every check re-reports on its next interval
```
johnsimons, line 168: is this related to heartbeats?
I am asking becasue there is no mention of those?
One thing to watch out, is an instance is down, we will not get a heartbeat so we want to ensure that we somehow assume that everything is ok.

docs/migration/ravendb-to-sql-migration-overview.md
```diff
@@ -0,0 +1,294 @@
...
+- Failed error imports, the record of errors that could not be ingested
```
johnsimons, line 169: I don't know is we can ignore these at all, at the end of the day they are error messages that could not be ingested, we may need to notify the customer about them!

docs/migration/ravendb-to-sql-migration-overview.md
```diff
@@ -0,0 +1,294 @@
...
+- Group comments, **copied last of everything**, after archived and resolved messages. A comment survives only once the failed messages its group is built from have arrived, so on a large archive the comments are the last thing to appear. An empty comment field partway through a migration is the copy still running, not data loss
```
johnsimons, line 170: This is important to bring in, as soon as we create the group, becasue there may be info there about informing things like "do not retry this group...."

docs/migration/ravendb-to-sql-migration-overview.md
```diff
@@ -0,0 +1,294 @@
...
+- Failed message edits
```
johnsimons, line 171: What is this?

docs/migration/ravendb-to-sql-migration-overview.md
warwickschroeder: Throughput history is in the required set of categories, so it will be migrated before SC starts up. This means that when the migration starts, the throughput table in SQL is empty so nothing to merge with. Set rather than add makes the migration resumable and the checkpoint stops a second pass overwriting days the collectors have written since.

I've update the overview doc toclarify this.

docs/migration/ravendb-to-sql-migration-overview.md
warwickschroeder: Looking into what exactly raven sets to expire, its only resolved or archived FailedMessages, and EventLogItems. Both arent in the required categories, so during the migration of the required set, nothing is expired from under us. If the optional records are past their retention already, they are skipped. If a row is swept away by raven mid-copy, it is an absence rather than a skip.

I'll update the overview doc to clarify this.

src/ServiceControl.Persistence/DataMigration/MigrationEngine.cs
warwickschroeder: Fixed in https://github.com/Particular/ServiceControl/pull/5894/changes/909e4bfeaa36503e62c1d56aaccd11d8c6ab1300#diff-14df616bb97776e63067960a909ea5a788d33570699a00737b6053d4708b9f34.

I've added a sequence diagram to the overview doc to clearly show what the migration engine does and how the checkpoint/resume functionality works.

johnsimons (approved):
A few question and comments but overall, lets keep going

src/ServiceControl.Persistence/DataMigration/MigrationEngine.cs
warwickschroeder: A small number cannot halt anything, because the floor gates the percentage entirely. But a cluster of more than 100 bad rows near the start does halt a category, which is probably correct.

One bug I did notice though is that legitimate skips (past retention, orphaned group comments, etc) are being counted towards a halt. These types of skips should not count towards it. Will fix.

I've also added a section to the overview doc with a diagram to illustrate this.

src/ServiceControl.UnitTests/Migration/MigrationEngineCategorySelectionTests.cs
warwickschroeder: done

src/ServiceControl.UnitTests/Migration/MigrationEngineCategorySelectionTests.cs
warwickschroeder: done

src/ServiceControl.Persistence/DataMigration/MigrationEngine.cs
warwickschroeder: Have also added some more tests to test all paths of the halting logic

src/ServiceControl.Persistence/DataMigration/MigrationEngine.cs
warwickschroeder: Have added more tests for all paths and edge cases of the checkpoint/resume logic

docs/migration/ravendb-to-sql-migration-overview.md
rbev: Probably worth just marking them as errors in the pre-flight and force them to resolve before migrate!

docs/migration/ravendb-to-sql-migration-instructions.md
warwickschroeder: These just list the current settings for raven, which the migration engine uses to set its "Source", but yes, any new setting thats migration specific will be "ServiceControl/Migration/xxxxx"

docs/migration/ravendb-to-sql-migration-overview.md
warwickschroeder: https://github.com/Particular/ServiceControl/blob/7dd5105f295f81f3db23d8b0a44dec28e96d072f/src/ServiceControl.Persistence.RavenDB/Editing/EditFailedMessagesDataStore.cs#L9

I've expanded the dot point to add more clarity.

PR #5899, opened 2026-09-16
Move hardcoded rowlimit out of the storage implementation

rbev:
Moves the existing 200-group limit from persistence implementation into the API, changed it to explicitly request a single 200 group page instead of relying the the storage tier to limit it.

- Adds `PagingInfo` to unresolved and archived group-list datastore methods.
- Requests page 1 with a page size of 200 from both APIs
- Applies paging within RavenDB and EF Core queries. 
- Removes persistence-owned `.Take(200)` calls and `FailureGroupQueries.MaxGroups`.
- Adds shared contract tests for unresolved and archived group paging.
- Preserves existing response metadata and `GroupFetcher` merge behavior.


src/ServiceControl.Persistence.Tests/Recoverability/ArchivedGroupVersionTests.cs
```diff
@@ -28,7 +29,7 @@ public async Task Version_changes_when_group_counts_move_but_the_total_and_the_s
         await Insert(oldest, middle, newest);
         await Archive(oldest, middle, newest);
 
-        var before = await GroupsStore.GetArchivedGroupsByClassifier(Classifier);
+        var before = await GroupsStore.GetArchivedGroupsByClassifier(Classifier, new PagingInfo(page: 1, pageSize: 200));
```
warwickschroeder: create this once and pass in? Arguably more correct as the page size is currently hard capped at 200 in code. same on the other test files

src/ServiceControl.Persistence.RavenDB/Recoverability/GroupsDataStore.cs
```diff
@@ -49,8 +50,8 @@ public async Task<QueryResult<IList<FailureGroupView>>> GetArchivedGroupsByClass
                 .Where(v => v.Type == classifier);
 
             var results = await groups
-                .OrderByDescending(x => x.Last)
-                .Take(200) // only show 200 groups
+                .OrderByDescending(view => view.Last)
+                .Paging(pagingInfo)
                 .ToListAsync(cancellationToken);
 
             return new QueryResult<IList<FailureGroupView>>(results, stats.ToQueryStatsInfo());
```
warwickschroeder, line 57: Raven will return `QueryStatistics.TotalResults` from `stats`, which the total number of records from the query above `results`, not the paged value. This is correct as the total count of the non paged query should go into the `Total-Count` response header which SP uses. EF [sets the header](https://github.com/Particular/ServiceControl/pull/5899/changes#diff-1f17dfffe7a5a95918638227a563054c25a7a5f956a5ca1616b01f82092fd2b1R45) to the paged maximum but should be total records.

rbev: Done

warwickschroeder (approved):

johnsimons:
It is unclear to me what is the benefit of moving the cheese this way.
I actually think that the next person touching this will look at it and say, ah the api already supports paging, so let me add paging to the frontend! Without really taking into consideration lots of other factors.

I think a better option, is to document what we spoke about in the slack thread in the code itself.

warwickschroeder:
@johnsimons - i think im missing context here. Could you elaborate on what the other factors are?

warwickschroeder (approved):

johnsimons (approved):

PR #5903, opened 2026-09-17
Add custom check for failing retention

rbev:
This pull request introduces a new custom check for monitoring retention sweep failures in the EF Core persistence layer. It tracks and reports failures in the retention process, ensuring that any issues are surfaced via the custom checks infrastructure and are only cleared after a fully successful sweep. The changes include the implementation of the custom check, supporting logic in the sweeper, updates to configuration, and comprehensive tests.

**Retention Sweep Monitoring and Reporting:**

* Introduced a new `RetentionSweepCustomCheck` class that reports failures in the retention sweep process as a custom check, failing after any retention pass fails and recovering after the next fully successful sweep. This check summarizes failures per entity and integrates with the custom check infrastructure. [[1]](diffhunk://#diff-9ec06849561c43502454e981c942b456ef5a298f71e5cd714080e7b544840384R1-R26) [[2]](diffhunk://#diff-cf186271d78dd3ad85666cee22bbdeb18b962bcd1750b3a11e098b74239accc1R50-R51)
* Updated `RetentionSweeper` to track active failures per entity, provide methods for recording and retrieving failures, and clear failures only after all passes succeed in a sweep. The sweeper now returns pass/fail status for each retention entity and records reasons for failures. [[1]](diffhunk://#diff-da84788333b253b21e0cc71e12446fad927888f74abb8672d9b563ada96efdd4R34-R35) [[2]](diffhunk://#diff-da84788333b253b21e0cc71e12446fad927888f74abb8672d9b563ada96efdd4R134-R149) [[3]](diffhunk://#diff-da84788333b253b21e0cc71e12446fad927888f74abb8672d9b563ada96efdd4L155-R204)

**Testing and Verification:**

* Added a new test suite `RetentionSweepCustomCheckTests` to verify the behavior of the custom check, including its response to failures, recovery, time persistence, and deduplication of state change events.
* Extended existing retention sweep tests to verify that failing passes are correctly tracked and reported.

**Configuration and Integration:**

* Registered the new custom check in the DI container and updated the list of internal custom checks to include "ServiceControl Retention". [[1]](diffhunk://#diff-cf186271d78dd3ad85666cee22bbdeb18b962bcd1750b3a11e098b74239accc1R50-R51) [[2]](diffhunk://#diff-71766e4f1428f6f4966f415c33df457c8d910cb06584ae957b16f17ee4f08bcfR46) [[3]](diffhunk://#diff-3b71c2a353196227c00e8534d4171846eaab781263e5fcb2ce8a441a6c0dc314R36)
* Updated app configuration to use SQL Server persistence and provide necessary connection strings and storage paths for local development. [[1]](diffhunk://#diff-55aabd0cc6efa7efa1632d7827e55852f7d8a32a2143b1028718f63c7b21e7d2L26-R33) [[2]](diffhunk://#diff-55aabd0cc6efa7efa1632d7827e55852f7d8a32a2143b1028718f63c7b21e7d2R95-R96)

**Test Artifacts:**

* Updated approval files for PostgreSQL and SQL Server to reflect the presence of the new custom check in the list of available checks. [[1]](diffhunk://#diff-fc0c7dd67409bc7a1513e7675e27c332b8ac798b8f8098209d5963e57aaa24e8R1) [[2]](diffhunk://#diff-fc0c7dd67409bc7a1513e7675e27c332b8ac798b8f8098209d5963e57aaa24e8R1)


src/ServiceControl.Persistence.EFCore/Infrastructure/RetentionSweeper.cs
```diff
@@ -152,28 +180,38 @@ async Task Sweep(DateTime? errorCutoff, DateTime? eventsCutoff, bool pace, Cance
...
-
             cycle.Complete();
+            return true;
         }
 #pragma warning disable PS0019 // The filter already excludes OperationCanceledException, so
         // cancellation propagates; PS0019 only recognises a cancellationToken guard.
         catch (Exception ex) when (ex is not OperationCanceledException)
         {
             logger.LogError(ex, "Error during the {RetentionEntity} retention pass", entity);
+            RecordFailure(entity, ex.Message);
```
johnsimons, line 213: We are in essence displaying `ex.Message` to customers, is that going to be useful for them to solve the problem?
Will it prevent them from raising a support case with us?
What action should they take if they see such error in ServicePulse?

johnsimons (approved):
From a coding perspective this PR is fine, no issues.
But displaying a custom check without an action plan for the customer to address it is not going to help, we may as well tell them to raise a support case.

This may need more thought.

src/ServiceControl.Persistence.EFCore/Infrastructure/RetentionSweeper.cs
```diff
@@ -152,28 +180,38 @@ async Task Sweep(DateTime? errorCutoff, DateTime? eventsCutoff, bool pace, Cance
...
+        var eventLogSucceeded = await RunPass(RetentionEntity.EventLog, token => SweepEventLogItems(pace, eventsCutoff, token), cancellationToken);
+        var groupCommentsSucceeded = await RunPass(RetentionEntity.GroupComments, SweepOrphanedGroupComments, cancellationToken);
+
+        if (failedMessagesSucceeded && eventLogSucceeded && groupCommentsSucceeded)
+        {
+            lock (failures)
+            {
+                failures.Clear();
+            }
+        }
```
warwickschroeder, line 193: Should we be clearing individual failures as they become successful rather than waiting for all 3? If 1 is fixed, it'l still show all 3 failing until all 3 are resolved.

Take a look at `RentetionMetrics` which is also tracking retention failures. Could we re-use at all?

src/ServiceControl.Persistence.EFCore/Infrastructure/RetentionSweeper.cs
```diff
@@ -129,6 +131,32 @@ async Task SweepWithoutAcquiringLock()
 
     public RetentionSweepCurrentStatus GetStatus() => new(isRunning, lastStartedAt, lastFinishedAt, lastErrorCutoff, lastEventsCutoff);
```
warwickschroeder, line 132: Will the custom check pass if the retention never starts? Could we also raise a failure for this as well?

src/ServiceControl.Persistence.EFCore/Infrastructure/RetentionSweeper.cs
warwickschroeder, line 213: Take a look at the pattern used in the below. Could we use this to move some of the failure logic out of the retention sweeper? https://github.com/Particular/ServiceControl/blob/c41e48bd3455b7b0977ded1d3702a12f689639f4/src/ServiceControl/Operations/ErrorIngestionCustomCheck.cs#L8

rbev: Do you have any suggestions to what action we could give them? 

The main thing that I'd see this failing on would either be deadlocks or timeouts from an overloaded DB, a completely unreachable DB takes ServicePulse offline equivalent to the SC instance disappearing.

src/ServiceControl.Persistence.EFCore/Infrastructure/RetentionSweeper.cs
rbev: Yeah that's much better

src/ServiceControl.Persistence.EFCore/Infrastructure/RetentionSweeper.cs
rbev: Done

rbev: Added a link pointing to the troubleshooting docs

src/ServiceControl.Persistence.EFCore/Infrastructure/RetentionSweepCustomCheck.cs
```diff
@@ -0,0 +1,60 @@
...
+    : CustomCheck("ServiceControl Retention", "ServiceControl Health", TimeSpan.FromMinutes(1))
+{
+    public override Task<CheckResult> PerformCheck(CancellationToken cancellationToken = default)
+    {
+        var failures = state.GetFailures();
+        var failureSummary = string.Join("; ", failures.Select(failure => $"{failure.Key}: {failure.Value}"));
+
+        return Task.FromResult(state.ConsecutiveFailedSweeps < 3
+            ? CheckResult.Pass
+            : CheckResult.Failed($"Retention processing has failures. Last failure per entity: {failureSummary}. See https://docs.particular.net/servicecontrol/troubleshooting for guidance on resolving the issue."));
```
johnsimons, line 19: We may need to adjust ServicePulse to render these links correctly

johnsimons (approved):

PR #5904, opened 2026-09-17
Add Roslyn analyser to block DateTime Now properties

rbev:
Follow up from https://github.com/Particular/ServiceControl/pull/5843

Adds new entries to the `BannedSymbols.txt` file to discourage direct usage of certain date and time properties. Instead, it recommends using `TimeProvider` as a replacement.


johnsimons (approved):

warwickschroeder (commented):
Is the BannedSymbols file even used/referenced? There should be warnings but there arnt. https://github.com/Particular/ServiceControl/blob/1dc6fb870d7fb81425fc52abf3dff1eb4673b503/src/ServiceControl.LicenseManagement/LicenseDetails.cs#L61

Also, once wired up would throw errors in CI as it treats warnings as errors.

Should "DateTime.Today" be banned as well?

rbev:
> Is the BannedSymbols file even used/referenced? There should be warnings but there arnt.
> 
> https://github.com/Particular/ServiceControl/blob/1dc6fb870d7fb81425fc52abf3dff1eb4673b503/src/ServiceControl.LicenseManagement/LicenseDetails.cs#L61
> 
> Also, once wired up would throw errors in CI as it treats warnings as errors.
> 
> Should "DateTime.Today" be banned as well?

That's so strange, I was surprised I got no errors so I added a reference, it wasn't a warning it completely failed the build locally without a `#pragma`. I didn't go searching, I just assumed you did an amazing job removing them all.

I'll look into why those ones don't create errors.

johnsimons (commented):
Shouldn't we use 
```
dotnet_diagnostic.RS0030.severity = hint
```
So that devs are aware of it?

Actually, to make sure devs do the right thing, should we use inline pragmas, and not disable it in `.editorconfig`?

rbev:
Abandoning as it was decided that this analyser did not provide enough value while there are so many exceptions in the project.

PR #5911, opened 2026-09-21
[Migration Engine Part 3] Implement the RavenDB to SQL migration machinery

warwickschroeder:
## What this adds

The machinery that copies an instance's data out of RavenDB into SQL Server or PostgreSQL while ServiceControl is closed, which is the only window in which a copy can still be thrown away. **Only `KnownEndpoints` and `EndpointSettings` can be copied, so a real instance with `Migration/Enabled` set still refuses to start**: the check that demands all twelve required categories has an escape hatch only the acceptance tests register. Message bodies are not read yet, and nothing acts on `Migration/AllowIncompleteExit`.

- **The copy runs before the host opens.** It finishes inside host startup, before a port is bound or anything is ingested, and the host does not open if a category did not finish.
- **Progress is saved with the data it describes.** Each batch commits its rows and its checkpoint together, so a restart carries on from the last committed batch, and a second instance writing the same checkpoints is refused rather than allowed to overwrite them.
- **Only categories both ends support are attempted.** The registry names twelve required and six optional categories, the source and the target each say what they can handle, and the rest never reach the engine.
- **Checks refuse before anything is copied.** The supported pair, this build's coverage, the settings, the target's schema, its retry history depth, its body storage and the source's data version are each checked first, in a fixed order.
- **A copy that stops getting anywhere is stopped and explained.** A stall, too many skipped rows or an unfinished category each end in a refusal naming the category, with what to fix and how to resume.
- **Going back stays free until the host opens on the target.** That moment is stamped in the target database, and every refusal says how to return the instance to RavenDB while it is still true.
- **`--migration-source-report` reports what a migration would read, and writes nothing.**
- **Docs.** `docs/migration/` gains an overview, a system design and operator instructions, linked from `README.md`.

## Tests

- `ServiceControl.UnitTests/Migration`: the engine's copy, resume, halt and skip paths, each startup check, the stall watchdog and the refusal messages, against in-memory fakes and a fake clock.
- `ServiceControl.Persistence.Tests/EFCore/Migration`: the target and its two writers on SQL Server and PostgreSQL, plus a claim that fails when an entity is added without a decision about migrating it.
- `ServiceControl.Persistence.Tests.RavenDB/DataMigration`: the readers, the source report and the read-only lifecycle against an embedded server.
- `ServiceControl.Persistence.Tests.SqlServer`: two keys differing only in case landing on one row, and the retry history depth reaching the persister.
- `ServiceControl.Migration.AcceptanceTests` (new): a killed copy, a restart, each refusal and the host opening on the target, run against both providers.
- `ServiceControl.Migration.Tests` (new project): category coverage, the source report command, and two persisters loaded in one process.


src/ServiceControl.Persistence.EFCore/DataMigration/Writers/KnownEndpointsWriter.cs
```diff
@@ -0,0 +1,65 @@
...
+    // Opens a scope of its own, because BatchSize is handed no DbContext, unlike Prepare and Count.
+    public int BatchSize
+    {
+        get
+        {
+            using var scope = scopeFactory.CreateScope();
+            var entityType = scope.ServiceProvider.GetRequiredService<ServiceControlDbContext>().Model.FindEntityType(typeof(KnownEndpointEntity))
+                ?? throw new InvalidOperationException($"{nameof(KnownEndpointEntity)} is not an entity in the EF Core model.");
+
+            return migrationDialect.RowsPerStatement(entityType.GetProperties().Count());
```
johnsimons: This pattern seems to repeat in other writers

PR #5914, opened 2026-09-22
Using spy endpoints for testing for messages to ServiceControl

afprtclr:
Replacing ScenarioContext with endpoint to assert the logic.


src/ServiceControl.Audit.AcceptanceTests/Monitoring/When_a_new_endpoint_is_detected.cs
```diff
@@ -1,31 +1,31 @@
...
-            var context = await Define<InterceptedMessagesScenarioContext>()
+            var context = await Define<MyContext>()
                 .WithEndpoint<Receiver>(b => b.When((bus, c) => bus.SendLocal(new MyMessage())))
-                .Done(c => c.SentRegisterEndpointCommands.Any())
+                .WithEndpoint<SpyEndpoint>()
+                .Done(c => c.ReceivedRegisterEndpointCommands.Any())
                 .Run();
 
-            var command = context.SentRegisterEndpointCommands.Single();
+            var command = context.ReceivedRegisterEndpointCommands.Single();
```
mchamberlin77: Why does this change from a SentRegisterEndpointCommands to a ReceivedRegisterEndpointCommands?

afprtclr: This is just a collection name, not a "logic" change. See that it's defined below in line 79. Originally, it was defined in InterceptedMessageScenarioContext which is now removed.

Why the rename? Given that the handler *receives* this command and stores it, it makes sense to call it *received* to indicate that "sending is not enough, it must be also delivered and received". Anyway, I'll rename back to *Sent* as it doesn't really matter.

mchamberlin77: Thanks for the explanation.  No need to change it back, I was just not clear on what drove that change.

mchamberlin77 (approved):

